PR #6, opened 2020-06-20
Add datetime types

Andrew-Chen-Wang:
Added dumpers for date time types. Not sure why the decoding/Loaders aren't working. The Current test case for loaders:

```
@pytest.mark.parametrize(
    "val, pgtype, want",
    [
        ("2000-12-12", "date", datetime.date(year=2000, month=12, day=12)),
    ],
)
@pytest.mark.parametrize("fmt_out", [Format.TEXT, Format.BINARY])
def test_load_datetime(conn, val, pgtype, want, fmt_out):
    cur = conn.cursor(binary=fmt_out == Format.BINARY)
    cur.execute(f"select %s::{pgtype}", (val,))
    assert cur.pgresult.fformat(0) == fmt_out
    assert cur.pgresult.ftype(0) == builtins[pgtype].oid
    result = cur.fetchone()[0]
    assert result == want
    assert type(result) is type(want)

    # arrays work too
    cur.execute(f"select array[%s::{pgtype}]", (val,))
    assert cur.pgresult.fformat(0) == fmt_out
    assert cur.pgresult.ftype(0) == builtins[pgtype].array_oid
    result = cur.fetchone()[0]
    assert result == [want]
    assert type(result[0]) is type(want)

```

Looking at the decoder, we know the encoding is set to UTF8, but when I print the bytes, I get some bytes that apparently can't be decoded; for example: `b'\xff\xf4\xdb\xf9'` and `b"\x00\x00\x01Z"`

At first, I thought this was  my machine's error, but the pytest was working fine with the text loaders and others. (That was until I switched to using machine/endian.h and everything stopped working :P Moved to gitpod's linux based for now.)

Build: https://travis-ci.org/github/psycopg/psycopg3/builds/700446307 Some reason not showing on gh. Lint failed; build was fine


psycopg3/types/datetime.py
```diff
@@ -0,0 +1,112 @@
...
+@Loader.binary(builtins["timestamptz"].oid)
+class DateTimeLoader(Loader):
+
+    decode: Optional[DecodeFunc]
+
+    def __init__(self, oid: int, context: AdaptContext):
+        super().__init__(oid, context)
+
+        if self.connection is not None:
+            if self.connection.encoding != "SQL_ASCII":
```
dvarrazzo, line 83: We don't need special-casing sqlascii database. datetime formats are ascii only, unless I'm mistaken.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,112 @@
...
+@Loader.text(builtins["date"].oid)
+@Loader.binary(builtins["date"].oid)
+@Loader.text(builtins["time"].oid)
+@Loader.binary(builtins["time"].oid)
+@Loader.text(builtins["timetz"].oid)
+@Loader.binary(builtins["timetz"].oid)
+@Loader.text(builtins["timestamp"].oid)
+@Loader.binary(builtins["timestamp"].oid)
+@Loader.text(builtins["timestamptz"].oid)
+@Loader.binary(builtins["timestamptz"].oid)
```
dvarrazzo, line 74: You can't use the same functions for binary and text decoding. I'm sure text format and binary format for dates are different.

Most likely these decorator will go each one on a different function, each one handling its own format.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,112 @@
...
+            self.decode = codecs.lookup("utf8").decode
+
+    def load(self, data: bytes) -> Union[datetime.datetime, datetime.date, datetime.time]:
+        obj = datetime.datetime.strptime(
+            self.decode(data)[0],
+            f"{'%Y-%m-%d' if self.oid == builtins['date'].oid else ''}"
+            f"{'%H:%M:%S' if self.oid in _time_oids else ''}"
+            f"{'%z' if self.oid in _timezone_oids else ''}"
+        )
+        if self.oid == builtins['date'].oid:
```
dvarrazzo, line 97: You shouldn't have ifs in the decoder. This function wants to decode 10 different formats. You should have 10 different functions each one decoding a single format. If two format happens to be the same then that's cool, but that's more of a special case.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,112 @@
...
+            return obj.date()
+        elif self.oid in _timezone_oids:
+            return obj
+        else:
+            return obj.time()
+
+
+@Loader.text(builtins["interval"].oid)
+def load_interval(data: bytes) -> datetime.timedelta:
+    pass
```
dvarrazzo, line 107: It should be `raise NotImplementedError` here. The function doesn't respect the signature.

psycopg3/types/datetime.pyx
```diff
@@ -0,0 +1,17 @@
...
+
+from libc.stdint cimport *
+
+from psycopg3.types.text cimport load_text, get_context_text
+
+
+cdef void register_text_c_loaders():
+    logger.debug("registering optimised datetime c loaders")
+    from psycopg3.types import datetime
+    register_c_loader(datetime.DateTimeLoader.load, load_text, get_context_text)
```
dvarrazzo, line 16: I'd rather these to be undefined than set to the wrong ones. Probably better to work on the python side of it and then write an optimised implementation.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,112 @@
+"""
+Adapters for datetime types.
+https://www.postgresql.org/docs/9.4/datatype-datetime.html#DATATYPE-DATETIME-TABLE
```
dvarrazzo, line 3: 9.4 is not supported yet. Use the `current` url.

psycopg3/types/datetime.py
Andrew-Chen-Wang: Ah slipped in...

psycopg3/types/datetime.py
Andrew-Chen-Wang: I see, thanks!

psycopg3/types/datetime.py
Andrew-Chen-Wang: I see. I wasn't quite sure what it was for in the text type since I'm not too well versed in jumping too far into PostgreSQL.

Andrew-Chen-Wang:
Thanks for the review @dvarrazzo! Yea, not very knowledgable about binary formats. Will update with suggestions tomorrow! Hopefully with test cases that somewhat resemble that of psycopg2 so that I don't miss anything.

psycopg3/types/datetime.py
dvarrazzo: For text, it's a special case for the `SQL_ASCII` format, because, rather than being "the database is encoded in ASCII", it means "the database has no encoding". That means it's pretty much a binary database: people can throw any byte between 1 and 255 at it and expect to find it back as it was. As such, for text, we don't decode the bytes received from database to an unicode `str`, like we normally do for `text` data type but we return them as python `bytes`.

This only happens for text types though: dates are only in formats including numbers and symbols < 128, so for them we don't really care about the encoding: we can expect that decoding from ASCII will work, and if it doesn't there is some other error somewhere.

dvarrazzo:
Yes, in psycopg2 there are a few test cases, including some peculiar ones with weird timezones, and it's a perfect place to start from. However in psycopg2 test suite you will find only:

- text format
- datestyle iso

which is all psycopg2 supports, whereas for psycopg3 I wish we can support binary format and different (text) datestyles. It's a good start though.

Andrew-Chen-Wang:
Ah the different date styles kind of confused me since I couldn't find any differing things formatting-wise in psycopg2. Makes sense now. So PostgreSQL supports [three + four default formats for input and output](https://www.postgresql.org/docs/12/runtime-config-client.html#GUC-DATESTYLE) that one specifies in the postgresql.conf file, with `SET` command, or the `PAGEDATESTYLE` environment variable.

Are the latter two methods -- specified [here](https://www.postgresql.org/docs/12/datatype-datetime.html#DATATYPE-TIMEZONES) -- changed together? Would we want the dev to set that via postgresql.conf or via psycopg3? How would we detect if postgresql.conf changes? Thanks!

Hopefully I can learn enough from the date times to make the rest of the types files. And soon enough figure out why this doesn't work on OS X... (yes I changed psycopg3/types/endian.pxd to machine/endian.h, but it completely broke my setup, so I had to resort to something else haha. I'm not sure how I set it up in the first place).

dvarrazzo:
Note that we need only to support different output formats, because postgres output becomes the loader input, which we parse into a python date object. In output my understanding is that if the dates always start with YYYY postgres figures out the format. Of course tests in the test suite would prove all the combinations work properly.

The way to test those things is to use `set` in the connection and then verify parsing works ok. When you want to read the current format you can use `conn.pgconn.parameter_status(b"DateStyle")`.

However, in terms of priority, supporting German datestyle comes later. First let's try to get some good basic text casters assuming ISO format, then let's get the binary ones, later we can look at non-standard ones.

Andrew-Chen-Wang:
@dvarrazzo Learned a lot figuring out the date loaders! I have a question regarding `datetime.pyx`. Although you said not to worry about the optimizations in that file, all the binary outputs for date and time are actually in DateADT and TimeADT which are Int32 and Int64 respectively. With what I have now, does it make sense? I feel like I could've also imported from `numeric.pyx` (if possible. I've never properly used cython for a package before).

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,146 @@
...
+@Loader.text(builtins["date"].oid)
+def load_date(data: bytes) -> datetime.date:
+    # TODO Multiple formats perhaps could do a try-except for loop?
+    return datetime.datetime.strptime(_decode(data)[0], "%Y-%m-%d").date()
+
+
+@Loader.binary(builtins["date"].oid)
+def load_date_binary(data: bytes) -> datetime.date:
+    dateADT = struct.unpack("!i", data)[0]  # num days since 2020-01-01
+    return datetime.date(year=2000, month=1, day=1) + datetime.timedelta(
```
dvarrazzo, line 63: You can have `datetime.date(year=2000, month=1, day=1)` as a module constant and avoid this function call at every decode.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,146 @@
...
+    return (
+        datetime.datetime(year=1, month=1, day=1)
+        + datetime.timedelta(microseconds=struct.unpack("!q", data)[0])
+    ).time()
+
+
+@Loader.text(builtins["timetz"].oid)
+def load_time_tz(data: bytes) -> datetime.time:
+    decoded: str = _decode(data)[0]
+    if decoded[::-1][2] in ("+", "-"):
```
dvarrazzo, line 84: You meant `decoded[-3]` maybe? :) No need to create a new string here. `if decoded[-3] in "+-"` is legit.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,146 @@
...
+        # Python usually expects +HHMM format for TZ
+        decoded += "00"
```
dvarrazzo, line 86: Please note that you will get `[+-]HH[:MM]` if the timezone has minutes offset. There are seconds too for extra fun, but only in historical cases.

```
piro=# SET TIME ZONE 'Asia/Calcutta';
SET
piro=# SELECT '1930-01-01 10:30:45'::timestamptz::timetz;
     timetz     
----------------
 10:30:45+05:30
(1 row)
```

the psycopg2 test suite should be pretty complete as funny cases.

dvarrazzo:
> Learned a lot figuring out the date loaders!

I'm pleased :)

> I have a question regarding datetime.pyx. Although you said not to worry about the optimizations in that file, all the binary outputs for date and time are actually in DateADT and TimeADT which are Int32 and Int64 respectively. With what I have now, does it make sense? I feel like I could've also imported from numeric.pyx (if possible. I've never properly used cython for a package before).

I'm not sure I understand. Do you mean about reusing the parsing from an int32/64 from the network into a Python object? I seem to understand that a cython package has a flat namespace and you can just use functions defined elsewhere (but I struggle with it sometimes).

> Edit: https://travis-ci.org/github/psycopg/psycopg3/builds/701004428 not sure why it's failing since the test is just copied from test_numeric.py... and test_numeric.py is passing. Local machine is passing as well.

Because of my change binary -> format or there is something else?

> Edit3: Working on timestamptz loader but the timezone is off by an entire second:

> tzinfo=datetime.timezone(datetime.timedelta(microseconds=-timezone)),
) + datetime.timedelta(seconds=-seconds)

timezone = microseconds sounds wrong. IIRC binary timestamptz is utc only, there is no TZ info.

> Also note, it might be better to use travis-ci.com rather than .org. Idk why but .org banners don't show up in my own repos, so I migrated them.

I'm not sure I understand. What banner?

psycopg3/types/datetime.py
Andrew-Chen-Wang: Ah right. Still haven't taken the time to copy the psycopg2 test suite. Added the weird ones to me first.

The `decoded[::-1][2]` detects if the timezone returned is only in `[+-]HH` format. Python's strptime with %z can work with both HHMM and HH:MM format according to test cases with:
```
(
        "00:00:01-1030",
        "timetz",
        datetime.time(
            hour=0,
            minute=0,
            second=1,
            tzinfo=datetime.timezone(
                datetime.timedelta(hours=-10, minutes=-30)
            ),
        ),
    ),
```
and
```
        "00:00:01-10:30"
```
Regarding the historical seconds, I'm not sure if strptime can work with that... then again we're looking to support multiple formats, so strptime may also not be the future of the datetime w/ tz loaders.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,146 @@
...
+
+@Loader.text(builtins["date"].oid)
+def load_date(data: bytes) -> datetime.date:
+    # TODO Multiple formats perhaps could do a try-except for loop?
+    return datetime.datetime.strptime(_decode(data)[0], "%Y-%m-%d").date()
+
+
+@Loader.binary(builtins["date"].oid)
+def load_date_binary(data: bytes) -> datetime.date:
+    dateADT = struct.unpack("!i", data)[0]  # num days since 2020-01-01
```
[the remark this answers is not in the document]
Andrew-Chen-Wang: Settled with that var name because I was programming in Swift simultaneously. Sorry, different styles get me confused sometimes.

I was honestly not sure what ADT is. I took a look at the source code (finally figured out that recv suffixes were the functions I needed to find), did some math finagling to figure out the `datetime.date(year=2000, month=1, day=1)` thing and finally ended up just using the class name as the variable name.

I did notice the numeric.py stuff, and initially I wanted to copy the style you did it. But the file seemed more easily navigable for me since I was still trying to figure out structs, unpacking, and figuring out/searching the Postgres source code. Fixing.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,146 @@
...
+
+from ..adapt import Dumper, Loader
+from .oids import builtins
+
+_encode = codecs.lookup("ascii").encode
+_decode = codecs.lookup("ascii").decode
+
+
+@Dumper.text(datetime.date)
+@Dumper.binary(datetime.date)
```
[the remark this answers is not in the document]
Andrew-Chen-Wang: I'm... stupid haha.

Andrew-Chen-Wang:
> Do you mean about reusing the parsing from an int32/64 from the network into a Python object?
Basically, should I try using Cython for these areas? And, if so, could I import from numeric.pyx? Not entirely sure the way I went about datetime.pyx was correct...
> Because of my change binary -> format or there is something else?
Ah I didn't realize that'd break on my branch though since I started first commit before your commit. Plus, it was working on my local machine so didn't think much of it until then.
> timezone = microseconds sounds wrong. IIRC binary timestamptz is utc only, there is no TZ info.
Interesting... when I was looking through the source code, the returned thing was TimeADT and Timezone with each taking up 4 bytes (I can't believe I lost the source code of where I found this...). Although, the other loaders that used microseconds were in 8 bytes, so I'm not sure... I guess more Googling Postgres doc pages until I stumble upon it again...
> I'm not sure I understand. What banner?
The checkmark or X next to a successful/failed build.

psycopg3/types/datetime.py
```diff
@@ -0,0 +1,147 @@
...
+    decoded: str = _decode(data)[0]
+    if decoded[-3] in "+-":
+        # Python usually expects +HHMM format for TZ
+        decoded += "00"
+    return datetime.datetime.strptime(decoded, "%H:%M:%S%z").timetz()
+
+
+@Loader.binary(builtins["timetz"].oid)
+def load_time_tz_binary(data: bytes) -> datetime.time:
+    microseconds, timezone = struct.unpack("!q i", data)
```
Andrew-Chen-Wang, line 93: Gee whizz... missed one.

dvarrazzo:
Looks like what's biting you on Python 3.6 is the lack of support for non whole minutes: https://docs.python.org/3/library/datetime.html#datetime.datetime.dst

dvarrazzo:
FYI, if you know a type in postgres you can look at its send and receive function: these are pretty much the loader/parser of the binary formats as implemented in postgres. For instance:

```
piro=# select typsend, typreceive from pg_type where oid = 'timetz'::regtype;
   typsend   | typreceive  
-------------+-------------
 timetz_send | timetz_recv
(1 row)
```

you can find these functions in postgres code, starting from `src/backend` and usually the name of the function starts from column 0 where it's defined, so:

```
piro@makkuro:~/dev/fs/postgres$ ag ^timetz_send src/backend/
src/backend/utils/adt/date.c
2063:timetz_send(PG_FUNCTION_ARGS)
```

dvarrazzo:
FYI, if you know a type in postgres you can look at its send and receive function: these are pretty much the loader/parser of the binary formats as implemented in postgres. For instance:

```
piro=# select typsend, typreceive from pg_type where oid = 'timetz'::regtype;
   typsend   | typreceive  
-------------+-------------
 timetz_send | timetz_recv
(1 row)
```

you can find these functions in postgres code, starting from `src/backend` and usually the name of the function starts from column 0 where it's defined, so:

```
piro@makkuro:~/dev/fs/postgres$ ag ^timetz_send src/backend/
src/backend/utils/adt/date.c
2063:timetz_send(PG_FUNCTION_ARGS)
```

dvarrazzo:
> Edit: Found the recv for it: https://doxygen.postgresql.org/backend_2utils_2adt_2timestamp_8c.html#ab16b570ee458902dec4844a8b9b23748. But I can't find the constructor...

The constructor is `timestamptz_send`, [just nearby](https://github.com/postgres/postgres/blob/7ce461560159948ba0c802c767e42c5f5ae08b4a/src/backend/utils/adt/timestamp.c#L819) in the source code.

dvarrazzo:
> Not sure. If timestamptz didn't carry tzinfo, then what would make the difference between timestamp and timestamptz?

None, in the internal format. The difference is only how it's interpreted back. Naive ts don't get a timezone, tz-aware ts get converted in the client timezone. Little experiment: this is the internal representation of 2020-01-01 midnight utc:
```
piro=# select timestamptz_send('2020-01-01T00:00+00'::timestamptz);
  timestamptz_send  
--------------------
 \x00023e0786c26000
```

it's exactly the same of 1am of the timezone one hour ahead - because it's actually the same moment:
```
piro=# select timestamptz_send('2020-01-01T01:00+01'::timestamptz)
       = timestamptz_send('2020-01-01T00:00+00'::timestamptz);
 ?column? 
----------
 t
(1 row)
```

and it happens to be the same binary representation of the naive timestamp:

```
piro=# select timestamptz_send('2020-01-01T01:00+01'::timestamptz)
       = timestamp_send('2020-01-01T00:00'::timestamp);
 ?column? 
----------
 t
(1 row)
```

The difference is set by the session timezone, and whether the client decides to make use of it:
```
piro=# set timezone to utc;
SET

piro=# select '2020-01-01T00:00+00'::timestamptz;
      timestamptz       
------------------------
 2020-01-01 00:00:00+00
(1 row)

piro=# set timezone to 'America/New_York';
SET

piro=# select '2020-01-01T00:00+00'::timestamptz;
      timestamptz       
------------------------
 2019-12-31 19:00:00-05
(1 row)
```

The timezone setting is kept in consideration by `timestamptz_out` (the function converting to the text format) but not by `timestamptz_send` (the one converting the internal format to network format).

So, one correct way to transform this in python is to parse the tz-aware exactly like the naive ts and slap it an utc tzinfo. That's already a correct result.

Another is do first do the above, then to convert it to the client offset by using pytz, which then would become a psycopg dependency. From the connection info you can get the session timezone name:
```
In [43]: conn.pgconn.exec_(b"set timezone to 'America/New_York'")                                              
In [46]: conn.pgconn.parameter_status(b'TimeZone')                                                             
Out[46]: b'America/New_York'
```

Then if you have your utc timestamp:
```
In [57]: dt.datetime(2020,1,1, tzinfo=dt.timezone.utc)                                                         
Out[57]: datetime.datetime(2020, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
```
it can be converted in the client timezone:
```
In [48]: import pytz                                                                                           
In [50]: pytz.timezone(_46)                                                                                    
Out[50]: <DstTzInfo 'America/New_York' LMT-1 day, 19:04:00 STD>

In [59]: _57.astimezone(_50)                                                                                   
Out[59]: datetime.datetime(2019, 12, 31, 19, 0, tzinfo=<DstTzInfo 'America/New_York' EST-1 day, 19:00:00 STD>)
```

Of course this is extra work, but I guess it's what people expect.

dvarrazzo:
Closing for inactivity.

date types are implemented in Python/text, they need to be implemented in C and binary.

Andrew-Chen-Wang:
Ah it was fun trying and learning. Thanks!

PR #7, opened 2020-07-26
Tests: Create svcconn with autocommit on

asqui:
So that you don't deadlock the test run if you forget to call commit()

### DIFF
--- a/tests/fix_db.py
+++ b/tests/fix_db.py
@@ -62,6 +62,6 @@ def svcconn(dsn):
     """
     from psycopg3 import Connection
 
-    conn = Connection.connect(dsn)
+    conn = Connection.connect(dsn, autocommit=True)
     yield conn
     conn.close()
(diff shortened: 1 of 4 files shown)


dvarrazzo:
I guess it makes sense. Or one can remember to not forget ;) How does a deadlock happen with this connection?

There's another occurrence of `svcconn.commit()` in `tests/types/test_composite.py`: could you fix it too please?

thank you!

asqui:
e.g. If you use svcconn to create a test table for use by tests, and 
forget to call svcconn.commit(), the first test to (try to) touch the
table will block forever waiting on the svcconn transaction to commit.

(And I'm using svcconn to manage the test table rather than conn because I want to be able to drop it after the test. If I use conn then tests can interfere with this by, for example, closing conn during the course of the test.)

I've force-pushed a change to include `test_composite.py` and add the first part of this comment to the commit message.

dvarrazzo:
cherry-picked

PR #8, opened 2020-08-07
Don't be so eager to execute 'begin'

asqui:
Only execute an automatic 'begin' statement if the TransactionStatus is
IDLE, so that a failure to execute 'begin' doesn't overshadow a failure
to execute a user query.

### DIFF
--- a/psycopg3/psycopg3/connection.py
+++ b/psycopg3/psycopg3/connection.py
@@ -241,7 +241,7 @@ class Connection(BaseConnection):
         if self._autocommit:
             return
 
-        if self.pgconn.transaction_status == TransactionStatus.INTRANS:
+        if self.pgconn.transaction_status != TransactionStatus.IDLE:
             return
 
         self.pgconn.send_query(b"begin")
@@ -349,7 +349,7 @@ class AsyncConnection(BaseConnection):
         if self._autocommit:
             return
 
-        if self.pgconn.transaction_status == TransactionStatus.INTRANS:
+        if self.pgconn.transaction_status != TransactionStatus.IDLE:
             return
 
         self.pgconn.send_query(b"begin")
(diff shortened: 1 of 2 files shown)


psycopg3/psycopg3/connection.py
```diff
@@ -349,7 +349,7 @@ async def _start_query(self) -> None:
         if self._autocommit:
             return
 
-        if self.pgconn.transaction_status == TransactionStatus.INTRANS:
+        if self.pgconn.transaction_status == TransactionStatus.IDLE:
```
dvarrazzo, line 352: Maybe != IDLE?

asqui: Derp! Maybe I need to get some sleep.

Also, I should look at making async work on Windows so I can actually run these tests locally :-/

dvarrazzo:
Cherry-picked

asqui:
Just out of interest, why did you need to cherry-pick this commit? (I purposely rebased my change off the lastest psycopg:master before raising this PR so that it would merge/fast-forward cleanly -- just wondering if I should do something differently in future.)

dvarrazzo:
I think merging from github creates a
```
o
| \
|  o
| /
o
```
anyway, and from command line `git merge your-branch` may or may not create a merge, whereas `git cp your-branch` will not for sure. The information that your changeset was a single commit was more readily available than the information that the branch would have fast-forwarded.

In general I don't have strong policies around always FF/never FF. In other projects I tend to `merge --no-ff` branches with some independent history (multiple commits on the same topic), even if they would ff, and to ff one-commit changeset which are "ordinary work".

So just applying muscle memory here. My suggestion to rebase your yesterday was because I changed the whole project structure a few days ago and I preferred you to check if that hadn't caused issues to your transactions work, but it's not a general rule of mine.

Is that ok or do you think we'd rather review some practice?

Thank you very much!

asqui:
Thanks for the comprehensive response! That all makes perfect sense.

It sounds it might be useful to enable "Allow rebase merging" at https://github.com/psycopg/psycopg3/settings to facilitate the "ff one-commit changeset" workflow with one-click in github.

PR #11, opened 2020-11-24
Adapt find_library() call to find libpq.dll on Windows

asqui:
Open to suggestions on:

1. How to 'detect' Windows: There are a number of options: `os.name`, `sys.platform`, `os.uname()`, and I have no strong opinion on which is best to use.
2. How to de-duplicate this code (if deemed necessary, since there are only two occurrences, one of which is in test code): I looked around for places where to put a helper method (and fixed a typo in `pq.__init__` while I was at it) but wasn't sure where it would best fit.

### DIFF
--- a/tests/fix_pq.py
+++ b/tests/fix_pq.py
@@ -1,5 +1,7 @@
 import re
 import operator
+import sys
+
 import pytest
 
 
@@ -38,7 +40,10 @@ def libpq():
 
     try:
         # Not available when testing the binary package
-        libname = ctypes.util.find_library("pq")
+        if sys.platform == "win32":
+            libname = ctypes.util.find_library("libpq.dll")
+        else:
+            libname = ctypes.util.find_library("pq")
         assert libname, "libpq libname not found"
         return ctypes.pydll.LoadLibrary(libname)
     except Exception as e:
--- a/psycopg3/psycopg3/pq/__init__.py
+++ b/psycopg3/psycopg3/pq/__init__.py
@@ -37,7 +37,7 @@ PGcancel: Type[proto.PGcancel]
 
 def import_from_libpq() -> None:
     """
-    Import pq objects implementation from the best libpw wrapper available.
+    Import pq objects implementation from the best libpq wrapper available.
 
     If an implementation is requested try to import only it, otherwise
     try to import the best implementation available.
--- a/psycopg3/psycopg3/pq/_pq_ctypes.py
+++ b/psycopg3/psycopg3/pq/_pq_ctypes.py
@@ -6,13 +6,17 @@ libpq access using ctypes
 
 import ctypes
 import ctypes.util
+import sys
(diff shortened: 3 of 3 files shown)


pmav99:
> How to 'detect' Windows: There are a number of options: os.name, sys.platform, os.uname(), and I have no strong opinion on which is best to use.

I think that nowadays `platform.platform` is the preferred way: https://docs.python.org/3/library/platform.html

tests/fix_pq.py
```diff
@@ -38,7 +40,10 @@ def libpq():
 
     try:
         # Not available when testing the binary package
-        libname = ctypes.util.find_library("pq")
+        if sys.platform == "win32":
+            libname = ctypes.util.find_library("libpq.dll")
+        else:
+            libname = ctypes.util.find_library("libpq")
```
dvarrazzo, line 46: I guess this has to be `"pq"`

asqui:
> > How to 'detect' Windows: There are a number of options: os.name, sys.platform, os.uname(), and I have no strong opinion on which is best to use.
> 
> I think that nowadays `platform.platform()` is the preferred way: https://docs.python.org/3/library/platform.html

Yay, more options! :-)

So `platform.platform()` is documented as "The output is intended to be human readable rather than machine parseable." so I'm not sure that's the best option, but maybe `platform.system() == 'Windows'`?

tests/fix_pq.py
asqui: I've force-pushed a fix for this typo (in both places).

dvarrazzo:
> The output is intended to be human readable rather than machine parseable

then I think `sys.platform` is a better choice. If any `platform.system()` could be useful but I don't see advantage on top of `sys.platform` yet.

dvarrazzo:
Is there any chance to set up tests on Windows, on Appvveyor?  (and even better would be to generate the packages)

Andrew-Chen-Wang:
GitHub actions should allow you to use Windows runners. It has Windows 2019, latest, and 2016: https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners

dvarrazzo:
Thank you @Andrew-Chen-Wang I should actually explore to move to it for Linux too, because Travis is being terrible lately.

Andrew-Chen-Wang:
Yea, I've been noticing a slow down too. As a heads up, it seems like GitHub actions' services use Docker which can only be run on linux runners right now. This workflow might be helpful so you can setup both linux and windows (i.e. a direct download to the archive): https://github.com/npgsql/npgsql/blob/main/.github/workflows/build.yml

I think the one issue you may have is that your travis workflow is in a diagonal pattern whereas their's is full-fledged.

dvarrazzo:
Merging this, as it does what it says on the tin. Thank you very much :+1:

PR #15, opened 2020-11-26
Windows compatibility for tests

asqui:
* Skip test_connect_timeout() on Windows for the time being as
  Connection.connect() hangs for some reason. This skip can probably
  go away once connection timeouts are implemented.
* test_multiprocess_close(): Turn LocalPath instance back into a str
  before putting it back into env["PYTHONPATH"] to avoid failure in:
  _winapi.CreateProcess(): TypeError: environment can only contain str

### DIFF
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -94,7 +94,7 @@ t.join()
     with (tmpdir / "mptest.py").open("w") as f:
         f.write(module)
     env = dict(os.environ)
-    env["PYTHONPATH"] = tmpdir + os.pathsep + env.get("PYTHONPATH", "")
+    env["PYTHONPATH"] = str(tmpdir + os.pathsep + env.get("PYTHONPATH", ""))
     # TODO: debug this. Importing c module fails on travis in this scenario
     env.pop("PSYCOPG3_IMPL", None)
     out = sp.check_output(
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1,4 +1,5 @@
 import gc
+import sys
 import time
 import socket
 import pytest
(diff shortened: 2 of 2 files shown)


tests/test_connection.py
```diff
@@ -35,9 +36,10 @@ def test_connect_bad():
 
 @pytest.mark.slow
 @pytest.mark.xfail
+@pytest.mark.skipif(sys.platform == "win32", "connect() hangs on Windows.")
```
asqui, line 39: Is it time to introduce a constant like `WINDOWS = sys.platform == "win32"` somewhere?

dvarrazzo: For the moment I don't see a huge reason, it's only used twice and it's pretty self-explanatory. A new const is cognitive load. But as things go on, if we start feeling the need, sure

tests/test_concurrency.py
```diff
@@ -94,7 +94,7 @@ def process():
     with (tmpdir / "mptest.py").open("w") as f:
         f.write(module)
     env = dict(os.environ)
-    env["PYTHONPATH"] = tmpdir + os.pathsep + env.get("PYTHONPATH", "")
+    env["PYTHONPATH"] = str(tmpdir + os.pathsep + env.get("PYTHONPATH", ""))
     # TODO: debug this. Importing c module fails on travis in this scenario
```
dvarrazzo, line 98: Ha ha, the above might be the reason for this failure :D

dvarrazzo: ...nope, it doesn't. https://travis-ci.org/github/psycopg/psycopg3/jobs/746070487

tests/test_connection.py
asqui: This is the third occurrence...
Only reason for suggesting it is I had to go look at one of the previous uses to remind myself of which approach for determining we are on Windows we settled on. :-)
Let's see how it goes. Hopefully not too many more Windows-specific things to worry about 🤞

tests/test_concurrency.py
asqui: Yeah I've always found it a bit of a nuisance that `CalledProcessError` doesn't include stdout and stderr in its string representation. I know it might be quite lengthy, but it's a pain to have to add explicit output of `CalledProcessError.stdout/stderr` to work out what is going on in cases like this.

tests/test_connection.py
dvarrazzo: A `@pytest.mark.skip_win32` is easy to implement :)

PR #32, opened 2021-02-10
row factory proposal

dlax:
Following-up on the discussion on the ML, here's a proposal to add a row_factory argument to `connection.cursor()` allowing to return custom values for rows of the result set.
 
---
We add a row_factory keyword argument in connection.cursor() and cursor
classes that will be used to produce individual rows of the result set.

A RowFactory can be implemented as a class with a `__call__` method
accepting raw values and initialized with a cursor instance; the
RowFactory instance is created when results are available. Type
definitions for RowFactory (and its respective RowMaker) are defined as
callback protocols so as to allow user to define a row factory without
the need for writing a class.

The default row factory returns values unchanged.


dvarrazzo:
Thank you very much, really appreciated!

psycopg3/psycopg3/cursor.py
```diff
@@ -43,13 +44,17 @@
     execute = generators.execute
 
 
+def default_row_factory(cursor: Any) -> RowMaker:
```
dvarrazzo, line 47: Not sure how much it gets in the way with the type system, or how much a micro-optimisation it is, but I would rather avoid a Python call per row in case the client doesn't set a row factory. So maybe I wouldn't have a no-op default row-maker but rather a None and a `return self._make_row(t) if self._make_row else t` sort of thing.

psycopg3/psycopg3/cursor.py
```diff
@@ -364,6 +373,7 @@ def _execute_results(self, results: Sequence["PGresult"]) -> None:
 
         self._results = list(results)
         self.pgresult = results[0]
+        self._make_row = self._row_factory(self)
```
dvarrazzo, line 376: Need a test to verify it works ok and doesn't regress for queries returning multiple results.

psycopg3/psycopg3/cursor.py
```diff
@@ -478,7 +488,7 @@ def executemany(self, query: Query, params_seq: Sequence[Params]) -> None:
 
     def stream(
         self, query: Query, params: Optional[Params] = None
-    ) -> Iterator[Sequence[Any]]:
+    ) -> Iterator[Row]:
```
dvarrazzo, line 491: Nice TypeVar use! :)

psycopg3/psycopg3/cursor.py
```diff
@@ -487,9 +497,10 @@ def stream(
             while self._conn.wait(self._stream_fetchone_gen()):
                 rec = self._tx.load_row(0)
                 assert rec is not None
-                yield rec
+                assert self._make_row is not None
```
dvarrazzo, line 500: This is where I would "if not self._make_row" as commented above

psycopg3/psycopg3/cursor.py
```diff
@@ -516,19 +529,21 @@ def fetchmany(self, size: int = 0) -> Sequence[Sequence[Any]]:
...
         """
         Return all the remaining records from the current recordset.
         """
         self._check_result()
         assert self.pgresult
         records = self._tx.load_rows(self._pos, self.pgresult.ntuples)
         self._pos += self.pgresult.ntuples
-        return records
+        assert self._make_row is not None
+        return [self._make_row(r) for r in records]
```
dvarrazzo, line 544: `list(map(self_make_row, records))` is maybe more efficient (one lookup less per entry), and entirely avoidable with the default make_row.

psycopg3/psycopg3/proto.py
```diff
@@ -115,3 +116,18 @@ def load_sequence(
...
     def get_loader(self, oid: int, format: pq.Format) -> "Loader":
         ...
+
+
+# Row factories
+
+Row = TypeVar("Row")
+
+
+class RowMaker(Protocol):
```
dvarrazzo, line 126: Not sure about these names, but not that I have in mind better ones.

dvarrazzo (changes requested):
Thank you very much for this. I like the idea of having a factory function better than having a class.

A few items I think missing or worth a conversation:

- [ ] `copy ... to stdout` should make use of the same factory
- [x] the adapter makes use of cursors internally (for instance in `TypeInfo.fetch()` and subclasses): let's make sure they still work with customised adapters
- [x] `connection.execute()` might also use connection.row_factory
- [ ] should the `row_factory` attribute be writable?
- [ ] docs needed
- [ ] we should have a `psycopg3.rows` module exposing "obvious" row factories (dicts, namedtuple)

dlax:
@dvarrazzo thanks for the review. One thing before I get back to coding: how would you prefer changes to be made? By adding fixup commits (and then ultimately squashing them all) or by amending the one already there?

dvarrazzo:
Adding further commits to the branch is fine, thank you :) I don't think a final squash will be required either, if each commit is clean enough.

psycopg3/psycopg3/connection.py
```diff
@@ -448,15 +448,20 @@ def close(self) -> None:
         """Close the database connection."""
         self.pgconn.finish()
 
-    def cursor(self, name: str = "", binary: bool = False) -> "Cursor":
+    def cursor(
+        self,
+        name: str = "",
+        binary: bool = False,
+        row_factory: RowFactory = cursor.default_row_factory,
```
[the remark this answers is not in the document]
dlax: I change the default to `None` in commit "Make row factory optional"; I'll add the attribute to connection later on.

psycopg3/psycopg3/cursor.py
dlax: Done in commit "Make row factory optional" I just pushed; and dropped the `default_row_factory()` accordingly. This turns out to be `return self._make_row(t) if self._make_row else cast(Row, t)` in order to pass type checking but maybe there's another solution?

dvarrazzo: Not that I know. I tried dabbling with a factory_function cheating and returning None instead of a no-op callable but got nowhere.

If we cannot make mypy happy, while I'm ok to use `cast()`, probably a `# type: ignore` is good enough.

I was thinking just now that maybe it would be good to add a `row_factory` attribute to the `Transformer`, so that `load_row()` and `load_rows()` may return directly a `Row`? I think it would require to intervene in way less code paths. Or to give the transformer just a `RowMaker`.

dvarrazzo: ...The advantage of giving it a RowMaker is that we would still be able to create it in the cursor with a `self._tx.row_maker = self.row_factory(self)` at the right point.

dlax: I took my a bit of work, but I finally made this idea working. It's indeed cleaner (no change in cursor code, besides type annotations).

psycopg3/psycopg3/cursor.py
dlax: Done in commit 'Add tests for cursor's row_factory with nextset()'.

psycopg3_c/psycopg3_c/_psycopg3.pyi
```diff
@@ -22,6 +22,10 @@ class Transformer(proto.AdaptContext):
     @property
     def adapters(self) -> AdaptersMap: ...
     @property
+    def make_row(self) -> Optional[proto.RowMaker]: ...
+    @make_row.setter
+    def make_row(self, row_maker: proto.RowMaker) -> None: ...
```
dvarrazzo, line 27: It seems the setter might take a None as well?

psycopg3/psycopg3/cursor.py
dvarrazzo: Yes thank you: it does seem cleaner indeed! Now the row_maker is used in only two places and other code paths (e.g. copy) should already be able to make use of it. Looks good!

dvarrazzo:
I have merged to master the server-side cursors branch, merged master into this branch, solved the conflicts, and added a test that verifies the row factory works already with server side cursors too :)

tests/test_cursor_async.py
```diff
@@ -270,10 +270,9 @@ async def test_iter_stop(aconn):
 
 async def test_row_factory(aconn):
     def my_row_factory(cursor):
-        assert cursor.description is not None
-        titles = [c.name for c in cursor.description]
-
         def mkrow(values):
+            assert cursor.description is not None
+            titles = [c.name for c in cursor.description]
```
dvarrazzo: This change invalidates the test: the row_factory is not called again on pgresult switch in nextset() and this change hides it. Fixed in 7d341d0

dvarrazzo:
I merged from master again after making a change that affects your changeset. There is now a code path where the pgresult is updated in the transformer, but the loaders are not updated because it is guaranteed that the new result has exactly the same columns of the previous one. This happens in `stream()` in single-row mode (because every row is a new result from the same query) and on fetch in server-side cursor (because the output columns are fetched right after DECLARE). In these cases there is no need to update the transformer's row maker either (although now that I think about that I've fixed `stream()` but not the server-side cursor... and there is a small correction to do indeed, not depending on this branch)

tests/test_cursor_async.py
dlax: I made this change¹ because I was thinking that the row factory should in fact not be called again on `nextset()` but only the row maker should. In other words, I assumed that the "lifetime" of the row maker (the function returned by the row factory) was that of the cursor, whereas it should be that of a result set, if I understand correctly?

¹: and should have mentioned it, sorry

psycopg3_c/psycopg3_c/_psycopg3.pyi
dlax: That's correct. Fixed in dda10aabd4556a4121f89f8c3a9678fb21bdb1cd.

psycopg3/psycopg3/connection.py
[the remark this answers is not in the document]
dlax: > I think it would be good to have the row factory an attribute of the connection too, which would provide a connection default as a preference of the global default.

Done in f99461c835cb3c2effcd81042b0d98945201cb72.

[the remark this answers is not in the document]
dlax: Hm. The way it's done is not quite right, I'll rework that.

tests/test_cursor_async.py
dvarrazzo: I understand, no worries :) No, a cursor can manage more than one result in its lifetime: either using execute() more than once (which was already handled correctly), or using more than one query and scrolling them with `nextset()`: in these cases results are independent. The other case, the one I fixed in master and merge here, is that _sometimes_ it's guaranteed that the new result object has the same columns of the previous one (in `stream()` there is one result per row, in server-side cursors there is one result per `FETCH`), so I have created a code path where the per-columns-set initialisation is not repeated. I did it to to skip instantiating the `Loader` objects for the data conversion, but it's exactly the same code path we can use to e.g. avoiding creating a new NamedTuple class if the previous one was already good.

psycopg3/psycopg3/connection.py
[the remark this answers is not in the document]
dlax: Fixed in f4e874ca04acec37a37e46b9b6a228a463c0cf95 (replacing f99461c835cb3c2effcd81042b0d98945201cb72), hopefully.

dlax:
Back to the initial "todo" list:
> * `copy ... to stdout` should make use of the same factory

Still TODO.

> * the adapter makes use of cursors internally (for instance in `TypeInfo.fetch()` and subclasses): let's make sure they still work with customised adapters

Done in afd8bffbbd083174db546607b31c3d8879280b02 I think.

> * `connection.execute()` might also use connection.row_factory
> 
> * should the `row_factory` attribute be writable?

Both implemented.

> * docs needed

Haven't start this yet. Do you mean docstrings or docs/tutorial in rst?

> * we should have a `psycopg3.rows` module exposing "obvious" row factories (dicts, namedtuple)

I have started that (see https://github.com/dlax/psycopg3/commit/1f214215092d700d14c90ecbd0a90688348e0ce6, not in this branch); but I'm not sure how to handle namedtuple. Should we reuse psycopg2's approach (I only had a quick look)?

Let me know if there's anything else (or something I missed in other comments).

---

Unfortunately, I won't be available to work further on this next week (I'll be mostly on vacations), but I'll get back to this later on. Feel free to complete if you want to move quicker. Thanks.

dvarrazzo:
- I can deal with COPY myself: I'm more practical with that codebase
- as documentation I mean mostly the `.rst` public docs, examples, etc. I can deal with them too if you prefer.
- for namedtuples, the psycopg2 approach (using a lru cache) was necessary to limit the performance impact (see https://github.com/psycopg/psycopg2/issues/838). You are welcome to merge your branch containing `rows.py` in this one and I can use your `dict_row()` as a blueprint for a `namedtuple_row()`

Thank you very much for your work so far! if you confirm you are not about to add any more flesh to these points I will take over when I have time.

dlax:
I just pushed the commit about `rows.py`. I confirm I'm done now and you can take this over, thanks.

dlax:
I finally got back to this and added the documentation and a namedtuple row factory.

dvarrazzo:
Thank you! Both changesets look good!

I tried to write a small client program to use the factories and checked if mypy plays well with it. I have a few small changes around the objects exposed by the main packaged. Unfortunately it's pretty difficult to get type validation from the client: the cursor should be done parametric on two types (like `BaseCursor[ConnectionType, RowType]`) and it seems both difficult to use and very invasive. So I guess for the moment the types system is good enough, with its limitations.

About COPY support, I think rows cannot be applied there, because the COPY ... TO STDOUT information from the libpq is too limited (we can't have a fully populated `cursor.description`). However COPY TO as rows is a pretty specialised use case, you can use a query most of the time: I think we are good leaving that use case out.

So it seems we are feature complete with this! I'll take a final look at everything and at the rendered docs and I'll merge if you think we have done :) Further adjustments can be made in master or in other branches.

dvarrazzo:
Is there a specific reason to not make `cursor.row_factory` writable?

docs/connection.rst
```diff
@@ -56,11 +64,21 @@ The `!Connection` class
     .. autoattribute:: closed
         :annotation: bool
 
-    .. automethod:: cursor
 
-        Calling the method without a *name* creates a client-side cursor,
-        specifying a *name* crates a server-side cursor. See
-        :ref:`cursor-types` for the details.
+    .. method:: cursor(*, binary: bool = False, row_factory: Optional[RowFactory] = None) -> Cursor
+    .. method:: cursor(name: str, *, binary: bool = False, row_factory: Optional[RowFactory] = None) -> ServerCursor
```
dlax, line 69: I'm not sure about the intent of this change. As is, it seems to me that `conn.cursor(row_factory=None)` does not work anymore as a way to reset the row factory that's been defined in `connect()` or `conn.row_factory`.

For instance, this crashes:
```diff
diff --git a/tests/test_connection.py b/tests/test_connection.py
index 8ab62ee..8290c8b 100644
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -498,7 +498,7 @@ def test_row_factory(dsn):
         cur.execute("select 1, 1, 2")
         assert cur.fetchall() == [{1, 2}]
 
-    with conn.cursor(row_factory=tuple_row) as cur:
+    with conn.cursor(row_factory=None) as cur:
         cur.execute("select 1, 1, 2")
         assert cur.fetchall() == [(1, 1, 2)]
```

psycopg3/psycopg3/proto.py
```diff
@@ -51,14 +51,12 @@
 
 
 class RowMaker(Protocol):
-    def __call__(self, __values: Sequence[Any]) -> Row:
+    def __call__(self, __values: Sequence[Any]) -> Any:
```
dlax, line 54: There was indeed a problem with this this type definition, as evidenced by mypy reporting an invalid argument type in `psycopg3.connect(row_factory=dict_row)` for instance. Not sure we can do better than `Any`, though... Or maybe this is a misusage of `TypeVar`? 

(Just thinking out load now, I keep thinking about that...)

dlax:
> Is there a specific reason to not make `cursor.row_factory` writable?

No. It could be, indeed.

docs/connection.rst
dvarrazzo: Yes, it crashes but just because it uses `my_row_factory`, which assumes values to be string. It fails exactly the same way as:
```diff
@@ -498,7 +498,7 @@ def test_row_factory(dsn):
         cur.execute("select 1, 1, 2")
         assert cur.fetchall() == [{1, 2}]
 
-    with conn.cursor(row_factory=tuple_row) as cur:
+    with conn.cursor() as cur:
         cur.execute("select 1, 1, 2")
         assert cur.fetchall() == [(1, 1, 2)]
```

The sense of that change is to give `None` the meaning of default: in `connect()` it means `tuple_row`, in `conn.cursor()` it means `conn.row_factory`. The value `None` itself is no more valid as `row_factory` attribute, which is no more `Optional`.

So now there is an explicit `tuple_row` which represents the default behaviour. It behaves _like_ `lambda cur: lambda t: t`, but it is actually defined as `lambda cur: None`, because `None` is valid as `Transformer.make_row`. This avoids the need of having a sentinel object and having the `conn.row_factory` non optional makes the paths a bit simpler - no if at every use. As performance is probably influential, calling a function returning a constant at every query is not a heavy thing as it would calling a function on any row (as it would be the case if the default row factory was `lambda cur: lambda t: t`. 

I had to relax the `RowFactory` signature to allow it to return `None`. But now that I am looking at the code we could also make `Transform.make_row` non-nullable and just define:

```python
# row.py

def tuple_factory(cur: Cursor) -> Callable[[Sequence[Any]], Tuple[Any, ...]]:
    return tuple
```

So they python transformer can just use it
```diff
@@ -38,7 +38,7 @@ class Transformer(AdaptContext):
     __module__ = "psycopg3.adapt"
     _adapters: "AdaptersMap"
     _pgresult: Optional["PGresult"] = None
-    make_row: Optional[RowMaker] = None
+    make_row: RowMaker = tuple
 
     def __init__(self, context: Optional[AdaptContext] = None):
 
@@ -174,17 +174,13 @@ class Transformer(AdaptContext):
 
         records: List[Row]
         records = [None] * (row1 - row0)  # type: ignore[list-item]
-        if self.make_row:
-            mkrow = self.make_row
-        else:
-            mkrow = tuple
         for row in range(row0, row1):
             record: List[Any] = [None] * self._nfields
             for col in range(self._nfields):
                 val = res.get_value(row, col)
                 if val is not None:
                     record[col] = self._row_loaders[col](val)
-            records[row - row0] = mkrow(record)
+            records[row - row0] = self.make_row(record)
 
         return records
``` 

and in the Pyrex code, to skip the no-op call, it can be:
```diff
index ce88491..261bde9 100644
--- a/psycopg3_c/psycopg3_c/_psycopg3/transform.pyx
+++ b/psycopg3_c/psycopg3_c/_psycopg3/transform.pyx
@@ -385,7 +385,7 @@ cdef class Transformer:
             Py_INCREF(pyval)
             PyTuple_SET_ITEM(record, col, pyval)
 
-        if self.make_row:
+        if self.make_row is not tuple:
             return self.make_row(record)
         return record
```

What do you think?

psycopg3_c/psycopg3_c/_psycopg3/transform.pyx
```diff
@@ -332,9 +342,11 @@ cdef class Transformer:
                     Py_INCREF(pyval)
                     PyTuple_SET_ITEM(<object>brecord, col, pyval)
 
+        if self.make_row:
+            return list(map(self.make_row, records))
```
dvarrazzo, line 346: Note to self: this doesn't generate good C: refactor

psycopg3_c/psycopg3_c/_psycopg3/transform.pyx
```diff
@@ -82,6 +83,7 @@ cdef class Transformer:
     cdef int _nfields, _ntuples
     cdef list _row_dumpers
     cdef list _row_loaders
+    cdef object _make_row
```
dvarrazzo, line 86: This can be a `cdef public object` and do without the `make_row` property

psycopg3/psycopg3/proto.py
dvarrazzo: I think we are stretching what mypy can do here. I have noticed that if you change `Row` definition it explodes a bit... I wouldn't dare touching it.

I suspect the cursor should be defined as a Generic on the Row too, but so the Connection should, and because the row can be changed at runtime maybe it's just not a thing you can do :)

dvarrazzo:
If you don't think these changes are terrible (the making row_cursor and then making make_row non optional) I can make them later, together with the other notes I took now that I looked better at the Cython code

docs/connection.rst
dlax: > The sense of that change is to give None the meaning of default [...]

I see now, thanks for explaining. It makes sense.

> But now that I am looking at the code we could also make Transform.make_row non-nullable and just define:
> ```python
> # row.py
> def tuple_factory(cur: Cursor) -> Callable[[Sequence[Any]], Tuple[Any, ...]]:
>     return tuple
> ```

This looks better, indeed.

dlax:
> If you don't think these changes are terrible (the making row_cursor and then making make_row non optional) I can make them later, together with the other notes I took now that I looked better at the Cython code

Ok; fine by me. Let me know if anything else is needed here.

docs/connection.rst
dvarrazzo: Done, and also made the attribute writable on the cursor

dvarrazzo:
Merged! Thank you very much @dlax !

dlax:
Excellent, thank you!

PR #33, opened 2021-02-12
add a tox environment to build the docs

dlax:
This adds a tox environment to build the docs.
Also included one change to make the docs build with python 3.7.

### DIFF
--- a/docs/lib/libpq_docs.py
+++ b/docs/lib/libpq_docs.py
@@ -131,7 +131,7 @@ class LibpqReader:
         return self.app.config.libpq_docs_version
 
 
-@lru_cache
+@lru_cache()
 def get_reader():
     return LibpqReader()
 
--- a/tox.ini
+++ b/tox.ini
@@ -17,6 +17,13 @@ commands = mypy
 deps = mypy >= 0.800
 skip_install = true
 
+[testenv:docs]
+commands = sphinx-build -W -T -b html docs docs/_build/html
+deps =
+    ./psycopg3
+    -r docs/requirements.txt
+skip_install = true
+
 [flake8]
 max-line-length = 85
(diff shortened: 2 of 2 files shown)


dvarrazzo:
Do built docs end up in the `docs/_build` dir in psycopg3 or do they stay in some `.tox` dir?

dlax:
I put them in docs/_build as I noticed this was used for that already. 
But I can move them in .tox if you prefer.

dlax:
Hm, perhaps it should rather go to _docs/build/html?

dvarrazzo:
I think the best place is the one where they would go with a normal `make html`, so `docs/_build/html/`, so that then something down the pipeline knows where to find them instead of being inside a Tox hidden path. But as you are making this feature you probably have your use case in mind?

dlax:
> I think the best place is the one where they would go with a normal `make html`, so `docs/_build/html/`, so that then something down the pipeline knows where to find them instead of being inside a Tox hidden path.

Done.

> But as you are making this feature you probably have your use case in mind?

At the moment, it's just to be able to build the docs in one command so 
I can certainly use 'make serve' for that.

In general, as a casual user, I am often happy to find a tox entry point 
to build the documentation locally without bothering with setting up the 
environment myself. Many projects have this.

The tox entry point might also be worth plugging into travis-ci.

dvarrazzo:
> The tox entry point might also be worth plugging into travis-ci.

Yes, that was already up. It broke when I migrated from travis-ci.org to travis-ci.com. Maybe it's only an auth token problem though.

Anyway, happy to merge. Is the branch ready?

dlax:
> Anyway, happy to merge. Is the branch ready?

Yes it is. Thanks!

PR #36, opened 2021-02-28
ci: move to github actions

xrmx:
Started to poke a bit with converting CI to github actions, tests failing because confused travis addons as service when they are local postgresql server from postgres apt repo :) Will try to fix later this week but sharing my work anyway if anyone want to fix that up.


dvarrazzo:
Thank you: it seems a good start!

Version support  can be extended to Python 3.9 and Postgres 13 (and 14 when available). I believe updating it from the docker services will be easier - Travis didn't make it easy, hence the big difference between Postgres 10 and the others.

PR #39, opened 2021-03-05
Fix binary argument of conn.cursor() in tests

dlax:


dvarrazzo:
Thank you, cherry-picked.

I have improved over your commit, in the following  fa2cd24, to enable leak test on both the text and the binary output path, which is what my original code wanted to do.

PR #44, opened 2021-04-07
Re-export names from psycopg3.types

dlax:
This is useful to silent mypy errors when implicit reexport is disabled
(e.g. in strict mode) like:

    error: Module 'psycopg3.types' does not explicitly export attribute
    'Jsonb'; implicit reexport disabled  [attr-defined]

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

Accordingly, we import Int4 from psycopg3.types directly in
test_copy.py.

### DIFF
--- a/psycopg3/psycopg3/types/__init__.py
+++ b/psycopg3/psycopg3/types/__init__.py
@@ -11,115 +11,128 @@ from ..proto import AdaptContext
 from . import array, composite, range
 
 # Wrapper objects
-from ..wrappers.numeric import Int2, Int4, Int8, IntNumeric, Oid
-from .json import Json, Jsonb
-from .range import Range
+from ..wrappers.numeric import (
+    Int2 as Int2,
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Aww, that's so cumbersome to maintain... But I don't see other options here, right?

dlax:
Another option is to write each import as:

from .module import Symbol as Symbol

dvarrazzo:
I was thinking about that. It' still quite the duplication but at least it's local and there is not a long unstructured list at the bottom of the file.

dlax:
Ok; changed to `import X as X`.

Also, there's an open question at https://github.com/python/mypy/issues/10198 for that "issue".

dvarrazzo:
Yes, redundant but I feel better about the more localised duplication. Thank you: will merge!

PR #46, opened 2021-04-15
Pass 'dsn' value to ConnectionPool in tests

dlax:
Otherwise, those tests fail when there's no database named after current
user (typically, 'connection failed: database "denis" does not exist').

### DIFF
--- a/tests/pool/test_pool.py
+++ b/tests/pool/test_pool.py
@@ -146,7 +146,7 @@ def test_configure(dsn):
         with conn.transaction():
             conn.execute("set default_transaction_read_only to on")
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you, good catch.

I have cherry-picked your changeset, so I'll close this (github doesn't recognise it as a merge). However the same problem is also in the async pool: did you test the sync pool only? I have pushed a fix for that test module too.

dlax:
> However the same problem is also in the async pool: did you test the sync pool only?

I guess I just missed or forgot the errors on async pool. I confirm your 
follow-up commit solves the issue there as well, thanks.

PR #47, opened 2021-04-16
Cursor and Connection generic on Row

dlax:
This adds support for having Cursor's fetch*() return type match whatever user-supplied row factory produces. For instance:
```python
class R:
    ...

def my_row_factory(cursor: BaseCursor[Any, R]) -> Callable[[Sequence[Any]], R]:
    ...

with conn.cursor(row_factory=my_row_factory) as cur:
    cur.execute("select 1")
    r = cur.fetchone()
    reveal_type(r)
    # Revealed type is 'Union[R`-1, None]'
```

* the big commits are "Make Cursor generic on Row" and "Make Connection generic on Row"
* other commits are either refactoring to ease the latter one or minor improvements
* an example file `tests/typing_example.py` is added and used as a mypy test in tox and ci


dvarrazzo:
Hey hello, thank you very much for this

It is something I tried to do after I merged your row_factory branch, but I found it tricky indeed, and found the same problem of propagating the type from the connection to the cursor. Good idea dropping that and checking if something less complete but simpler can be done! :+1: 

Will review soon. Cheers!

dlax:
It's not quite right yet because, when `row_factory` is not specified in connection's methods, cursor type is not correctly declared. E.g.:
```python
with conn.cursor() as cur:  # or conn.cursor(row_factory=None)
   ...
```
mypy complains with `Need type annotation for 'cur'  [var-annotated]`.

I suspect we'll need some (more) overload on `cursor()` methods, and possibly `execute()`...

dlax:
About the 'typing_example.py' script that could serve as a regression test for type annotations, it could be plugged into tox as https://github.com/dlax/psycopg3/commit/fe59ad75bff4319c0261cd2d163372db771d3d9f (the script itself needs polishing, though).

dlax:
I'll take a stab at making Connection also generic on Row as I think it might be doable without too much trouble. I'll probably rework some of the commits here accordingly, so changing the status of this PR to "draft". Will report back when it's ready to review.

dlax:
I think I managed to get something suitable and useful. So ready for review, now.

There's still one thing missing: calls to `connect()` (or `*Connection.connect()`) without a `row_factory` argument do not return a typed Connection (but `Connection[Any]`):
```python
reveal_type(connect(row_factory=dict_row))
# Revealed type is 'psycopg3.connection.Connection[builtins.dict*[builtins.str, Any]]'
reveal_type(connect())
# Revealed type is 'psycopg3.connection.Connection[<nothing>]'
```
 One possible solution for this would be to define overloads of `connect()`, but I'm currently stuck there because `@overload` does not work with `**kwargs`. See https://github.com/dlax/psycopg3/commit/d13568bd708ffee48be53e3bff4fb71f3f97b9f4 for a (breaking) attempt.

psycopg3/psycopg3/connection.py
```diff
@@ -498,8 +498,28 @@ def close(self) -> None:
...
+        row_factory: RowFactory[Row],
+    ) -> Cursor[Row]:
+        ...
+
+    @overload
+    def cursor(
+        self,
+        name: str,
+        *,
+        binary: bool = False,
```
dvarrazzo, line 521: Black is being overzelous here. If you drop the comma off the last arguments and ask to reformat it will get these line in a better shape, like:

```python
    @overload
    def cursor(self, *, binary: bool = False) -> Cursor[RowConn]:
        ...

    @overload
    def cursor(
        self, *, binary: bool = False, row_factory: RowFactory[Row]
    ) -> Cursor[Row]:
        ...

    @overload
    def cursor(
        self, name: str, *, binary: bool = False
    ) -> ServerCursor[RowConn]:
        ...

    @overload
    def cursor(
        self, name: str, *, binary: bool = False, row_factory: RowFactory[Row]
    ) -> ServerCursor[Row]:
        ...
```

dvarrazzo:
Thank you very much: will review this soon.

These type annotations are useful for people using psycopg3. As such I assume they will need a certain level of documentation, right? Up to what detail do you think we should provide?

I see `tests/typing_example.py` is called separately in tox. Can it become part of pytest run? Or is there a specific reason for it to be that way?

dlax:
> These type annotations are useful for people using psycopg3. As such I assume they will need a certain level of documentation, right? Up to what detail do you think we should provide?

I haven't thought about this too much yet. It seems to me that we should 
at least extend the "advanced/rows.rst" page in the doc with type 
annotations in all examples.

> I see `tests/typing_example.py` is called separately in tox. Can it become part of pytest run? Or is there a specific reason for it to be that way?

This file is to be checked with mypy (it's not functional at runtime, 
and not really meant to be). So I just added "mypy" commands in tox 
environments, to make sure there's no regression.
To run it with pytest, I suppose we'd need a plugin like 
https://pypi.org/project/pytest-mypy/ or 
https://pypi.org/project/pytest-mypy-plugins/ (I have no experience with 
any of those).

dvarrazzo:
It seems easy enough to run subprocess. How about thiis?
https://github.com/psycopg/psycopg3/commit/19d7baca2030bd0eff6205a2fa1e61f2dd5ec12f

dvarrazzo:
Force-pushed https://github.com/psycopg/psycopg3/commit/3584b72978df1adcc488e066802583b56beed9de to skip the test with Python 3.6

dlax:
Sure, that's fine by me.

dvarrazzo:
> Sure, that's fine by me.

Cool! I'll leave you to cherry-pick the changeset as I see you are pretty active on your branch: I'd rather not push there :)

dlax:
Just pushed a change in commit https://github.com/psycopg/psycopg3/pull/47/commits/412af3419bb19d1615a66dc5e159a9c5393b715d as previous version was breaking the documentation build.

dvarrazzo:
Hi @dlax Is this ready to review now? :)

dlax:
Yes, ready.

dvarrazzo:
Playing a bit as an user, a simple line such as

```python
    cnn = psycopg3.connect(os.environ["DSN"])
```

has mypy complaining of "Need type annotation for 'cnn'". But we know that its' a connection returning tuples. It seems it can be stated by overloading the signature of connect:

```diff
diff --git a/psycopg3/psycopg3/connection.py b/psycopg3/psycopg3/connection.py
index 7c3135f..6a903b9 100644
--- a/psycopg3/psycopg3/connection.py
+++ b/psycopg3/psycopg3/connection.py
@@ -22,7 +22,7 @@ from . import waiting
 from . import encodings
 from .pq import ConnStatus, ExecStatus, TransactionStatus, Format
 from .sql import Composable
-from .rows import tuple_row
+from .rows import tuple_row, TupleRow
 from .proto import AdaptContext, ConnectionType, Params, PQGen, PQGenConn
 from .proto import Query, Row, RowConn, RowFactory, RV
 from .cursor import Cursor, AsyncCursor
@@ -446,7 +446,26 @@ class Connection(BaseConnection[RowConn]):
         super().__init__(pgconn, row_factory)
         self.lock = threading.Lock()
 
+    @overload
     @classmethod
+    def connect(
+        cls,
+        conninfo: str = "",
+        *,
+        autocommit: bool = False,
+        row_factory: RowFactory[RowConn],
+        **kwargs: Any,
+    ) -> "Connection[RowConn]":
+        ...
+
+    @overload
+    @classmethod
+    def connect(
+        cls, conninfo: str = "", *, autocommit: bool = False, **kwargs: Any
+    ) -> "Connection[TupleRow]":
+        ...
+
+    @classmethod  # type: ignore[misc]
     def connect(
         cls,
         conninfo: str = "",
```

With this change, the typing_example can drop a few ignores:

```diff
diff --git a/tests/typing_example.py b/tests/typing_example.py
index fa59a21..f03116a 100644
--- a/tests/typing_example.py
+++ b/tests/typing_example.py
@@ -32,7 +32,7 @@ class Person:
 
 def check_row_factory_cursor() -> None:
     """Type-check connection.cursor(..., row_factory=<MyRowFactory>) case."""
-    conn = connect()  # type: ignore[var-annotated] # Connection[Any]
+    conn = connect()
 
     cur1: Cursor[Any]
     cur1 = conn.cursor()
@@ -81,7 +81,7 @@ def check_row_factory_connection() -> None:
 
     cur3: Cursor[Tuple[Any, ...]]
     r3: Optional[Tuple[Any, ...]]
-    conn3 = connect()  # type: ignore[var-annotated]
+    conn3 = connect()
     cur3 = conn3.execute("select 3")
     with conn3.cursor() as cur3:
         cur3.execute("select 42")
```

Which seems a positive thing, doesn't it?

Without the `type: ignore[misc]` in the `connect()` implementation mypy reports "Overloaded function implementation cannot produce return type of signature 2": I may be doing something wrong but it seems a mypy false positive.

What do you think?

dvarrazzo:
Ah the type produced by `connect(..., row_factory=whatever)` seems `Connection[Any]` though. It can be restricted by specifying a type for the connection, but I think that the default of not raising a mypy complaint is better (and maybe, if mypy people are able to solve the confusion that probably comes from **kwargs, it might improve in the future...)

dvarrazzo:
Found the reason for the required `type: ignore`: I had left `Connection[RowConn]` as return value for the implementation. This doesn't give warnings:
```diff
@@ -446,15 +446,34 @@ class Connection(BaseConnection[RowConn]):
         super().__init__(pgconn, row_factory)
         self.lock = threading.Lock()
 
+    @overload
     @classmethod
     def connect(
         cls,
         conninfo: str = "",
         *,
         autocommit: bool = False,
-        row_factory: Optional[RowFactory[RowConn]] = None,
+        row_factory: RowFactory[RowConn],
         **kwargs: Any,
     ) -> "Connection[RowConn]":
+        ...
+
+    @overload
+    @classmethod
+    def connect(
+        cls, conninfo: str = "", *, autocommit: bool = False, **kwargs: Any
+    ) -> "Connection[TupleRow]":
+        ...
+
+    @classmethod
+    def connect(
+        cls,
+        conninfo: str = "",
+        *,
+        autocommit: bool = False,
+        row_factory: Optional[RowFactory[RowConn]] = None,
+        **kwargs: Any,
+    ) -> "Connection[Any]":
         """
         Connect to a database server and return a new `Connection` instance.
 
```

however it doesn't fix the type:
```python
    tcnn = psycopg3.connect(os.environ["DSN"])
    reveal_type(tcnn)
    dcnn = psycopg3.connect(os.environ["DSN"], row_factory=dict_row)
    reveal_type(dcnn)
```
tcnn is revealed as `Connection[Tuple[Any, ...]]`, but dcnn as `Connection[Any]`.

dlax:
(That's the issue I was mentioning earlier https://github.com/psycopg/psycopg3/pull/47#issuecomment-825732812 :))

The overloads definition should be close to your last proposal, see also https://github.com/dlax/psycopg3/commit/d13568bd708ffee48be53e3bff4fb71f3f97b9f4 without `**kwargs` (which makes everything work, IIRC). But can we drop `**kwargs` or make them explicit? Or maybe have them only in `connect()` function but not in `Connection.connect()` (not even sure this will work)?

dlax:
It seems that if we restrict `kwargs` from being `Any`, and use `Union[str, int]` for instance, types are corrects even though we still get the mypy error about the overload:
```diff
@@ -446,15 +446,38 @@ class Connection(BaseConnection[RowConn]):
         super().__init__(pgconn, row_factory)
         self.lock = threading.Lock()
 
+    @overload
     @classmethod
     def connect(
         cls,
         conninfo: str = "",
         *,
         autocommit: bool = False,
-        row_factory: Optional[RowFactory[RowConn]] = None,
-        **kwargs: Any,
+        **kwargs: Union[str, int],
+    ) -> "Connection[TupleRow]":
+        ...
+
+    @overload
+    @classmethod
+    def connect(
+        cls,
+        conninfo: str = "",
+        *,
+        autocommit: bool = False,
+        row_factory: RowFactory[RowConn],
+        **kwargs: Union[str, int],
     ) -> "Connection[RowConn]":
+        ...
+
+    @classmethod  # type: ignore[misc]
+    def connect(
+        cls,
+        conninfo: str = "",
+        *,
+        autocommit: bool = False,
+        row_factory: Optional[RowFactory[Any]] = None,
+        **kwargs: Union[str, int],
+    ) -> "Connection[Any]":
         """
         Connect to a database server and return a new `Connection` instance.
```
```python
conn = connect()
reveal_type(conn)
# Revealed type is 'psycopg3.connection.Connection[builtins.tuple[Any]]'
tconn = connect(row_factory=rows.tuple_row)
reveal_type(tconn)
# Revealed type is 'psycopg3.connection.Connection[builtins.tuple*[Any]]'
dconn = connect(row_factory=rows.dict_row)
reveal_type(dconn)
# Revealed type is 'psycopg3.connection.Connection[builtins.dict*[builtins.str, Any]]'
```

dvarrazzo:
> It seems that if we restrict kwargs from being Any, and use Union[str, int] for instance, types are corrects even though we still get the mypy error about the overload:

This seems a good solution. I'm adding some reveal_type() tests in pytest to have a full picture of the behaviour

dlax:
https://github.com/dlax/psycopg3/commit/5888eb0cf9c607396b2b014fb473b7ebc2459c5d implements this idea, but there are other issues in pool code (I haven't looked close at these yet).

dvarrazzo:
> [dlax@5888eb0](https://github.com/dlax/psycopg3/commit/5888eb0cf9c607396b2b014fb473b7ebc2459c5d) implements this idea, but there are other issues in pool code (I haven't looked close at these yet).

I've pushed in this branch something very similar to this and add exhaustive tests, which seem to suggest that the types work in a very natural way (`connect()`, `cursor()`, `fetch*()` are all able to declare the type they return). 

Looking into the pool (to check if it can be made generic on the type too) and I would like to see if it's possible somehow to avoid the dependency on `BaseCursor` in the row factory signature (it would be nice if Cursor/AsyncCursor or their union could be used, and BaseCursor to be left out of the interface).

dlax:
Agreed about "hiding" BaseCursor. See https://github.com/dlax/psycopg3/commit/db2cc379101b44c7fea083d8f9ab3499035e4f90 for a possible solution.

dlax:
> Agreed about "hiding" BaseCursor. See dlax@db2cc37 for a possible solution.

Added that commit here, let me know if it's suitable.

dvarrazzo:
I tried to define AnyCursor[Row] as an Union[Cursor, AsyncCursor] but the base class wasn't amused, so I think it's ok this way.

I have added some documentation, including documenting the protocols.

dlax:
> I have added some documentation, including documenting the protocols.

Nice!

dvarrazzo:
I think further docs improvements can be made to the main branch. I'm happy about the state of this branch at the moment. Shall we merge, @dlax ?

dlax:
Yes!

PR #48, opened 2021-04-27
Check documentation build in CI

dlax:
This is useful to prevent regressions during pull requests as the docs build uses `-W` flag of sphinx.


dlax:
It's currently run as part of the "lint" workflow, but it could be elsewhere if better.

dvarrazzo:
No, as a linting step sounds good, thank you!

PR #49, opened 2021-04-29
Add an adaptation example for infinity date

dlax:
Reporting the example from https://www.psycopg.org/articles/2020/11/24/psycopg3-adaptation/ in docs, thus resolving a todo.

### DIFF
--- a/docs/advanced/adapt.rst
+++ b/docs/advanced/adapt.rst
@@ -57,9 +57,52 @@ configuration.
     subclass `!Dumper` or `!Loader` you should call the ``.register()`` on the
     class you created.
 
+For example, suppose you want to work with the "infinity" date which is
+available in PostgreSQL but not handled by Python:
+
+.. code:: python
+
+    >>> conn.execute("'infinity'::date").fetchone()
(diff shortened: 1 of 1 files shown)


dvarrazzo:
This makes me think about a problem. Soon I will add binary adapters for date/time types. If the adaptation system stays as it is, when a binary adapter is available, it is used in precedence of a text adapter. The rationale is that usually the binary adapter is more efficient.

There are an exception or two to this rule, which at the moment are managed by [an ugly if in the code](https://github.com/psycopg/psycopg3/blob/master/psycopg3/psycopg3/adapt.py#L222-L237). Strings are adapted to text mode even and with unknown oid, otherwise Postgres will fail to cast.

Probably there should rather be an explicit set of dumpers for the "auto" format (a `self._dumpers[pq.Format.AUTO]` in the code above) instead of being "binary with exceptions".

dvarrazzo:
Cherry-picked the other commit too after solving the adaptation problem mentioned (see 9d49bf8). Also added a test to make sure that when there will be a binary adapter, even if preferred, the customisation will still work.

PR #50, opened 2021-05-05
Upgrade testing libraries

dlax:
I was trying to run with python 3.10, so needed pytest 6.2.4. (There another issue after that.)
It seems that nothing breaks with more recent versions of all of these.

### DIFF
--- a/psycopg3/setup.py
+++ b/psycopg3/setup.py
@@ -35,9 +35,9 @@ extras_require = {
         "mypy >= 0.812",
         "pproxy >= 2.7, < 2.8",
         "pytest >= 6.2.4, < 6.3",
-        "pytest-asyncio >= 0.14.0, < 0.15",
-        "pytest-randomly >= 3.5, < 3.6",
-        "tenacity >= 6.3, < 6.4",
+        "pytest-asyncio >= 0.15.0, < 0.16",
+        "pytest-randomly >= 3.7, < 3.8",
+        "tenacity >= 7, < 7.1",
     ],
     "dev": [
         "black",
--- a/psycopg3/setup.py
+++ b/psycopg3/setup.py
@@ -34,7 +34,7 @@ extras_require = {
     "test": [
         "mypy >= 0.812",
         "pproxy >= 2.7, < 2.8",
-        "pytest >= 6, < 6.1",
+        "pytest >= 6.2.4, < 6.3",
         "pytest-asyncio >= 0.14.0, < 0.15",
         "pytest-randomly >= 3.5, < 3.6",
         "tenacity >= 6.3, < 6.4",


dvarrazzo:
If you _need_ those versions you should peg the lower limit too to what is needed (e.g. pytest >= 6.2.4, < 6.3; tenacity >= 7, < 7.1 etc) :)

dvarrazzo:
I understand you are adding upcoming versions to the testing grid. Are you also able to add Postgres 14? There is a bunch of changes on the way in that version (infinity numeric, multi-interval range, pipeline mode...) that might be useful to preview.

Just yesterday I have committed (5cb1a6e59b192d82a38e13864dccb8a3248054b5) a pytest marker in the (binary-numeric)[https://github.com/psycopg/psycopg3/commits/binary-numeric) branch to write markers such as `@pytest.mark.pg(">= 14")` to write selective tests. We can merge it to master if it's useful.

dlax:
> If you need those versions you should peg the lower limit too to what is needed

Not sure we need any of those now; pytest >= 6.2.4 will certainly be needed for python 3.10 (along with, at least pytest-randomly). I was rather suggesting to increase the upper bounds when there does not seem to have incompatibilities. But that's certainly not important now.

> Are you also able to add Postgres 14?

No, I haven't tried that.

dvarrazzo:
The low limit is "what we have tested with, and works". The high limit is there to avoid a regression if there is a new release which happens to be non-backward compatible. So, if we have a constrain like ">=6.4, <6.5", but we happen to test with 6.5.2 and it works well, then the allowed range can be bumped up no problem, to ">=6.5, <6.6".

In your case, as you have verified that 6.2.4 is needed for Python 3.10 and that it works with the other versions, the range can be bumped to >=6.2.4, <6.3 no problem.

Does it make sense? :)

dlax:
Yes, makes sense; and I agree there's no problem on our side to bump de lower bound.

However, there's another point of view, typically that of a distro packager, where the lower bound would appear unnecessarily strict. For instance, Debian bullseye (next-stable) will ship python 3.9 and pytest 6.0.2.

Then, there is semver... where it seems to me we should use "pytest >= 6, < 7".

I don't mean we should favor one point of view over the current one, rather wanted to understand the current "policy" :)

So, shall I bump?

dvarrazzo:
I would rather not be pegged to what distros package, especially when it comes to dev packages.

Using the minor release number instead of the major is more a way to make sure to not start using "accidentally" new features on the depending packages. The scenario is that, if you have >=6, <7, and you install on one site 6.1, then later on your laptop you install 6.3, and you start using a feature available in 6.2 (you can add a new feature in a semver minor release), then the site with 6.1 gets broken even if it respects the range.

Of course this is sort of fuzzy and not really robust. For production installations I've started using Poetry to enforce things more robustly (not really happy about it, still cumbersome and immature, but it's better than nothing). But that's the principle why I prefer to peg a range on a minor release rather than the major one (and Poetry is not a tool to use on libraries).

dlax:
Ok, all ranges are now narrow.

PR #51, opened 2021-05-05
Add support for connection timeout

dlax:


psycopg3/psycopg3/waiting.py
```diff
@@ -71,23 +71,24 @@ def wait_conn(gen: PQGenConn[RV], timeout: Optional[float] = None) -> RV:
...
         while 1:
             sel.register(fileno, s)
-            ready = None
-            while not ready:
-                ready = sel.select(timeout=timeout)
+            ready = sel.select(timeout=timeout)
             sel.unregister(fileno)
-            fileno, s = gen.send(ready[0][1])
+            if not ready:
+                raise e.DatabaseError("timeout expired")
```
dvarrazzo, line 90: Probably `OperationalError` is more precise

dvarrazzo:
There is a failure in the test, but it seems a fluke: I have repeated the test in the same condition (`pytest --randomly-seed=261706959` in Python 3.7, C implementation) and doesn't fail. Will rerun.

dlax:
Just rebased on today's master.

dvarrazzo:
Hi @dlax, thank you! I've just gone back on the project after a couple of weeks of other works. Will review this soon!

psycopg3/psycopg3/waiting.py
```diff
@@ -71,23 +71,24 @@ def wait_conn(gen: PQGenConn[RV], timeout: Optional[float] = None) -> RV:
...
             sel.register(fileno, s)
-            ready = None
-            while not ready:
-                ready = sel.select(timeout=timeout)
+            ready = sel.select(timeout=timeout)
             sel.unregister(fileno)
-            fileno, s = gen.send(ready[0][1])
+            if not ready:
+                raise e.OperationalError("timeout expired")
+            fileno, s = gen.send(ready[0][1])  # type: ignore[arg-type]
```
dvarrazzo, line 91: Curious that this ignore is required. Or better, curious it wasn't required already. It seems a mypy bug: adding `reveal_type(ready)` after the innermost while on the master branch doesn't report anything.

dlax: Agreed, the ignore should have been required already before.
We can get rid of it by changing to `fileno, s = gen.send(Ready(ready[0][1]))`, is it worth?

dvarrazzo: Probably worth. I'm adding a small changeset just because I don't like these var names but I will leave the ignore around.

dvarrazzo:
Rebased and merged, thank you!

PR #60, opened 2021-06-10
minor changes/cleanups to the 'sql' module

dlax:
I was mostly looking at the TODO of `Composable.as_bytes()` and started changing things here and there.


dvarrazzo:
Merged, thank you

PR #61, opened 2021-06-20
Add macOS support

c0nn3r:
This pull request has the two changes I had to make to get psycopg3 running on macOS.

1. First, `<endian.h>` is `<machine/endian.h>` on macOS, but sadly, even with this change, the crippled Mac version lacks `htobe16` functions etc. The workaround is using some functions from `libkern` to define these functions.
2. Second, `ctypes.util.find_library` needs to search with a `.dylib` extension.

Hope this is useful and thank you for this great library!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -95,8 +95,6 @@ jobs:
           --health-retries 5
 
     steps:
-      - name: print username
-        run: whoami
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
@@ -126,7 +124,7 @@ jobs:
     env:
       PSYCOPG3_IMPL: python
       PGHOST: 127.0.0.1
-      PGUSER: postgres
+      PGUSER: runner 
       PSYCOPG3_TEST_DSN: "dbname=postgres"
 
     steps:
@@ -163,7 +161,7 @@ jobs:
     env:
       PSYCOPG3_IMPL: c
       PGHOST: 127.0.0.1
-      PGUSER: postgres
+      PGUSER: runner 
       PSYCOPG3_TEST_DSN: "dbname=postgres"
       # skip tests failing on importing psycopg3_c.pq on subprocess
       # they only fail on Travis, work ok locally under tox too.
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Hello, thank you very much, this is really appreciated.

Could you please add MacOS to the test grid? https://github.com/psycopg/psycopg3/blob/master/.github/workflows/tests.yml

https://github.com/psycopg/psycopg3/blob/master/.github/workflows/tests.yml

For an example you can take a look at the psycopg2 workflow: the packages workflow has a MacOS stanza.

Cheers!

c0nn3r:
I haven't given up don't worry 😛 , but it looks like I still have my work cut out for me. I think part of the reason there is a discrepancy is that I am using conda as my package manager which installs `libpg` itself. I'm not sure why so many Python implementation tests are failing, somewhat worried about that.

c0nn3r:
I think the main reason the tests are failing is that I cannot setup a health check like the docker container has on Linux. I will try to think on how to emulate this for Mac, or perhaps we should disable those tests for Mac for now?

.github/workflows/tests.yml
```diff
@@ -103,3 +103,80 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Install postgres
+        run: brew install postgresql@${{ matrix.postgres }}
+      - name: Start postgres
+        run: brew services start postgresql
+      - name: Install tox
+        run: pip install tox
+      - name: Run tests
+        run: tox -c psycopg3 -e ${{ matrix.python }} -- --color yes
```
dvarrazzo, line 182: Please configure your editor to terminate all the files with a newline

dvarrazzo:
Very strange that pretty much all the sched tests fail. Why there is so much delay on MacOS? :thinking:

dvarrazzo:
Master now has MacOS support (test and packages build). I have copied some of the things done here, but the codebase had diverged because of the psycopg3 -> psycopg renaming and because of Windows compatibility too.

Thank you very much!

c0nn3r:
Thanks for getting that done! I've been busy and appreciate the help.

PR #62, opened 2021-06-21
add pypy3.7 CI run

mattip:
Putting this out there as a test to see if this works.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
           - {impl: python, python: 3.7, postgres: 11}
           - {impl: python, python: 3.8, postgres: 12}
           - {impl: python, python: 3.9, postgres: 13}
-          - {impl: python, python: pypy-3.7, postgres: 11}
+          - {impl: python, python: pypy-3.7, postgres: 13}
           - {impl: c, python: 3.6, postgres: 13}
           - {impl: c, python: 3.7, postgres: 12}
           - {impl: c, python: 3.8, postgres: 11}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,7 @@ jobs:
           - {impl: python, python: 3.7, postgres: 11}
           - {impl: python, python: 3.8, postgres: 12}
           - {impl: python, python: 3.9, postgres: 13}
+          - {impl: python, python: pypy-3.7, postgres: 11}
           - {impl: c, python: 3.6, postgres: 13}
           - {impl: c, python: 3.7, postgres: 12}
           - {impl: c, python: 3.8, postgres: 11}
--- a/psycopg/setup.py
+++ b/psycopg/setup.py
@@ -37,7 +37,7 @@ extras_require = {
     ],
     # Requirements to run the test suite
     "test": [
-        "mypy >= 0.812; implementation_name !='pypy'",
+        "mypy >= 0.812; implementation_name != 'pypy'",
         "pproxy >= 2.7, < 2.8",
         "pytest >= 6.2.4, < 6.3",
         "pytest-asyncio >= 0.15.0, < 0.16",
--- a/psycopg/setup.py
+++ b/psycopg/setup.py
@@ -37,7 +37,7 @@ extras_require = {
     ],
     # Requirements to run the test suite
     "test": [
-        "mypy >= 0.812",
+        "mypy >= 0.812 ; implementation_name!='pypy'",
         "pproxy >= 2.7, < 2.8",
         "pytest >= 6.2.4, < 6.3",
         "pytest-asyncio >= 0.15.0, < 0.16",
@@ -48,7 +48,7 @@ extras_require = {
     "dev": [
         "black",
         "flake8 >= 3.8, < 3.9",
-        "mypy >= 0.812",
+        "mypy >= 0.812; implementation_name != 'pypy'",
         "wheel",
     ],
     # Requirements needed to build the documentation
--- a/psycopg/setup.py
+++ b/psycopg/setup.py
@@ -37,7 +37,7 @@ extras_require = {
     ],
     # Requirements to run the test suite
     "test": [
-        "mypy >= 0.812 ; implementation_name!='pypy'",
+        "mypy >= 0.812; implementation_name !='pypy'",
         "pproxy >= 2.7, < 2.8",
         "pytest >= 6.2.4, < 6.3",
         "pytest-asyncio >= 0.15.0, < 0.16",
--- a/tests/pool/test_pool.py
+++ b/tests/pool/test_pool.py
@@ -1,3 +1,4 @@
+import gc
 import sys
 import logging
 import weakref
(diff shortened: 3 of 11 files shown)


mattip:
Nope, it [failed on my fork](https://github.com/mattip/psycopg3/runs/2879277582?check_suite_focus=true#step:6:76) since typed-ast has declared [they can not support PyPy](https://github.com/python/typed_ast/issues/111#issuecomment-493279393). Is there a way to work around that, by disabling mypy checks for this CI run?

mattip:
I will ping here when I get this working on my fork

tests/pool/test_pool.py
```diff
@@ -12,6 +13,11 @@
 from psycopg3.pq import TransactionStatus
 
 
+def gc_collect():
```
dvarrazzo, line 16: This seems useful, as the gc tests seem somewhat flaky on cpython too.

dvarrazzo: I have incorporated a similar changeset in master.

thedrow:
@mattip Could you rebase this PR?

psycopg3/setup.py
```diff
@@ -32,7 +32,7 @@
         f"psycopg3-binary == {version}",
     ],
     "test": [
-        "mypy >= 0.812",
+        "mypy >= 0.812 ; implementation_name!='pypy'",
```
dvarrazzo, line 35: Please use consistent spacing around the operator

.github/workflows/tests.yml
```diff
@@ -17,6 +17,8 @@ jobs:
             postgres: 10
           - python: 3.7
             postgres: 11
+          - python: pypy-3.7
+            postgres: 11
```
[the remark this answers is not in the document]
mattip: changed to 13

psycopg3/setup.py
mattip: removed space before ';'

dvarrazzo (changes requested):

tests/test_copy.py
```diff
@@ -583,9 +589,14 @@ def work():
                 gc_collect()
                 n.append(len(gc.get_objects()))
 
-            assert (
-                n[0] == n[1] == n[2]
-            ), f"objects leaked: {n[1] - n[0]}, {n[2] - n[1]}"
+            if sys.implementation.name == "pypy":
+                assert (
+                    n[0] >= n[1] >= n[2]
+                ), f"objects leaked: {n[1] - n[0]}, {n[2] - n[1]}"
```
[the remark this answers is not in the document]
mattip: Sorry, I think this came from a bad merge. fixing.

psycopg3/setup.py
dvarrazzo: I meant the spaces around the operator, as you used in the other line :)

mattip:
Workflow is awaiting approval to run since I am a new contributor

tests/test_cursor.py
```diff
@@ -13,6 +14,11 @@
 from .utils import gc_collect
 
 
+def gc_collect():
```
dvarrazzo, line 17: Note that now there is a definition of `gc_collect()` in `.utils`, see line above

dvarrazzo:
@mattip 

> Workflow is awaiting approval to run since I am a new contributor

I have poked a bit at the settings, you should be able to run workflows without approval needed now

mattip:
Three failures:
```
FAILED tests/test_connection.py::test_connection_warn_close - assert not Warn...
FAILED tests/test_connection_async.py::test_connection_warn_close - Assertion...
FAILED tests/test_server_cursor_async.py::test_warn_close - psycopg.errors.In...
```

The first warns when it shouldn't, the second does not warn when it should (`gc_collect` might help) and I don't know what is going on in the third one.

dvarrazzo:
I think those tests about resources warnings should be just skipped on pypi: they assume the behaviour of the cpython garbage collector.

In order to see a clearer error message for the third, please rebase on master to include https://github.com/psycopg/psycopg/commit/c8e895281151a95da607541515e07e164f5fac7b

tests/pool/test_pool.py
```diff
@@ -12,6 +13,11 @@
 from psycopg.pq import TransactionStatus
 
 
+def gc_collect():
+    for i in range(3):
+        gc.collect()
```
dvarrazzo, line 18: This can be imported from the `.utils` module too

psycopg/setup.py
```diff
@@ -37,7 +37,7 @@
     ],
     # Requirements to run the test suite
     "test": [
-        "mypy >= 0.812",
+        "mypy >= 0.812; implementation_name !='pypy'",
```
dvarrazzo, line 40: Still not formatted ok :)

dvarrazzo (changes requested):

psycopg/setup.py
mattip: do I need a space before the semicolon?

dvarrazzo:
```suggestion
        "mypy >= 0.812; implementation_name != 'pypy'",
```

tests/test_connection.py
```diff
@@ -89,19 +89,25 @@ def test_broken(conn):
     assert conn.broken
 
 
+@pytest.mark.skipif(
```
dvarrazzo, line 92: If you are skipping this test on pypy, please drop the `gc_collect()` you added

dvarrazzo (changes requested):

tests/pool/test_pool.py
```diff
@@ -11,6 +11,8 @@
 import psycopg_pool as pool
 from psycopg.pq import TransactionStatus
 
+from .utils import gc_collect
```
dvarrazzo, line 14:
```suggestion
from ..utils import gc_collect
```

mattip:
I don't understand the `psycopg.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block` error. Is that supposed to raise differently or not happen at all?

dvarrazzo:
> I don't understand the psycopg.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block error. Is that supposed to raise differently or not happen at all?

Ah, it's still in `tests/test_server_cursor.py:99`? That' strange. I assumed the test could have been put in bad state by an assert but probably it's not the case.

You can try and drop the retry loop to see what error happen. If you find and fix it, probably the transaction_status check I added in c8e8952 is not needed and can be removed.

om26er:
I would like to help with this if there is a chance we can revive this effort. We currently use `asyncpg` with SQLAlchemy but PyPy support is a hard requirement for us and there is little hope in `asyncpg` getting that. OTH, psycopg seems to almost work https://github.com/psycopg/psycopg/issues/391#issuecomment-1368471790 -- My first goal is to get psycopg working with ctypes on PyPy and then as a next step, I would like to explore cffi support.

mattip:
Feel free to start over from the beginning with addding pypy to the CI. Note that the PyPy team is no longer developing pypy3.7, and pypy3.8 is also nearing EOL, so it would be best to do this with pypy3.9.

ngnpope:
Superseded by #686.

PR #80, opened 2021-09-10
WIP: Add PostGIS Geometry adapter based on Shapely

jacopofar:
As mentioned in https://github.com/rustprooflabs/pgosm-flex/pull/165 I tried to write an adapter for PostGIS geometries based on Shapely.

The usage looks like this:
<details>
<summary>Example usage</summary>

```python3

import psycopg


from psycopg.types.geometry import register_shapely_adapters

CONN_STR = "postgres://postgres:testpassword@localhost:15432/osm_data"

# get the Reichstag building geometry from a pgosm-flex import of Berlin
# it's a multiploygon with holes in it
# https://www.openstreetmap.org/relation/2201742
READ_QUERY = """
    select osm_id, name, st_area(geom), geom
    from osm.building_polygon
    where osm_id = -2201742"""

with psycopg.connect(CONN_STR) as conn:
    register_shapely_adapters(conn)

    with conn.cursor(binary=False) as cur:
        cur.execute(READ_QUERY)
        row = cur.fetchone()
        print('Binary protocol:')
        print(row)
        print('area from shape:', row[-1].area)
        print('area from DB:', row[-2])


    with conn.cursor(binary=True) as cur:
        cur.execute(READ_QUERY)
        row = cur.fetchone()
        print('Text protocol:')
        print(row)
        print('area from shape:', row[-1].area)
        print('area from DB:', row[-2])
        from shapely import affinity
        cur.execute("insert into osm.building_polygon(osm_id, osm_type, address, geom) VALUES(999999, 'fake', 'fake', %s)", (affinity.rotate(row[-1], 30),))
        conn.commit()


with psycopg.connect(CONN_STR) as conn:
    # this connection has no adapters
    with conn.cursor(binary=False) as cur:
        cur.execute(READ_QUERY)
        row = cur.fetchone()
        print(str(row)[:400] + '...(truncated)')
```

</details>
it works both in binary and text mode, and I can see the rotated shape in QGIS

<details>
<summary>Screenshot of the polygons in QGIS</summary>

a map showing the polygon and the rotated one inserted using the adapter

</details>

There are a few things missing:

* the geometry is a type from postGIS, so I think it can only be retrieved at runtime
* how to write a test foir this? I ran it against a test PostGIS DB
* I'm not sure about the import for shapely, now it's simply attempted and there's no error handling on it
* documentation, once the usage is defined


psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,63 @@
...
+    After invoking this function on a connection, the queries retrieving
+    PostGIS geometry objects will return Shapely's shape object instances
+    in text and binary mode.
+
+    Similarly, shape onjects can be sent to the database.
+
+    :param conn: the connection on which to set up the loader and dumper
+
+    """
+    import shapely.wkb
```
dvarrazzo, line 28: Better warn the user if they need an extra package: see
https://github.com/psycopg/psycopg/blob/2c991429cc28ca50b5ad292f5e42ad5a85b4cf46/psycopg/psycopg/_dns.py#L21-L24
for example.

I wonder if we shouldn't add an extra, so that people may install `psycopg[shapely]` to get a shapely of the right version bracket. Or is this best left to the user?

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,63 @@
...
+    from shapely.geometry.base import BaseGeometry
+
+    class GeometryBinaryLoader(Loader):
```
dvarrazzo, line 31: You should define these functions top level in the module (you can make `shapely` a global var so that you can lazy import it and still use it  from these classes.

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,63 @@
...
+            return shapely.wkb.loads(data)
+
+    class GeometryTextLoader(Loader):
+        format = Format.TEXT
+
+        def load(self, data: bytes) -> BaseGeometry:
+            # it's a hex string in binary
+            return shapely.wkb.loads(data.decode(), hex=True)
+
+    class GeometryBinaryDumper(Dumper):
```
dvarrazzo, line 44: Once we know the `geometry` oid, we should create subclasses which set the `oid` class attribute of the dumpers. You can use `type()` to dynamically create a class. For example:
https://github.com/psycopg/psycopg/blob/2c991429cc28ca50b5ad292f5e42ad5a85b4cf46/psycopg/psycopg/types/hstore.py#L117-L120

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,63 @@
...
+
+        def dump(self, obj: BaseGeometry) -> bytes:
+            return shapely.wkb.dumps(obj, hex=True).encode()  # type: ignore
+
+    t = TypeInfo.fetch(conn, "geometry")
+    if t is None:
+        raise ValueError("The database has no geometry type. Is it PostGIS?")
+    t.register(conn)
+    conn.adapters.register_loader("geometry", GeometryBinaryLoader)
+    conn.adapters.register_loader("geometry", GeometryTextLoader)
```
dvarrazzo, line 61: For consistency with other adapter, please call the text ones `GeometryLoader` and `GeometryDumper`

dvarrazzo (changes requested):
Testing and documentation are needed.

For testing:
- do we have to add the PostGIS package to postgres?
- one important criteria is that psycopg imports ok without shapely installed, so the basic test suite should skip the shapely tests
  - however, I'd like to test it too (I have the same problem with the `dns` module, depending on a module I don't want to depend on, so the test are skipped
  - I think we should add a new tox target which would install shapely in its environment and run the shapely tests only (we can add a pytest tag for that)

dvarrazzo:
Please take a look at https://github.com/psycopg/psycopg/commit/ef9cb2b718140e787e7257de1df5a9a2251ad610 as an example about isolating tests requiring an optional dependency.

dvarrazzo:
Please call the module `psycopg.types.shapely`, not `geometry`. There might come out other ways to adapt geometry types to different Python objects.

psycopg/psycopg/types/geometry.py
jacopofar: I think it's best left to the user, installing it requires GEOS and usually is done with extra libraries to handle projections. Also there are different versions actively used, 1.7, the beta for 1.8 and the future 2.0 (I tested using 1.7 and 1.8 beta)

jacopofar:
I applied the suggested changes, and wrote a doc (I never used RST and was unable to serve the docs locally, am not totally sure about the syntax), will look at the tests later.

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,63 @@
...
+from psycopg.adapt import Dumper, Loader
+from psycopg.pq import Format
+from psycopg.types import TypeInfo
+
+if TYPE_CHECKING:
+    from psycopg.connection import Connection
+    from psycopg.rows import Row
+
+
+def register_shapely_adapters(conn: Connection[Row]) -> None:
```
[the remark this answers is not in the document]
jacopofar: Nice, done

dvarrazzo:
> I applied the suggested changes, and wrote a doc (I never used RST and was unable to serve the docs locally, am not totally sure about the syntax), will look at the tests later.

I can help you with the docs, no worries.

If you would like to give it a shot you can run `pip install -e ./psycopg[docs]` to install the dependencies in the current virtualenv and `make -C docs serve` to run a web server serving the docs locally.

Thank you very much :)

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,86 @@
...
+    def load(self, data: bytes) -> "BaseGeometry":
+        return wkb.loads(data)
+
+
+class GeometryLoader(Loader):
+    format = Format.TEXT
+
+    def load(self, data: bytes) -> "BaseGeometry":
+        # it's a hex string in binary
+        return wkb.loads(data.decode(), hex=True)
```
dvarrazzo, line 37: I didn't notice that `bytes.decode()` and `str.encode()` default to utf8 in Python 3! I was still of the Python 2 mindset that it would have tried an unpredictable random encooding. Not passing a parameter is also slightly faster, so I guess I'll drop the `"utf8"` param from the rest of the project :yum:

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,86 @@
...
+        "GeometryBinaryDumper", (GeometryBinaryDumper,), {"oid": info.oid}
+    )
+    dumper: Type[GeometryDumper] = type(
+        "GeometryDumper", (GeometryDumper,), {"oid": info.oid}
+    )
+
+    adapters.register_loader(info.oid, GeometryBinaryLoader)
+    adapters.register_loader(info.oid, GeometryLoader)
+    adapters.register_dumper(BaseGeometry, binary_dumper)
+    adapters.register_dumper(BaseGeometry, dumper)
```
dvarrazzo, line 86: Registering dumpers on the same type, the last one becomes the default dumper. The way you registered them here make the text dumper the default one. I assume you would prefer the binary one to be the default, right? If that's the case you may want to swap these two `register_dumper()`

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,86 @@
...
+    def load(self, data: bytes) -> "BaseGeometry":
+        # it's a hex string in binary
+        return wkb.loads(data.decode(), hex=True)
+
+
+class GeometryBinaryDumper(Dumper):
+    format = Format.BINARY
+
+    def dump(self, obj: "BaseGeometry") -> bytes:
+        return wkb.dumps(obj).encode()  # type: ignore
```
dvarrazzo, line 44: I don't think you want `encode()` here. AFAICS, wkb.dumps() dumps already bytes.
```python
>>> wkb.dumps(point)
b'\x01\x01\x00\x00\x00333333\xf3?333333\x0b@'
```

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,86 @@
...
+        "The module psycopg.types.geometry requires the package 'Shapely'"
+        " to be installed"
+    )
+
+
+class GeometryBinaryLoader(Loader):
+    format = Format.BINARY
+
+    def load(self, data: bytes) -> "BaseGeometry":
+        return wkb.loads(data)
```
dvarrazzo, line 29: I've made a few manual tests and this doesn't seem to work.

The input data should be `psycopg.abc.Buffer`, which is `Union[bytes,bytearray,memoryview]`. Most often, the parameters from a query are passed as `memoryview`. It doesn't seem that wkb.loads accepts bytearray and memoryview, so you should convert this to bytes:
```diff
--- a/psycopg/psycopg/types/geometry.py
+++ b/psycopg/psycopg/types/geometry.py
@@ -5,7 +5,7 @@ Adapters for PostGIS geometries
 from typing import Optional, Type
 
 from .. import postgres
-from ..abc import AdaptContext
+from ..abc import AdaptContext, Buffer
 from ..adapt import Dumper, Loader
 from ..pq import Format
 from .._typeinfo import TypeInfo
@@ -25,14 +25,18 @@ except ImportError:
 class GeometryBinaryLoader(Loader):
     format = Format.BINARY
 
-    def load(self, data: bytes) -> "BaseGeometry":
+    def load(self, data: Buffer) -> "BaseGeometry":
+        if not isinstance(data, bytes):
+            data = bytes(data)
         return wkb.loads(data)
```

psycopg/psycopg/types/geometry.py
dvarrazzo, line 37: Likewise, if a memoryview ends in this function (as it normally does), it will not have a decode() method available.

```diff
 class GeometryLoader(Loader):
     format = Format.TEXT
 
-    def load(self, data: bytes) -> "BaseGeometry":
+    def load(self, data: Buffer) -> "BaseGeometry":
+        if isinstance(data, memoryview):
+            data = bytes(data)
         # it's a hex string in binary
         return wkb.loads(data.decode(), hex=True)
```

dvarrazzo (changes requested):

dvarrazzo:
I have pushed a changeset in this branch to fix documentation, both the reST syntax and adding shapely as a doc dependency so that docs introspection works

psycopg/psycopg/types/geometry.py
```diff
@@ -0,0 +1,86 @@
...
+
+from .. import postgres
+from ..abc import AdaptContext
+from ..adapt import Dumper, Loader
+from ..pq import Format
+from .._typeinfo import TypeInfo
+
+
+try:
+    import shapely.wkb as wkb
```
dvarrazzo, line 15: Micro optimization:
```suggestion
    from shapely.wkb import loads, dumps
```
and then use directly the functions, without looking them up from the module.

dvarrazzo (changes requested):

psycopg/psycopg/types/geometry.py
jacopofar: Isn't it also stored based on the format ([here](https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/_adapters_map.py#L142)) ? I'm not sure what that `type` wizardry does here exactly but I expect the `format` attribute to be defined by the two different base classes so that the corresponding one is used when a cursor is created with `binary=True` or not

psycopg/psycopg/types/geometry.py
jacopofar: Right, wkb means literally _Well Known Binary_ :-)
Checking this I found out that Shapely already returns str or bytes depending on the `hex` parameter

jacopofar:
> I can help you with the docs, no worries.
> 
> If you would like to give it a shot you can run `pip install -e ./psycopg[docs]` to install the dependencies in the current virtualenv and `make -C docs serve` to run a web server serving the docs locally.
> 
> Thank you very much :)

Great, I'm very curious about trying RST and will give it a second try when I have time

psycopg/psycopg/types/geometry.py
dvarrazzo: `type()` is just a way to create a class using a function instead of using the `class` syntax. In your case tbh the class would have worked as well (in other places I think I have, or had, definitions in a loop or something like that). Changing the base class name to disambiguate, the call:
```python
dumper = type("GeometryDumper", (BaseGeometryDumper,), {"oid": info.oid})
```
is equivalent to:
```python
class GeometryDumper(BaseGeometryDumper):
    oid = info.oid

dumper = GeometryDumper
```

About the formats, psycopg behaviour is [explained here](https://www.psycopg.org/psycopg3/docs/basic/params.html#binary-data). When passing parameters to the client, each value can be sent as text or binary, independently. When the `%s` placeholder is used, psycopg uses the most convenient format: usually but not always binary. In order to force the text or binary format in a query, the user can use `%t` or `%b` placeholders.

On the way back, postgres will return all the data either as text or as binary: that's what `binary=True` on the cursor or query controls.

So, Dumpers can be text or binary, yes, but there is also the choices of "which one to use by default when the user doesn't care?" I think in this case the binary one works better, because it's complete and more efficient. Which one is the default one? The last one registered.

https://github.com/psycopg/psycopg/blob/b70e9da3bedc408e4e4f88ab68b892f5d4d0fd25/psycopg/psycopg/_adapters_map.py#L134-L135

dvarrazzo:
it looks like you have cherry-picked the changes from master, so now it seems that this branch has changed 51 files. Please rebase on master, thank you :)

psycopg/psycopg/types/geometry.py
dvarrazzo: It's too bad, anyway, that the hex string is not accepted as bytes: it would be more efficient to parse. I also see a lot of ctypes: [for experience](https://www.varrazzo.com/blog/2020/05/19/a-trip-into-optimisation/) that's a great loss of performance.

psycopg/psycopg/types/geometry.py
jacopofar: Fixed, thanks

psycopg/psycopg/types/geometry.py
jacopofar: done

jacopofar:
Thanks for the explanation, I was now able to serve the documentation locally and it seems OK, also wrote some test based on the example above. The tests store and retrieve a few Shapely objects, and also generate geometries in the DB using GeoJSON. I'm not used to tox nor GH actions so likely there are some issues.

I still have an issue with the Multipolygon object (the big geoJSON at the beginning of the test), when parsing the wkb Shapely says it's invalid but PostGIS and a few other tools are fine with it, I'm still investigating

psycopg/psycopg/types/geometry.py
jacopofar: I see, it makes sense now.

jacopofar:
Hmm it's failing on matching the shapely dependency with the implementation.
If I set "postgis" as an implementation in `.github/workflows/tests.yml` it doesn't work because it has to match an implementation defined in `psycopg/pq/__init__.py`. If I set it to "python" and use a separate parameter to differentiate the service (and run postgis instead of postgres) then shapely is not installed in the environment, even though I'm passing the "postgis" environment using the `-e` flag and tox.ini has it.
Locally I run it with `PSYCOPG_TEST_DSN=postgres://postgres:testpassword@localhost:15432/osm_data tox -c psycopg -e postgis -- tests/test_shapely.py --color yes` and it works.

jacopofar:
I also just noticed that the doc step fails because it cannot find the Shapely library when importing the adapter code (to generate its docs I assume), I tried to set `autodoc_mock_imports` to mock shapely in `docs/conf.py` without results.

I would avoid installing Shapely only to generate the documentation, but could not find a way to prevent it.

dvarrazzo:
> I also just noticed that the doc step fails because it cannot find the Shapely library when importing the adapter code

That's strange: didn't I add shapely to the `docs` extra? It seems it run it ok in the CI (the lint step checks for errors there). Looking at the changeset you might have lost those changes in a rebase. I've pushed it again on our repos side: see the changes in `setup.py` in https://github.com/psycopg/psycopg/commit/49341f6aa1a674d2f65ab8c724663f0d1f803d3a You can integrate it back

dvarrazzo:
> Now it works, I'm not sure about using `pytest.importorskip` rather than doing as in here:

I am pretty ok with what you did here, we can do without importskip :)

tests/test_shapely.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.mark.parametrize("fmt_out", [Format.TEXT, Format.BINARY])
+def test_write_read_shape(shapely_conn, fmt_out):
+    from shapely.geometry import Point, Polygon
+
+    SAMPLE_POINT = Point(1.2, 3.4)
+    SAMPLE_POLYGON = Polygon([(0, 0), (1, 1), (1, 0)])
+
+    fmt_placeholder = "%b" if fmt_out == Format.BINARY else "%t"
```
dvarrazzo, line 101: For better coverage of the param format combination you can use a `fmt_in` passing `adapt.Format` as parameter. See for instance:

https://github.com/psycopg/psycopg/blob/7fe7d90c4c35054636cf7b0be18ea410037e5ad5/tests/types/test_uuid.py#L12-L17

Explanation: [`PyFormat`](https://www.psycopg.org/psycopg3/docs/api/adapt.html#psycopg.adapt.PyFormat) is an enum whose values (`s`, `b`, `t`) are the placeholders letters. This is unlike [`pq.Format`](https://www.psycopg.org/psycopg3/docs/api/pq.html#psycopg.pq.Format) which is only "text" and "binary".

Using both `fmt_in` and `fmt_out` you test all the possible 6 combinations, although maybe it's more appropriate to test the dumpers with the `fmt_in` range and the loaders with the `fmt_out` one. The [uuid test](https://github.com/psycopg/psycopg/blob/master/tests/types/test_uuid.py) is a good sample.

Note that I've just simplified the formats iteration in many tests: take a look at the current state of master now.

tests/test_shapely.py
```diff
@@ -0,0 +1,143 @@
...
+
+@pytest.mark.parametrize("fmt_out", [Format.TEXT, Format.BINARY])
+def test_match_geojson(shapely_conn, fmt_out):
+    from shapely.geometry import Point
+
+    SAMPLE_POINT = Point(1.2, 3.4)
+    with shapely_conn.cursor(binary=fmt_out) as cur:
+        cur.execute(
+            f"""
+            select ST_GeomFromGeoJSON('{SAMPLE_POINT_GEOJSON}')
```
dvarrazzo, line 138: Maybe better to pass this as normal parameter instead of f-string? Same for others?

dvarrazzo (changes requested):

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           - {impl: c, python: 3.8, postgres: 11}
           - {impl: c, python: 3.9, postgres: 10}
           - {impl: dns, python: 3.9, postgres: 13}
-          - {impl: python, python: 3.9, postgres: "13-master", postgis: true}
+          - {impl: python, python: 3.9, postgres: "13-master", postgis: "yes"}
```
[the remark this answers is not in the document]
jacopofar: yes, I didn't want to touch existing tests config but it makes sense

.github/workflows/tests.yml
```diff
@@ -51,14 +52,18 @@ jobs:
           PYTEST_ADDOPTS: "-m 'not subprocess'"
 
       - name: Run DNS-related tests
-        run: tox -c psycopg -e dns -- --color yes -m dns
+        run: tox -c psycopg -e dns -- --color yes
```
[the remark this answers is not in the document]
jacopofar: my mistake, will restore it

.github/workflows/tests.yml
```diff
@@ -102,9 +110,17 @@ jobs:
...
         run: brew install postgresql@13
 
+      - name: Install PostGIS extension on the runner
+        run: brew install postgis
+        if: ${{ matrix.postgis == 'yes' }}
+
       - name: Start PostgreSQL service for test
         run: brew services start postgresql
 
+      - name: Enable PostGIS extension
```
dvarrazzo, line 120: In other tests, setting prerequisites is done by the unit test, using a fixture. See hstore for an example.

Not sure it's better in absolute terms. If it fails, tests are skipped instead of failing, which is not really optimal. However it keeps the pipeline simpler.

I'll take a look myself at whether it can be improved.

.github/workflows/tests.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         run: brew services start postgresql
 
       - name: Enable PostGIS extension
-        run: psql -c "CREATE EXTENSION postgis;" postgres
+        run: psql -h 127.0.0.1 -U runner -c "CREATE EXTENSION postgis;" postgres
```
dvarrazzo, line 121: If you are finding problems with macOS but postgis is known to just work with that platform maybe we can avoid running these tests?

Also note that creating the extension from pytest rather than from runner would probably need no configuration tweaking

.github/workflows/tests.yml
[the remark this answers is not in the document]
jacopofar: done

.github/workflows/tests.yml
jacopofar: Yes, and installing it seems quite heavy, maybe the best solution is to test it only on Linux (either using the postgis image or setting it up somehow in the unit test itself)

dvarrazzo: If postgis image and Linux work, I'm happy with that :)

tests/test_shapely.py
jacopofar: Done

.github/workflows/tests.yml
jacopofar: Removed, now it's only on linux (where it's already configured in the postgis image so no CREATE EXTENSION step is needed)

.github/workflows/tests.yml
jacopofar: I think now that macOS postgis test is gone it's not needed anymore

tests/test_shapely.py
jacopofar: Nice, done!

dvarrazzo:
> I still have an issue with the Multipolygon object (the big geoJSON at the beginning of the test), when parsing the wkb Shapely says it's invalid but PostGIS and a few other tools are fine with it, I'm still investigating

Hi @jacopofar 

Have you resolved this issue? I guess it was related to the argument passing style. If so, you can hand me over this branch: I will rebase it, uniform the tests to the rest of the sure and merge. 

I mostly care that you pass us your experience with the postgis objects, and so that the test suite has representative tests; we can take care of the pytest/tox/GitHub integration.

Than very much for your work so far!

dvarrazzo:
I have cleaned up a bit the history of this branch and merged to master. Thank you very much for this contribution :slightly_smiling_face:

PR #84, opened 2021-09-26
Improve async story in the test on windows

### LINKED ISSUE #76: Support async on Windows
Windows asyncio tests don't pass and they are [currently skipped](https://github.com/psycopg/psycopg/blob/5b92c16e3b151dc57ecbbcabddfda27a1751e8c2/tests/conftest.py#L52-L55).

Does async work on Windows at all?

CaselIT:
Make test work under windows, by requiring a compatible event loop.

Also raise an exception when trying to use an async connection with an incompatible event loop on windows.
I'm not sure about what the most appropriate exception is in this case, InterfaceError?

- TODO: test of the error

Fixes #76

### DIFF
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -6,6 +6,7 @@ psycopg async connection objects
 
 import asyncio
 import logging
+import sys
 from types import TracebackType
 from typing import Any, AsyncIterator, Dict, Optional, Type, Union
 from typing import cast, overload, TYPE_CHECKING
@@ -90,6 +91,18 @@ class AsyncConnection(BaseConnection[Row]):
         row_factory: Optional[AsyncRowFactory[Row]] = None,
         **kwargs: Any,
     ) -> "AsyncConnection[Any]":
+
+        if sys.platform == "win32":
+            loop = asyncio.get_running_loop()
+            if isinstance(loop, asyncio.ProactorEventLoop):
+                raise e.InterfaceError(
+                    "psycopg does not currently support running in async mode "
+                    "on windows on the 'ProactorEventLoop'. "
+                    "Please ensure that a compatible event loop is used, like "
+                    "by setting ``asyncio.set_event_loop_policy`` to "
+                    "WindowsSelectorEventLoopPolicy"
+                )
+
         params = await cls._get_connection_params(conninfo, **kwargs)
         conninfo = make_conninfo(**params)
 
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -93,7 +93,10 @@ class AsyncConnection(BaseConnection[Row]):
     ) -> "AsyncConnection[Any]":
 
         if sys.platform == "win32":
-            loop = asyncio.get_running_loop()
+            if sys.version_info < (3, 7):
+                loop = asyncio.get_event_loop()
+            else:
+                loop = asyncio.get_running_loop()
             if isinstance(loop, asyncio.ProactorEventLoop):
                 raise e.InterfaceError(
                     "psycopg does not currently support running in async mode "
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ __pycache__/
 /docs/.venv/
 *.html
 /psycopg_binary/
+.vscode
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,8 +87,16 @@ def event_loop(request):
     else:
         assert loop == "default"
 
+    loop = None
     if sys.platform == "win32":
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
-    loop = asyncio.get_event_loop_policy().new_event_loop()
+        if sys.version_info < (3, 7):
+            loop = asyncio.SelectorEventLoop()
+            asyncio.set_event_loop(loop)
+        else:
+            asyncio.set_event_loop_policy(
+                asyncio.WindowsSelectorEventLoopPolicy()
+            )
+    if not loop:
+        loop = asyncio.get_event_loop_policy().new_event_loop()
     yield loop
     loop.close()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -58,13 +58,6 @@ def pytest_report_header(config):
     return [f"asyncio loop: {loop}"]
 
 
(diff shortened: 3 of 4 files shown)


psycopg/psycopg/connection_async.py
```diff
@@ -90,6 +91,20 @@ async def connect(
...
     ) -> "AsyncConnection[Any]":
+
+        if sys.platform == "win32":
+            loop = asyncio.get_running_loop()
+            if isinstance(loop, asyncio.ProactorEventLoop):
+                # TODO: is there a general psycopg error not related
+                # to db operations? InterfaceError
+                raise NotImplementedError(
+                    "psycopg does not currently support running in async mode "
+                    "on windows on the ``ProactorEventLoop`` event loop. "
```
dvarrazzo, line 102: `'ProactorEventLoop' event loop`  is redundant in a message already quite long :)

psycopg/psycopg/connection_async.py
```diff
@@ -90,6 +91,20 @@ async def connect(
         row_factory: Optional[AsyncRowFactory[Row]] = None,
         **kwargs: Any,
     ) -> "AsyncConnection[Any]":
+
+        if sys.platform == "win32":
+            loop = asyncio.get_running_loop()
+            if isinstance(loop, asyncio.ProactorEventLoop):
+                # TODO: is there a general psycopg error not related
```
dvarrazzo, line 98: Probably `InterfaceError` is good, as it's not a `DatabaseError` (see [inheritance tree](https://www.psycopg.org/psycopg3/docs/api/module.html))

dvarrazzo:
The failed test on Linux is caused by something that I have fixed on master a few minutes ago, so please rebase

CaselIT:
Interesting, here all test on windows pass (other than on py3.6 because it's missing something) locally I had some other failures.

Thanks for the review, I'll update later

dvarrazzo:
> Interesting, here all test on windows pass (other than on py3.6 because it's missing something) locally

It's a random test :) I enlarged random tests coverage a few days ago and it started failing on certain exotic types (like arrays on numeric ranges).

CaselIT:
Updated

> 
> 
> > Interesting, here all test on windows pass (other than on py3.6 because it's missing something) locally
> 
> It's a random test :) I enlarged random tests coverage a few days ago and it started failing on certain exotic types (like arrays on numeric ranges).

I also get some other errors locally, like a single
```py
_________________________________________________________ test_connect_timeout __________________________________________________________

kwargs = {}, coro = <coroutine object test_connect_timeout at 0x000002BB758B0BC0>
task = <Task pending name='Task-20' coro=<test_connect_timeout() running at \GitHub\psycopg\tests\test_connection_async.py:65> wait_for=<_GatheringFuture pending cb=[<TaskWakeupMethWrapper object at 0x000002BB755B48B0>()]>>

    @functools.wraps(func)
    def inner(**kwargs):
        coro = func(**kwargs)
        if coro is not None:
            task = asyncio.ensure_future(coro, loop=_loop)
            try:
>               _loop.run_until_complete(task)

..\..\..\miniconda3\envs\psy\lib\site-packages\pytest_asyncio\plugin.py:195:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\..\miniconda3\envs\psy\lib\asyncio\base_events.py:629: in run_until_complete
    self.run_forever()
..\..\..\miniconda3\envs\psy\lib\asyncio\base_events.py:596: in run_forever
    self._run_once()
..\..\..\miniconda3\envs\psy\lib\asyncio\base_events.py:1854: in _run_once
    event_list = self._selector.select(timeout)
..\..\..\miniconda3\envs\psy\lib\selectors.py:324: in select
    r, w, _ = self._select(self._readers, self._writers, [], timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <selectors.SelectSelector object at 0x000002BB755B4310>, r = {848}, w = {924}, _ = [], timeout = 0

    def _select(self, r, w, _, timeout=None):
>       r, w, x = select.select(r, w, w, timeout)
E       OSError: [WinError 10038] An operation was attempted on something that is not a socket

..\..\..\miniconda3\envs\psy\lib\selectors.py:315: OSError
======================================================== short test summary info ========================================================
FAILED tests/test_connection_async.py::test_connect_timeout - OSError: [WinError 10038] An operation was attempted on something that is...
===================================================== 1 failed, 76 passed in 6.55s ======================================================
```

no clue why the github action works, but it's not very important I guess.

dvarrazzo:
As far as I can see, async tests on 3.6 can just be skipped. The version is on its last leg and the async pool doesn't work (on any platform) because of a bug that won't be fixed, so I think let's just reinstate the skip that was there before, just limited to Python < 3.7.

CaselIT:
> 
> 
> As far as I can see, async tests on 3.6 can just be skipped. The version is on its last leg and the async pool doesn't work (on any platform) because of a bug that won't be fixed, so I think let's just reinstate the skip that was there before, just limited to Python < 3.7.

last try to see if it's ok now

CaselIT:
all green note, sorry for the noise before

dvarrazzo:
Thank you very much! Merging

CaselIT:
Hi,

I'll try taking a look. 

> note that the error might be misleading here: it might have happened in a previous test which generated a warning and this one caught it.

I think that may be the case. I think the issue is the `def test_windows_error(dsn):` that does not close the loop and I guess sometimes the gc fill fire, raising the warning, in this test.

The warning comes from here https://github.com/python/cpython/blob/4e605666b08b8f863cbbbdaa34bb06988e648d26/Lib/asyncio/base_events.py#L674-L678

CaselIT:
I think the snipped in this docs link should solve it

https://docs.python.org/3/library/asyncio-eventloop.html?highlight=close#asyncio.loop.shutdown_asyncgens

PR #87, opened 2021-09-29
Adjust test on PGconn.tty for libpq 14

dlax:
Quoting https://www.postgresql.org/docs/14/libpq-status.html:

>  [PQtty] no longer does anything, but it remains for backwards
>  compatibility. The function always return an empty string.

Or perhaps we should drop 'tty' property (and binding), since this is at
least obsolete since PostgreSQL 7.4?

### DIFF
--- a/tests/pq/test_pgconn.py
+++ b/tests/pq/test_pgconn.py
@@ -224,12 +224,19 @@ def test_port(pgconn):
         pgconn.port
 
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Yes, it's pretty useless. Dropping it from the info object and leaving only the pq binding is surely enough.

Issue #74, opened 2021-08-29
Support Pipeline/Batch mode


dvarrazzo:
In this mode, several cursor may send commands to the same connection and then sit in a queue to receive the result.

https://www.postgresql.org/docs/14/libpq-pipeline-mode.html

dlax:
@dvarrazzo, as I'm progressing on this, I'm wondering where we should put tests for this feature. Would they live alongside other tests in `test_connection.py`, `test_cursor.py`, `test_transaction.py`, `test_server_cursor.py` (and async siblings)? Or should we just add a `test_pipeline.py` file?

dvarrazzo:
Certain major features have their own test file, because it made more sense working focused on that (e.g. `test_prepared.py`). If you find more convenient to have `test_pipeline.py` and `test_pipeline_async.py`. I think it makes sense so.

dvarrazzo:
Finally closed in #175.

Thank you very much, @dlax. This was an epic work!

PR #89, opened 2021-10-01
add libpq interface for pipeline mode

dlax:
First step towards #74 adding  libpq bindings and extending CI checks with libpq 14.

* * *
Next working on implementing a Python interface; as a first (simple) draft, I was thinking of something like:
```python
with conn.pipeline() as pipeline:
   cur.execute(...)
   cur2.execute(...)
   pipeline.sync()  # calls PQpipelineSync
   r1 = cur.fetchone()
   r2 = cur2.fetchall()
```

If I understand correctly, we need to keep track of cursors executing queries during a pipeline session and process their result when asked; which mean `fetch*()` methods would emit `PQgetResult` calls (instead of `execute()` currently). Since multiple cursors might be involved, we need a queue of active cursors to later bind pipeline results to their respective cursor.


dvarrazzo:
Hello, thank you for looking into this. I think it would be an amazing feature to provide.

As a first MR, please feel free to propose one only adding the `psycopg.pq` objects, because their implementation (which you already have is straighforward). What should be built on top of it will be much more open to discussion...

.github/workflows/lint.yml
```diff
@@ -11,6 +11,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
+      - name: Install libpq 14
+        run: |
+          echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -c -s)-pgdg main 14" | sudo tee /etc/apt/sources.list.d/pgdg14.list
+          sudo apt-get -qq update
+          sudo apt-get -y install libpq-dev>=14 libpq5>=14
```
dvarrazzo, line 18: Is libpq 14 needed in the lint pipeline?

dlax: Possibly not. I think I added that before making bindings conditional on libpq version, and these checks failed otherwise.

dlax:
> As a first MR, please feel free to propose one only adding the `psycopg.pq` objects, because their implementation (which you already have is straighforward).

Ok, sounds good.
Do you want only the first commit with bindings only? Or also libpq 14 installation in CI?

.github/workflows/lint.yml
dlax: Removed

tests/pq/test_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+
+def test_pipeline_abort(pgconn, table):
+    pgconn.enter_pipeline_mode()
+    pgconn.send_query_params(b"insert into pipeline values ($1)", [b"1"])
+    pgconn.send_query_params(b"select no_such_function($1)", [b"1"])
+    pgconn.send_query_params(b"insert into pipeline values ($1)", [b"2"])
+    pgconn.pipeline_sync()
+    pgconn.send_query_params(b"insert into pipeline values ($1)", [b"3"])
+    pgconn.pipeline_sync()
+    r = pgconn.get_result()
```
dvarrazzo, line 62: Could you please make this test a bit more readable by dividing the following lines in stanzas and comment to which one of the statements above they match? Thank you :)

dvarrazzo (changes requested):

dvarrazzo:
> Do you want only the first commit with bindings only? Or also libpq 14 installation in CI?

I think having a test in the CI is important but maybe let's have a few jobs with libpq 14 and a few with whatever the system throws in. Adding a few comments to the MR...

.github/workflows/tests.yml
```diff
@@ -36,6 +36,12 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
+      - name: Install libpq 14
```
dvarrazzo, line 39: What I would do is to add a "libpq" entry to the grid above, and mix a few 14 and a few null to the existing steps. We can use conditional instructions and decide to install libpq 14 only on that value. This would allow to verify there is no regression when an older libpq is used, and that we can use pipeline mode also with older server versions.

Please rebase your MR on current master, because yesterday I've changed the grid massively, after introducing PG14 and Py3.10 support.

tests/pq/test_pipeline.py
```diff
@@ -0,0 +1,87 @@
+import pytest
+
+import psycopg
+from psycopg import pq
+
+pytestmark = pytest.mark.libpq(">= 14")
```
dvarrazzo, line 6: Could you please add tests to verify that on libpq < 14 the functions raise the expected exception? Thank you

dlax: Done.

.github/workflows/tests.yml
dlax: Good idea, done.

Also took the liberty to use "postgres:14" docker image, which is now available.

tests/pq/test_pipeline.py
dlax: Done.

PR #91, opened 2021-10-04
Define and use send(), fetch_many() and fetch() generators in C implementation

dlax:
It seems this would be useful for pipeline mode #74 and can already be used in `cursor.stream()`.


dvarrazzo:
Thank you. I'll look at this with more care later (the diff as shown by github is confusing: I'll look at it on my computer)

dlax:
(Just rebased on master, no change otherwise.)

dvarrazzo:
Also dropping that spurious `continue` at the end of the loop from the Python implementation...

PR #92, opened 2021-10-04
Explicitly disallow some mypy checks for tests/ (to improve developer experience)

dlax:
When running mypy on the tests/ directory, the top-level mypy
configuration is used and strict mode is thus enforced whereas tests
code is not typed. As mypy might be run in people editors by default,
the current situation makes this uncomfortable.

By disabling obvious checks, we get a nicer developer experience. A more
straightforward option would be to set 'check_untyped_defs = False', but
this would prevent valid code to be checked, so the current proposal is
a trade-off.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -13,3 +13,7 @@ ignore_missing_imports = True
 
 [mypy-shapely.*]
 ignore_missing_imports = True
+
+[mypy-tests.*]
+disallow_untyped_defs = False
+disallow_untyped_calls = False


dvarrazzo:
Why run mypy on the tests dir? They are not meant to be type-safe (certain tests just involve throwing wrong types at functions, plus I don't think mypy gets fixtures at all...)

I would be more ok with a test that disables mypy in the test dir altogether. I just didn't have the need to do it so far: for me these entries work and mypy doesn't run on other files:
```
[mypy]
files = psycopg/psycopg
mypy_path = psycopg_c, psycopg_pool
```

dlax:
In my experience (from other projects), running mypy in test files is quite useful as it would catch things early and it helps while writing tests. Thus I tend to enable this anyways.

Typically, here, with just `check_untyped_defs = False` and proper usage of the API (e.g. avoiding `cur.fetchone()[0]`), I can easily write "type-safe" tests and have my editor not complain, which is nice from my developer point of view.

Sure mypy sometimes spots deliberate errors, but they are not so common and we could just `#type: ignore` case-by-case, or process by gradual typing by disabling respective checks overall. (I even started doing that in psycopg a while ago, but did not finished.)

Fixtures (injected) are not getting typed (or maybe this has improved since I checked?), but this is not a problem in my experience because they would be inferred as `Any` and their usage would not generate errors.

The current configuration specifies which file to inspect when `mypy` is invoked without any path argument, but it does not prevent `mypy` to be run with other files (try running `mypy tests`), just as editors would do. There's an `exclude` configuration option, but I did not manage to have working.

dvarrazzo:
You are right: I forgot I did that but I actually configured my editor to not run mypy in `tests` directories.

I will give a go to your configuration and merge. I am positive to make the codebase friendly for you and for other devs :)

dvarrazzo:
Taking a look at it. I don't think I want to go about the chore of fixing all the subtle typing error in the tests: if they give an error that's ok, they are tests. In this moment mypy reports 162 errors.

image

I don't feel ok with these warning in my editor: they clobber other problems (e.g. flake8) and I am ok with `cur.fetchone()[0]` in the tests: if it raises an exception the test failed.

162 false positives in the tests code seems sloppy. For you, Is really better than avoiding to run mypy in the tests altogether? (which I wouldn't know how to do, by the way, because it doesn't seem there is a way to ignore all the errors in the package, and `exclude` doesn't work if a file name is generated).

If you think this improves the current state I will merge it, but probably I will personally go back to not run mypy in the tests dir.

By the way, I think that `.mypy.ini` can be improved with:

```diff
diff --git a/.mypy.ini b/.mypy.ini
index 46466c5..2c6008c 100644
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,9 +1,8 @@
 [mypy]
-files = psycopg/psycopg
+files = psycopg/psycopg, psycopg_pool/psycopg_pool, psycopg_c/psycopg_c
 warn_unused_ignores = True
 show_error_codes = True
 strict = True
-mypy_path = psycopg_c, psycopg_pool
 
 [mypy-pytest]
 ignore_missing_imports = True
```

because at the moment it doesn't run checks in the pool code and C extension (where there's probably just a handful of files anyway). I also see that now mypy config can be written in `pyproject.toml`. But I'll make these changes after we decide what to do with this MR, to avoid to get in your way :)

dlax:
For the test file I'm writing now, mypy is happy with just these two checks disabled; so, yes, this would improve the situation for me :)

I cannot (easily) disable mypy for `tests/` directories in my editor config because I want it enabled in other projects.

I'm also fine with other changes you suggest; thank you!

dvarrazzo:
Here you go :)

dvarrazzo:
@dlax heads up: I have just merged https://github.com/psycopg/psycopg/commit/d8cdbbcf6d9ca567f07a119623f002c906030752 so, after you pull from master, you might want to run `pip install -e ./psycopg[dev]` again to install the typing extensions.

dlax:
Great!
Thanks for the heads up.

PR #93, opened 2021-10-05
Pipeline mode

dlax:
This adds support for connection pipeline mode #74 on Python side.

- the API is to be discussed
- works with sync and async connection/cursor
- at the moment, the user needs to call `PQpipelineSync()` themselves at least once in the pipeline; perhaps we'd want to do this ourselves (e.g. at transaction scope as suggested in postgres doc)
- some things do not work (yet) in this mode, like server-side cursor, `cursor.stream()`, update of query prepare state
- there's propbably some code duplication that can benefit from a refactoring
- I'm not sure yet about `PQflush()` calls (already there) vs. `PQpipelineSync()`, so I need to investigate more about this

Still draft, but happy to get feedback anyways. I'd suggest to review commit-by-commit.

### DIFF
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -7,8 +7,9 @@ psycopg connection objects
 import logging
 import warnings
 import threading
+from collections import deque
 from types import TracebackType
-from typing import Any, Callable, cast, Dict, Generic, Iterator, List
+from typing import Any, Callable, cast, Deque, Dict, Generic, Iterator, List
 from typing import NamedTuple, Optional, Type, TypeVar, Union
 from typing import overload, TYPE_CHECKING
 from weakref import ref, ReferenceType
@@ -37,6 +38,7 @@ from .server_cursor import ServerCursor
 
 if TYPE_CHECKING:
     from .pq.abc import PGconn, PGresult
+    from .cursor import BaseCursor
     from psycopg_pool.base import BasePool
 
 logger = logging.getLogger("psycopg")
@@ -81,16 +83,24 @@ Notify.__module__ = "psycopg"
 NoticeHandler = Callable[[e.Diagnostic], None]
 NotifyHandler = Callable[[Notify], None]
 
+PipelineQueue = Deque[Optional["BaseCursor[Any, Any]"]]
+
 
 class BasePipeline:
-    def __init__(self, pgconn: "PGconn") -> None:
+    def __init__(self, pgconn: "PGconn", queue: PipelineQueue) -> None:
         self.pgconn = pgconn
+        self._queue = queue
+
+    def __len__(self) -> int:
+        return len(self._queue)
 
     def _status(self) -> pq.PipelineStatus:
         return self.pgconn.pipeline_status()
 
     def _sync(self) -> None:
         self.pgconn.pipeline_sync()
+        self._queue.append(None)
+        logger.debug("sent a sync message to pipeline")
 
 
 class Pipeline(BasePipeline):
@@ -153,6 +163,11 @@ class BaseConnection(Generic[Row]):
 
         # Attribute set when the connection enters pipeline mode.
         self._pipeline_mode = False
+
+        # Attribute only set when the connection is in pipeline mode; this
+        # contains either a cursor, if queued query is expected to have data
+        # in its results or None otherwise.
+        self._pipeline_queue: PipelineQueue = deque()
 
         # Time after which the connection should be closed
         self._expire_at: float
@@ -455,6 +470,8 @@ class BaseConnection(Generic[Row]):
 
         if self._pipeline_mode:
             yield from send(self.pgconn)
+            self._pipeline_queue.append(None)
+            logger.debug("sent command '%s' in pipeline", command.decode())
             return None
 
         result = (yield from execute(self.pgconn))[-1]
@@ -534,7 +551,32 @@ class BaseConnection(Generic[Row]):
     def _fetch_many_gen(self) -> PQGen[List["PGresult"]]:
         """Return results of one query from the database."""
         results = yield from fetch_many(self.pgconn)
+        logger.debug(
+            "received '%s' result(s) from pipeline",
+            ", ".join(pq.ExecStatus(r.status).name for r in results),
+        )
         return results
+
+    def _end_pipeline_gen(self) -> PQGen[None]:
+        """Empty the pipeline queue from unfetched results, usually those from
+        commands (e.g. COMMIT) or pipeline synchronization points which
+        results are not expected to be fetched by the user.
+        """
+        while self._pipeline_queue:
+            queued = self._pipeline_queue.popleft()
+
+            if queued is not None:
+                raise e.ProgrammingError(
+                    f"{queued} has unfetched results in the pipeline"
+                )
+
+            results = yield from self._fetch_many_gen()
+
+            (result,) = results
+            if result.status == ExecStatus.FATAL_ERROR:
+                raise e.error_from_result(
+                    result, encoding=pgconn_encoding(self.pgconn)
+                )
 
 
 class Connection(BaseConnection[Row]):
@@ -797,11 +839,14 @@ class Connection(BaseConnection[Row]):
 
         :rtype: Pipeline
         """
+        assert not self._pipeline_queue
         self.pgconn.enter_pipeline_mode()
         self._pipeline_mode = True
         try:
-            yield Pipeline(self.pgconn)
+            yield Pipeline(self.pgconn, self._pipeline_queue)
         finally:
+            with self.lock:
+                self.wait(self._end_pipeline_gen())
             self.pgconn.exit_pipeline_mode()
             self._pipeline_mode = False
 
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -78,6 +78,28 @@ NoticeHandler = Callable[[e.Diagnostic], None]
 NotifyHandler = Callable[[Notify], None]
 
 
+class BasePipeline:
+    def __init__(self, pgconn: "PGconn") -> None:
+        self.pgconn = pgconn
+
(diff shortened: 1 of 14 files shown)


psycopg/psycopg/connection.py
```diff
@@ -125,6 +147,9 @@ def __init__(self, pgconn: "PGconn"):
         # apart a connection in the pool too (when _pool = None)
         self._pool: Optional["BasePool[Any]"]
 
+        # Attribute set when the connection enters pipeline mode.
+        self._pipeline_mode = False
```
dvarrazzo, line 165: Why do you say that the code would be more complicated using `PGconn.pipeline_status()`? It doesn't seem to create problems in async too. I would prefer to do that: where we could, I tried to avoid to add Python state and rely on the pgconn state only (e.g. there are no explicit status, transaction_status, closed...)

psycopg/psycopg/connection_async.py
```diff
@@ -34,6 +34,17 @@
 logger = logging.getLogger("psycopg")
 
 
+class AsyncPipeline(BasePipeline):
+    """Handler for async connection in pipeline mode."""
+
+    async def status(self) -> PipelineStatus:
```
dvarrazzo, line 40: Why is this an async method? `PQpipelineStatus` is only a C member lookup, it doesn't do I/O. I think we should make `status` a property of `BasePipeline`.

Actually, looking at the `pq` wrappers, I think that `pq.pipeline_status` might be a property and we could actually wrap all the three status/enter/exit under a writable property. I've looked at the source code and none of them should do any I/O.

tests/test_pipeline_async.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+async def test_pipeline_status(aconn):
+    assert not aconn._pipeline_mode
+    async with aconn.pipeline() as p:
+        assert await p.status() == pq.PipelineStatus.ON
+        assert aconn._pipeline_mode
+        await p.sync()
+        r = aconn.pgconn.get_result()
+        assert r.status == pq.ExecStatus.PIPELINE_SYNC
```
dvarrazzo: The reason to make `status` a property: we would have `pipeline.status()` but `result.status` :)

psycopg/psycopg/cursor.py
```diff
@@ -275,6 +287,9 @@ def _stream_send_gen(
         self._last_query = query
 
     def _stream_fetchone_gen(self, first: bool) -> PQGen[Optional["PGresult"]]:
+        assert (
+            not self._conn._pipeline_mode
+        ), "pipeline mode not supported"  # TODO
```
dvarrazzo, line 302: I would just deny the use of `stream()` in pipeline mode. They seem two different use cases:
- pipeline is to send several operations in parallel from separate threads
- stream is for a single slow operation that doesn't produce all the result at the time of call.
So I don't see why people should try to call stream() in a pipeline() block :)

So I would just make calling `stream()` under `pipeline()` to raise a `ProgrammingError`.

psycopg/psycopg/server_cursor.py
```diff
@@ -108,6 +108,7 @@ def _close_gen(self, cur: BaseCursor[ConnectionType, Row]) -> PQGen[None]:
                 "SELECT 1 FROM pg_catalog.pg_cursors WHERE name = {}"
             ).format(sql.Literal(self.name))
             res = yield from cur._conn._exec_command(query)
+            assert res is not None, "pipeline mode not supported"  # TODO
```
dvarrazzo, line 111: server-side cursors in pipeline mode. mmm... It might be useful but it might pretty complicated maybe?

I'd be tempted to raise `NotSupportedError` to leave the door open to think about it in the future but leave it out of the first implementation.

psycopg_c/psycopg_c/pq/libpq.pxd
```diff
@@ -294,6 +294,8 @@ cdef extern from *:
 #endif
 
 #if PG_VERSION_NUM < 140000
+int PGRES_PIPELINE_SYNC = 10;
```
dvarrazzo, line 297: This becomes a module variable, so I think it should be declared static... But probably a `#define PGRES_PIPELINE_SYNC 10` is even better.

psycopg/psycopg/connection.py
```diff
@@ -534,8 +551,33 @@ def _rollback_gen(self) -> PQGen[None]:
...
 
+    def _end_pipeline_gen(self) -> PQGen[None]:
+        """Empty the pipeline queue from unfetched results, usually those from
+        commands (e.g. COMMIT) or pipeline synchronization points which
+        results are not expected to be fetched by the user.
+        """
+        while self._pipeline_queue:
+            queued = self._pipeline_queue.popleft()
+
+            if queued is not None:
```
dvarrazzo, line 568: Is it really an error condition to have unfetched results? What if the user doesn't care (e.g. UPDATE or SELECT for a side effect)? I could also see race conditions in case the cursor is managed by a thread while the pipeline block is managed by the main thread (which sounds like a likely way to use it).

psycopg/psycopg/cursor.py
```diff
@@ -502,6 +514,47 @@ def _scroll(self, value: int, mode: str) -> None:
             raise IndexError("position out of bound")
         self._pos = newpos
 
+    def _fetch_pipeline_gen(self) -> PQGen[None]:
```
dvarrazzo, line 517: In this method, if I understand correctly, when a cursor receives an execute*() call, it will go in the connection queue, look for itself in the cursors to wait, dispatching previously received results to the respective other cursors, and then proceed on.

It doesn't feel the most straightforward implementation to me. There is the initial check if the cursor is in the pipeline which is O(n) (in a likely small n, sure). But the entire idea of getting the result to be driven by the user seems backwards compared to the flow of information: who is in charge here is the connection and the results it receives.

The way I was thinking about this algorithm is more:

- the connection has a queue of cursors that have executed a statement
- the cursor has a `queued` [`threading.Event`](https://docs.python.org/3/library/threading.html#event-objects). This object has `clear()` and `set()` methods to change its state and a `wait()` method. If `wait()` is called when the event is cleared, it will block, until another thread calls `set()`. If it waits when set, it doesn't block.

Things that happen:

- on cursor.execute():
  - the connection executes the query
  - the connection puts the cursor in the queue
  - `cursor.queued.clear()` is called.

- on cursor.fetch()
  - the cursor `wait()` on the queued event, meaning that it will blocked if its result hasn't arrived yet, otherwise will waste no time
  - the cursor processes the result

- when the connection receives the result:
  - it takes the first item from the cursors queue
  - sets the result on the cursor
  - calls `set()` the queued event to release a thread eventually stuck on a fetch() or to mark it ready to fetch if the client hasn't requested a result yet.

I feel this organisation more natural, with the connection "dealing" results as they come and every cursor only in charge of its own state, not changing the state of the connection and of other cursors.

What do you think?

tests/test_pipeline_async.py
```diff
@@ -1,34 +1,47 @@
...
 from psycopg import pq
+from psycopg.errors import UndefinedTable
 
 pytestmark = [
     pytest.mark.asyncio,
     pytest.mark.libpq(">=14"),
 ]
 
 
+@pytest.fixture(autouse=True)
```
dvarrazzo, line 15: What are these fixtures for?

dvarrazzo:
>     * the API is to be discussed

Controversial idea: what if psycopg was _always_ in pipeline mode, instead of managing a context? Would that be a possibility?

The pipeline mode is only a libpq construct, it is not a protocol concept, and async-oriented clients that don't use the libpq pretty much do that (a rust driver I read some conversations about does that, at least)


> Still draft, but happy to get feedback anyways. I'd suggest to review commit-by-commit.

I wanted to tell you: thank you very much for this work so far and for the organisation of your patches: your changeset was very readable so far :)

psycopg/psycopg/connection_async.py
dlax: You're totally right. I'm still kind of looking for the way in libpq API so my reasoning is sometimes wrong. Thanks for the suggestion, I'll apply that pattern.

tests/test_pipeline.py
```diff
@@ -17,3 +18,176 @@ def test_pipeline_status(conn):
...
+        assert pgconn.get_result() is None
+
+        r = pgconn.get_result()
+        assert r.status == pq.ExecStatus.PIPELINE_SYNC
+
+
+@pytest.mark.xfail
+def test_auto_prepare(conn):
+    # Auto prepare does not work because cache maintainance requires access to
+    # results at the moment.
```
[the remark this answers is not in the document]
dlax: Yes, that's doable I think (hence xfail tests already there). I had the same idea basically.

psycopg/psycopg/cursor.py
```diff
@@ -242,13 +247,20 @@ def _maybe_prepare_gen(
...
+            else:
+                (result,) = yield from execute(self._pgconn)
+                if result.status == ExecStatus.FATAL_ERROR:
+                    raise e.error_from_result(
+                        result, encoding=pgconn_encoding(self._pgconn)
+                    )
             self._send_query_prepared(name, pgq, binary=binary)
 
+        if pipeline_mode:
+            yield from send(self._pgconn)
```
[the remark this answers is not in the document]
dlax: That `send()` in the "else" case is for `PQsendPrepare` but there is then a `PQsendQueryPrepared` issued later, which is sent by this one. Unless I'm mistaken, the logic did not change as there already was (possibly) two `execute()` before.

dlax:
Hm, windows builds fail with `LINK : fatal error LNK1181: cannot open input file 'libpq.lib'`.

dlax:
@dvarrazzo, I am worried that this might lead nowhere if I keep iterating over the initial implementation, as I did until now. So I would rather start over based on the design discussions we had here and resubmit another (or several other) PR.

dlax:
Replaced by #116

PR #97, opened 2021-10-07
Drop AnyCursor typevar

dvarrazzo:
It must have been made redundant some time ago by some type system
refactoring.

@dlax does this work ok with your branch?


dlax:
Makes things simpler on my branch, indeed. But the generic "self" is needed, according to tests failure, it seems.

PR #98, opened 2021-10-07
Properly close the at hock event loop in test_windows_error

CaselIT:
This should fix https://github.com/psycopg/psycopg/pull/84#issuecomment-937894055


CaselIT:
The fails in the pipeline do not seem related to this. Also I've seen this one `AssertionError: assert Range(empty=True) == NumericRange(Decimal('-0.0000'), Decimal('0E+6'), '[)')` also in a run on my fork, so maybe there is some other issue at play?

PR #102, opened 2021-10-11
Add PGconn.send_flush_request()

dlax:
This function is associated with the pipeline mode (#74).


dvarrazzo:
The test failure is unrelated. Fixed in 0317bb8.

PR #107, opened 2021-10-13
waiting: futures, Ready.RW

dlax:
The second change appears to be needed for the pipeline mode #74 in order to have a generator waiting on RW-read event in async case. I tried to explain things in commit messages. For more context, the generator I'm working on is visible at https://github.com/dlax/psycopg3/commit/98f741a95b2beb24208639a539e4d74c40d26fd2.

### DIFF
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -13,7 +13,13 @@ import select
 import selectors
 from enum import IntEnum
 from typing import Optional
-from asyncio import get_event_loop, wait_for, TimeoutError
+from asyncio import (
+    get_event_loop,
+    wait as asyncio_wait,
+    wait_for,
+    FIRST_COMPLETED,
+    TimeoutError,
+)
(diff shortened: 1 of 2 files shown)


dlax:
> I wasn't aware of a thing like ready for RW. I thought the kernel tells you either one or the other.

On that topic, https://github.com/dlax/psycopg3/commit/8881aa49a11eeb465ea1a5f76f10a19cbaf95b61 adds a test to demonstrate that this is possible with various select() strategies. (See the CI job at https://github.com/dlax/psycopg3/actions/runs/1340787826.)

Also, simpler:
```python
>>> import select
>>> import tempfile
>>> fd, fpath = tempfile.mkstemp()
>>> fd
3
>>> select.select([fd], [fd], [])
([3], [3], [])
```

(Not an expert in this field, so my reasoning might be wrong...)

dlax:
Amended second commit to use the simpler test mentioned above and apply to all `wait()` implementations.

Issue #113, opened 2021-10-14
Pipenv can't install psycopg[binary]


Polectron:
When installing psycopg-binary through pipenv (`pipenv install psycopg[binary]`) I get the following error:
```
Pipfile.lock not found, creating...
Locking [dev-packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success! 
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✘ Locking Failed! 
CRITICAL:pipenv.patched.notpip._internal.index.package_finder:Could not find a version that satisfies the requirement psycopg-binary==3.0 (from psycopg[binary]==3.0->-r /tmp/pipenvzd_vdabkrequirements/pipenv-g7_y782g-constraints.txt (line 6)) (from versions: none)
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/resolver.py", line 741, in _main
[ResolutionFailure]:       resolve_packages(pre, clear, verbose, system, write, requirements_dir, packages, dev)
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/resolver.py", line 702, in resolve_packages
[ResolutionFailure]:       results, resolver = resolve(
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/resolver.py", line 684, in resolve
[ResolutionFailure]:       return resolve_deps(
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/utils.py", line 1397, in resolve_deps
[ResolutionFailure]:       results, hashes, markers_lookup, resolver, skipped = actually_resolve_deps(
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/utils.py", line 1110, in actually_resolve_deps
[ResolutionFailure]:       resolver.resolve()
[ResolutionFailure]:   File "/home/daniel/.local/lib/python3.9/site-packages/pipenv/utils.py", line 835, in resolve
[ResolutionFailure]:       raise ResolutionFailure(message=str(e))
[pipenv.exceptions.ResolutionFailure]: Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  First try clearing your dependency cache with $ pipenv lock --clear, then try the original command again.
[...]

PR #115, opened 2021-10-15
Manylinux2014 support

Polectron:
Add workflow job for creating manylinux2014 images for x86_64 and i686, based on psycopg2_binary workflow. Necessary for versions older than Ubuntu 21.04 and letting pipenv build locks correctly (#113 )

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -83,7 +83,7 @@ jobs:
   # }}}
 
   manylinux2014: # {{{
-    name: Build Linux packages
+    name: Build manylinux2014 packages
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you, but no.

Manylinux 2014 builds on Centos 7, which includes too outdated libpq and openssl libraries.

Psycopg is perfectly usable on current and older ubuntu version: it is enough to upgrade pip, as pip itself continuously repeats to do.

We are also not responsible of pipenv bugs. psycopg works fine with poetry, with pip-tools, so if they have problems they can surely fix them.

PR #116, opened 2021-10-15
pipeline mode (v2)

dlax:
Second attempt to implement the pipeline mode (#74), replacing #93. This is on top of PR #120 (first 5 commits).
Still suggest to read commit by commit, though the last one is large.

The core algorithm implements what's described at https://www.postgresql.org/docs/14/libpq-pipeline-mode.html#LIBPQ-PIPELINE-INTERLEAVE on `execute()` steps. In `fetch*()` steps during the pipeline, we might force to flush the output buffer with `PQsendFlushRequest()`; this also happens at the end of pipeline. 

This is tested with a similar strategy as in postgres test at https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/test/modules/libpq_pipeline/libpq_pipeline.c;h=c27c4e0adaf052749574681dc4d2c50ccab9b08b;hb=refs/tags/REL_14_0#l632

Automatic prepared statement is handled now (in contrast with #93), thanks to some refactoring in `PrepareManager`.

The "lock" idea suggested at https://github.com/psycopg/psycopg/pull/93#discussion_r722726088 and https://github.com/psycopg/psycopg/pull/93#issuecomment-937751429 (require a `fetch*()` call before a subsequent `execute()` call on a cursor) is not implemented. In fact, I'm not quite sure it's needed or, more specifically, not sure why it'd be needed in pipeline mode but not in normal mode... To be discussed (as everything else).

### DIFF
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -110,7 +110,6 @@ class BasePipeline:
 
     def _sync(self) -> None:
         self.pgconn.pipeline_sync()
-        logger.debug("sending sync message to pipeline")
         self._queue.append(None)
 
 
@@ -481,7 +480,6 @@ class BaseConnection(Generic[Row]):
 
         if self.pgconn.pipeline_status:
             self._pipeline_queue.append(None)
-            logger.debug("sending command '%s' in pipeline", command.decode())
             return None
 
         result = (yield from execute(self.pgconn))[-1]
@@ -661,11 +659,9 @@ class BaseConnection(Generic[Row]):
         fetching.
         """
         if not self._pipeline_queue:
-            logger.debug("pipeline queue is empty, skipping fetch")
             return
 
         if flush:
-            logger.debug("sending flush request")
             self.pgconn.send_flush_request()
             yield from send(self.pgconn)
 
@@ -684,10 +680,6 @@ class BaseConnection(Generic[Row]):
         cache. Otherwise, results are attached to their respective cursor.
         """
         queued = self._pipeline_queue.popleft()
-        logger.debug(
-            "received %s results from pipeline",
-            ", ".join(pq.ExecStatus(r.status).name for r in results),
-        )
         if queued is None:
             (result,) = results
             if result.status in (
--- a/psycopg/psycopg/cursor.py
+++ b/psycopg/psycopg/cursor.py
@@ -4,7 +4,6 @@ psycopg cursor objects
 
 # Copyright (C) 2020-2021 The Psycopg Team
 
-import logging
 import sys
 from types import TracebackType
 from typing import Any, Callable, Generic, Iterable, Iterator, List
@@ -29,8 +28,6 @@ if TYPE_CHECKING:
     from .abc import Transformer
     from .pq.abc import PGconn, PGresult
     from .connection import Connection, PipelineQueueItem
-
-logger = logging.getLogger("psycopg")
 
 execute: Callable[["PGconn"], PQGen[List["PGresult"]]]
 
@@ -266,11 +263,6 @@ class BaseCursor(Generic[ConnectionType, Row]):
             # The query must be prepared and executed
             self._send_prepare(name, pgq)
             if pipeline_mode:
-                logger.debug(
-                    "sending prepared statement creation request for '%s' "
-                    "in pipeline",
-                    pgq.query.decode(),
-                )
                 self._conn._pipeline_queue.append(None)
             else:
                 (result,) = yield from execute(self._pgconn)
@@ -283,7 +275,6 @@ class BaseCursor(Generic[ConnectionType, Row]):
         key = self._conn._prepared.maybe_add_to_cache(pgq, prep, name)
 
         if pipeline_mode:
-            logger.debug("sending query '%s' in pipeline", pgq.query.decode())
             if key is not None:
                 queued: "PipelineQueueItem" = (self, (key, prep, name))
             else:


psycopg/psycopg/generators.py
```diff
@@ -149,6 +149,63 @@ def fetch(pgconn: PGconn) -> PQGen[Optional[PGresult]]:
...
+        ready = yield Wait.RW
+
+        if ready & Ready.R:
+            # This requires that the server has its output buffer flushed,
+            # which happens either automatically upon PQpipelineSync() or by
+            # calling PQsendFlushRequest().
+            pgconn.consume_input()
+
+            # Consume notifies
+            while 1:
```
dvarrazzo, line 189: I guess you copied this from my code. `while 1` is a pattern I can't take off my muscle memory...

dlax: Yes :) /me trying to be consistent with existing style... I wouldn't mind if we'd change all these!

psycopg/psycopg/cursor.py
```diff
@@ -396,9 +426,10 @@ def _execute_results(
         if not results:
             raise e.InternalError("got no result from the query")
 
-        for res in results:
-            if res.status not in self._status_ok:
-                self._raise_from_results(results)
+        if not self._pgconn.pipeline_status:
```
dvarrazzo, line 429: Maybe we can set a `results = ()` in the `if pipeline_mode` branch above to avoid this `if` here. Not for some mircro-optimisation but because this method is pretty complicated.

I still don't have a holistic view over the pipeline changes so I don't know if this will be changed later in the changeset I haven't seen yet, or if it can be simplified somehow.

tests/test_pipeline_async.py
```diff
@@ -56,3 +96,117 @@ async def test_server_cursor(aconn):
...
+        c4 = await aconn.execute("select 2")
+        await pipeline.sync()
+
+        (r,) = await c1.fetchone()
+        assert r == 1
+
+        with pytest.raises(UndefinedTable):
+            await c2.fetchone()
+
+        with pytest.raises(psycopg.OperationalError, match="pipeline aborted"):
```
dvarrazzo, line 165: This can probably undergo localization: it would be safer to check the SQLSTATE of the exception

tests/test_pipeline_async.py
```diff
@@ -56,3 +96,117 @@ async def test_server_cursor(aconn):
...
+    async with aconn.pipeline():
+        cursors = [
+            await aconn.execute("select count(*) from pg_prepared_statements")
+            for i in range(10)
+        ]
+
+        assert len(aconn._prepared._prepared) == 1
+
+    res = [(await c.fetchone())[0] for c in cursors]
+    assert res == [0] * 5 + [1] * 5
```
dvarrazzo, line 196: `conn.prepare_threshold = 5` is the default but maybe it's more explicit to specify it at the start of the test

psycopg/psycopg/cursor.py
dvarrazzo: (ah just noticed this is the last changeset...)

dvarrazzo (changes requested):
Hey @dlax, amazing work!

I've gone through the patches step by step, I've just highlighted small things along the way. Now I will check out the branch to see the code all together, follow the flow, play a little bit with it.

Thank you very much!

psycopg/psycopg/connection.py
```diff
@@ -125,6 +168,11 @@ def __init__(self, pgconn: "PGconn"):
         # apart a connection in the pool too (when _pool = None)
         self._pool: Optional["BasePool[Any]"]
 
+        # Attribute only set when the connection is in pipeline mode; this
```
dvarrazzo, line 171: This comment is out of date now

dvarrazzo (changes requested):

dvarrazzo:
I have rebased your branch. I won't force-push as it feels impolite: you can fetch it back it from [the `pipeine2` branch in our repo](https://github.com/psycopg/psycopg/tree/pipeline2).

dlax:
> I have rebased your branch. I won't force-push as it feels impolite: you can fetch it back it from [the `pipeine2` branch in our repo](https://github.com/psycopg/psycopg/tree/pipeline2).

I'm not sure how we should proceed then:
- what to do with #120 (see my last comment https://github.com/psycopg/psycopg/pull/120#issuecomment-966156078)?
- how should I account for review change request? add more commits here?

dvarrazzo:
> > I have rebased your branch. I won't force-push as it feels impolite: you can fetch it back it from [the `pipeine2` branch in our repo](https://github.com/psycopg/psycopg/tree/pipeline2).
> 
> I'm not sure how we should proceed then:
> 
>     * what to do with [support RW ready in waiting functions #120](https://github.com/psycopg/psycopg/pull/120) (see my last comment [support RW ready in waiting functions #120 (comment)](https://github.com/psycopg/psycopg/pull/120#issuecomment-966156078))?
> 
>     * how should I account for review change request? add more commits here?

Rebasing on master seems independent from the decision about what to do. You can `git reset --hard upstream/pipeline2` to have the conflicts resolved and then we can think about it.

Or am I missing something?

dlax:
Yes, I'm fine with rebasing, it's done for both PR now.
It's just that I find it unhandy to work with two dependent PR at the same time...

dvarrazzo:
>  It's just that I find it unhandy to work with two dependent PR at the same time...

Yes, I understand. I apologise for having left open for so long and hadn't realised until yesterday you were working on the anyio front too.

tests/test_pipeline.py
```diff
@@ -0,0 +1,30 @@
...
+
+import pytest
+
+from psycopg import waiting
+
+pytestmark = pytest.mark.libpq(">= 14")
+
+
+@pytest.mark.slow
+def test_pipeline_communicate(pgconn, demo_pipeline, generators):
```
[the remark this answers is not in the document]
dlax: Agreed. These were necessary for me to come up with a working algorithm but I've now moved them as scripts.
Incidentally, running `python tests/scripts/pipeline-demo.py  -n 1 [--async]` produces a deadlock, so there seem to be a problem in the high level API (no problem with the generator AFAICT): to be investigated...

tests/test_pipeline_async.py
dlax: The message comes from us :)

tests/test_pipeline.py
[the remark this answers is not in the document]
dlax: > Incidentally, running `python tests/scripts/pipeline-demo.py -n 1 [--async]` produces a deadlock, so there seem to be a problem in the high level API

Strangely, the error is gone now after rebase. Perhaps because the waiting function (`wait_async()`) changed in the meantime.

psycopg/psycopg/generators.py
```diff
@@ -149,6 +149,63 @@ def fetch(pgconn: PGconn) -> PQGen[Optional[PGresult]]:
...
+                    pgconn.notify_handler(n)
+
+            res: List[PGresult] = []
+            while not pgconn.is_busy():
+                r = pgconn.get_result()
+                if r is None:
+                    if not res:
+                        break
+                    results.append(res[:])
+                    del res[:]
```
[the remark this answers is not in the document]
dlax: I can change as:
```diff
-                    results.append(res[:])
-                    del res[:]
+                    results.append(res)
+                    res = []

```
is that what you had in mind?

tests/test_pipeline.py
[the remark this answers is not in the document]
dlax: Hm, now it fails with `python tests/scripts/pipeline-demo.py  -n10 [--async]`; investigating...

psycopg/psycopg/generators.py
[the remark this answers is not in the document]
dvarrazzo: yes :)

dlax:
Hi @dvarrazzo, I have rebase this PR on your "pipeline-over-prepare" branch and included back the last commit (change from review) as it was missing. Ready to go as far as I can tell!

dvarrazzo:
Hello @dlax 

Playing a bit with the current state of the pipeline. There is this thing:

```python
def test_error(conn):
    conn.autocommit = True
    with conn.pipeline() as pipeline:
        conn.execute("select error")
        conn.execute("select 1").fetchone()
```

this test fails on the fetchone, with "pipeline aborted". I can see that we do receive the error, we associate it to the cursor, but, because we never fetch, we never execute check_result.

If there is no fetch at all, errors are pretty much swallowed and there is no report that something failed, and what. This test passes with no exception but no table is created.

```python
def test_error(conn):
    conn.autocommit = True
    with conn.pipeline() as pipeline:
        conn.execute("select error")
        conn.execute("create table voila ()")
```

This test instead blocks:

```python
def test_error(conn):
    with conn.transaction():
        with conn.pipeline() as pipeline:
            conn.execute("select error")
            conn.execute("create table voila ()")
```

I can see, on the wire, that COMMIT is send no result is received.

I wonder if all these artifacts are related and how we can deal with them better, especially the no silent failures.

psycopg/psycopg/connection.py
```diff
@@ -494,10 +547,72 @@ def _rollback_gen(self) -> PQGen[None]:
...
+
+        This matchs 'results' with its respective element in the pipeline
+        queue. For commands (None value in the pipeline queue), results are
+        checked directly. For prepare statement creation requests, update the
+        cache. Otherwise, results are attached to their respective cursor.
+        """
+        queued = self._pipeline_queue.popleft()
+        logger.debug(
+            "received %s results from pipeline",
+            ", ".join(pq.ExecStatus(r.status).name for r in results),
```
dvarrazzo, line 597: This is a relatively heavy operation which happens regardless of the log level. I think we should remove the excess of debug before merging

dlax: I agree, they are no longer useful. Removed them, now.

dvarrazzo:
Thank you very much @dlax! Very appreciated :)

dlax:
Replaced by #175

PR #120, opened 2021-10-18
support RW ready in waiting functions

dlax:
Replaces #107 and would be needed for the pipeline mode #93.

### DIFF
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -29,6 +29,7 @@ class Wait(IntEnum):
 class Ready(IntEnum):
     R = EVENT_READ
     W = EVENT_WRITE
+    RW = EVENT_READ | EVENT_WRITE
 
 
 def wait_selector(
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -119,29 +119,27 @@ async def wait_async(gen: PQGen[RV], fileno: int) -> RV:
 
     def wakeup(state: Ready) -> None:
         nonlocal ready
-        ready = state
+        ready |= state  # type: ignore[assignment]
         ev.set()
 
     try:
         s = next(gen)
         while 1:
+            reader = s & Wait.R
+            writer = s & Wait.W
+            if not reader and not writer:
+                raise e.InternalError(f"bad poll status: {s}")
             ev.clear()
-            if s == Wait.R:
+            ready = 0  # type: ignore[assignment]
+            if reader:
                 loop.add_reader(fileno, wakeup, Ready.R)
-                await ev.wait()
-                loop.remove_reader(fileno)
-            elif s == Wait.W:
+            if writer:
                 loop.add_writer(fileno, wakeup, Ready.W)
-                await ev.wait()
-                loop.remove_writer(fileno)
-            elif s == Wait.RW:
-                loop.add_reader(fileno, wakeup, Ready.R)
-                loop.add_writer(fileno, wakeup, Ready.W)
-                await ev.wait()
+            await ev.wait()
+            if reader:
                 loop.remove_reader(fileno)
+            if writer:
                 loop.remove_writer(fileno)
-            else:
-                raise e.InternalError("bad poll status: %s")
             s = gen.send(ready)
 
     except StopIteration as ex:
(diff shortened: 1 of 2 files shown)


psycopg/psycopg/waiting.py
```diff
@@ -232,11 +232,12 @@ def wait_epoll(
             while not fileevs:
                 fileevs = epoll.poll(timeout)
             ev = fileevs[0][1]
+            ready = 0
             if ev & ~select.EPOLLOUT:
-                s = Ready.R
-            else:
-                s = Ready.W
-            s = gen.send(s)
+                ready |= Ready.R
```
dvarrazzo, line 222: Here we can drop the `|`

tests/test_waiting.py
```diff
@@ -21,6 +22,11 @@
 ]
 
 
+linux = pytest.mark.skipif(
```
dvarrazzo, line 25: Maybe `skip_if_not_linux`?

psycopg/psycopg/waiting.py
```diff
@@ -110,38 +110,29 @@ async def wait_async(gen: PQGen[RV], fileno: int) -> RV:
...
+                rf.add_done_callback(lambda f: loop.remove_reader(fileno))
+                fs.append(rf)
+            if s & Wait.W:
+                wf = loop.create_future()
+                loop.add_writer(fileno, wf.set_result, Ready.W)
+                wf.add_done_callback(lambda f: loop.remove_writer(fileno))
+                fs.append(wf)
+            if not fs:
+                raise e.InternalError(f"bad poll status: {s}")
+            done, _ = await asyncio_wait(fs, return_when=FIRST_COMPLETED)
```
dvarrazzo, line 133: use `[0]` instead of unpacking the tuple and discarding the element

dvarrazzo (changes requested):
Out of curiosity, have you run any benchmark on your async loop changed? I assume it would be even a tad more performing than before...

psycopg/psycopg/waiting.py
dlax: I copied this from the implementation of [EpollSelector.select()](https://github.com/python/cpython/blob/3.10/Lib/selectors.py#L473) because I believe it's good to follow this reference.
But yes, the `|` is not needed. Micro-optimization?

dlax:
> Out of curiosity, have you run any benchmark on your async loop changed? I assume it would be even a tad more performing than before...

No, I haven't.

But in fact, while working on anyio support #146 I think we'll need to get back to the `Event` approach (and no longer use `Future` objects). Now, as timed passed, I think that handling `Ready.RW` is also achievable with the `Event` approach. So, maybe it's worth reverting to the previous approach? What do you think? (I fine with either options.)

dvarrazzo:
Let's recap:

- in order to implement the pipeline we need to handle `Ready.RW`. No RW no pipeline.
- you have an implementation ready (Futures-based)
- There could be an anyio implementation (does it work already with the pipelines? I haven't read that code yet)
- There could be an Event-based implementation too (but not written yet)

Is this right?

Afaics this choice is important but limited in scope: it only affects the wait loop and we can replace implementation any moment. I think we can proceed with the Futures so we have the interface set up and then we can benchmark to figure out the implementation we prefer.

Also, the Event implementation is the first I wrote and you might have seen from the comments that I've never been entirely sold on it: happy to know if there's a better way.

psycopg/psycopg/waiting.py
dvarrazzo: Yes, micro-optimization: this is pretty much the tightest loop. I'm sure there's more to squeeze here but this was visible.

dlax:
> There could be an Event-based implementation too (but not written yet)

Actually, I have written that one too, see https://github.com/dlax/psycopg3/commit/e0bea08d63eae723ff4bc9ecea6f01dd088c175e

> Also, the Event implementation is the first I wrote and you might have seen from the comments that I've never been entirely sold on it: happy to know if there's a better way.

Yes, seen that. On the other hand, I then saw that anyio's implementation was also using Event object (see https://github.com/agronholm/anyio/blob/3.3.4/src/anyio/_backends/_asyncio.py#L1564); so perhaps it's not so bad?

dlax:
> There could be an anyio implementation (does it work already with the pipelines? I haven't read that code yet)

The proposed anyio implementation does not support RW-ready at it is based on current master (which does not support this), but I can easily adjust it after this change (either `Future`- or `Event`-based). Not a big deal.

dlax:
So my point was that, if there are chances that the anyio proposal gets accepted at some point, then there is no real advantage in picking the `Future`-based implementation and benchmarking it since we'd eventually need to move back to an `Event`-based approach.

dlax:
Suggested changes applied in the last push:
```diff
diff --git a/psycopg/psycopg/waiting.py b/psycopg/psycopg/waiting.py
index eeafe519..7bdf9f5c 100644
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -219,7 +219,7 @@ def wait_epoll(
             ev = fileevs[0][1]
             ready = 0
             if ev & ~select.EPOLLOUT:
-                ready |= Ready.R
+                ready = Ready.R
             if ev & ~select.EPOLLIN:
                 ready |= Ready.W
             assert s & ready
diff --git a/tests/test_waiting.py b/tests/test_waiting.py
index 092ab42d..10e99bbb 100644
--- a/tests/test_waiting.py
+++ b/tests/test_waiting.py
@@ -22,7 +22,7 @@ timeouts = [
 ]
 
 
-linux = pytest.mark.skipif(
+skip_if_not_linux = pytest.mark.skipif(
     not sys.platform.startswith("linux"), reason="non-Linux platform"
 )
 
@@ -62,7 +62,7 @@ waits, wids = list(zip(*waits_and_ids))
 
 @pytest.mark.parametrize("waitfn", waits, ids=wids)
 @pytest.mark.parametrize("wait, ready", zip(waiting.Wait, waiting.Ready))
-@linux
+@skip_if_not_linux
 def test_wait_ready(waitfn, wait, ready):
     def gen():
         r = yield wait
@@ -131,7 +131,7 @@ async def test_wait_async(pgconn):
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize("wait, ready", zip(waiting.Wait, waiting.Ready))
-@linux
+@skip_if_not_linux
 async def test_wait_ready_async(wait, ready):
     def gen():
         r = yield wait
```

pipeline2 branch is out of sync from this one, now; I can rebase it as well or wait until we decide.

dlax:
Ok; I've pushed this implementation here.
Also added https://github.com/psycopg/psycopg/pull/120/commits/c01762372dc81eec2220c60b8e697e302cf6f99f to always remove reader/writer even in case of wait failure.

psycopg/psycopg/waiting.py
dlax: No longer applies as we've moved back to Event-based implementation.

PR #132, opened 2021-11-02
Let mypy check tests/types files

dlax:
First step towards having the test suite type-checked.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,12 @@ testpaths=[
 
 
 [tool.mypy]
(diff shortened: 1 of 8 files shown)


dvarrazzo:
Wow, this (and the other changesets) are massive! I guess we are in the direction of having the test suite statically typed then :smile:

dlax:
Added two more PR #136 and #137.

With those and a couple local changes I have, running mypy in `tests/` leaves:
```
Found 121 errors in 19 files (checked 138 source files)
```
So I think the goal to have the test suite statically typed checked is achievable.

I'll wait for review/integration before moving on :)

PR #133, opened 2021-11-02
Let mypy check tests/pq files

dlax:
Follow-up, based on #132.


dvarrazzo:

PR #134, opened 2021-11-02
Let mypy check tests/test_sql.py

dlax:
Followup #132, #133.


dvarrazzo:

PR #135, opened 2021-11-02
Let mypy check tests/*dbapi20.py

dlax:
Follow-up #132, #133, #134

### DIFF
--- a/tests/dbapi20.py
+++ b/tests/dbapi20.py
@@ -20,6 +20,7 @@ __author__ = 'Stuart Bishop <stuart@stuartbishop.net>'
 import unittest
 import time
 import sys
+from typing import Any, Dict
 
 
 # Revision 1.12  2009/02/06 03:35:11  kf7xm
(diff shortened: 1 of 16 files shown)


tests/dbapi20.py
```diff
@@ -191,8 +192,8 @@ def test_Exceptions(self):
             self.failUnless(issubclass(self.driver.Warning,Exception))
             self.failUnless(issubclass(self.driver.Error,Exception))
         else:
-            self.failUnless(issubclass(self.driver.Warning,StandardError))
-            self.failUnless(issubclass(self.driver.Error,StandardError))
+            self.failUnless(issubclass(self.driver.Warning,StandardError))  # type: ignore[name-defined]
```
dvarrazzo, line 195: My intention was to not fix this file, which is a sort of historical drop-in. But if we have to work on it maybe it's better to do it properly, in case mypy complains? The Python 2 branch might be probably just removed, and I'm thinking about the deprecated method names.

dlax: I wasn't sure about this file, so went on with the least invasive changes for now.

dvarrazzo:

PR #136, opened 2021-11-02
Let mypy check tests/pool code

dlax:


dvarrazzo:

PR #137, opened 2021-11-02
Let mypy check tests/scripts code

dlax:
Along the way, fix the type annotation of cursor.fetchmany() to allow an `Iterable` as `params` (as used in `dectest.py` already).


dvarrazzo:

PR #140, opened 2021-11-04
Let mypy check tests code

dlax:
The goal is to have the whole `tests/` directory type-checked.
I'll add commits as I make progress on this task. Happy to get feedback on them along the way.

### DIFF
--- a/psycopg/setup.py
+++ b/psycopg/setup.py
@@ -47,6 +47,7 @@ extras_require = {
     # Requirements needed for development
     "dev": [
         "black",
+        "dnspython ~= 2.1.0",
         "flake8 ~= 4.0.1",
         "mypy >= 0.910",
         "pytest-mypy >= 0.8.1",
--- a/tests/test_dns.py
+++ b/tests/test_dns.py
@@ -2,6 +2,7 @@ import pytest
 
 import psycopg
 from psycopg.conninfo import conninfo_to_dict
+from psycopg.rows import Row
 
 pytestmark = [pytest.mark.dns]
 
@@ -131,7 +132,7 @@ async def test_resolve_hostaddr_conn(monkeypatch, fake_resolve):
     )
 
     # TODO: not enabled by default, but should be usable to make a subclass
-    class AsyncDnsConnection(psycopg.AsyncConnection):
+    class AsyncDnsConnection(psycopg.AsyncConnection[Row]):
         @classmethod
         async def _get_connection_params(cls, conninfo, **kwargs):
             params = await super()._get_connection_params(conninfo, **kwargs)
--- a/tests/test_dns.py
+++ b/tests/test_dns.py
@@ -47,7 +47,9 @@ async def test_resolve_hostaddr_async_no_resolve(
         for k, v in env.items():
             monkeypatch.setenv(k, v)
     params = conninfo_to_dict(conninfo)
-    params = await psycopg._dns.resolve_hostaddr_async(params)
+    params = await psycopg._dns.resolve_hostaddr_async(  # type: ignore[attr-defined]
+        params
+    )
     assert conninfo_to_dict(want) == params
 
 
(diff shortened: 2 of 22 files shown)


tests/fix_dns.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def psycopg_dns():
```
dvarrazzo, line 5: Probably it's ok to have these fixtures in the `fix_psycopg` module.

tests/test_dns.py
```diff
@@ -40,13 +40,13 @@
...
 async def test_resolve_hostaddr_async_no_resolve(
-    monkeypatch, conninfo, want, env, fail_resolve
+    monkeypatch, conninfo, want, env, fail_resolve, psycopg_dns
 ):
     if env:
         for k, v in env.items():
             monkeypatch.setenv(k, v)
     params = conninfo_to_dict(conninfo)
-    params = await psycopg._dns.resolve_hostaddr_async(params)
+    params = await psycopg_dns.resolve_hostaddr_async(params)
```
dvarrazzo, line 50: Honestly I don't understand why this is reported as error by mypy, adding the import in this function:

```python
    from psycopg._dns import resolve_hostaddr_async
```
which says that the module doesn't have that function. However the dependencies are installed and the function works ok in this environment.

It seems a mypy problem, and it takes quite some work to work around it, no?

tests/test_dns.py
```diff
@@ -117,7 +117,7 @@ async def test_resolve_hostaddr_async_bad(
         for k, v in env.items():
             monkeypatch.setenv(k, v)
     params = conninfo_to_dict(conninfo)
-    with pytest.raises((TypeError, psycopg.Error)):
+    with pytest.raises(psycopg.Error):
```
dvarrazzo, line 120: Maybe I had some test in the past raising a legit TypeError but they didn't make to the commit. I can't remember.

psycopg/setup.py
```diff
@@ -47,6 +47,7 @@
     # Requirements needed for development
     "dev": [
         "black",
+        "dnspython ~= 2.1.0",
```
dvarrazzo, line 50: I think this would make more difficult than expected to set up the test environment. Not so much for this module, but shapely might need some external dependencies to come up. Honestly it's a direction I'd prefer not to go.

tests/test_dns.py
dlax: You're right. In fact, I misread the error message initially and thought that the problem was that `psycopg._dns` was not imported in test files (so, that `psycopg` was missing the `_dns` attribute).

Just spend an hour investigating the mypy issue, but it seems tricky, so I'd rather add some `# type: ignore[attr-defined]` and possibly report an issue.

psycopg/setup.py
dlax: I think shapely would not need to be added here because it is not typed (in contrast with dnspython) and thus already has an override defined in  pyproject.toml. Are we fine then? Alternatively, I can add this to `tox.ini` only in the `mypy` environment.

tests/fix_dns.py
dlax: Dropped the fixture, no needed now with `# type: ignore[attr-defined]`.

dvarrazzo:
Heads up: https://github.com/psycopg/psycopg/commit/1a5b09b2533fc4000c01699d4572066e602eb0d2 

I have seen you ifversion the import of TypedDict, but it's only for tests and not for the psycopg code so I'm not 100% sure it's worth to have it in `psycopg._compat`.

dlax:
> I have seen you ifversion the import of TypedDict, but it's only for 
> tests and not for the psycopg code so I'm not 100% sure it's worth to 
> have it in |psycopg._compat|.

Yes, I refrained from adding this in _compat for this reason.

dlax:
(Rebased to resolve conflicts.)

psycopg/setup.py
dvarrazzo: Not entirely sold on it, but ok...

dvarrazzo:
This work was massive! Thank you very much!

dvarrazzo:
Rebased and merged.

Issue #124, opened 2021-10-22
Publish manylinux2014 wheels


rafmagns-skepa-dreag:
Hi Daniele, thanks so much for your hard work on this! Glad to see it out of beta! I work for a company that's still on CentOS 7 (and will be for a while) which means that we can't currently install any of the wheels from PyPI because they're all manylinux_2_24 (we're on libc 2.17). Would you consider publishing manylinux2014 wheels as well?

rafmagns-skepa-dreag:
Sorry to disappear for a month. Been traveling and working on other things. I see that psycopg includes a far more comprehensive matrix for wheels than psycopg2. Would you rather I limit the manylinux2014 build to x86 and i686 or add it as another item in the matrix which will push the number of linux wheels from 40 up to 80. I'm testing out the image right now to make sure the build works as expected

dvarrazzo:
Hello Richard, and thank you very much, no need to apologise.

> limit the manylinux2014 build to x86 and i686

This would be better, yes: a psycopg release is already in the order of 150Mb on PyPI.

Could you please let me know, in the Centos image used for the build:
- what libpq library is available? Ideally it should have 14.1, the important is that it has one of the versions [not affected by CVE-2021-23222](https://www.postgresql.org/about/news/postgresql-141-135-129-1114-1019-and-9624-released-2349/) (14.1, 13.5, 12.9, 11.14, 10.19, and 9.6.24). If not, we need to compile the libpq from source.
- what openssl library is available on the system? 1.1.x or 1.0.x? 1.0 has been long untested and has concurrency problems that have affected psycopg2 for long time. I don't think people had problems with 1.1.x. This should be built from source too if not available.
- There is a `build_libpq.sh` script in psycopg2 `scripts/build` directory which you are free to use and should probably work mostly out-of-the-box.
- If possible, it would be amazing if you could also fix the sysconfdir in order to solve https://github.com/psycopg/psycopg2/issues/1365 - I assume it's `./configure --sysconfdir=/etc/postgresql-common` at libpq configure time. Or, probably, the value used for Centos would be more appropriate... What does `pg_config --sysconfdir` tells you on Centos?

Cheers!

rafmagns-skepa-dreag:
Looks like we'll have to build openssl (1.0.2 installed) and libpq (none installed, 5.5 available from postgres 9.2.24 in the default repositories). I'll check out the sysconfig stuff though I'm not confident about getting a real answer. 

[...]

PR #141, opened 2021-11-08
Update cibuildwheel to v2.2.2 to build `musllinux` wheels

lithammer:
The v2.2.x version of cibuildwheel adds supports for building `musllinux`[^1] wheels. Meaning Alpine users can finally take advantage of binary installations.

[^1]: https://www.python.org/dev/peps/pep-0656/

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -93,6 +93,7 @@ jobs:
       matrix:
         arch: [x86_64, i686, ppc64le, aarch64]
         pyver: [cp36, cp37, cp38, cp39, cp310]
+        platform: [manylinux, musllinux]
 
     steps:
       - uses: actions/checkout@v2
@@ -112,7 +113,7 @@ jobs:
           CIBW_MANYLINUX_I686_IMAGE: manylinux_2_24
           CIBW_MANYLINUX_AARCH64_IMAGE: manylinux_2_24
           CIBW_MANYLINUX_PPC64LE_IMAGE: manylinux_2_24
-          CIBW_BUILD: ${{matrix.pyver}}-manylinux_${{matrix.arch}}
+          CIBW_BUILD: ${{matrix.pyver}}-${{matrix.platform}}_${{matrix.arch}}
           CIBW_ARCHS_LINUX: auto aarch64 ppc64le
           CIBW_BEFORE_ALL_LINUX: ./tools/build/wheel_linux_before_all.sh
           CIBW_REPAIR_WHEEL_COMMAND: >-
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -104,7 +104,7 @@ jobs:
         run: python3 ./tools/build/copy_to_binary.py
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.1.2
+        uses: pypa/cibuildwheel@v2.2.2
         with:
           package-dir: psycopg_binary
         env:
@@ -167,7 +167,7 @@ jobs:
         run: python3 ./tools/build/copy_to_binary.py
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.1.2
+        uses: pypa/cibuildwheel@v2.2.2
         with:
           package-dir: psycopg_binary
         env:
@@ -211,7 +211,7 @@ jobs:
         run: python3 ./tools/build/copy_to_binary.py
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.1.2
+        uses: pypa/cibuildwheel@v2.2.2
         with:
           package-dir: psycopg_binary
         env:
--- a/tools/build/wheel_linux_before_all.sh
+++ b/tools/build/wheel_linux_before_all.sh
@@ -6,23 +6,33 @@
 set -euo pipefail
 set -x
 
-dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+. /etc/os-release
 
-# Install libpq
-# Note that the pgdg doesn't have an aarch64 repository so wheels are build
-# with the libpq packaged with Debian 9, which is 9.6.
-if [[ ! "$AUDITWHEEL_ARCH" = "aarch64" ]]; then
-    source /etc/os-release
-    echo "deb http://apt.postgresql.org/pub/repos/apt ${VERSION_CODENAME}-pgdg main" \
-        > /etc/apt/sources.list.d/pgdg.list
-    # TODO: on 2021-09-30 curl fails with
-    # curl: (60) SSL certificate problem: certificate has expired
-    # Test again later if -k can be removed.
-    curl -sk https://www.postgresql.org/media/keys/ACCC4CF8.asc \
-        | apt-key add -
-fi
-
-apt-get update
-
-# zip is required by strip_wheel.sh
-apt-get -y install libpq-dev zip
+# Install PostgreSQL development files.
+case "$ID" in
+    alpine)
+        # zip is required by strip_wheel.sh
+        # tzdata is required for datetime tests.
+        apk add --no-cache postgresql-dev tzdata zip
+        ;;
+    debian)
+        # Note that the pgdg doesn't have an aarch64 repository so wheels are
+        # build with the libpq packaged with Debian 9, which is 9.6.
+        if [ "$AUDITWHEEL_ARCH" != 'aarch64' ]; then
+            echo "deb http://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main" \
+                > /etc/apt/sources.list.d/pgdg.list
+            # TODO: On 2021-11-09 curl fails on 'ppc64le' with:
+            #   curl: (60) SSL certificate problem: certificate has expired
+            # Test again later if -k can be removed.
+            curl -skf https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+                > /etc/apt/trusted.gpg.d/postgresql.asc
+        fi
+        apt-get update
+        # zip is required by strip_wheel.sh
+        apt-get -y install libpq-dev zip
+        ;;
+    *)
+        echo "Unexpected Linux distribution: '$ID'" >&2
+        exit 1
+        ;;
+esac


lithammer:
Converting to draft since I forgot to update `tools/build/wheel_linux_before_all.sh`.

lithammer:
Or actually, maybe it's not necessary...? 🤔

lithammer:
@dvarrazzo I'm a bit perplexed how the `tools/build/wheel_linux_before_all.sh` even works? According to the cibuildwheel documentation, the `CIBW_BEFORE_ALL_LINUX` script runs _inside_ a container, not the host system:

> This option is very useful for the Linux build, where builds take place in isolated Docker containers managed by cibuildwheel. This command will run inside the container before the wheel builds start.
>
> – https://cibuildwheel.readthedocs.io/en/stable/options/#before-all

The pinned images can be found [here](https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/resources/pinned_docker_images.cfg). And if you pull one of them and start, you'll notice it's a CentOS 7 image, not Debian. So the `wheel_linux_before_all.sh` shouldn't work at all given that it tries to run `apt-get` and similar things.

```console
$ docker run --rm -it --entrypoint bash quay.io/pypa/manylinux2014_x86_64:2021-10-10-790306f
[root@306b13e1d126 /]# cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

Are you sure this works correct in `master`?

lithammer:
Ah, after further digging I see that `manylinux_2_24` is using Debian 9. So that explains that. GH-124 also contains some background as we to why `manylinux_2_24` is used over `manylinux2010` and `manylinux2014`.

tools/build/wheel_linux_before_all.sh
```diff
@@ -1,25 +1,27 @@
-#!/bin/bash
+#!/bin/sh
```
lithammer, line 1: Normally Bash isn't installed out-of-the-box on Alpine images. But it seems like the `quay.io/pypa/musllinux_*` has it. I could revert back to `/bin/bash` if you want? Otherwise this script is POSIX now, which is probably a good thing since it needs to run on multiple platforms.

tools/build/wheel_linux_before_all.sh
```diff
@@ -1,25 +1,27 @@
...
-# Install libpq
-# Note that the pgdg doesn't have an aarch64 repository so wheels are build
-# with the libpq packaged with Debian 9, which is 9.6.
-if [[ ! "$AUDITWHEEL_ARCH" = "aarch64" ]]; then
-    source /etc/os-release
-    echo "deb http://apt.postgresql.org/pub/repos/apt ${VERSION_CODENAME}-pgdg main" \
-        > /etc/apt/sources.list.d/pgdg.list
-    # TODO: on 2021-09-30 curl fails with
-    # curl: (60) SSL certificate problem: certificate has expired
-    # Test again later if -k can be removed.
```
lithammer, line 20: I've verified that this is no longer necessary.

dvarrazzo:
Hello, thank you for this

I have just released 3.0.2, and just noticed that the binary packages are very bloated and experimenting with stripping the debug symbols. For a linux package it seems the download size shrink around 30% and installed file > 90%.

Could you look into that for the musl packages too please? I'll keep you up to date.

tools/build/wheel_linux_before_all.sh
```diff
@@ -1,25 +1,27 @@
...
+    alpine)
+        apk add --no-cache postgresql-dev
+        ;;
+    debian)
+        # Note that the pgdg doesn't have an aarch64 repository so wheels are
+        # build with the libpq packaged with Debian 9, which is 9.6.
+        if [ "$AUDITWHEEL_ARCH" != 'aarch64' ]; then
+            echo "deb http://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main" \
+                > /etc/apt/sources.list.d/pgdg.list
+            curl -sf https://www.postgresql.org/media/keys/ACCC4CF8.asc \
```
dvarrazzo, line 21: What is `-f` for? It seems that it makes curl fail silently. Don't we want to fail noisely?

tools/build/wheel_linux_before_all.sh
```diff
@@ -1,25 +1,27 @@
...
-    # Test again later if -k can be removed.
-    curl -sk https://www.postgresql.org/media/keys/ACCC4CF8.asc \
-        | apt-key add -
-fi
-apt-get update
-apt-get -y install libpq-dev
+# Install PostgreSQL development files.
+case "$ID" in
+    alpine)
+        apk add --no-cache postgresql-dev
```
dvarrazzo, line 13: What libpq version do you get?

tools/build/wheel_linux_before_all.sh
dvarrazzo: It seems like you have dropped the checks for undeclared variables and failing pipes. The script contains variables and pipes. If you don't mind I would prefer it to run in safe mode :)

tools/build/wheel_linux_before_all.sh
```diff
@@ -1,25 +1,27 @@
...
+        if [ "$AUDITWHEEL_ARCH" != 'aarch64' ]; then
+            echo "deb http://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main" \
+                > /etc/apt/sources.list.d/pgdg.list
+            curl -sf https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+                > /etc/apt/trusted.gpg.d/postgresql.asc
+        fi
+        apt-get update
+        apt-get -y install libpq-dev
+        ;;
+esac
```
dvarrazzo, line 27: I think we would like an else case to explode with incommensurable noise here

dvarrazzo (changes requested):

tools/build/wheel_linux_before_all.sh
lithammer: The documentation is actually poorly phrased here in my opinion. Normally, curl only exists with non-zero if there's a network related error. But if the server returns a 4xx/5xx status code, that's considered "ok".

What the `-f` flag does is change this behaviour, so a 4xx/5xx (maybe others as well?) results in the exit code `22`.

```console
$ curl -i --http1.1 https://httpbin.org/status/500
HTTP/1.1 500 INTERNAL SERVER ERROR
Date: Mon, 08 Nov 2021 16:33:28 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 0
Connection: keep-alive
Server: gunicorn/19.9.0
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true

$ echo $?
0
```
```console
$ curl -if --http1.1 https://httpbin.org/status/500
HTTP/1.1 500 INTERNAL SERVER ERROR
Date: Mon, 08 Nov 2021 16:33:37 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 0
Connection: keep-alive
Server: gunicorn/19.9.0
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true

curl: (22) The requested URL returned error: 500
$ echo $?
22
```

tools/build/wheel_linux_before_all.sh
lithammer: 12.8.

```console
# apk info postgresql-dev
postgresql-dev-12.8-r0 description:
A sophisticated object-relational DBMS (development files)

postgresql-dev-12.8-r0 webpage:
https://www.postgresql.org/

postgresql-dev-12.8-r0 installed size:
9117696
```

tools/build/wheel_linux_before_all.sh
lithammer: Is this good enough?

```diff
diff --git a/tools/build/wheel_linux_before_all.sh b/tools/build/wheel_linux_before_all.sh
index 11570d6..a7fd911 100755
--- a/tools/build/wheel_linux_before_all.sh
+++ b/tools/build/wheel_linux_before_all.sh
@@ -24,4 +24,8 @@ case "$ID" in
         apt-get update
         apt-get -y install libpq-dev
         ;;
+    *)
+        echo "Unknown Linux distribution: '$ID'" >&2
+        exit 1
+        ;;
 esac
```

tools/build/wheel_linux_before_all.sh
lithammer: I can bring back `uo` (shouldn't have removed those!). But I can't add back `pipefail` without reverting back to Bash. Which is why I changed the code to not use any pipes. But it's your call!

dvarrazzo: I prefer to have pipefail over useing sh.

tools/build/wheel_linux_before_all.sh
dvarrazzo: Ok, that seems more sensible indeed.

tools/build/wheel_linux_before_all.sh
dvarrazzo: "Unexpected" is more correct than "Incorrect", for the rest yes, that will do :)

tools/build/wheel_linux_before_all.sh
lithammer: Fair enough!

tools/build/wheel_linux_before_all.sh
dvarrazzo: Let's go with it for the moment. Getting v14 would be better in the near future (3.1 should ship with some features depending on it, which would be unavailable in these packages otherwise).

lithammer: Unfortunately even the latest Alpine version (3.14) only has PostgreSQL 13. So even if [pypa/manylinux](https://github.com/pypa/manylinux) updates the base image from Alpine 3.12 to 3.14, it won't get us to PostgreSQL 14.

I'm not super well versed in Alpine's release cycle. But I suspect we'd have to wait for Alpine 3.16 for PostgreSQL 14 (or newer).

lithammer:
> Hello, thank you for this
> 
> I have just released 3.0.2, and just noticed that the binary packages are very bloated and experimenting with stripping the debug symbols. For a linux package it seems the download size shrink around 30% and installed file > 90%.
> 
> Could you look into that for the musl packages too please? I'll keep you up to date.

This seems related: https://github.com/pypa/cibuildwheel/issues/331

tools/build/wheel_linux_before_all.sh
lithammer: Did at bit of experimenting, and it seems like it's possible to install packages from Alpine's 'edge' repository (presumably their upcoming 3.15 release). And it includes `postgresql-dev-14.0-r6`.

So it worst case scenario, that could probably be hacked together for a psycopg 3.1 release. But until then it's probably best to stay with the 'main' repositoy since using 'edge' is likely a bit brittle.

```console
$ apk add --repository https://dl-cdn.alpinelinux.org/alpine/edge/main 'postgresql-dev>=14'
...
(21/21) Installing postgresql-dev (14.0-r6)
Executing busybox-1.31.1-r20.trigger
OK: 322 MiB in 33 packages
```

dvarrazzo: I think it's ok to wait for when those package will become stable :)

dvarrazzo:
Just merged to master script to reduce binary package size. Please rebase and it would be nice to know if it's beneficial for these builds too.

lithammer:
> Just merged to master script to reduce binary package size. Please rebase and it would be nice to know if it's beneficial for these builds too.

It seems like it's _very_ beneficial, unless I did something wrong. But basically I copied your `strip_wheel.sh` and built a psycopg-c wheel by hand:

```console
$ cd psycopg_c
$ python3 setup.py bdist_wheel
$ ./strip_wheel.sh dist/psycopg_c-3.0.2-cp38-cp38-linux_x86_64.whl
...
```

This reduced the wheel size from 3594967 → 489219 (3.4M → 477.8K).

lithammer:
I've rebased no as well. And ensured that `zip` is installed in the Alpine image as well (because of `strip_wheel.sh`).

tools/build/wheel_linux_before_all.sh
dvarrazzo: I am not sure this is working as expected: https://github.com/psycopg/psycopg/runs/4145304671?check_suite_focus=true#step:6:216

dvarrazzo:
I am trying to run a workflow but it doesn't seem packages are being created: https://github.com/psycopg/psycopg/actions/runs/1436953943

Maybe cibuildwheel need to be configured explicitly?

dvarrazzo (commented):
Disregard this message: just trying to dismiss github being a nuisance.

lithammer:
> I am trying to run a workflow but it doesn't seem packages are being created: https://github.com/psycopg/psycopg/actions/runs/1436953943
> 
> Maybe cibuildwheel need to be configured explicitly?

I suspect maybe it's the `CIBW_BUILD` environment variable. I think maybe something like this is needed:

```diff
diff --git a/.github/workflows/packages.yml b/.github/workflows/packages.yml
index 29cd0cb..76fd3ca 100644
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -93,6 +93,7 @@ jobs:
       matrix:
         arch: [x86_64, i686, ppc64le, aarch64]
         pyver: [cp36, cp37, cp38, cp39, cp310]
+        platform: [manylinux, musllinux]
 
     steps:
       - uses: actions/checkout@v2
@@ -112,7 +113,7 @@ jobs:
           CIBW_MANYLINUX_I686_IMAGE: manylinux_2_24
           CIBW_MANYLINUX_AARCH64_IMAGE: manylinux_2_24
           CIBW_MANYLINUX_PPC64LE_IMAGE: manylinux_2_24
-          CIBW_BUILD: ${{matrix.pyver}}-manylinux_${{matrix.arch}}
+          CIBW_BUILD: ${{matrix.pyver}}-${{matrix.platform}}_${{matrix.arch}}
           CIBW_ARCHS_LINUX: auto aarch64 ppc64le
           CIBW_BEFORE_ALL_LINUX: ./tools/build/wheel_linux_before_all.sh
           CIBW_REPAIR_WHEEL_COMMAND: >-
```

lithammer:
I pushed the above change. Could you try again and see if wheels are being properly produced?

.github/workflows/packages.yml
```diff
@@ -93,6 +93,7 @@ jobs:
       matrix:
         arch: [x86_64, i686, ppc64le, aarch64]
         pyver: [cp36, cp37, cp38, cp39, cp310]
+        platform: [manylinux, musllinux]
```
lithammer, line 96: Honestly don't know what to call this. I picked "platform" from [PEP 656](https://www.python.org/dev/peps/pep-0656/) 🤷‍♂️

dvarrazzo:
Building: https://github.com/psycopg/psycopg/actions/runs/1439615327

Why, can't you build it in your clone? There should be a workflow "build packages"  which you can trigger manually. Or alternatively change the "on: workflow_dispach" to "on: push" in `packages.yml`

lithammer:
> Building: https://github.com/psycopg/psycopg/actions/runs/1439615327
> 
> Why, can't you build it in your clone? There should be a workflow "build packages" which you can trigger manually. Or alternatively change the "on: workflow_dispach" to "on: push" in `packages.yml`

Riiight. I didn't consider that 😓 I'll do that if more experimentation is required.

tools/build/wheel_linux_before_all.sh
lithammer: I take this back. It's still a problem on the ppc64le architecture.

lithammer:
So two things from the [latest run ](https://github.com/psycopg/psycopg/actions/runs/1439615327):

1. The unit tests fail on Alpine. I feel like I'm a bit out of my depth there. Seems related to `datetime`.
2. The SSL cert problem is still present on the `ppc64le` arch. So that needs to be added back.

dvarrazzo:
> The unit tests fail on Alpine. I feel like I'm a bit out of my depth there. Seems related to `datetime`.

From the look of it it lacks the Olson timezone database. Maybe an `apk add --no-cache tzdata` would solve it?

lithammer:
> > The unit tests fail on Alpine. I feel like I'm a bit out of my depth there. Seems related to `datetime`.
> 
> From the look of it it lacks the Olson timezone database. Maybe an `apk add --no-cache tzdata` would solve it?

Yes that solves the test issue, thanks!

.github/workflows/packages.yml
dvarrazzo: Seems ok, and it's a local name anyway. If we find a better naming it will be easy to fix :+1:

lithammer:
And now we finally have wheels! 🎉 

https://github.com/lithammer/psycopg/actions/runs/1440255644

dvarrazzo:
Congratulations! :smile: 

I will bump up the dev version and upload them on testpypi to verify they download ok, then we can release them in 3.0.3.

dvarrazzo:
Looking good!

```
$ docker run --rm -ti python:3-alpine sh
# pip install -i https://test.pypi.org/simple/ psycopg[binary]
...
Successfully installed psycopg-3.0.3.dev1 psycopg-binary-3.0.3.dev1

# python
Python 3.10.0 (default, Oct 26 2021, 23:20:47) [GCC 10.3.1 20210424] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import psycopg
>>> psycopg.connect("host=172.19.0.1 user=piro dbname=psycopg3_test").execute("select now()").fetchone()[0]
datetime.datetime(2021, 11, 9, 23, 15, 39, 330955, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
```

Thank you very much!

lithammer:
Excellent! 🥳 Thanks for taking your time to review!

PR #143, opened 2021-11-09
Update dep on pytest-asyncio, works fine with 0.16.0

mikelolasagasti:


psycopg/setup.py
```diff
@@ -40,7 +40,7 @@
         "mypy >= 0.910",
         "pproxy ~= 2.7.8",
         "pytest ~= 6.2.5",
-        "pytest-asyncio ~= 0.15.1",
+        "pytest-asyncio >= 0.15.1, <0.17.0",
```
dvarrazzo, line 43: I would prefer to use a single minor release, because we might start using something from 0.16 and fail to notice that 0.15 doesn't work anymore.

How about... ?

```suggestion
        "pytest-asyncio ~= 0.16.0",
```

dvarrazzo (changes requested):

psycopg/setup.py
mikelolasagasti: Fine for me. I was testing with 0.16.0 in Fedora and as it was working with both versions, I just extended to add 0.16.0.

dvarrazzo:
Cool, thank you!

Issue #29, opened 2021-01-23
Trio support


jab:
From https://www.psycopg.org/psycopg3/docs/async.html it looks like the async support is tightly coupled to asyncio. Would you be willing to use anyio so that Trio users could also use psycopg3 async?

Thanks for your consideration and for your work on psycopg3!

dvarrazzo:
Of course non-blocking psycopg3 is built on asyncio: asyncio is the primitive Python standard way to do write concurrent I/O bound code. Why it should be tightly coupled to anything else?

[I tried anyio a few months ago](https://github.com/psycopg/psycopg2/issues/1057): I found it buggy, not up to the task, and adding quite some overhead. Surely not ready for its prime, definitely not a foundation to build on.

Just this morning, in order to answer somebody else's email, I took a look at Trio and I just can't understand what is its point. It claims to allow to call asyncio code from sync code. I struggle to understand this use case; however, if Trio does allow to call asyncio code, why can't it call asyncio psycopg3? If people use trio because they don't like async/await keywords, why don't they use multithread- or greenthread- libraries?

psycopg3 supports blocking/multithread code, can be monkeypatched by eventled and gevent, and supports asyncio: does it really need anything on top of it?

Furthermore psycopg3 is built around a core of non-blocking functionalities, implemented as generators, yielding when it's time to block and maintaining the objects state (BaseConnection, BaseCursor...): on top of them the public interfaces are built in a blocking variant (Connection, Cursor) and an asyncio variant (AsyncConnection, AsyncCursor): which are just thin layer describing how to block. From my reply this morning, they may look this way:

```python
class BaseCursor:
    def _execute_gen(self, query, params) -> PQgen[Resutl]:
        # a lot of work, yielding on wait

class Cursor(BaseCursor):
    def execute(self, query, params): -> Result
        with self._lock: # a sync lock
            return self._wait(self.execute_gen(query, params))  # wait based on epoll

class AsyncCursor(BaseCursor):
    async def execute(self, query, params) -> Result:
        async with self._lock:  # an async lock
            return await self._wait(self.execute_gen(query, params))  # wait based on the asyncio loop
```

[...]

PR #146, opened 2021-11-11
use anyio in "core" psycopg

dlax:
This adds support for anyio in psycopg "core" (i.e. not the psycopg_pool) through an `AnyIOConnection` class, associated waiting functions ~~and a custom `AnyIOLibpqWriter` implementation~~.

This uses conditional imports and pulls no explicit dependency on anyio.

The test suite now uses [anyio pytest plugin](https://anyio.readthedocs.io/en/stable/testing.html), instead of pytest-asyncio (this change has beed submitted as another PR #352, which this one is based on).

ticket #29

### DIFF
--- a/psycopg/psycopg/__init__.py
+++ b/psycopg/psycopg/__init__.py
@@ -25,6 +25,7 @@ from .cursor_async import AsyncCursor
 from .server_cursor import AsyncServerCursor, ServerCursor
 from .client_cursor import AsyncClientCursor, ClientCursor
 from .connection_async import AsyncConnection
+from ._anyio.connection import AnyIOConnection
 
 from . import dbapi20
 from .dbapi20 import BINARY, DATETIME, NUMBER, ROWID, STRING
@@ -59,6 +60,7 @@ types.array.register_all_arrays(adapters)
 # this is the canonical place to obtain them and should be used by MyPy too,
 # so that function signatures are consistent with the documentation.
 __all__ = [
+    "AnyIOConnection",
     "AsyncClientCursor",
     "AsyncConnection",
     "AsyncCopy",
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -189,7 +189,8 @@ class AsyncConnection(BaseConnection[Row]):
             params["connect_timeout"] = None
 
         # Resolve host addresses in non-blocking way
-        params = await resolve_hostaddr_async(params)
+        loop = asyncio.get_running_loop()
+        params = await resolve_hostaddr_async(params, getaddrinfo=loop.getaddrinfo)
 
         return params
 
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -345,7 +345,7 @@ class AsyncConnection(BaseConnection[Row]):
 
     async def wait(self, gen: PQGen[RV]) -> RV:
         try:
-            return await waiting.wait_async(gen, self.pgconn.socket)
+            return await waiting.wait_asyncio(gen, self.pgconn.socket)
         except KeyboardInterrupt:
             # TODO: this doesn't seem to work as it does for sync connections
             # see tests/test_concurrency_async.py::test_ctrl_c
@@ -356,14 +356,14 @@ class AsyncConnection(BaseConnection[Row]):
             c = self.pgconn.get_cancel()
             c.cancel()
             try:
-                await waiting.wait_async(gen, self.pgconn.socket)
+                await waiting.wait_asyncio(gen, self.pgconn.socket)
             except e.QueryCanceled:
                 pass  # as expected
             raise
 
     @classmethod
     async def _wait_conn(cls, gen: PQGenConn[RV], timeout: Optional[int]) -> RV:
-        return await waiting.wait_conn_async(gen, timeout)
+        return await waiting.wait_conn_asyncio(gen, timeout)
 
     def _set_autocommit(self, value: bool) -> None:
         self._no_set_async("autocommit")
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -53,6 +53,7 @@ class AsyncConnection(BaseConnection[Row]):
     row_factory: AsyncRowFactory[Row]
     _pipeline: Optional[AsyncPipeline]
     _Self = TypeVar("_Self", bound="AsyncConnection[Any]")
+    _lockcls = asyncio.Lock
 
     def __init__(
         self,
@@ -61,9 +62,13 @@ class AsyncConnection(BaseConnection[Row]):
     ):
         super().__init__(pgconn)
         self.row_factory = row_factory
-        self.lock = asyncio.Lock()
+        self.lock = self._lockcls()
         self.cursor_factory = AsyncCursor
         self.server_cursor_factory = AsyncServerCursor
+
+    @staticmethod
+    def _async_library() -> str:
+        return "asyncio"
 
     @overload
     @classmethod
@@ -107,7 +112,7 @@ class AsyncConnection(BaseConnection[Row]):
         cursor_factory: Optional[Type[AsyncCursor[Row]]] = None,
         **kwargs: Any,
     ) -> "AsyncConnection[Any]":
-        if sys.platform == "win32":
+        if sys.platform == "win32" and cls._async_library() == "asyncio":
             loop = asyncio.get_running_loop()
             if isinstance(loop, asyncio.ProactorEventLoop):
                 raise e.InterfaceError(
@@ -167,6 +172,11 @@ class AsyncConnection(BaseConnection[Row]):
         if not getattr(self, "_pool", None):
             await self.close()
 
+    @staticmethod
+    def _getaddrinfo() -> Any:
+        loop = asyncio.get_running_loop()
+        return loop.getaddrinfo
+
     @classmethod
     async def _get_connection_params(
         cls, conninfo: str, **kwargs: Any
@@ -189,8 +199,7 @@ class AsyncConnection(BaseConnection[Row]):
             params["connect_timeout"] = None
 
         # Resolve host addresses in non-blocking way
-        loop = asyncio.get_running_loop()
-        params = await resolve_hostaddr_async(params, getaddrinfo=loop.getaddrinfo)
+        params = await resolve_hostaddr_async(params, getaddrinfo=cls._getaddrinfo())
 
         return params
 
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -96,7 +96,7 @@ def wait_conn(gen: PQGenConn[RV], timeout: Optional[float] = None) -> RV:
         return rv
 
 
-async def wait_async(gen: PQGen[RV], fileno: int) -> RV:
+async def wait_asyncio(gen: PQGen[RV], fileno: int) -> RV:
     """
     Coroutine waiting for a generator to complete.
 
@@ -146,7 +146,7 @@ async def wait_async(gen: PQGen[RV], fileno: int) -> RV:
         return rv
 
 
-async def wait_conn_async(gen: PQGenConn[RV], timeout: Optional[float] = None) -> RV:
+async def wait_conn_asyncio(gen: PQGenConn[RV], timeout: Optional[float] = None) -> RV:
     """
     Coroutine waiting for a connection generator to complete.
 
--- a/tests/test_client_cursor_async.py
+++ b/tests/test_client_cursor_async.py
@@ -433,11 +433,13 @@ async def test_rownumber(aconn):
     await cur.fetchmany(10)
     assert cur.rownumber == 12
     rns: List[int] = []
-    async for i in cur:
+    itcur = cur.__aiter__()
+    async for i in itcur:
         assert cur.rownumber
         rns.append(cur.rownumber)
         if len(rns) >= 3:
             break
+    await itcur.aclose()
     assert rns == [13, 14, 15]
     assert len(await cur.fetchall()) == 42 - rns[-1]
     assert cur.rownumber == 42
@@ -455,13 +457,17 @@ async def test_iter(aconn):
 async def test_iter_stop(aconn):
     cur = aconn.cursor()
     await cur.execute("select generate_series(1, 3)")
-    async for rec in cur:
+    itcur = cur.__aiter__()
+    async for rec in itcur:
         assert rec == (1,)
         break
+    await itcur.aclose()
 
-    async for rec in cur:
+    itcur = cur.__aiter__()
+    async for rec in itcur:
         assert rec == (2,)
         break
+    await itcur.aclose()
 
     assert (await cur.fetchone()) == (3,)
     async for rec in cur:
--- a/tests/test_waiting.py
+++ b/tests/test_waiting.py
@@ -6,8 +6,10 @@ import pytest
 
 import psycopg
 from psycopg import waiting
+from psycopg._anyio import waiting as waiting_anyio
 from psycopg import generators
 from psycopg.pq import ConnStatus, ExecStatus
+
 
 skip_if_not_linux = pytest.mark.skipif(
     not sys.platform.startswith("linux"), reason="non-Linux platform"
@@ -113,47 +115,61 @@ def test_wait_large_fd(dsn, waitfn):
             f.close()
 
 
+@pytest.fixture
+def wait_async(anyio_backend_name):
+    return {
+        "asyncio": waiting.wait_asyncio,
+        "trio": waiting_anyio.wait,
+    }[anyio_backend_name]
+
+
+@pytest.fixture
+def wait_conn_async(anyio_backend_name):
+    return {
+        "asyncio": waiting.wait_conn_asyncio,
+        "trio": waiting_anyio.wait_conn,
+    }[anyio_backend_name]
+
+
 @pytest.mark.parametrize("timeout", timeouts)
-@pytest.mark.anyio
-async def test_wait_conn_async(dsn, timeout):
+async def test_wait_conn_async(dsn, timeout, wait_conn_async):
     gen = generators.connect(dsn)
-    conn = await waiting.wait_conn_asyncio(gen, **timeout)
+    conn = await wait_conn_async(gen, **timeout)
     assert conn.status == ConnStatus.OK
 
 
-@pytest.mark.anyio
-async def test_wait_conn_async_bad(dsn):
+async def test_wait_conn_async_bad(dsn, wait_conn_async):
     gen = generators.connect("dbname=nosuchdb")
     with pytest.raises(psycopg.OperationalError):
-        await waiting.wait_conn_asyncio(gen)
+        await wait_conn_async(gen)
 
 
 @pytest.mark.anyio
 @pytest.mark.parametrize("wait, ready", zip(waiting.Wait, waiting.Ready))
 @skip_if_not_linux
-async def test_wait_ready_async(wait, ready):
+async def test_wait_ready_async(wait, ready, wait_async):
     def gen():
         r = yield wait
         return r
 
     with socket.socket() as s:
-        r = await waiting.wait_asyncio(gen(), s.fileno())
+        r = await wait_async(gen(), s.fileno())
     assert r & ready
 
 
 @pytest.mark.anyio
-async def test_wait_async(pgconn):
+async def test_wait_async(pgconn, wait_async):
     pgconn.send_query(b"select 1")
     gen = generators.execute(pgconn)
-    (res,) = await waiting.wait_async(gen, pgconn.socket)
+    (res,) = await wait_async(gen, pgconn.socket)
     assert res.status == ExecStatus.TUPLES_OK
 
 
 @pytest.mark.anyio
-async def test_wait_async_bad(pgconn):
+async def test_wait_async_bad(pgconn, wait_async):
     pgconn.send_query(b"select 1")
     gen = generators.execute(pgconn)
     socket = pgconn.socket
     pgconn.finish()
     with pytest.raises(psycopg.OperationalError):
-        await waiting.wait_asyncio(gen, socket)
+        await wait_async(gen, socket)
--- a/tests/test_waiting.py
+++ b/tests/test_waiting.py
@@ -117,7 +117,7 @@ def test_wait_large_fd(dsn, waitfn):
 @pytest.mark.anyio
 async def test_wait_conn_async(dsn, timeout):
     gen = generators.connect(dsn)
-    conn = await waiting.wait_conn_async(gen, **timeout)
+    conn = await waiting.wait_conn_asyncio(gen, **timeout)
     assert conn.status == ConnStatus.OK
 
 
@@ -125,7 +125,7 @@ async def test_wait_conn_async(dsn, timeout):
 async def test_wait_conn_async_bad(dsn):
     gen = generators.connect("dbname=nosuchdb")
     with pytest.raises(psycopg.OperationalError):
-        await waiting.wait_conn_async(gen)
+        await waiting.wait_conn_asyncio(gen)
 
 
 @pytest.mark.anyio
@@ -137,7 +137,7 @@ async def test_wait_ready_async(wait, ready):
         return r
 
     with socket.socket() as s:
-        r = await waiting.wait_async(gen(), s.fileno())
+        r = await waiting.wait_asyncio(gen(), s.fileno())
     assert r & ready
 
 
@@ -156,4 +156,4 @@ async def test_wait_async_bad(pgconn):
     socket = pgconn.socket
     pgconn.finish()
     with pytest.raises(psycopg.OperationalError):
-        await waiting.wait_async(gen, socket)
+        await waiting.wait_asyncio(gen, socket)
--- a/tests/test_waiting.py
+++ b/tests/test_waiting.py
@@ -116,19 +116,13 @@ def test_wait_large_fd(dsn, waitfn):
 
 
 @pytest.fixture
-def wait_async(anyio_backend_name):
-    return {
-        "asyncio": waiting.wait_asyncio,
-        "trio": waiting_anyio.wait,
-    }[anyio_backend_name]
+def wait_async(use_anyio):
+    return waiting_anyio.wait if use_anyio else waiting.wait_asyncio
 
 
 @pytest.fixture
-def wait_conn_async(anyio_backend_name):
-    return {
-        "asyncio": waiting.wait_conn_asyncio,
-        "trio": waiting_anyio.wait_conn,
-    }[anyio_backend_name]
+def wait_conn_async(use_anyio):
+    return waiting_anyio.wait_conn if use_anyio else waiting.wait_conn_asyncio
 
 
 @pytest.mark.parametrize("timeout", timeouts)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Include dnspython to the packages to install
         if: ${{ matrix.ext == 'dns' }}
         run: |
-          echo "DEPS=$DEPS dnspython" >> $GITHUB_ENV
+          echo "DEPS=$DEPS anyio[trio] dnspython" >> $GITHUB_ENV
           echo "MARKERS=$MARKERS dns" >> $GITHUB_ENV
 
       - name: Include shapely to the packages to install
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,28 +26,29 @@ jobs:
       matrix:
         include:
           # Test different combinations of Python, Postgres, libpq.
-          - {impl: python, python: "3.7", postgres: "postgres:10", libpq: newest}
-          - {impl: python, python: "3.8", postgres: "postgres:12"}
-          - {impl: python, python: "3.9", postgres: "postgres:13"}
-          - {impl: python, python: "3.10", postgres: "postgres:14"}
-          - {impl: python, python: "3.11", postgres: "postgres:15", libpq: oldest}
+          - {impl: python, python: "3.7", postgres: "postgres:10", libpq: newest, pytest_opts: "--anyio=asyncio"}
+          - {impl: python, python: "3.8", postgres: "postgres:12", pytest_opts: ""}
+          - {impl: python, python: "3.9", postgres: "postgres:13", pytest_opts: "--anyio=trio"}
+          - {impl: python, python: "3.10", postgres: "postgres:14", pytest_opts: ""}
+          - {impl: python, python: "3.11", postgres: "postgres:15", libpq: oldest, pytest_opts: ""}
 
-          - {impl: c, python: "3.7", postgres: "postgres:15", libpq: newest}
-          - {impl: c, python: "3.8", postgres: "postgres:13"}
-          - {impl: c, python: "3.9", postgres: "postgres:14"}
-          - {impl: c, python: "3.10", postgres: "postgres:11", libpq: oldest}
-          - {impl: c, python: "3.11", postgres: "postgres:10", libpq: newest}
+          - {impl: c, python: "3.7", postgres: "postgres:15", libpq: newest, pytest_opts: ""}
+          - {impl: c, python: "3.8", postgres: "postgres:13", pytest_opts: "--anyio=asyncio"}
+          - {impl: c, python: "3.9", postgres: "postgres:14", pytest_opts: ""}
+          - {impl: c, python: "3.10", postgres: "postgres:11", libpq: oldest, pytest_opts: "--anyio=trio"}
+          - {impl: c, python: "3.11", postgres: "postgres:10", libpq: newest, pytest_opts: ""}
 
-          - {impl: python, python: "3.9", ext: dns, postgres: "postgres:14"}
-          - {impl: python, python: "3.9", ext: postgis, postgres: "postgis/postgis"}
+          - {impl: python, python: "3.9", ext: dns, postgres: "postgres:14", pytest_opts: ""}
+          - {impl: python, python: "3.9", ext: postgis, postgres: "postgis/postgis", pytest_opts: ""}
 
           # Test with minimum dependencies versions
-          - {impl: c, python: "3.7", ext: min, postgres: "postgres:15"}
+          - {impl: c, python: "3.7", ext: min, postgres: "postgres:15", pytest_opts: ""}
 
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
       PSYCOPG_TEST_DSN: "host=127.0.0.1 user=postgres password=password"
+      PYTEST_ADDOPTS: ${{ matrix.pytest_opts }}
       MARKERS: ""
 
     steps:
@@ -110,21 +111,22 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {impl: python, python: "3.7"}
-          - {impl: python, python: "3.8"}
-          - {impl: python, python: "3.9"}
-          - {impl: python, python: "3.10"}
-          - {impl: python, python: "3.11"}
-          - {impl: c, python: "3.7"}
-          - {impl: c, python: "3.8"}
-          - {impl: c, python: "3.9"}
-          - {impl: c, python: "3.10"}
-          - {impl: c, python: "3.11"}
+          - {impl: python, python: "3.7", pytest_opts: ""}
+          - {impl: python, python: "3.8", pytest_opts: "--anyio=asyncio"}
+          - {impl: python, python: "3.9", pytest_opts: ""}
+          - {impl: python, python: "3.10", pytest_opts: "--anyio=trio"}
+          - {impl: python, python: "3.11", pytest_opts: ""}
+          - {impl: c, python: "3.7", pytest_opts: "--anyio=trio"}
+          - {impl: c, python: "3.8", pytest_opts: ""}
+          - {impl: c, python: "3.9", pytest_opts: "--anyio=asyncio"}
+          - {impl: c, python: "3.10", pytest_opts: ""}
+          - {impl: c, python: "3.11", pytest_opts: ""}
 
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
       PSYCOPG_TEST_DSN: "host=127.0.0.1 user=runner dbname=postgres"
+      PYTEST_ADDOPTS: ${{ matrix.pytest_opts }}
       # MacOS on GitHub Actions seems particularly slow.
       # Don't run timing-based tests as they regularly fail.
       # pproxy-based tests fail too, with the proxy not coming up in 2s.
@@ -165,21 +167,22 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {impl: python, python: "3.7"}
-          - {impl: python, python: "3.8"}
-          - {impl: python, python: "3.9"}
-          - {impl: python, python: "3.10"}
-          - {impl: python, python: "3.11"}
-          - {impl: c, python: "3.7"}
-          - {impl: c, python: "3.8"}
-          - {impl: c, python: "3.9"}
-          - {impl: c, python: "3.10"}
-          - {impl: c, python: "3.11"}
+          - {impl: python, python: "3.7", pytest_opts: "--anyio=asyncio"}
+          - {impl: python, python: "3.8", pytest_opts: ""}
+          - {impl: python, python: "3.9", pytest_opts: "--anyio=trio"}
+          - {impl: python, python: "3.10", pytest_opts: ""}
+          - {impl: python, python: "3.11", pytest_opts: ""}
+          - {impl: c, python: "3.7", pytest_opts: ""}
+          - {impl: c, python: "3.8", pytest_opts: "--anyio=trio"}
+          - {impl: c, python: "3.9", pytest_opts: ""}
+          - {impl: c, python: "3.10", pytest_opts: "--anyio=asyncio"}
+          - {impl: c, python: "3.11", pytest_opts: ""}
 
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
       PSYCOPG_TEST_DSN: "host=127.0.0.1 dbname=postgres"
+      PYTEST_ADDOPTS: ${{ matrix.pytest_opts }}
       # On windows pproxy doesn't seem very happy. Also a few timing test fail.
       NOT_MARKERS: "timing proxy mypy"
 
@@ -235,12 +238,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {impl: c, crdb: "latest-v22.1", python: "3.10", libpq: newest}
-          - {impl: python, crdb: "latest-v22.2", python: "3.11"}
+          - {impl: c, crdb: "latest-v22.1", python: "3.10", libpq: newest, pytest_opts: "--anyio=trio"}
+          - {impl: python, crdb: "latest-v22.2", python: "3.11", pytest_opts: ""}
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
       PSYCOPG_TEST_DSN: "host=127.0.0.1 port=26257 user=root dbname=defaultdb"
+      PYTEST_ADDOPTS: ${{ matrix.pytest_opts }}
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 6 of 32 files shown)


dvarrazzo:
What is the advantage of using anyio?

Last time I checked anyio it was adding a lot of overhead. Can we see flameshot graphs to show what changes during querying and during copy?

It wouldn't be bad to put up something to create them automatically actually. When I play with them they were a great guide to improve performances.

Ref: https://www.varrazzo.com/blog/2020/05/19/a-trip-into-optimisation/

dlax:
> What is the advantage of using anyio?

The advantage is that we have a single implementation on our side that's independent of the underlying async library (asyncio or trio or anything that anyio might support in the future). For users, it works transparently: they can use `asyncio.run(main())` or `trio.run(main)`.

> Last time I checked anyio it was adding a lot of overhead.

I haven't checked.

But... FastAPI recently moved to anyio as well (https://github.com/tiangolo/fastapi/releases/tag/0.69.0), following starlette move https://www.starlette.io/release-notes/#0150
I suppose this would not have happened if they were performance issues (https://github.com/encode/starlette/pull/1157 mentions a 4.5% drop in performance).

> Can we see flame graphs to show what changes during querying and during copy?
> 
> It wouldn't be bad to put up something to create them automatically actually. When I play with them they were a great guide to improve performances.
> 
> Ref: https://www.varrazzo.com/blog/2020/05/19/a-trip-into-optimisation/

That would be very helpful, indeed.

dlax:
I have run [MagickStack's benchmark](https://github.com/MagicStack/pgbench) against 3.0.3 and this branch. The source code of the benchmark tool handling psycopg 3 is available at https://github.com/dlax/pgbench (branch `psycopg`).

Running `./pgbench python-psycopg-async`, on master (tag 3.0.3):
```
Initializing temporary PostgreSQL cluster...
Running python-psycopg-async benchmarks...
==========================================

1-pg_type.json, concurrency 10
------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/1-pg_type.json
12833 queries in 30.02 seconds
Latency: min 10.25ms; max 28.65ms; mean 23.369ms; std: 0.421ms (1.8%)
Latency distribution: 25% under 23.226ms; 50% under 23.34ms; 75% under 23.47ms; 90% under 23.611ms; 99% under 24.637ms; 99.99% under 28.104ms
Queries/sec: 427.5
Rows/sec: 256073.72

2-generate_series.json, concurrency 10
--------------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/2-generate_series.json
22759 queries in 30.01 seconds
Latency: min 3.9ms; max 50.88ms; mean 13.171ms; std: 4.914ms (37.31%)
Latency distribution: 25% under 9.371ms; 50% under 13.162ms; 75% under 13.431ms; 90% under 17.753ms; 99% under 31.327ms; 99.99% under 49.834ms
Queries/sec: 758.36
Rows/sec: 758357.32

3-large_object.json, concurrency 10
-----------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/3-large_object.json
24550 queries in 30.01 seconds
Latency: min 7.2ms; max 24.93ms; mean 12.207ms; std: 1.226ms (10.04%)
Latency distribution: 25% under 12.262ms; 50% under 12.352ms; 75% under 12.46ms; 90% under 12.788ms; 99% under 14.929ms; 99.99% under 22.861ms
Queries/sec: 818.11
Rows/sec: 81810.64

4-arrays.json, concurrency 10
-----------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/4-arrays.json
3706 queries in 30.07 seconds
Latency: min 13.31ms; max 119.47ms; mean 81.05ms; std: 13.833ms (17.07%)
Latency distribution: 25% under 74.269ms; 50% under 75.136ms; 75% under 76.479ms; 90% under 111.65ms; 99% under 115.749ms; 99.99% under 119.473ms
Queries/sec: 123.23
Rows/sec: 12322.83

5-copyfrom.json, concurrency 10
-------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/5-copyfrom.json
11710000
956 queries in 30.62 seconds
Latency: min 108.04ms; max 1844.63ms; mean 305.438ms; std: 111.382ms (36.47%)
Latency distribution: 25% under 251.7ms; 50% under 290.71ms; 75% under 329.63ms; 90% under 396.308ms; 99% under 535.209ms; 99.99% under 1844.638ms
Queries/sec: 31.22
Rows/sec: 314449.84

6-batch.json, concurrency 10
----------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=57458 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/6-batch.json
130 queries in 30.39 seconds
Latency: min 731.93ms; max 1799.65ms; mean 1186.183ms; std: 357.06ms (30.1%)
Latency distribution: 25% under 773.54ms; 50% under 1201.73ms; 75% under 1568.21ms; 90% under 1695.09ms; 99% under 1765.544ms; 99.99% under 1799.66ms
Queries/sec: 4.28
Rows/sec: 5922.59

Geometric mean results
----------------------

3858.99 queries in 30.19 seconds
Latency: min 25.915ms; max 156.0ms; mean 69.255ms; std: 10.571ms (15.26%)
Latency distribution: 25% under 58.132ms; 50% under 68.084ms; 75% under 73.314ms; 90% under 85.91ms; 99% under 103.93ms; 99.99% under 152.743ms
Queries/sec: 127.85
Rows/sec: 84522.12
```

And then, on this branch, with anyio:
```
Initializing temporary PostgreSQL cluster...
Running python-psycopg-async benchmarks...
==========================================

1-pg_type.json, concurrency 10
------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/1-pg_type.json
22539 queries in 30.01 seconds
Latency: min 5.75ms; max 27.48ms; mean 13.306ms; std: 1.263ms (9.49%)
Latency distribution: 25% under 12.699ms; 50% under 12.937ms; 75% under 13.872ms; 90% under 14.229ms; 99% under 17.172ms; 99.99% under 27.275ms
Queries/sec: 751.03
Rows/sec: 449868.48

2-generate_series.json, concurrency 10
--------------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/2-generate_series.json
33502 queries in 30.02 seconds
Latency: min 2.95ms; max 29.62ms; mean 8.953ms; std: 3.055ms (34.12%)
Latency distribution: 25% under 8.034ms; 50% under 8.218ms; 75% under 8.345ms; 90% under 8.733ms; 99% under 21.577ms; 99.99% under 27.413ms
Queries/sec: 1116.0
Rows/sec: 1116004.02

3-large_object.json, concurrency 10
-----------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/3-large_object.json
19780 queries in 30.0 seconds
Latency: min 11.46ms; max 23.31ms; mean 15.16ms; std: 0.659ms (4.35%)
Latency distribution: 25% under 14.78ms; 50% under 14.893ms; 75% under 15.619ms; 90% under 15.949ms; 99% under 17.089ms; 99.99% under 23.19ms
Queries/sec: 659.24
Rows/sec: 65923.6

4-arrays.json, concurrency 10
-----------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/4-arrays.json
3218 queries in 30.09 seconds
Latency: min 11.86ms; max 98.6ms; mean 93.365ms; std: 2.913ms (3.12%)
Latency distribution: 25% under 92.681ms; 50% under 93.458ms; 75% under 94.27ms; 90% under 94.986ms; 99% under 96.259ms; 99.99% under 98.604ms
Queries/sec: 106.95
Rows/sec: 10694.72

5-copyfrom.json, concurrency 10
-------------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/5-copyfrom.json
880 queries in 30.28 seconds
Latency: min 229.69ms; max 617.65ms; mean 343.868ms; std: 32.653ms (9.5%)
Latency distribution: 25% under 333.63ms; 50% under 337.943ms; 75% under 345.45ms; 90% under 353.507ms; 99% under 542.424ms; 99.99% under 617.658ms
Queries/sec: 29.06
Rows/sec: 290612.64

6-batch.json, concurrency 10
----------------------------

/home/denis/src/pgbench/pgbench_python --warmup-time=5 --duration=30 --timeout=2 --pghost=/tmp --pgport=63536 --pguser=postgres --output-format=json --concurrency=10 psycopg-async /home/denis/src/pgbench/queries/6-batch.json
160 queries in 31.37 seconds
Latency: min 1595.57ms; max 1897.04ms; mean 1698.134ms; std: 94.568ms (5.57%)
Latency distribution: 25% under 1615.24ms; 50% under 1650.63ms; 75% under 1790.2ms; 90% under 1846.58ms; 99% under 1895.788ms; 99.99% under 1897.05ms
Queries/sec: 5.1
Rows/sec: 5738.42

Geometric mean results
----------------------

4349.14 queries in 30.29 seconds
Latency: min 30.747ms; max 113.974ms; mean 67.953ms; std: 5.328ms (7.84%)
Latency distribution: 25% under 64.985ms; 50% under 65.986ms; 75% under 68.73ms; 90% under 70.51ms; 99% under 92.509ms; 99.99% under 112.277ms
Queries/sec: 143.57
Rows/sec: 91589.3
```

dvarrazzo:
> I have run [MagickStack's benchmark](https://github.com/MagicStack/pgbench) against 3.0.3 and this branch. The source code of the benchmark tool handling psycopg 3 is available at https://github.com/dlax/pgbench (branch `psycopg`).

Why is your branch using internal functions instead of the public interface?

About 1y ago I made a psycopg3 pgbench branch too: https://github.com/dvarrazzo/pgbench/tree/psycopg3. I don't know if it's still good or if it's rotten.

> Running `./pgbench python-psycopg-async`, on master (tag 3.0.3):

> Queries/sec: 427.5
> Queries/sec: 758.36
> Queries/sec: 818.11
> Queries/sec: 123.23
> Queries/sec: 31.22
> Queries/sec: 4.28
> ----------------------
> Queries/sec: 127.85

> And then, on this branch, with anyio:
> Queries/sec: 751.03
> Queries/sec: 1116.0
> Queries/sec: 659.24
> Queries/sec: 106.95
> Queries/sec: 29.06
> Queries/sec: 5.1
> ----------------------
> Queries/sec: 143.57

Curious results, no? Do tests hit different paths, so that anyio is faster on the first two and slower on the other ones? Or am I reading something wrong?

Without changing too much, but only dropping in your Futures patch in the 3.0.3 wait loop, how does it compare?

Thank you for this!

dlax:
> Why is your branch using internal functions instead of the public interface?

You mean `psycopg.pq`? Yes I added benchmark runners implemented in this low-level API (benchmark names: `python-pq` and `python-pq-async` to compare with the normal psycopg API (benchmark names: `python-psycopg` and `python-psycopg-async`).
But *here*, I'm using the normal API (benchmark: `python-psycopg-async`).
 
> About 1y ago I made a psycopg3 pgbench branch too: https://github.com/dvarrazzo/pgbench/tree/psycopg3. I don't know if it's still good or if it's rotten.

Didn't know about this; it's close to my branch, only the "copy" benchmarks seem different (you used an `io.StringIO()`, I used a loop with `copy.write_row()`).

How about we coordinate with a common fork in psycopg organization?
At some point, having a benchmark suite run automatically and following the git history would be really nice. Perhaps using [asv](https://asv.readthedocs.io/en/stable/)?

> Curious results, no? 

Yeah, that looked strange to me as well. But I did not investigate much yet.

> Without changing too much, but only dropping in your Futures patch in the 3.0.3 wait loop, how does it compare?

Good idea, will do.

dlax:
> Curious results, no? Do tests hit different paths, so that anyio is faster on the first two and slower on the other ones? Or am I reading something wrong?

In fact, I cannot get reliable results with this benchmarks. For instance, running them twice on 3.0.3 tag, I get:
```
Queries/sec: 1266.77
Queries/sec: 1242.04
Queries/sec: 1277.79
Queries/sec: 112.31
Queries/sec: 29.83
Queries/sec: 8.16
Queries/sec: 194.99
```
and
```
Queries/sec: 597.51
Queries/sec: 1094.14
Queries/sec: 1664.3
Queries/sec: 106.7
Queries/sec: 31.63
Queries/sec: 8.64
Queries/sec: 177.93
```
Even running each benchmark alone does not produce consistent results... That's all I can tell with the hardware I have.

> Without changing too much, but only dropping in your Futures patch in the 3.0.3 wait loop, how does it compare?

I did not observe any speedup with the Future-based waiting loop (see commit dlax/psycopg3@4630e68f8b73b4f1ee28d2579b8615c13f2b10dc) rather a slight slowdown actually. In addition to the above benchmarks (and their inconsistency), I ran the "pipeline demo" test (https://github.com/psycopg/psycopg/pull/116/files#diff-7c5c69215e01c35a4a21c087ec14a2581abf18209386d7522b685ca8af20961dR46, takes about 2.5s to complete on my machine) as a benchmark with the Event-based and Future-based loop: no obvious winner/looser there either...

dlax:
Reworked this to make the dependency on anyio optional.

(Still needs some work on copy.)

dlax:
`tests/types/test_net.py::test_lazy_load` fails in many cases; not sure how it's related.

dvarrazzo:
> `tests/types/test_net.py::test_lazy_load` fails in many cases in CI; not sure how it's related.

Probably anyio uses that module. The test runs with the assumption that the network module is not imported when psycopg is.

dlax:
I see, makes sense; thanks.
I'll see how to adjust this test then.

Kludex:
I understand why this PR was created... But does it still make sense to have `anyio` now that it's wanted for it to be optional dependency? 🤔

dlax:
@Kludex, I'm afraid I don't quite get your point; would you mind reformulating or expanding a bit?

Kludex:
The idea of `anyio` is to give freedom to package users to choose between `asyncio` and `trio`, and while doing it, leveraging _trio style_ (structured concurrency).

If we add `anyio` as optional dependency, the above is not achieved, it's just going to add another layer. Instead, if we want to allow package users to use `trio`, then implementing an optional flow with `trio` itself would do.

@graingert does it makes sense what I'm saying here?

dhirschfeld:
> *If we add anyio as optional dependency, the above is not achieved*

Isn't the parallel implementation of `asyncio` and `anyio` intended to be temporary until `anyio` has proven itself, after which `asyncio` can be ripped out in favour of `anyio` so the maintainers only have to support 1 async backend rather than 2.

dlax:
So I have spent a significant amount of time on this work of using anyio as an alternative to asyncio in psycopg (core only, not speaking about the pool).

For the record, this involves the following functional changes:
1. implementing alternative "waiting" functions using anyio,
2. implementing an `AnyIOConnection`, as an alternative to `AsyncConnection`, using an `anyio.Lock`, and,
3. implementing an `AnyIOCopy` using anyio streams instead of asyncio queues.
This part is mostly done (though it'd take a significant rebase work due to recent changes in the copy code in master).

But I'm still struggling with tests. Currently, we use `pytest-asyncio` (in "auto" mode, if that matters). And, unless I'm missing something, we'd need to use anyio's pytest plugin to test anyio implementations. From this point, I'm not sure how to proceed:
- using both plugins together seems very hard and fragile (for instance, I'm never quite sure which plugin will be used), and
- while replacing pytest-asyncio with anyio plugin is doable it is an invasive change (on which I'd need an agreement from @dvarrazzo).

So I'd appreciate some help from the community supporting anyio/trio here. Do we have examples of other projects haven successfully implemented the asyncio->anyio transition?

dhirschfeld:
> *So I'd appreciate some help from the community supporting anyio/trio here*

@graingert is definitely the expert here so hopefully he can shed some light on the best way forward. If it's not obvious we can always reach out to the broader community, including the AnyIO author, on how best to test multiple async implementations.

graingert:
> But I'm still struggling with tests. Currently, we use pytest-asyncio (in "auto" mode, if that matters). And, unless I'm missing something, we'd need to use anyio's pytest plugin to test anyio implementations

You shouldn't need to drop pytest-asyncio if you don't want to, the anyio pytest plugin should just make your life easier

You can also switch to pytest-asyncio manual mode and mix and match pytest plugins so you'd be able to use pytest-trio too

Let me know what you need here and I'm happy to help any way I can

agronholm:
Is there any reason *not* to use the AnyIO plugin for asyncio or trio tests in any project?

graingert:
> Is there any reason _not_ to use the AnyIO plugin for asyncio or trio tests in any project?

I had a bit of trouble migrating from pytest-asyncio where anyio forces you to do the right thing and pytest-asyncio lets you do the wrong thing, and so switching to the anyio pytest plugin has a bit of friction but I think it's worth doing

dlax:
I moved on with using anyio instead pytest-asyncio in the test suite, which indeed really helps for adding support for anyio in psycopg; this is done in the first commit of this PR (also submitted as #352, to illustrate that there is no regression in CI).

Made good progress on application code, besides; let's see how CI goes now...

dlax:
I don't understand the test failure on Windows as it complains about the loop policy being incorrect (Proactor), *but only upon retry*.

graingert:
> I don't understand the test failure on Windows as it complains about the loop policy being incorrect (Proactor), _but only upon retry_.

Ah probably because of a bad interaction with set_event_loop_policy and anyio resetting your event loop policy

dhirschfeld:
I've got a project coming up where I'd love to be able to use both `psycopg` and `trio`.

@dlax, is there anything left to do here (other than a rebase), or is this ready for review/merge from your perspective?

dlax:
It's ready for review; I'll rebase when I get a moment.

psycopg/psycopg/copy.py
```diff
@@ -629,14 +645,74 @@ async def write(self, data: Buffer) -> None:
...
+
+
+try:
+    import anyio
+    import anyio.abc
+except ImportError:
+    pass
+else:
+
+    class AnyIOLibpqWriter(AsyncLibpqWriter):
```
dvarrazzo, line 670: In 3.1, this seems more the behaviour of `AsyncQueuedLibpqWriter` than `AsyncLibpqWriter`, because it spawn separate workers, as opposite as `AsyncLibpqWriter`, which works in-task.

Afaics, `AsyncLibpqWriter` would work out-of-the-box with anyio, because it only uses `await self.connection.wait(<some generator>)` to send control back to the loop.

Originally, using a separate thread/task seemed to yield some results, but in following tests I couldn't verify a real benefit for using `AsyncQueuedLibpqWriter`. As such, psycopg 3.1 uses `AsyncLibpqWriter` by default, and maybe it's not worth to have an `AnyIOLibpqWriter`.

psycopg/psycopg/__init__.py
```diff
@@ -26,6 +26,13 @@
 from .client_cursor import AsyncClientCursor, ClientCursor
 from .connection_async import AsyncConnection
 
+try:
+    from .connection_async import AnyIOConnection  # noqa: F401
+except ImportError:
+    _anyio = False
+else:
+    _anyio = True
+
```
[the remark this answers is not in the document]
dvarrazzo: Or even making `AnyIOConnection` a class failing with ImportError on initialization, and giving it a public, documented classmethod `AnyIOConnection.is_supported()` returning True of False.

psycopg/psycopg/connection_async.py
```diff
@@ -434,3 +446,54 @@ async def tpc_recover(self) -> List[Xid]:
             await self.rollback()
 
         return res
+
+
+try:
+    import anyio
```
dvarrazzo, line 452: I think it's worth adding an extra so that people can `pip install psycopg[trio]` and get the dependencies too (anyio, sniffio, trio). It would also establish the min required versions.

psycopg/psycopg/waiting.py
```diff
@@ -198,14 +203,147 @@ def wakeup(state: Ready) -> None:
...
-    except TimeoutError:
+    except AsyncIOTimeoutError:
         raise e.ConnectionTimeout("connection timeout expired")
 
     except StopIteration as ex:
         rv: RV = ex.args[0] if ex.args else None
         return rv
 
 
+try:
```
dvarrazzo, line 214: ...same here: I would prefer to rely on the anyio objects implemented in separate modules. This woudn't even need conditional import, because it would be imported by `connection_anyio.py`

tests/test_waiting.py
```diff
@@ -104,47 +104,58 @@ def test_wait_epoll_bad(pgconn):
     assert res.status == ExecStatus.TUPLES_OK
 
 
+@pytest.fixture
+def wait_async(anyio_backend_name):
+    return {
+        "asyncio": waiting.wait_asyncio,
+        "trio": waiting.wait_anyio,
+    }[anyio_backend_name]
```
dvarrazzo, line 112: This way the test depends on trio/anyio. It means that we are not able to verify anymore that psycopg works if trio is not installed.

I think that we should run the trio tests in a separate environment that would install `psycopg[trio]` and would test the anyio objects, whereas in the base tests we would test taht AnyIOConnection etc. "not work as expected".

tests/test_waiting.py
```diff
@@ -104,47 +104,58 @@ def test_wait_epoll_bad(pgconn):
...
-    (res,) = await waiting.wait_async(gen, pgconn.socket)
+    (res,) = await wait_async(gen, pgconn.socket)
     assert res.status == ExecStatus.TUPLES_OK
 
 
-@pytest.mark.asyncio
 @pytest.mark.parametrize("wait, ready", zip(waiting.Wait, waiting.Ready))
 @skip_if_not_linux
-async def test_wait_ready_async(wait, ready):
+async def test_wait_ready_async(wait, ready, wait_async):
```
dvarrazzo, line 145: Is the `wait_async` fixture only used in this couple of tests? Maybe we could just test the `wait_asyncio()/wait_anyio()` functions in separate tests instead.

dvarrazzo (changes requested):
Thank you for this work, @dlax!

What would be the test strategy? I assume, at the moment, every test runs in "two copies" because of parametrized fixtures. As such there is no change in the pipeline.

As suggested, I would prefer to test that the base psycopg would work without trio etc. installed, and that then we had a separate test run which would run in a tox env where `psycopg[trio]` was installed and that tests forced to run on the trio code (for which I don't think there is a mechanism? Or am I wrong?)

psycopg/psycopg/copy.py
```diff
@@ -528,15 +528,20 @@ async def write_row(self, row: Sequence[Any]) -> None:
...
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
         if self._direction == COPY_IN:
             data = self.formatter.end()
             if data:
                 await self._write(data)
-            await self.writer.finish(exc)
+            await self.writer.finish(exc_type, exc_val, exc_tb)
```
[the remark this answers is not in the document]
dlax: No; it's because we're calling the `__aexit__()` of `TaskGroup` context manager, which needs all three values.

psycopg/psycopg/__init__.py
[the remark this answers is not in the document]
dlax: The issue with `AnyIOConnection: Optional[Type[AsyncConnection]]` is that this will leads to typing error in user code such as:
```
Item "None" of "Optional[Type[AsyncConnection[Any]]]" has no attribute "connect"
```

I'm not sure about the `is_supported()` idea. On the one hand, it feels right for the pipeline mode where "supported" means having installed the correct libpq version (not a Python package), on the other hand, for a plain Python dependency it seems a bit strange... I'll take a stab on the `ImportError` on initialization though.

psycopg/psycopg/copy.py
dlax: Yes, I noticed that lately. I agree we should use `AsyncLibpqWriter` by default with anyio and `AnyIOLibpqWriter` (possibly renamed) would be the optional counterpart of `AsyncQueuedLibpqWriter`.

psycopg/psycopg/__init__.py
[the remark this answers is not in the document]
dvarrazzo: Probably you are right: we could assume the library always supported because we assume that the dependencies were installed correctly. We can also always create a class, but making it fail __init__ with ImportError if anyio is missing, which would make typing easier.

What do you think about it?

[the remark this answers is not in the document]
dlax: Yes, that's what I'll try to do.

psycopg/psycopg/connection_async.py
dlax: Yes, agreed; I'll do that.

But rather `psycopg[anyio]`, no? Because people might be interested in using AnyIO with the `asyncio` backend. (And we have nothing specific to `trio` on our side.)

dvarrazzo: If you think that `anyio` is a better name I trust that. You know the relation among the projects better than I do :slightly_smiling_face:

psycopg/psycopg/copy.py
```diff
@@ -629,14 +645,74 @@ async def write(self, data: Buffer) -> None:
...
 
         if self._worker:
             await asyncio.gather(self._worker)
             self._worker = None  # break reference loops if any
 
-        await super().finish(exc)
+        await super().finish(exc_type, exc_val, exc_tb)
+
+
+try:
```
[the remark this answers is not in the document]
dlax: Ok; moved all anyio related code to dedicated modules.

psycopg/psycopg/connection_anyio.py
```diff
@@ -0,0 +1,81 @@
...
+
+if TYPE_CHECKING:
+    import anyio
+    import sniffio
+else:
+    anyio = None
+    sniffio = None
+
+
+def _import_anyio() -> None:
```
dvarrazzo, line 25: I propose an `@lru_cache()` here to avoid triggering the import machinery at every connection.

graingert: If the module is in `sys.modules` it's already as fast as a dictionary lookup

dvarrazzo: The function is not slow, but with a cache it's 4x faster for free.

```
In [4]: %timeit _import_anyio()
229 ns ± 4.45 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: from functools import lru_cache

In [6]: _import_anyio2 = lru_cache()(_import_anyio)

In [7]: %timeit _import_anyio2()
52.4 ns ± 0.685 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

tests/test_waiting.py
dlax: We'd need to duplicate three test functions for `wait_async` and two more for `wait_conn_async`. That's beyond my (subjective) limit for code repetition :)

psycopg/setup.py
```diff
@@ -31,16 +31,20 @@
     "binary": [
         f"psycopg-binary == {version}",
     ],
+    # Requirements to use AnyIO and Trio async library.
+    "anyio": [
+        "anyio",
```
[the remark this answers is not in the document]
dlax: Indeed, fixed.

(Also fixed crdb tests, and now rebased to drop the fixups.)

tests/test_waiting.py
dlax: The test suite needs `anyio`'s pytest plugin, per first commit (and discussions above) where it replaces `pytest-asyncio`. But since the plugin comes along with the library, the test suite actually depends on the whole `anyio` package.

Maybe we could make the `trio` dependency optional in the test suite (to be caught in `anyio_backend` fixture, I'd suppose). But I'm not sure I get why we'd want this. Can you elaborate? Is this a developer experience issue, to avoid running async tests with `trio`?

Or maybe we could introduce a pytest option (or a mark) to pick the `anyio` backend? I would work as:
- no option, run tests with plain `asyncio` (but through anyio pytest plugin), i.e. uses `AsyncConnection`
- `--anyio-backend=asyncio`, run tests with the `asyncio` backend of `anyio`, i.e. uses `AnyIOConnection` but run the `asyncio` event loop etc.
- `--anyio-backend=trio`, run test with the `trio` backend of `anyio`, i.e. still uses `AnyIOConnection` but with `trio` event loop etc.

dvarrazzo: > The test suite needs anyio's pytest plugin, per first commit (and discussions above) where it replaces pytest-asyncio. But since the plugin comes along with the library, the test suite actually depends on the whole anyio package.

I see, you are right.

> But I'm not sure I get why we'd want this. Can you elaborate?

I just would like to have some test that guarantees that anyio is an optional dependency. I understand that we can't do it anymore, but it would be useful to be able to test without trio.

> Is this a developer experience issue, to avoid running async tests with trio?

Now all the async tests run in two copies, which adds some time to the test run. I think it would be a good idea for the test suite to run with a single backend, yes. I think that would make tests configuration easier, as the backend can be selected globally on tests init.

I think it's a good idea to parametrize the test run from command lines or env vars. I understand we have choices. We can use `--anyio={asyncio|trio|no}` and we also have `--loop={uvloop|default}` (we can change it)... so a whopping 6 choices of async flavous if I'm correct? (as in you can go trio+uvloop too etc?) Maybe we don't need to cover the grid entirely and can trust that certain anyio+uvloop are well tested upstream.

So, I'd make a test run only operate with a given backend selection, and use the CI grid to test things more extensively on push :)

dlax: Last commit "test: allow to select async backend and implementation" implements this plan.
```
$ pytest --help
[...]
custom options:                                                                                                                         
  --anyio={asyncio,trio}    
                        Use AnyIO implementation of the async API, run tests with specified backend. If unset, use plain asyncio
                        implementation, and run with asyncio backend.
[...]
```

dhirschfeld:
It seems this PR is pretty close to being ready to merge (modulo the conflicts)?

Sorry, I don't mean to seem pushy, I'm just excited by the possibility to use `trio` with `psycopg`! :heart::rocket:

agronholm:
FYI, there was an additional, fatal compatibility issue noticed in v3.6.1 so I released v3.6.2 to pin the trio extra to < 0.22. This will be fully rectified in v4.0.

dlax:
This PR was ready; but now again there are conflicts so it needs work.
So before I possibly get back to it:
* do we want the feature @dvarrazzo?
* is this still interesting to the community?

dhirschfeld:
> *is this still interesting to the community?*

As an anecdotal data point, this is *very* interesting to me!

I'm super excited about the prospect of being able to use `psycopg` in a `trio` app (without having to resort to using threads)! :rocket:

dlax:
I've reworked this PR to make it as minimal as possible by:
* removing the custom copy-writer class (similar to `AsyncQueuedLibpqWriter`, but using streams), as the default `AsyncLibpqWriter` works fine with anyio/trio (previous implementation [here](https://github.com/dlax/psycopg3/commit/3ceba3e129dc6ffedc2ec8fa18ffbc655ac6dc79));
* keeping concurrency tests (which use asyncio in their implementation) only run with asyncio backend (as I don't think porting them to also run the trio backend would bring much coverage w.r.t. to the [churn](https://github.com/dlax/psycopg3/commit/9be86f86068690ebe6fb54cc57197451823a800a)).

So should be easier to review now...

saolof:
What is the status on this?

saolof:
What happened to this? This was a really nice PR that I was excited about

dlax:
@dvarrazzo, can we make a decision about this? (The lack of response makes me feel you are not prone to integrate the feature but maybe not? I'd be a bit sad to give up on this work, sure; but staying in limbo is worse.)

dvarrazzo:
Hello @dlax 

I believe I have lost the ball on this: trying to figure out the advantage (because I don't know much about the frameworks outside asyncio). I think I was waiting on a followup about the inconsistencies in the benchmark.

Let's not consider this closed.

tests/test_client_cursor_async.py
```diff
@@ -433,11 +433,13 @@ async def test_rownumber(aconn):
...
     assert cur.rownumber == 12
     rns: List[int] = []
-    async for i in cur:
+    itcur = cur.__aiter__()
+    async for i in itcur:
         assert cur.rownumber
         rns.append(cur.rownumber)
         if len(rns) >= 3:
             break
+    await itcur.aclose()
```
dvarrazzo, line 442: Why is this pattern needed? It doesn't seem a very representative way to iterate on an object.

agronholm:
> Now, I see that anyio has a [concept of global backend](https://anyio.readthedocs.io/en/stable/basics.html#running-async-programs), which means that we don't expect people to use both `AsyncConnection` and `AnyIOConnection` in the same program.

Just so we're on the same page, what do you understand by the concept of "global backend"?

dvarrazzo:
@agronholm 

> Just so we're on the same page, what do you understand by the concept of "global backend"?

What I linked, i.e. that if someone run their program into a:
```python
from anyio import run
...
run(main, backend='trio')
```
then they don't expect to run `asyncio` resources in the same runtime. At least not via the `anyio` abstraction.

As a consequence, my understanding is that psycopg user using trio have no reason to use `AsyncConnection()` working on asyncio loop and `AnyIOConnection()` working on trio loop, in the same process. Hence my observation that we might not need an `AnyIO*` version of every `Async*` object we have that happens to use a lock, but we might have only one `AsyncConnection()` public object, for the user to use, and a global configuration calls allowing to select whether to use anyio or not (then what backend should anyio use is up to its way to configure - internally we would use sniffio as @dlax has implemented here).

Is my understanding on the page you expect to be?

agronholm:
> As a consequence, my understanding is that psycopg user using trio have no reason to use `AsyncConnection()` working on asyncio loop and `AnyIOConnection()` working on trio loop, in the same process. Hence my observation that we might not need an `AnyIO*` version of every `Async*` object we have that happens to use a lock, but we might have only one `AsyncConnection()` public object, for the user to use, and a global configuration calls allowing to select whether to use anyio or not (then what backend should anyio use is up to its way to configure - internally we would use sniffio as @dlax has implemented here).
> 
> Is my understanding on the page you expect to be?

`anyio.run()` is little more than a convenience which users may or may not be using. Trio based apps might use trio-asyncio to run embedded asyncio code. As such, it is at least theoretically possible for someone to be using both backends in the same thread.

The reason your wording piqued my interest is that there is indeed something _potentially_ coming in the AnyIO space called a "global event loop". This would allow one to provide both an async and a sync interface to a library by making the synchronous version into a shim that runs the async version in a dedicated event loop thread. This approach is demonstrated in the latest alpha of [APScheduler](https://pypi.org/project/APScheduler/4.0.0a3/).

dhirschfeld:
> *my understanding is that psycopg user using trio have no reason to use AsyncConnection() working on asyncio loop and AnyIOConnection() working on trio loop*

I don't think it's necessary (and I think we want to avoid)  having duplicate implementations. My understanding is that if you implement `AsyncConnection` using *`anyio`* primitives then a user can import it and use it directly in *either* an `asyncio` or a `trio` program.

i.e. if you do have an `AnyIOConnection` then an `asyncio` user could just as well use that in their `asyncio` program as an `AsyncConnection` implemented directly using `asyncio` primitives.

This makes duplicate implementations redundant at best, from an end-user perspective, and in practice, doubling the api surface area and forcing the user to choose between implementations is a worse UX.

I'm not an `anyio` expert so perhaps @agronholm can correct me if that's not the case?

agronholm:
Sure, ideally that is how it would work. But, I've seen some prominent downstream projects that are averse to having an unconditional dependency on AnyIO (at least [httpcore](https://github.com/encode/httpcore) comes to mind).

dvarrazzo:
> My understanding is that if you implement AsyncConnection using anyio primitives then a user can import it and use it directly in either an asyncio or a trio program

@dhirschfeld read my message above: anyio has its own release cycle, on which I don't want psycopg to depend on. I don't want to drop Python 3.x support when anyio decides it's the time, I don't want to put a upper limit to the anyio version used because a non-backward-compatible change they may make makes our support difficult.

To further on it, If anyio project changes in direction we don't like, if anyio goes out of fashion because another abstraction api comes up making it easier to work with N > 2 async libraries, I don't want psycopg to depend on it.

tests/test_client_cursor_async.py
dlax: This is to resolve a `ResourceWarning` raised by trio; see the message of commit 88f85700b3a95dee89f8a15947fc067e9605a460.

agronholm:
> @dhirschfeld read my message above: anyio has its own release cycle, on which I don't want psycopg to depend on. I don't want to drop Python 3.x support when anyio decides it's the time, I don't want to put a upper limit to the anyio version used because a non-backward-compatible change they may make makes our support difficult.

AnyIO supports the Python releases supported by the PSF. That is to say, it won't drop support until a Python release reaches EOL. If you intend to keep supporting old Python versions longer, then this could potentially be an issue.
 
> To further on it, If anyio project changes in direction we don't like, if anyio goes out of fashion because another abstraction api comes up making it easier to work with N > 2 async libraries, I don't want psycopg to depend on it.

I am quite skeptical that another such abstraction API would surface. It's a tremendous amount of work.

dlax:
I'm totally in favour of avoiding classes duplication for each async library. (In fact, I seem to remember this was an option initially but we decided not to go that way; anyway...)

> A cleaner option would be to create an `AsyncBackend` class, with a global instance exposed as `psycopg.backends.async_backend`, which may be replaced by an instance of a subclass `psycopg_anyio.AnyIOBackend`; if deemed useful we could do the same with a `Backend` class for the different ways of doing sync things (default threading, eventlet, gevent).

This option is worth a try; it's also quite similar to what's done by httpcore I think.

> of course this branch now has several conflict with master, and I have no problem to help rebasing it, if we think it's the right thing to do

I think the first commits of this branch (or some of them), i.e. all but the last two, might be worth considering for integration now as they are just preparatory refactorings; they need a rebase and they indeed conflict. The last-but-one commit, the one introducing the feature, would need a complete rework if we actually go the "AsyncBackend" route, so it's not worth rebasing IMO. @dvarrazzo, if you can help rebasing (possibly only some of) the first commits, that'd indeed be appreciated. I would then be happy to take over the rest.

dhirschfeld:
> *`anyio` has its own release cycle, on which I don't want psycopg to depend on*

Unconditionally depending on `anyio` does to some extent couple your programs - that's the same cost/benefit calculation for taking on any dependency. In practice it would only become an issue if older versions weren't supported as otherwise you can just pin the upper bound and upgrade in your own time.

Having two separate code paths, one for an `asyncio` implementation and one for `anyio` does mean that if `anyio` release a backward-incompatible version 5 and immediately stop supporting version 4, you could still release `psycopg` with `asyncio` support; but there's a maintenance cost in doing so - it's more work maintaining two code paths and it exposes you to two different sets of bugs.

Maintaining separate code paths is pessimistically paying a cost for an eventuality which may never happen. It's like an insurance policy against `anyio` making breaking changes or stopping being supported. It might be less work to optimistically just support `anyio` and accept the potential cost of a rewrite if and when `anyio` support became untenable.

At the end of the day, I'll be happy if I can just use `psycopg` natively in my `trio` apps. As an end-user I don't care how that's implemented so I'm happy with whatever path you choose. I'm just piping up here as I care about the long-term health of the project and so don't want the maintenance burden increasing if it's not necessary/beneficial. That's entirely your decision to make though so I'll just leave it at that - I'll be happy with whatever you decide is best for your project!

dvarrazzo:
@dhirschfeld thank you for your input.

API breakage with anyio already happened in 3 -> 4; luckily it just affected the test suite. Together with the change of interface, anyio dropped support for Python 3.7. This is all very legitimate, nothing bad from their part.

We plan to drop Python 3.7 support too, but in Python 3.2, possibly in a few months time. So we would have had to do something awkward to support possibly incompatible 3 and 4, or drop support for 3.7 before planned, or set an upper boundary < 4, which is [a very bad idea](https://iscinumpy.dev/post/bound-version-constraints/). Incidentally, in the test suite for the moment I did exactly that, but just because it's a test dependency, and hopefully temporary (#662 easier as Python 3.7 is now out of CI).

> At the end of the day, I'll be happy if I can just use psycopg natively in my trio apps

I have also been considering that AFAICS the real feature is not supporting anyio, it is supporting trio.

- anyio supports N backends, but at the moment N = 2 (asyncio, trio). Will there be a third? Is there anything there which might become that?
- anyio offers remarkable features, but we don't use them.
- we are now thinking about having our own backends system for the minimal objects we need that would be different between async framework. We would have an asyncio backend in core and we can have external backends in external modules with decoupled release cycles and dependencies. In this case, is it worth to make an anyio backend or shouldn't we just make a trio backend? I understand that, if a program depends on anyio and runs either asyncio or trio backend, psycopg would still work in that environment, in both cases.

agronholm:
> anyio supports N backends, but at the moment N = 2 (asyncio, trio). Will there be a third? Is there anything there which might become that?

AnyIO started with 3 back-ends, ended up dropping Curio as it was untenable to support and its author explicitly requested this action. One prospect has been supporting Twisted, but in all honesty, I don't think that's ever going to happen as work on sniffio support in Twisted has pretty much stalled, and there seems to be comparably little interest in Twisted support in AnyIO in the first place.

> we are now thinking about having our own backends system for the minimal objects we need that would be different between async framework. We would have an asyncio backend in core and we can have external backends in external modules with decoupled release cycles and dependencies. In this case, is it worth to make an anyio backend or shouldn't we just make a trio backend? I understand that, if a program depends on anyio and runs either asyncio or trio backend, psycopg would still work in that environment, in both cases.

Whether using AnyIO makes sense depends on what you're doing. It offers a single API for task management, networking, IPC, subprocesses, async file access among other things. I personally consider the structural concurrency APIs perhaps the most important bit.

dvarrazzo:
> Whether using AnyIO makes sense depends on what you're doing. It offers a single API for task management, networking, IPC, subprocesses, async file access among other things. I personally consider the structural concurrency APIs perhaps the most important bit.

The concurrency is mostly left to the users of psycopg: the core psycopg query loop is a sequence of non-blocking calls interspersed with waiting for sockets to be ready, which is where we try to play fair with whatever is running concurrently. We allow connections to be used by more than one thread, but they get effectively serialized in their request-response cycle. So, to implement all this we need a wait function and a Lock. This part of psycopg is a library, not a program that is massively concurrent on its own. I think there was an issue in the past with anyio not being able to wait for read and write at the same time on a socket, and that's a deal breaker for us (https://github.com/psycopg/psycopg2/issues/1057#issuecomment-607186694). I assume that if @dlax has gone so far with this MR then that's a solved problem?

There is a bit more action in the pool, which has a scheduling task and a few worker tasks waiting on a queue. In the pool we create and destroy tasks, use Queues and Conditions. I see that Trio doesn't implement a Queue object, it uses sending/receiving channels, and that anyio doesn't offer a Queue-type object either, so we should create that interface ourselves anyway, if we want to keep the similarities with the sync code. Trio doesn't have a Condition object, anyio wraps it, but it's [a simple object](https://github.com/python/cpython/blob/228b12493286a7cb959438cd5b1a018e0afb17a8/Lib/asyncio/locks.py#L218).

Comparatively, there is way more concurrency in the test suite to simulate many different concurrency scenarios.

AFAICS the concurrency needs in our library seem simple; what we want the most is to make the library available to people who have made their architecture considerations and have decided that anyio is the tool they want to use, or that trio is.

dlax:
Upon reflection, and another look at the code and discussions recently, I don't see a clear way forward for this change proposal. As it does not seem actionable by me, as the author, I am closing the PR rather than keeping it in my todo list.

PR #150, opened 2021-11-14
Configure pytest warning control in pyproject.toml

dlax:
This reduces duplication and makes the configuration available when
testing outside CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,27 +65,23 @@ jobs:
 
       - name: Run tests (Python implementation)
         if: ${{ matrix.impl == 'python' }}
-        # The zoneinfo warning ignore is only required on Python 3.6
         run: >
           tox -c psycopg -e ${{ matrix.python }} --
-          --color yes -W error
-          -W ignore::DeprecationWarning:backports.zoneinfo._common
+          --color yes
 
       - name: Run tests (C implementation)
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Oh thank you. I was thinking that the errors on warning would be welcome at dev time indeed.

I was also thinking that the --color=yes can be added as an env var available to all the jobs in the CI. I'll make that change...

dlax:
> I was also thinking that the --color=yes can be added as an env var available to all the jobs in the CI. I'll make that change...

Maybe it's doable with PYTEST_ADDOPTS environment variable.

dvarrazzo:
> Maybe it's doable with PYTEST_ADDOPTS environment variable.

Yes, I'm pushing just that: 195b39b32765a57376f1c82fa1cbdd53f90b274c

dlax:
> > Maybe it's doable with PYTEST_ADDOPTS environment variable.
> 
> Yes, I'm pushing just that: [195b39b](https://github.com/psycopg/psycopg/commit/195b39b32765a57376f1c82fa1cbdd53f90b274c)

You might also need to tell tox to pass the variable (with `passenv` if I remember correctly); or does it work as is already?

dvarrazzo:
It knows already:

https://github.com/psycopg/psycopg/blob/5e4689522327631772f931d9b8ac0917f4b54895/psycopg/tox.ini#L9

PR #155, opened 2021-11-18
open pools outside __init__()

dlax:
Per discussions in #151, this PR removes pools opening (workers startup) from `__init__()` and instead performs this at context manager enter or at explicit call of the new `open()` method.
This is breaking change, thus targetting psycopg_pool version 4.0


dlax:
Replaced by #196.

PR #159, opened 2021-11-21
Add a minimal coverage configuration

dlax:
This does not check psycopg_c code, and I'm sure if/how this can be achieved, but this is enough to be useful to me for now.

### DIFF
--- a/psycopg/setup.py
+++ b/psycopg/setup.py
@@ -41,6 +41,7 @@ extras_require = {
         "pproxy ~= 2.7.8",
         "pytest ~= 6.2.5",
         "pytest-asyncio ~= 0.16.0",
+        "pytest-cov ~= 3.0.0",
         "pytest-randomly ~= 3.10.1",
         "tenacity ~= 8.0.1",
     ],
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ __pycache__/
 /psycopg_binary/
 .vscode
 .venv
+.coverage
+htmlcov
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,16 @@ testpaths=[
 # log_format = "%(asctime)s.%(msecs)03d %(levelname)-8s %(name)s:%(filename)s:%(lineno)d %(message)s"
 # log_level = "DEBUG"
 
+[tool.coverage.run]
+source = [
+    "psycopg/psycopg",
+    "psycopg_pool/psycopg_pool",
+]
+[tool.coverage.report]
+exclude_lines = [
+    "if TYPE_CHECKING:",
+    '\.\.\.$',
+]
 
 [tool.mypy]
 files = [


dvarrazzo:
I'm not really a fan of coverage but I know I'm a minority... :D

How do you run the coverage script?

Maybe it's worth adding the packages to the `psycopg[test]` extra?

dlax:
> How do you run the coverage script?

I use `pytest --cov --cov-report=html` and then `open htmlcov/index.html`. This requires `pytest-cov` plugin (which pulls `coverage`), it's handy to have a single command to run.

> Maybe it's worth adding the packages to the `psycopg[test]` extra?

Yes; added `pytest-cov` there.

psycopg/setup.py
```diff
@@ -41,6 +41,7 @@
         "pproxy ~= 2.7.8",
         "pytest ~= 6.2.5",
         "pytest-asyncio ~= 0.16.0",
+        "pytest-cov",
```
dvarrazzo, line 44: Let's pin the version like for the other packages:
```suggestion
        "pytest-cov ~= 3.0.0",
```

dvarrazzo (changes requested):
We don't do bad with coverage, considered that I never checked coverage before. It pays to be TDD from day 0 :nerd_face: 

For some reason, with `--cov` enabled, three tests fail for me:

```
FAILED tests/test_copy.py::test_copy_from_leaks[Format.TEXT-False] - AssertionError: objects leaked: 2, 2
FAILED tests/test_copy.py::test_copy_from_leaks[Format.BINARY-True] - AssertionError: objects leaked: 2, 2
FAILED tests/test_copy.py::test_copy_from_leaks[Format.TEXT-True] - AssertionError: objects leaked: 2, 2
```

Test setup is:
```
platform linux -- Python 3.8.10, pytest-6.2.5, py-1.10.0, pluggy-0.13.1
Using --randomly-seed=1713874353
libpq wrapper implementation: c
libpq used: 140000
libpq compiled: 140000
rootdir: /home/piro/dev/psycopg3, configfile: pyproject.toml, testpaths: tests
plugins: asyncio-0.16.0, rerunfailures-10.1, randomly-3.10.1, cov-3.0.0, mypy-0.8.1
```
That's not the only test for leaks or the only test for leaks involving threads in the test suite, so I wonder why. Do these tests fail for you? It seems they didn't in the workflow.

Also I see that `--cov` is tripped by any interface method `...` and any `if TYPE_CHECKING`. Is it possible to remove this noise? But I really don't want to pepper the source code with `# pragma nocover` and :hankey: like that...

dlax:
> For some reason, with `--cov` enabled, three tests fail for me:
> 
> ```
> FAILED tests/test_copy.py::test_copy_from_leaks[Format.TEXT-False] - AssertionError: objects leaked: 2, 2
> FAILED tests/test_copy.py::test_copy_from_leaks[Format.BINARY-True] - AssertionError: objects leaked: 2, 2
> FAILED tests/test_copy.py::test_copy_from_leaks[Format.TEXT-True] - AssertionError: objects leaked: 2, 2
> ```
> [...]
> Do these tests fail for you?

Yes, I see these as well. This occurs as well without the pytest plugin (e.g. `coverage run $(which pytest) tests/test_copy.py`). Appart from that, no idea what's going on...

> Also I see that `--cov` is tripped by any interface method `...` and any `if TYPE_CHECKING`. Is it possible to remove this noise? But I really don't want to pepper the source code with `# pragma nocover` and hankey like that...

Good point; I've excluded those lines from the report.

dvarrazzo:
FYI the leaking objects are:
```
<bound method Collector.disable_plugin of <Collector at 0x7f5022132d60: CTracer>>
<bound method Collector.switch_context of <Collector at 0x7f5022132d60: CTracer>>
```

dvarrazzo:
Merging this branch as it is, because it may be handy and the test issue is not a problem as we don't include coverage in the CI.

WIth the current excludes, running pytest with PSYCOPG_IMPL=python covers 95% of the codebase. Not bad considering that we never run coverage and there is code meant for different database versions, different pq implementations...

dvarrazzo:
Open https://github.com/nedbat/coveragepy/issues/1283 about the issue with refcount in tests.

Issue #158, opened 2021-11-20
macOS: kqueue file descriptor leak


qwesda:
I was trying to pipe data from one server to another like this:

```python
with src_cur.copy("""COPY public.src_table TO STDOUT (FORMAT BINARY);""") as copy_src:
    with dst_cur.copy("""COPY public.dst_table FROM STDIN (FORMAT BINARY);""") as copy_dst:
        while data_row := copy_src.read():
            copy_dst.write(bytes(data_row))
```

and got the following error on bigger tables:

```
sending query failed: another command is already in progress
Traceback (most recent call last):
  File ".env/lib/python3.9/site-packages/psycopg/cursor.py", line 681, in copy
    yield copy
  File "src/run_psycopg3.py", line 27, in transfer_table_data
    while data_row := copy_src.read():
  File ".env/lib/python3.9/site-packages/psycopg/copy.py", line 204, in read
    return self.connection.wait(self._read_gen())
  File ".env/lib/python3.9/site-packages/psycopg/connection.py", line 767, in wait
    return waiting.wait(gen, self.pgconn.socket, timeout=timeout)
  File ".env/lib/python3.9/site-packages/psycopg/waiting.py", line 53, in wait_selector
    sel = DefaultSelector()
  File "/usr/local/Cellar/python@3.9/3.9.8/Frameworks/Python.framework/Versions/3.9/lib/python3.9/selectors.py", line 512, in __init__
    self._selector = select.kqueue()
OSError: [Errno 24] Too many open files
```

macOS 12.0.1 (21A559)
Python 3.9.8
psycopg-binary 3.0.4

I was adapting my code for psycopg3, previously I used the following pattern for piping data between servers. Not sure if I'm using the v3 interface wrong ...

```python
dst_fd, src_fd = os.pipe()

def copy_dst():
    with os.fdopen(dst_fd) as read_f:
        dst_cur.copy_expert(f"""COPY public.dst_table FROM STDIN;""", read_f)

dst_thread = threading.Thread(target=copy_dst)
dst_thread.start()

with os.fdopen(src_fd, 'w') as write_f:
    src_cur.copy_expert("""COPY public.src_table TO STDOUT;""", write_f)

dst_thread.join()
```

dvarrazzo:
Are both cursors open on the same connection?

dvarrazzo:
Is this something that happens at the first read/write or after a few loop cycles? How large is the file, or wide? Does it matter?

[...]

PR #160, opened 2021-11-21
adds a test `copy_to_large.py` for issue #158

qwesda:
As discussed in issue #158. This test should fire if file descriptors leak in large `COPY ... TO` operations.


dvarrazzo:
Closing this because the issue doesn't seem reliably reproducible.

PR #166, opened 2021-12-03
Add _PQencryptPasswordConn() in _pq_ctypes.pyi

dlax:
Running the generation script produces this diff.


dvarrazzo:
I'm not sure the underscored functions should be in that file. However `_PQhostaddr` has been there from the early days...

I'll merge this for sure, then try to figure out better what we should do.

PR #167, opened 2021-12-03
Add bindings for PQ tracing functions

dlax:
I'm trying to add bindings for `PQtrace()` and friends, https://www.postgresql.org/docs/14/libpq-control.html.
I think this would be really useful to get insights of what happens at protocol level, especially when developing things like the pipeline mode.

~~Unfortunately, I'm struggling with ctypes (at the moment, then cython later probably) because of the `FILE` type that's used by libpq and which I don't know how to interface with Python.~~

~~@dvarrazzo, do you have any idea how to do this kind of bindings properly?~~


dvarrazzo:
@dlax maybe you can define it as int and pass it the `file.fileno()`?

dlax:
Seems to work, thanks; though we need to use `libc.fdopen()`, it seems?

dlax:
Added the cython implementation.

dlax:
On second thought, I won't be able to invest time making this working on windows (or macos), so I flagged this at linux-only. Since this is mostly a feature useful for development, maybe that's enough.

dvarrazzo:
That sounds good. If anyone is interested on those platforms they may test themselves

tests/fix_db.py
```diff
@@ -93,12 +146,13 @@ async def aconn(dsn, request):
...
         pytest.skip(msg)
-    yield conn
+    with maybe_trace(conn.pgconn, tracefile, request.function):
+        yield conn
     await conn.close()
 
 
 @pytest.fixture(scope="session")
-def svcconn(dsn):
+def svcconn(request, dsn):
```
dvarrazzo, line 155: Is this change spurious?

tests/fix_db.py
```diff
@@ -51,8 +60,48 @@ def dsn(request):
...
+@contextmanager
+def maybe_trace(pgconn, tracefile, function):
+    """Handle libpq client/server communication traces for a single test
+    function.
+    """
+    if tracefile is None:
+        yield None
+        return
+
+    title = f" {function.__module__}::{function.__qualname__} ".center(80, "=")
```
dvarrazzo, line 86: That's very nice :D

tests/fix_db.py
dlax: Yes, fixed now.

dvarrazzo:
Very useful, thank you :)

Issue #145, opened 2021-11-10
Possible executemany() optimization


sekrause:
So I was playing around with the new `psycopg` and Wireshark to learn about PostgreSQL's extended query protocol and looked at the behavior of `executemany()` at the procotol level.

Currently `executemany()` sends one Parse/Sync message at the beginning and then a Bind/Describe/Execute/Sync message for each params in the params sequence, waiting the the server response after each message. This greatly limits the performance of `executemany()` by the roundtrip time to the server (at least when the latency is not negligible).

In [this message](https://www.postgresql.org/message-id/542A3D9D.30403%402ndquadrant.com) on the PostgreSQL mailing list it was mentioned that:

> You can and should send parse/bind/describe/execute messages back-to-back without waiting for a server response. Just Sync and wait for server response at the end.
> 
> **You can even send a parse then a stream of bind/describe/execute messages for batch execution of a prepared statement against a list of params, then a single Sync at the end.**

The last sentence describes a possible optimization for `executemany()`: Don't end individual Bind/Describe/Execute message with a Sync and wait for a server-response, just send large batches of Bind/Describe/Execute messages, followed by a single Sync at the end. This should reduce the performance penalty due to server round-trip time to a minimum.

[asyncpg](https://magicstack.github.io/asyncpg/current/) does exactly that: It puts many Bind/Execute commands into a single extended query message, only the final message of the batch has a Sync at the end.

I know that you're generally supposed to use `COPY` for large batch inserts (and the new `psycopg` makes it really easy), but sometimes (e.g. for batch updates) `executemany()` is more convenient.

dvarrazzo:
Hello @sekrause thank you for your analysis and your input.

psycopg works at libpq level, which is the C library that handles the underlying protocol. So if the functionality is not exposed we cannot reach for it.

@dlax is currently working on the pipeline/batch mode (see #116, #74) which allows psycopg to make use of certain libpq functionalities included in PostgreSQL 14.

I seem to understand that at the moment what you suggest to do is not available in libpq but it would once the batch mode is available, is that right?

Actually, thinking about it, the work in #116 is mostly an high level construct on top of the libpq 14 functionalities, but Denis has already made them available (see dc18d9452731307270aa88c060069cbe2393ff45).

So, do I understand correctly that the right way to optimise executemany is to make use of the newly available pipeline/batch libpq calls?

dvarrazzo:
[...]

PR #175, opened 2021-12-08
pipeline mode (v4)

dlax:
Third attempt to implement the pipeline mode (#74), replacing #116.

The main difference (with #116) is, as discussed, that we now send all commands in the `pipeline_communicate()` generator instead of just flushing there previously. We thus maintain a queue of commands to dispatch and a queue of results to be fetched (for further processing).

Secondly, we now check cursor results as we process/receive them. 

Aside, as it is not allowed to emit multiple queries per command (separated with a `;`) in pipeline mode, we adjust transaction code accordingly.

Also included more tests, especially those discussed in  https://github.com/psycopg/psycopg/pull/116#issuecomment-980816719 combining transaction and errors.

### DIFF
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Any
 import concurrent.futures
 
@@ -47,6 +48,16 @@ def test_pipeline_broken_conn_exit(conn: psycopg.Connection[Any]) -> None:
             closed = True
 
     assert closed
+
+
+def test_pipeline_exit_error_noclobber(conn, caplog):
+    caplog.set_level(logging.WARNING, logger="psycopg")
+    with pytest.raises(ZeroDivisionError):
+        with conn.pipeline():
+            conn.close()
+            1 / 0
+
+    assert len(caplog.records) == 1
 
 
 def test_cursor_stream(conn):
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -4,6 +4,7 @@ commands pipeline management
 
 # Copyright (C) 2021 The Psycopg Team
 
+import logging
 from types import TracebackType
 from typing import Any, List, Optional, Union, Tuple, Type, TYPE_CHECKING
 
@@ -37,6 +38,8 @@ else:
 PendingResult: TypeAlias = Union[
     None, Tuple["BaseCursor[Any, Any]", Optional[Tuple[Key, Prepare, bytes]]]
 ]
+
+logger = logging.getLogger("psycopg")
 
 
 class BasePipeline:
@@ -168,13 +171,19 @@ class Pipeline(BasePipeline):
             with self._conn.lock:
                 self.sync()
                 try:
-                    # Send an pending commands (e.g. COMMIT or Sync);
+                    # Send any pending commands (e.g. COMMIT or Sync);
                     # while processing results, we might get errors...
                     self._conn.wait(self._communicate_gen())
                 finally:
                     # then fetch all remaining results but without forcing
                     # flush since we emitted a sync just before.
                     self._conn.wait(self._fetch_gen(flush=False))
+        except Exception as exc2:
+            # Don't clobber an exception raised in the block with this one
+            if exc_val:
+                logger.warning("error ignored exiting %r: %s", self, exc2)
+            else:
+                raise
         finally:
             self._exit()
 
@@ -202,12 +211,18 @@ class AsyncPipeline(BasePipeline):
             async with self._conn.lock:
                 self.sync()
                 try:
-                    # Send an pending commands (e.g. COMMIT or Sync);
+                    # Send any pending commands (e.g. COMMIT or Sync);
                     # while processing results, we might get errors...
                     await self._conn.wait(self._communicate_gen())
                 finally:
                     # then fetch all remaining results but without forcing
                     # flush since we emitted a sync just before.
                     await self._conn.wait(self._fetch_gen(flush=False))
+        except Exception as exc2:
+            # Don't clobber an exception raised in the block with this one
+            if exc_val:
+                logger.warning("error ignored exiting %r: %s", self, exc2)
+            else:
+                raise
         finally:
             self._exit()
--- a/tests/test_pipeline_async.py
+++ b/tests/test_pipeline_async.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 from typing import Any
 
 import pytest
@@ -50,6 +51,16 @@ async def test_pipeline_broken_conn_exit(aconn: psycopg.AsyncConnection[Any]) ->
             closed = True
 
     assert closed
+
+
+async def test_pipeline_exit_error_noclobber(aconn, caplog):
+    caplog.set_level(logging.WARNING, logger="psycopg")
+    with pytest.raises(ZeroDivisionError):
+        async with aconn.pipeline():
+            await aconn.close()
+            1 / 0
+
+    assert len(caplog.records) == 1
 
 
 async def test_cursor_stream(aconn):


dlax:
Rebased and adjusted to properly handle `executemany()` results.

dlax:
(Rebased, and fixed formatting with new black.)

psycopg/psycopg/connection.py
```diff
@@ -851,6 +989,32 @@ def notifies(self) -> Generator[Notify, None, None]:
                 n = Notify(pgn.relname.decode(enc), pgn.extra.decode(enc), pgn.be_pid)
                 yield n
 
+    @contextmanager
+    def pipeline(self) -> Iterator[None]:
+        """Context manager to switch the connection into pipeline mode."""
+        if self._pipeline is not None:
+            raise e.ProgrammingError("already in pipeline mode")
+
+        pipeline = self._pipeline = Pipeline(self.pgconn)
```
dvarrazzo, line 998: Setting the value outside the lock seems unsafe: guarding it with https://github.com/psycopg/psycopg/commit/b19e48bf4c7ea86fa64bd5cd15def5c272bb8523

psycopg/psycopg/connection.py
```diff
@@ -851,6 +989,32 @@ def notifies(self) -> Generator[Notify, None, None]:
                 n = Notify(pgn.relname.decode(enc), pgn.extra.decode(enc), pgn.be_pid)
                 yield n
 
+    @contextmanager
+    def pipeline(self) -> Iterator[None]:
+        """Context manager to switch the connection into pipeline mode."""
+        if self._pipeline is not None:
+            raise e.ProgrammingError("already in pipeline mode")
```
dvarrazzo, line 996: Is it necessary to make pipeline non-reentering? I'm playing with a trivial optimisation of executemany for #145 by just wrapping the `cursor.executemany()` in a `conn.pipeline()` and it would be made more complex by checking whether we are or not in a pipeline already. How about https://github.com/psycopg/psycopg/commit/2f50a415ff7543329e01045e6b8ac4082df53761 ?

psycopg/psycopg/connection.py
```diff
@@ -851,6 +989,32 @@ def notifies(self) -> Generator[Notify, None, None]:
                 n = Notify(pgn.relname.decode(enc), pgn.extra.decode(enc), pgn.be_pid)
                 yield n
 
+    @contextmanager
+    def pipeline(self) -> Iterator[None]:
```
dvarrazzo, line 993: Any reason not to return the Pipeline object? It might be useful... if any for our tests :P

psycopg/psycopg/connection_async.py
```diff
@@ -35,6 +35,22 @@
 logger = logging.getLogger("psycopg")
 
 
+class AsyncPipeline(BasePipeline):
```
dvarrazzo, line 38: It doesn't seem there is really a need for Sync/Async pipeline at the moment? I guess it would be different if there were Queues instead of Deque and some blocking was involved here, but it seems that blocking only happens in the connection wait.

psycopg/psycopg/connection.py
```diff
@@ -79,6 +86,119 @@ class Notify(NamedTuple):
...
+PipelinePendingResult = Union[
+    None,
+    Tuple[
+        "BaseCursor[Any, Any]",
+        Optional[Tuple[Key, Prepare, bytes]],
+    ],
+]
+
+
+class BasePipeline:
```
dvarrazzo, line 98: I am moving the pipeline in a separate module.

dvarrazzo: @dlax I have introduced a `_pipeline` module. However, that way I have "won" the whole code. I am happy if you keep the attribution of most of the pipeline code, so I invite you to --reset-author on that commit and push it to your branch :)

psycopg/psycopg/cursor.py
```diff
@@ -218,20 +224,28 @@ def _executemany_gen(
...
 
-            for res in results:
-                nrows += res.command_tuples or 0
+            if self._conn._pipeline:
+                yield from self._conn._pipeline._communicate_gen()
+            else:
+                assert results is not None
+                self._check_results(results)
+                if returning and results[0].status == ExecStatus.TUPLES_OK:
+                    self._results.extend(results)
```
dvarrazzo, line 234: The management of results and rowcount in the pipeline block is broken with this change. I have fixed most of the most obvious problems in the [executemany-pipeline](https://github.com/psycopg/psycopg/tree/executemany-pipeline) branch, but the problem persists if executemany() is executed under a pipeline().

Maybe we could add a `self._fetch_pipeline()` call to nextset/rowcount? mmm, nah, it would worn in async mode... Anyway I wouldn't be desperate if it turns out that rowcount is unreliable in this case.

psycopg/psycopg/connection.py
dlax: Yes; looks reasonable to me, thanks.

psycopg/psycopg/connection.py
dlax: Done; thanks.

psycopg/psycopg/connection.py
dvarrazzo: Playing with executemany tests, yesterday, I have noticed that the behaviour changes if we run:
```
cur.executemany()  # calls pipeline internally
cur.rowcount  # has received all the results
```
or
```
with conn.pipeline()
    cur.executemany()  # internal pipeline is no-o[
    cur.rowcount  # results are not returned
```
I would have expected that [triggering a flush](https://github.com/psycopg/psycopg/commit/a7ab0fdc615084846f0c65b86a2c2895f0f69488#diff-38c84614fefee615d2c96ae4cb1fec9e483f82ffc1d9606040f91e5fa8f2c809R237) would have made the results reach the cursor. Am I wrong expecting that? Do you think it's possible to reach that "state"? If not, maybe making `pipeline()` transparent to re-enter is not such a good idea.

psycopg/psycopg/connection.py
dlax: Done in 032c43b3a030ca01aa8ff63305d002920aef95b5

psycopg/psycopg/connection_async.py
dlax: Done in 5465b9248d331fc9871adc7a61ce88db08417aa1.

dlax:
@dvarrazzo, I need more time to think about the executemany() issues that you raised. In the meantime, in  a00a8e71562fba38a2ada5bb84f54cecb696ff71, I added a `--many` flag to `tests/scripts/pipeline-demo.py` to run the demo with `executemany()`, in case that helps.

psycopg/psycopg/connection_async.py
dvarrazzo: Cool! It's not impossible that in the future we will split them: it depends if they acquire blocking code

psycopg/psycopg/connection.py
```diff
@@ -851,6 +865,42 @@ def notifies(self) -> Generator[Notify, None, None]:
...
+            with pipeline:
+                try:
+                    yield pipeline
+                finally:
+                    with self.lock:
+                        pipeline.sync()
+                        try:
+                            # Send an pending commands (e.g. COMMIT or Sync);
+                            # while processing results, we might get errors...
+                            self.wait(pipeline._communicate_gen())
```
dvarrazzo, line 895: Talking about sync/async pipelines: maybe instead it would be better to move this code into the pipelines [a]exit? It uses private-ish pipeline methods and public connection ones...

dvarrazzo: What do you think about https://github.com/psycopg/psycopg/commit/fbb027c0f355ae0f2a02258b8795d9819cdeeae2 ?

tests/test_pipeline.py
```diff
@@ -0,0 +1,259 @@
...
+
+            # PQsendQuery[BEGIN], PQsendQuery
+            assert len(p.result_queue) == 2
+
+        assert cur.fetchone() == (1,)
+
+
+def test_pipeline_errors_processed_at_exit(conn):
+    conn.autocommit = True
+    with pytest.raises((e.OperationalError, e.UndefinedTable)):
```
dvarrazzo, line 65: It doesn't seem that we need to catch `OperationalError` here, right? Or is it raised in some condition?
```suggestion
    with pytest.raises(e.UndefinedTable):
```

psycopg/psycopg/connection.py
dlax: But then we need to distinguish back `Pipeline` and `AsyncPipeline` thus making 5465b9248d331fc9871adc7a61ce88db08417aa1 pointless?

dvarrazzo: Yes, I've based this commit on the commit before dropping sync/async pipeline. This change starts making the distinction of the two classes meaningful.

dlax: Ok; sounds good, pushed.

dvarrazzo:
I have pushed a few other changes to our side of the `pipeline4` branch. They are improvement/fixes found worked to the executemany-pipeline branch.

Which now works :tada: Should give it a benchmark...

psycopg/psycopg/connection.py
dlax: > I would have expected that [triggering a flush](https://github.com/psycopg/psycopg/commit/a7ab0fdc615084846f0c65b86a2c2895f0f69488#diff-38c84614fefee615d2c96ae4cb1fec9e483f82ffc1d9606040f91e5fa8f2c809R237) would have made the results reach the cursor. Am I wrong expecting that? Do you think it's possible to reach that "state"?

The flush tells the server that it can start returning result from now on, but the client still has to explicitly fetch those (calling `PQgetResult`). At the end of the pipeline, we fetch all results pending; this probably explains the difference you observed depending on whether the pipeline has exited or not when querying `.rowcount`. Maybe calling `cur.fetch*()` helps?

dlax:
Excellent!

Happy to review the docs, thanks!

psycopg/psycopg/cursor.py
dlax: I'm working on fixing `rowcount` with executemany() and pipeline; will probably send a PR on top of #260 today.

PR #195, opened 2022-01-02
Change type of `wait` in `pool_async.py`

sobolevn:
Hi!

Our CI in typeshed identified a future error in your code: https://github.com/python/typeshed/pull/6783

So, I am doing a pro-active PR to fix that 🙂 

Local test that it works (with the future stubs):
Снимок экрана 2022-01-02 в 15 00 39

### DIFF
--- a/psycopg_pool/psycopg_pool/pool_async.py
+++ b/psycopg_pool/psycopg_pool/pool_async.py
@@ -239,10 +239,11 @@ class AsyncConnectionPool(BasePool[AsyncConnection[Any]]):
 
         # Wait for the worker threads to terminate
         wait = asyncio.gather(self._sched_runner, *self._workers)
-        if timeout > 0:
-            wait = asyncio.wait_for(asyncio.shield(wait), timeout=timeout)
         try:
-            await wait
+            if timeout > 0:
+                await asyncio.wait_for(asyncio.shield(wait), timeout=timeout)
+            else:
+                await wait
         except asyncio.TimeoutError:
             logger.warning(
                 "couldn't stop pool %r tasks within %s seconds",


psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -238,7 +238,9 @@ async def close(self, timeout: float = 5.0) -> None:
             await conn.close()
 
         # Wait for the worker threads to terminate
-        wait = asyncio.gather(self._sched_runner, *self._workers)
+        wait: Awaitable[List[Any]] = asyncio.gather(
+            self._sched_runner, *self._workers
+        )
         if timeout > 0:
             wait = asyncio.wait_for(asyncio.shield(wait), timeout=timeout)
```
dlax, line 245: I'm not sure this is correct. The issue seems to be that the `wait` name is used in two assignments with different types. Namely, if I understand correctly, `asyncio.gather()` would indeed return a list of results, but `asyncio.wait_for()` would only return one result value. Perhaps a `# type: ignore[assignment]` would be more appropriate?

sobolevn: @dlax yes, this is quite complex 🙂 

Let's dig into it:
- `asyncio.gather()` returns `Future[List[...]]`, but `await asyncio.gather()` returns `List[...]` (there are some cases when `gather` might return `tuple[]` type, but this is a limitation of typeshed
- Then, we reassign the existing `wait` with `asyncio.wait_for()` which will be defined as `async def wait_for(fut: _FutureT[_T], timeout: float | None) -> _T: ...`, in other words the return type of `wait_for()` call will be `Coroutine[Any, Any, _T]` where `T=List[Any]` in our case

I don't agree that `# type: ignore[assignment]` is better. Here's a quick demo why:

```python
x = 1
x = 'a'  # type: ignore[assignment]

print(x.bit_count)  # passes (but should not!), because `x` is now `Any`
```

In our case `wait` is only used for `await wait`, so `Awaitable` should be fine. At least Mypy agrees with me 🙂

dvarrazzo: I agree with @dlax that the issue is related to the reuse of `wait`. How about:
```diff
diff --git a/psycopg_pool/psycopg_pool/pool_async.py b/psycopg_pool/psycopg_pool/pool_async.py
index 80dc00f8..597f306a 100644
--- a/psycopg_pool/psycopg_pool/pool_async.py
+++ b/psycopg_pool/psycopg_pool/pool_async.py
@@ -239,10 +239,11 @@ class AsyncConnectionPool(BasePool[AsyncConnection[Any]]):
 
         # Wait for the worker threads to terminate
         wait = asyncio.gather(self._sched_runner, *self._workers)
-        if timeout > 0:
-            wait = asyncio.wait_for(asyncio.shield(wait), timeout=timeout)
         try:
-            await wait
+            if timeout > 0:
+                await asyncio.wait_for(asyncio.shield(wait), timeout=timeout)
+            else:
+                await wait
         except asyncio.TimeoutError:
             logger.warning(
                 "couldn't stop pool %r tasks within %s seconds",
```

sobolevn: Done! 👍

dvarrazzo:
Thank you for the contribution, @sobolevn!

Squashing, rebasing, merging after tests pass.

sobolevn:
@dvarrazzo happy to help! I am a long-time-user of `psycopg`, thanks a lot for maintaining it! 👍

PR #196, opened 2022-01-03
Pool open

dvarrazzo:
Hello @dlax!

I have resurrected #151 because the `open()` method is good to have in the API. I have added an `open` parameter to init, defaulting to `true`, so the behaviour is not changed and we can release this in 3.1.

In a future 4.0, if we think it's useful, we could change the default for open to False. [As you mentioned](https://github.com/psycopg/psycopg/pull/151#discussion_r753846910), an anyio-based pool might even have a different default for it.

This MR is largely made of your commits for #155, with the addition of the init open param, docs, better error messages. Happy to hear your opinion about merging and releasing this for 3.1.

### DIFF
--- a/docs/api/pool.rst
+++ b/docs/api/pool.rst
@@ -145,6 +145,8 @@ The `!ConnectionPool` class
 
           # the connection is now back in the pool
    
+   .. automethod:: open
+
    .. automethod:: close
 
       .. note::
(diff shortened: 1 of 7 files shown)


dlax:
Looks good to me. Thanks for moving this forwards!

PR #197, opened 2022-01-03
Run pytest with increased verbosity in CI

dlax:
This way, we can watch individual tests progression (instead of per test
file previously) in github's CI report (making it easier to diagnose
possible blocking/timeouts for instance).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       PSYCOPG_TEST_DSN: "host=127.0.0.1 user=postgres"
       PGPASSWORD: password
-      PYTEST_ADDOPTS: --color yes
+      PYTEST_ADDOPTS: --color yes -vv
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


dvarrazzo:
We have more then 3k tests already. The output seems excessively long (512k)...

https://github.com/psycopg/psycopg/runs/4693530032?check_suite_focus=true

I don't think that so far I have had the need to diagnose a timeout in CI. Is it happening to you in some branch of yours?

Maybe we could do the same, when we need, using  `PYTEST_ADDOPTS=-v` as an env var configured in the workflow run, and do what we currently do with it in some other way? Or introducing another env var to use in tox? That's assuming that the problem you are diagnosing is reproducible of course.

dlax:
It's happening here https://github.com/psycopg/psycopg/actions/runs/1648974441 and has happened to me a number of times before.

In fact, I just always run tests with `-vv` myself in CI (and even locally) precisely because I've been trapped too many times with a truncated output.

> Maybe we could do the same, when we need, using `PYTEST_ADDOPTS=-v` as an env var configured in the workflow run, and do what we currently do with it in some other way? Or introducing another env var to use in tox?

I don't quite understand what you suggest here.

dlax:
In fact, it seems that the verbose mode does not really help; i.e. I still see "stalled" jobs without any way to check progress until timeout... So I agree this brings no benefit.

PR #209, opened 2022-01-16
Run tests with pip constraints on Sundays in CI

dlax:
We add a scheduled job in github actions at 18:19 UTC, on Sunday that
will run tests with pip constraints.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,7 @@ on:
   pull_request:
   schedule:
     - cron: '48 6 * * *'
+    - cron: '19 18 * * 0'
 
 jobs:
 
@@ -36,11 +37,15 @@ jobs:
       PSYCOPG_TEST_DSN: "host=127.0.0.1 user=postgres"
       PGPASSWORD: password
       PYTEST_ADDOPTS: --color yes
-      # Enable to run tests using the minumum version of dependencies.
-      # PIP_CONSTRAINT: ${{ github.workspace }}/tests/constraints.txt
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: "Set pip constraints"
+        if: github.event.schedule == '19 18 * * 0'
+        run: echo "Running with oldest version of dependencies (Pip constraints)"
+        env:
+          PIP_CONSTRAINT: ${{ github.workspace }}/tests/constraints.txt
 
       - uses: actions/setup-python@v2
         with:
@@ -132,7 +137,6 @@ jobs:
       # Don't run timing-based tests as they regularly fail.
       # pproxy-based tests fail too, with the proxy not coming up in 2s.
       PYTEST_ADDOPTS: -m 'not timing and not proxy and not mypy' --color yes
-      # PIP_CONSTRAINT: ${{ github.workspace }}/tests/constraints.txt
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -182,11 +192,16 @@ jobs:
...
-      # PIP_CONSTRAINT: ${{ github.workspace }}/tests/constraints.txt
 
     steps:
       - uses: actions/checkout@v2
 
+      - name: "Set pip constraints"
+        if: github.event.schedule == '19 18 * * 0'
+        run: echo "Running with oldest version of dependencies (Pip constraints)"
+        env:
+          PIP_CONSTRAINT: ${{ github.workspace }}/tests/constraints.txt
```
dvarrazzo, line 203: Are you sure that an env set in a step affects the other steps? I doubt it does, I would think they are isolated

.github/workflows/tests.yml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
   schedule:
     - cron: '48 6 * * *'
+    - cron: '19 18 * * 0'
```
dvarrazzo, line 12: If we do this we should change the postgis version of Python to 3.8, as explained in the floating commit to check the minimum version so far: the shapely version 1.7.0 specified as minimum requirements has only binary packages up to 3.8. Other Python versions require Shapely to be installed from source, but this requires the libgeos-dev library installed.

.github/workflows/tests.yml
```diff
@@ -143,6 +147,12 @@ jobs:
       - name: Start PostgreSQL service for test
         run: brew services start postgresql
 
+      - name: "Set pip constraints"
+        if: github.event.schedule == '19 18 * * 0'
```
dvarrazzo, line 151: This is clever, although brittle :smile:

dvarrazzo (changes requested):
Hi @dlax, thank you for this.

I was thinking to do it as well, but it wasn't immediately obvious for me how to do it. I am not sure that setting the env in a step also sets it in other step, I would honestly be surprised if it did. Did you try it? [Docs only mention](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsenv)  that a more specific env overrides a more generic.

The right place to set that env var - if I had defined the github action syntax myself - would be the `schedule` entry, but that [doesn't seem supported](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onschedule).

Another thing that would have been great would have been to define the wrokflow in a file to import from other files, overriding the env vars to customize different runs. Pretty easy to do in GitLab but it doesn't seem possible in Github either... Or maybe no, that's not true: I see [reusable workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows). Maybe those could work better?

Having this lower bound checked periodically is useful, but less critical than the upper bound. So, if we can schedule it easily, that's ok for me, but if it's not easily possible I would be ok to leave it in a floating branch to rebase periodically.

dlax:
> I was thinking to do it as well, but it wasn't immediately obvious for me how to do it. I am not sure that setting the env in a step also sets it in other step, I would honestly be surprised if it did. Did you try it? [Docs only mention](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsenv) that a more specific env overrides a more generic.

I tried it and thought it was working but now double-checking it actually does not. (I got trapped by the fact that most dependencies in the constraints file were already the latest -- only pytest-randomly has a newer version.)



> Having this lower bound checked periodically is useful, but less critical than the upper bound. So, if we can schedule it easily, that's ok for me, but if it's not easily possible I would be ok to leave it in a floating branch to rebase periodically.

I see; makes sense.

I'll think a bit more about this later on this week and, if no solution appears, I agree, we can live with that.

dlax:
All things considered, I'm withdrawing this.

PR #211, opened 2022-01-19
Codespell

The-One-And-Only-H:
- Added and configured Codespell to `tox.ini`
- Added Codespell to `lint.yml` pipeline
- Corrected spelling mistakes both with Codespell and manually

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = black, flake8, mypy
+envlist = black, flake8, mypy, codespell
 isolated_build = True
 
 [testenv]
@@ -26,6 +26,11 @@ commands = sphinx-build -W -T -b html docs docs/_build/html
 deps =
     ./psycopg[docs]
     ./psycopg_pool
+
+[testenv:codespell]
+commands = codespell -L "alot,ans,ba,fo,te" docs psycopg* tests
+deps =
+    codespell
 
 [flake8]
 max-line-length = 85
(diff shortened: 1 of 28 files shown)


dvarrazzo:

tox.ini
```diff
@@ -27,6 +27,11 @@ deps =
     ./psycopg[docs]
     ./psycopg_pool
 
+[testenv:codespell]
+commands = codespell -L "alot,ans,ba,fo,te" docs psycopg* tests
```
dlax, line 31: I've read about "alot" elsewhere, but what are those strange other words?

dvarrazzo: *alot* is a misspelling but I put there on purpose, just to have some fun during the long hours writing documentation... :smile: 

Others are variable names and the likes I guess. `ans` is the DNS answer, `ba` is for `bytearray`... The use is localised and obvious and I don't feel a compelling need to fix them just to make the tool happy. Other are small strings use in tests: they might be changed but again I feel no pressure about it.

One that was uglier was fixed in the first commit here, that `noe` which is better called `no_e`, a function that strips the prefix from postgres `E'string'` for the sake of tests.

This codespell tool is nice: it doesn't report errors on ay asdfasdf that it doesn't understand but still does a good job to tell apart misspelling of human words. I don't know its rate of false negatives but the false positives seem very few.

dlax: I see, thanks for clarifying; makes sense.

I'm totally on board with codespell as well.

PR #217, opened 2022-01-29
fix: confusing named parameters in example query

adelorn44:
In the named parameters example query, the name of the parameters are Python built in data types: `int` and `str`.

I think this is confusing because it looks like we are casting values passed as arguments instead of passing named arguments.

This MR replaces `int`, `str` and `date` arguments by keys `id`, `last_name` and `created_at` to remove the possible confusion.

### DIFF
--- a/docs/basic/params.rst
+++ b/docs/basic/params.rst
@@ -30,7 +30,7 @@ example the Python function call:
 .. code:: python
 
     cur.execute("""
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you, I appreciate the suggestion and thank you very much for the merge request :)

Issue #246, opened 2022-03-17
About detecting disconnections


octopoulpe:
Hi. Little question here.

It's really (really !) nice to provide a recipe using selectors to detect disconnections. Still something is unclear to me.
From what I understand, the `select 1` will trigger after every query done on the socket (eventually multiple times in the middle of a server-side cursor ?).
Am I wrong here ? If not, wouldn't it be good to warn about the overhead ?

Would a thread/coro-safe hook (handler, callback, other) on the connection (to be notified about every OperationalError) be a good pattern to manage application state, pool cleanup etc. ?

dvarrazzo:
Hi @octopoulpe 

I assume you mean [this recipe](https://www.psycopg.org/psycopg3/docs/advanced/async.html#detecting-disconnections)?

In that recipe it is assumed that a connection is only used in order to detect it breaking and not used for other operations. That's why it specifies "using a dedicated connection".

I see that this sentence is just thrown there towards the end and in parentheses... so probably it's not given the importance it deserves. Maybe you can suggest some rewording to make it clearer? 

By the way, this makes me think that maybe there is some space here to add a disconnection detection feature to the pool. Using a connection can be wasteful: if the operation is done from inside the pool then, with some care, the connection might be still given to a client... Added #247 to think about it.

octopoulpe:
Oh @dvarrazzo I'm so sorry it was indeed really clear.

PR #248, opened 2022-03-17
doc: give more context to disconnection detection recipe

octopoulpe:
following #246 to avoid further useless questions :)


dvarrazzo:
Thank you, very appreciated. And I don't think the question was useless: this could have totally been clarified better :)

PR #250, opened 2022-03-19
Fixed documentation wrong word

xenups:
Fixed variable names
converted _return_params to static function
converted sort_rfc2782 to static function
fixed local variable might be referenced before assignment
Fixed ambiguous argument name is_ip_address function

### DIFF
--- a/psycopg/psycopg/_dns.py
+++ b/psycopg/psycopg/_dns.py
@@ -272,7 +272,7 @@ class Rfc2782Resolver:
         return self._get_solved_entries(hp, ans)
 
     def _get_solved_entries(
-            self, hp: HostPort, entries: "Sequence[SRV]"
+        self, hp: HostPort, entries: "Sequence[SRV]"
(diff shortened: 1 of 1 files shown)


dvarrazzo:
I assume you run some sort of linter on the code?

xenups:
yes unfortunately I forgot to fix the black format, so is it ok to make pr again?

dvarrazzo:
Hi @xenups,

What I meant, did you run some refactoring tool that made those changes for you, such as renaming i -> index or assuming that, because self is not used in that function, it can be converted to a static one?

"fixed local variable might be referenced before assignment" is not a thing that can happen if you look at the code. That list comes from a defaultdict whose entries are created using append. This means that there is no empty list there, so it is not possible for index to be unassigned.

I am not interested in the code change proposed. If you can make a MR just to fix the documentation, I will accept it, but please leave the code as it is.

Issue #242, opened 2022-03-11
Properly catch OperationalError 


philpep:
Hi,

I'm trying to catch `OperationalError('connection failed: fe_sendauth: no password supplied')`

Exception message is unreliable for this since it might be localized. And I see no other field on exception to filter on:

```
>>> exc.sqlstate
None
```

And all fields on `exc.diag` are None, looks like a bug no ?

This is on `psycopg 3.0.9`

Thanks!

dvarrazzo:
No, it's not a bug. sqlstate, as well as other error fields, are not available for any error. The specific error you mention is generated by the client library, not by the server, and I doubt that the client returns any SQLSTATE (if it does, it is more an exception than a rule).

Looking at the libpq code, it doesn't seem that the message in question undergoes any l10n

```
piro@baloo:~/dev/fs/postgres$ ag -C2 PQnoPasswordSupplied src/interfaces/libpq/
src/interfaces/libpq/libpq-fe.h
557-/* Error when no password was given. */
558-/* Note: depending on this is deprecated; use PQconnectionNeedsPassword(). */
559:#define PQnoPasswordSupplied	"fe_sendauth: no password supplied\n"
560-
561-/* Create and manipulate PGresults */

src/interfaces/libpq/fe-auth.c
559-	{
560-		appendPQExpBufferStr(&conn->errorMessage,
561:							 PQnoPasswordSupplied);
562-		goto error;
563-	}
--
1031-				{
1032-					appendPQExpBufferStr(&conn->errorMessage,
1033:										 PQnoPasswordSupplied);
1034-					return STATUS_ERROR;
1035-				}
```

However relying on the message is certainly brittle.

The `PGconn` object offers functions to detect if a password was required but not provided. This interface is stable, although not documented yet (the `needs_password` attribute is a wrapper for the `PQconnectionNeedsPassword()` mentioned in the libpq comment above):

```python
>>> cnn = psycopg.pq.PGconn.connect(b"host=localhost user=foo")
>>> cnn
<psycopg_c.pq.PGconn [BAD] at 0x7f49d78db310>

>>> psycopg.pq.misc.error_message(cnn)
'fe_sendauth: no password supplied'

cnn.needs_password
True

cnn.password
b''
```

However, when `Connection.connect()` fails, the underlying `PGconn` object is discarded. We might attach the `PGconn` to the exception object instead, to make it available in this case, but this is a new feature, not a bugfix.

[...]

PR #251, opened 2022-03-19
docs: document PGconn needs_password and used_password attributes

dlax:
This improves the documentation of Error class, which refers to those
in the new pgconn attribute documentation.

(The base branch is `pgconn-in-exception`.)

* * *
The rest of `pgconn-in-exception` branch looks good to me and would certainly help with #242 (the OP is  a colleague of mine).

### DIFF
--- a/docs/api/errors.rst
+++ b/docs/api/errors.rst
@@ -53,7 +53,7 @@ These classes are exposed both by this module and the root `psycopg` module.
         Most likely it will be in `~psycopg.pq.ConnStatus.BAD` state;
         however it might be useful to verify precisely what went wrong, for
         instance checking the `~psycopg.pq.PGconn.needs_password` and
-        `~psycopg.pq.PGconn.used_password`.
+        `~psycopg.pq.PGconn.used_password` attributes.
 
         .. versionadded:: 3.1
 
--- a/psycopg/psycopg/pq/pq_ctypes.py
+++ b/psycopg/psycopg/pq/pq_ctypes.py
@@ -240,10 +240,19 @@ class PGconn:
 
     @property
     def needs_password(self) -> bool:
+        """True if the connection authentication method required a password,
+        but none was available.
+
+        See :pq:`PQconnectionNeedsPassword` for details.
+        """
         return bool(impl.PQconnectionNeedsPassword(self._pgconn_ptr))
 
     @property
     def used_password(self) -> bool:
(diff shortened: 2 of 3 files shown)


dvarrazzo:
Thank you for reviewing #242! My main doubt is whether that approach is the best one to expose "failed objects". Do you see alternative ways of doing it?

docs/api/errors.rst
```diff
@@ -52,8 +52,8 @@ These classes are exposed both by this module and the root `psycopg` module.
 
         Most likely it will be in `~psycopg.pq.ConnStatus.BAD` state;
         however it might be useful to verify precisely what went wrong, for
-        instance checking the `~psycopg.pq.PGconn.needs_password` and
-        `~psycopg.pq.PGconn.used_password`.
+        instance checking the :attr:`~psycopg.pq.PGconn.needs_password` and
+        :attr:`~psycopg.pq.PGconn.used_password` attributes.
```
dvarrazzo, line 56: The `:attr:` shouldn't be necessary, because the default interpreted text role is `:obj:`.

psycopg/psycopg/pq/pq_ctypes.py
```diff
@@ -240,10 +240,14 @@ def backend_pid(self) -> int:
...
     @property
     def needs_password(self) -> bool:
+        """True if the connection authentication method required a password,
+        but none was available.
+        """
         return bool(impl.PQconnectionNeedsPassword(self._pgconn_ptr))
 
     @property
     def used_password(self) -> bool:
+        """True if the connection authentication method used a password."""
```
dvarrazzo, line 250: Maybe we can add the links to the libpq documentation? There is a `:pq:` interpreted role for it, see `get_cancel()` docstring for details.

dvarrazzo (changes requested):

psycopg/psycopg/pq/pq_ctypes.py
dlax: Indeed, thanks for the tip!

dlax:
> My main doubt is whether that approach is the best one to expose "failed objects". Do you see alternative ways of doing it?
>
No, I don't have a better idea. But even if we find one later on, having 
those attributes exposed would not hurt I think.

dvarrazzo:
Perfect, thank you! I think this branch can be merged.

PR #256, opened 2022-03-24
Fix infinite loop in copy and worker thread swallowing exceptions

### LINKED ISSUE #255: `write_row` busy loop
Found some weird behavior in version 3.0.10 (in both C and python variant, installed from pip).

This seems never to return, it sticks at 100% python CPU with no postgres activity:
```python
def provoke(conn):
    txt = '\\' * 536870883
    c = conn.cursor()
    c.execute('create temporary table ttt(t1 text, t2 text, t3 text, t4 text, t5 text)')
    with c.copy('COPY ttt (t1,t2) FROM STDIN') as copy:
        copy.write_row([txt] * 2)
    c.execute('select length(t1), length(t2), length(t3) from ttt')
    print(c.fetchall())
    c.execute('drop table ttt')
provoke(connection)
```
Have not yet further debugged, maybe some sort of busy loop in text escaping solely on python side. The value size was chosen for typical postgres limits, which should raise here, but it never gets to postgres side of things.

dvarrazzo:
Close #255

### DIFF
--- a/psycopg/psycopg/copy.py
+++ b/psycopg/psycopg/copy.py
@@ -51,8 +51,22 @@ class BaseCopy(Generic[ConnectionType]):
     """
 
     # Max size of the write queue of buffers. More than that copy will block
-    # Each buffer around Formatter.BUFFER_SIZE size
+    # Each buffer should be around BUFFER_SIZE size.
     QUEUE_SIZE = 1024
+
+    # Size of data to accumulate before sending it down the network. We fill a
+    # buffer this size field by field, and when it passes the threshold size
+    # wee ship it, so it may end up being bigger than this.
+    BUFFER_SIZE = 32 * 1024
+
+    # Maximum data size we want to queue to send to the libpq copy. Sending a
+    # buffer too big to be handled can cause an infinite loop in the libpq
+    # (#255) so we want to split it in more digestable chunks.
+    MAX_BUFFER_SIZE = 4 * BUFFER_SIZE
+    # Note: making this buffer too large, e.g.
+    # MAX_BUFFER_SIZE = 1024 * 1024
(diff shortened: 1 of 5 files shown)


jerch:
@dvarrazzo 
From my side this gets a LGTM, thx for the fast fix. :rocket:

About the 2**20 chunk size limit - this might be too small to fully saturate the socket in the end - in case the network is very fast, but still has some latency (which I'd call a typical data center setup, also see [bandwidth delay product](https://en.wikipedia.org/wiki/Bandwidth-delay_product)). A safer bet might be 16MB or even 64MB, it also would reduce additional memcpy needs.

I am still somewhat unhappy with one aspect, with no easy solutions yet in sight:
- high mem consumption from out buffer limit, for no good reason with COPY FROM - Currently there is no way to force libpq not to fill up the buffer, if `pqSendSome` cannot keep up. This can be a show stopper for cluster applications with strict memory limits (this is actually one of our use cases). If you are interested in limiting this somehow, maybe we can discuss it in another issue or discussion.

psycopg/psycopg/copy.py
```diff
@@ -270,15 +271,20 @@ def finish(self, exc: Optional[BaseException]) -> None:
     def worker(self) -> None:
         """Push data to the server when available from the copy queue.
 
-        Terminate reading when the queue receives a None.
+        Terminate reading when the queue receives an false-y value, or in case
```
dlax, line 274: (typo) _a_ false-y value

jerch:
> Also, reduce the chunk appended to the libpq from 1Mb to 128K. This
makes an *immense* difference: the too large chunk probably triggers
some quadraticity in the libpq.

:scream: No clue what going there, guess its time to build libpq with debug flags...

Issue #164, opened 2021-11-23
Returning `executemany()`


dvarrazzo:
In #163 we brushed on the idea of having `executemany()` returning the resulting results.

Let's explore this idea.

One possible interface already available could be [`Cursor.nextset()`](https://www.psycopg.org/psycopg3/docs/api/cursors.html#psycopg.Cursor.nextset). The results might be retrieved only if the queries return `TUPLES_OK`: for instance INSERT... RETURNING or SELECT function() with a side effect, discarded for `COMMAND_OK` results.

Note that the user might accidentally consume a lot of memory if they are not aware of the feature. However it seems an unlikely occurrence. We might add a `returning: bool` argument to control it, but I'd be comfortable enough to default it to `True`.

dvarrazzo:
Like in #145, this feature might take advantage of the pipeline/batch mode for performance. Probably the pq-level functions are enough for it.

eheu:
sorry for nitpicking - just occured to me, maybe the default `returning` value should be `False` so there wont be a performance hit just from upgrading to 3.1? admittedly only however slight the cost of having the result set returned is, but still...

dvarrazzo:
Hello @eheu thank you.

We are still in time to change it. However the memory would be only used in case executemany runs a query returning a result, i.e. people use a SELECT with side effect or an INSERT... RETURNING.  There is not extra penalty when, for instance, a query with just an INSERT is executed.

So I think the application affected would be relatively few, ones executing massive INSERT... RETURNING or SELECT in an executemany and discarding the result anyway, because there wasn't a way to access it previously. The downside for these apps is that the memory of these results is held for the time between the executemany and the next time the cursor is used again, or it is closed. There is no performance impact because, even previously, the results were transmitted to the client anyway - just to be discarded immediately.

So all in all it seems a low risk change for the benefit of a more convenient interface. I don't see `returning=False` very used, unless people know that the executemany is returning massive but unneeded results. In that case, by the way, it would make much more sense to reformulate the query in order to not return the result at all, because regardless of the `returning` value, and both in psycopg 2 and 3, the result values are transferred back to the client.

dvarrazzo:
@eheu I have changed the default to be False, based on the fact that if the user didn't ask us for something we shouldn't do extra work :) We are improving executemany performance by using libpq batch mode (#260) and the extra fetch might also imply slowing the pipeline, so we will return the results only if users really want it.

Thank you again for your input!

PR #260, opened 2022-03-28
Executemany pipeline

dvarrazzo:
Optimization of `Cursor.executemany()` using pipelines (#145). 

This branch is based on @dlax pipeline4 branch (#175) and will be merged after the other MR is.

Speedup looks in the region of the 2.5x improvement.

```python
# test-execmany.py 
import psycopg
cnn = psycopg.connect("host=localhost", autocommit=True)
cur = cnn.cursor()
cur.execute("""
    create temp table testmany (
        id serial primary key, n1 int, n2 int, s1 text, s2 text)
    """)
cur.executemany(
    "insert into testmany (n1, n2, s1, s2) values (%s, %s, %s, %s)",
    [(20, 5000, "foobar", "foobar" * 100)] * 100000,
)
```

```
$ time python test-execmany.py 
real	0m8.586s
user	0m3.416s
sys	0m1.469s

$ git co -
Switched to branch 'executemany-pipeline'

$ time python test-execmany.py 
real	0m3.347s
user	0m2.432s
sys	0m0.903s
```

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -63,11 +63,6 @@ class BasePipeline:
     def status(self) -> pq.PipelineStatus:
         return pq.PipelineStatus(self.pgconn.pipeline_status)
 
-    def sync(self) -> None:
-        """Enqueue a PQpipelineSync() command."""
-        self.command_queue.append(self.pgconn.pipeline_sync)
-        self.result_queue.append(None)
-
     @staticmethod
     def is_supported() -> bool:
         """Return `True` if the psycopg libpq wrapper suports pipeline mode."""
@@ -84,6 +79,17 @@ class BasePipeline:
     def _exit(self) -> None:
         if self.pgconn.status != ConnStatus.BAD:
             self.pgconn.exit_pipeline_mode()
+
+    def _sync_gen(self) -> PQGen[None]:
+        self._enqueue_sync()
+        try:
+            # Send any pending commands (e.g. COMMIT or Sync);
+            # while processing results, we might get errors...
+            yield from self._communicate_gen()
+        finally:
+            # then fetch all remaining results but without forcing
+            # flush since we emitted a sync just before.
+            yield from self._fetch_gen(flush=False)
 
     def _communicate_gen(self) -> PQGen[None]:
         """Communicate with pipeline to send commands and possibly fetch
@@ -107,7 +113,8 @@ class BasePipeline:
             return
 
         if flush:
-            yield from self._flush_gen()
+            self.pgconn.send_flush_request()
+            yield from send(self.pgconn)
 
         to_process = []
         while self.result_queue:
@@ -121,10 +128,6 @@ class BasePipeline:
 
         for queued, results in to_process:
             self._process_results(queued, results)
-
-    def _flush_gen(self) -> PQGen[None]:
-        self.pgconn.send_flush_request()
-        yield from send(self.pgconn)
 
     def _process_results(
         self, queued: PendingResult, results: List["PGresult"]
@@ -150,6 +153,11 @@ class BasePipeline:
                 # Update the prepare state of the query.
                 cursor._conn._prepared.validate(key, prep, name, results)
 
+    def _enqueue_sync(self) -> None:
+        """Enqueue a PQpipelineSync() command."""
+        self.command_queue.append(self.pgconn.pipeline_sync)
+        self.result_queue.append(None)
+
 
 class Pipeline(BasePipeline):
     """Handler for connection in pipeline mode."""
@@ -160,7 +168,7 @@ class Pipeline(BasePipeline):
     def __init__(self, conn: "Connection[Any]") -> None:
         super().__init__(conn)
 
-    def communicate(self) -> None:
+    def sync(self) -> None:
         """Sync the pipeline, send any pending command and fetch and process
         all available results.
 
@@ -168,15 +176,7 @@ class Pipeline(BasePipeline):
         purposes (e.g. in nested pipelines).
         """
         with self._conn.lock:
-            self.sync()
-            try:
-                # Send any pending commands (e.g. COMMIT or Sync);
-                # while processing results, we might get errors...
-                self._conn.wait(self._communicate_gen())
-            finally:
-                # then fetch all remaining results but without forcing
-                # flush since we emitted a sync just before.
-                self._conn.wait(self._fetch_gen(flush=False))
+            self._conn.wait(self._sync_gen())
 
     def __enter__(self) -> "Pipeline":
         self._enter()
@@ -189,7 +189,7 @@ class Pipeline(BasePipeline):
         exc_tb: Optional[TracebackType],
     ) -> None:
         try:
-            self.communicate()
+            self.sync()
         except Exception as exc2:
             # Don't clobber an exception raised in the block with this one
             if exc_val:
@@ -209,7 +209,7 @@ class AsyncPipeline(BasePipeline):
     def __init__(self, conn: "AsyncConnection[Any]") -> None:
         super().__init__(conn)
 
-    async def communicate(self) -> None:
+    async def sync(self) -> None:
         """Sync the pipeline, send any pending command and fetch and process
         all available results.
 
@@ -217,15 +217,7 @@ class AsyncPipeline(BasePipeline):
         purposes (e.g. in nested pipelines).
         """
         async with self._conn.lock:
-            self.sync()
-            try:
-                # Send any pending commands (e.g. COMMIT or Sync);
-                # while processing results, we might get errors...
-                await self._conn.wait(self._communicate_gen())
-            finally:
-                # then fetch all remaining results but without forcing
-                # flush since we emitted a sync just before.
-                await self._conn.wait(self._fetch_gen(flush=False))
+            await self._conn.wait(self._sync_gen())
 
     async def __aenter__(self) -> "AsyncPipeline":
         self._enter()
@@ -238,7 +230,7 @@ class AsyncPipeline(BasePipeline):
         exc_tb: Optional[TracebackType],
     ) -> None:
         try:
-            await self.communicate()
+            await self.sync()
         except Exception as exc2:
             # Don't clobber an exception raised in the block with this one
             if exc_val:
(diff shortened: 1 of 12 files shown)


dvarrazzo:
We can see what the libpq is doing using a trace file (thank you @dlax for it too!):

```python
import psycopg

cnn = psycopg.connect("host=localhost", autocommit=True)
cur = cnn.cursor()
cur.execute(
    """
    create temp table testmany (
        id serial primary key, n1 int, n2 int)
    """
)
with open("execmany.trace", "w") as f:
    cnn.pgconn.trace(f.fileno())
    cnn.pgconn.set_trace_flags(psycopg.pq.Trace.SUPPRESS_TIMESTAMPS)

    cur.executemany(
        "insert into testmany (n1, n2) values (%s, %s)",
        [(64, 1024)] * 3,
    )
    cnn.close()
```

The trace file produced before this branch looks like:
```
F	67	Parse	 "_pg3_0" "insert into testmany (n1, n2) values ($1, $2)" 2 21 21
F	4	Sync
B	4	ParseComplete
B	5	ReadyForQuery	 I
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	4	Terminate

```

After this branch it is reduced to:

```
F	67	Parse	 "_pg3_0" "insert into testmany (n1, n2) values ($1, $2)" 2 21 21
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Flush
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	4	Terminate
```

psycopg/psycopg/cursor.py
```diff
@@ -483,11 +471,34 @@ def _set_current_result(self, i: int, format: Optional[Format] = None) -> None:
...
-            self._set_current_result(0)
-        else:
+        first_batch = not self._results
+
+        if self._execmany_returning is None:
+            # Received from execute()
+            # TODO: bug we also end up here on executemany() if run from inside
+            # a pipeline block. This causes a wrong rowcount. As it isn't so
+            # serious, currently leaving it this way.
+            first_batch = not self._results
```
dlax, line 522: This line is redundant with l. 474 above, no?

psycopg/psycopg/cursor.py
```diff
@@ -223,34 +229,16 @@ def _executemany_gen(
...
-            self._rowcount = nrows
+            yield from self._maybe_prepare_gen(pgq, prepare=True)
+            yield from pipeline._communicate_gen()
 
         self._last_query = query
 
         for cmd in self._conn._prepared.get_maintenance_commands():
             yield from self._conn._exec_command(cmd)
 
+        yield from pipeline._flush_gen()
```
dlax, line 243: Is this needed when not "returning"?

```python
        if self._execmany_returning:
            yield from pipeline._flush_gen()
```
maybe?

dlax (commented):
That's quite a nice win!
Changes look good to me, just a few nits.

psycopg/psycopg/cursor.py
dvarrazzo: Yes it is, thank you :)

psycopg/psycopg/cursor.py
dvarrazzo: Maybe it's not. However maybe it would be worth changing the default of returning to False, it would be a better win.

`returning=True` is useful only for select-like statements, in the sense that, even if `returning=True` but the statement doesn't return rows, the results are discarded from the cursor state (after getting their rowcount).

If there is an happy path which consists in not sending the flush, then it would be nice to hit it by default. Normally executemany is mostly used for updates and not to read the result, so it's reasonable for users to opt-it-in:

```python
cur.executemany("insert into tbl ... returning id". data, returning=True)
```

Also nice would have been to know upfront if the statement will return or not. We might do it by reading the result after sending the first record: if we see that it's a no-tuple result we can go not-returning, but this would stall the pipeline.

Someone had expressed [nervousness](https://github.com/psycopg/psycopg/issues/164#issuecomment-1044272063) in the default of returning being True. We may as well go conservative moving the default to False (it's not been released yet) and at the same time squeeze even more performance.

dlax: Independently of `returning` being `True` or not, we still have to retrieve results in order to check if the operations succeeded. But, yes, we could avoid storing them.

Whatever `returning`'s value is, a flush will occur either when calling `cur.fetch*()` or/and when exiting the pipeline.

So by symmetry with the non-executemany case, this flush (even conditional) here is not needed, I would say. (Or we should emit one as well in `_execute_gen()`.)

dvarrazzo: Thank you very much for the insight. I will test a bit with it :)

dvarrazzo: Yes, I have tried dropping it and the test suite passes as before. I think it was part of a failed attempt to fix rowcount.

The trace doesn't change from what posted above, except that now the 
```
F	4	Flush
```

is missing. This probably helps to keep the pipeline flowing if there are more than just one executemany operation in it.

psycopg/psycopg/_pipeline.py
```diff
@@ -197,7 +201,16 @@ def __exit__(
             else:
                 raise
         finally:
-            self._exit()
+            try:
+                self._exit()
+            except Exception as exc2:
+                # Notice that this error might be pretty irrecoverable. It
+                # happens on COPY, for insance: even if sync succeeds, exiting
+                # fails with "cannot exit pipeline mode with uncollected results"
```
dlax, line 209: The comment is out of date per previous commit, isn't it?

psycopg/psycopg/_pipeline.py
```diff
@@ -234,8 +247,14 @@ async def __aexit__(
         except Exception as exc2:
             # Don't clobber an exception raised in the block with this one
             if exc_val:
-                logger.warning("error ignored exiting %r: %s", self, exc2)
+                logger.warning("error ignored syncing %r: %s", self, exc2)
```
dlax, line 250: The same fix could be applied on `Pipeline.__exit__()` above.

dlax (commented):
Last commits look to me, the `trace` fixture is nice indeed.
Just a few nits while reading.

psycopg/psycopg/_pipeline.py
dvarrazzo: No, not really. It's still possible to put the connection in weird state, if you treat it weirdly:

```
In [1]: import psycopg
In [2]: cnn = psycopg.connect("host=localhost")

In [3]: with cnn.pipeline() as p:
   ...:     cur = cnn.cursor()
   ...:     with cur.execute("copy (select 1) to stdout") as copy:
   ...:         pass

ProgrammingError                          Traceback (most recent call last)
...
ProgrammingError: COPY cannot be used with this method; use copy() insead

OperationalError                          Traceback (most recent call last)
...
OperationalError: cannot exit pipeline mode with uncollected results

In [4]: cnn.execute("select 1")

OperationalError                          Traceback (most recent call last)


In [5]: with cnn.pipeline() as p:
   ...:     p.sync()

error ignored exiting <psycopg.Pipeline [ACTIVE] (host=localhost database=piro) at 0x7f365874d760>: connection not in pipeline mode
error ignored exiting <psycopg.Pipeline [ACTIVE] (host=localhost database=piro) at 0x7f365874d760>: cannot send pipeline while in COPY
internal error: cannot send pipeline while in COPY
cannot exit pipeline mode with uncollected results

OperationalError                          Traceback (most recent call last)

OperationalError: connection not in pipeline mode
```

And who knows, there might be other ways too. But this is not really under our control, that's why I removed the assert for pipeline off at the end of the block: it seems blaming us rather than blaming the user (this is a 400, not a 500 :).

I considered closing the connection there instead of leaving it in the inconsistent state, but the effect is the same (likely, the connection is now unusable) and the error messages sort of point at what broke it, more than what a "connection is close" message would say afterwards.

psycopg/psycopg/_pipeline.py
dvarrazzo: Of course! Fixing it. Thank you, good catch.

psycopg/psycopg/_pipeline.py
dlax: I see, probably an exotic usage; but can indeed happen, thanks for the clarification.

> that's why I removed the assert for pipeline off at the end of the block: it seems blaming us rather than blaming the user (this is a 400, not a 500 :).

Totally agreed; the message of 4820f361b50906e3ae147caa979020140b0fdf36 gives of good rationale for that choice.

dvarrazzo:
Merged to master, after merging pipeline4 before.

PR #261, opened 2022-03-29
follow-up executemany() & pipeline

dlax:
Based on #260 to resolve the `rowcount` issue in particular.

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -160,6 +160,24 @@ class Pipeline(BasePipeline):
     def __init__(self, conn: "Connection[Any]") -> None:
         super().__init__(conn)
 
+    def communicate(self) -> None:
+        """Sync the pipeline, send any pending command and fetch and process
+        all available results.
+
+        This is called when exiting the pipeline, but can be used for other
+        purposes (e.g. in nested pipelines).
+        """
+        with self._conn.lock:
+            self.sync()
+            try:
+                # Send any pending commands (e.g. COMMIT or Sync);
+                # while processing results, we might get errors...
+                self._conn.wait(self._communicate_gen())
+            finally:
+                # then fetch all remaining results but without forcing
+                # flush since we emitted a sync just before.
+                self._conn.wait(self._fetch_gen(flush=False))
+
     def __enter__(self) -> "Pipeline":
         self._enter()
         return self
@@ -171,16 +189,7 @@ class Pipeline(BasePipeline):
         exc_tb: Optional[TracebackType],
     ) -> None:
         try:
-            with self._conn.lock:
-                self.sync()
-                try:
-                    # Send any pending commands (e.g. COMMIT or Sync);
-                    # while processing results, we might get errors...
-                    self._conn.wait(self._communicate_gen())
-                finally:
-                    # then fetch all remaining results but without forcing
-                    # flush since we emitted a sync just before.
-                    self._conn.wait(self._fetch_gen(flush=False))
+            self.communicate()
         except Exception as exc2:
             # Don't clobber an exception raised in the block with this one
             if exc_val:
@@ -200,6 +209,24 @@ class AsyncPipeline(BasePipeline):
     def __init__(self, conn: "AsyncConnection[Any]") -> None:
         super().__init__(conn)
 
+    async def communicate(self) -> None:
+        """Sync the pipeline, send any pending command and fetch and process
+        all available results.
+
+        This is called when exiting the pipeline, but can be used for other
+        purposes (e.g. in nested pipelines).
+        """
+        async with self._conn.lock:
+            self.sync()
+            try:
+                # Send any pending commands (e.g. COMMIT or Sync);
+                # while processing results, we might get errors...
+                await self._conn.wait(self._communicate_gen())
+            finally:
+                # then fetch all remaining results but without forcing
+                # flush since we emitted a sync just before.
+                await self._conn.wait(self._fetch_gen(flush=False))
+
     async def __aenter__(self) -> "AsyncPipeline":
         self._enter()
         return self
@@ -211,16 +238,7 @@ class AsyncPipeline(BasePipeline):
         exc_tb: Optional[TracebackType],
     ) -> None:
         try:
-            async with self._conn.lock:
-                self.sync()
-                try:
-                    # Send any pending commands (e.g. COMMIT or Sync);
-                    # while processing results, we might get errors...
-                    await self._conn.wait(self._communicate_gen())
-                finally:
-                    # then fetch all remaining results but without forcing
-                    # flush since we emitted a sync just before.
-                    await self._conn.wait(self._fetch_gen(flush=False))
+            await self.communicate()
         except Exception as exc2:
             # Don't clobber an exception raised in the block with this one
             if exc_val:
--- a/psycopg/psycopg/cursor.py
+++ b/psycopg/psycopg/cursor.py
@@ -68,8 +68,6 @@ class BaseCursor(Generic[ConnectionType, Row]):
         self._closed = False
         self._last_query: Optional[Query] = None
         self._reset()
-        # None if executemany() not executing, True/False according to returning state
-        self._execmany_returning: Optional[bool] = None
 
     def _reset(self, reset_query: bool = True) -> None:
         self._results: List["PGresult"] = []
@@ -79,6 +77,8 @@ class BaseCursor(Generic[ConnectionType, Row]):
         self._rowcount = -1
         self._query: Optional[PostgresQuery]
         self._encoding = "utf-8"
+        # None if executemany() not executing, True/False according to returning state
+        self._execmany_returning: Optional[bool] = None
         if reset_query:
             self._query = None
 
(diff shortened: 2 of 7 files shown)


psycopg/psycopg/_pipeline.py
```diff
@@ -171,16 +189,7 @@ def __exit__(
...
-                self.sync()
-                try:
-                    # Send any pending commands (e.g. COMMIT or Sync);
-                    # while processing results, we might get errors...
-                    self._conn.wait(self._communicate_gen())
-                finally:
-                    # then fetch all remaining results but without forcing
-                    # flush since we emitted a sync just before.
-                    self._conn.wait(self._fetch_gen(flush=False))
+            self.communicate()
```
dvarrazzo, line 192: :heart:

psycopg/psycopg/cursor.py
```diff
@@ -79,6 +77,8 @@ def _reset(self, reset_query: bool = True) -> None:
         self._rowcount = -1
         self._query: Optional[PostgresQuery]
         self._encoding = "utf-8"
+        # None if executemany() not executing, True/False according to returning state
+        self._execmany_returning: Optional[bool] = None
```
dvarrazzo, line 81: A better place, yes

dvarrazzo:
I've been thinking about this: I'm not convinced 100% that is a net win.

If people use executemany in a pipeline, for instance:
```python
import psycopg

cnn = psycopg.connect("host=localhost", autocommit=True)
cur = cnn.cursor()
cur.execute(
    """
    create temp table testmany (
        id serial primary key, n1 int, n2 int)
    """
)
with open("execmany.trace", "w") as f:
    cnn.pgconn.trace(f.fileno())
    cnn.pgconn.set_trace_flags(psycopg.pq.Trace.SUPPRESS_TIMESTAMPS)

    with cnn.pipeline():
        cur.executemany(
            "insert into testmany (n1, n2) values (%s, %s)",
            [(64, 1024)] * 3,
        )
        cur.executemany(
            "insert into testmany (n1, n2) values (%s, %s)",
            [(64, 1024)] * 3,
        )
    cnn.close()
```

with the old version everything happens in a single roundtrip; with the new version there are at least two. I actually see three (there is a final Sync/ReadyForQuery which I don't know if it was expected?):

<details>
<summary>Previous version</summary>

```
F	67	Parse	 "_pg3_0" "insert into testmany (n1, n2) values ($1, $2)" 2 21 21
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	4	Terminate
```
</details>

<details>
<summary>This version version</summary>

```
F	67	Parse	 "_pg3_0" "insert into testmany (n1, n2) values ($1, $2)" 2 21 21
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	36	Bind	 "" "_pg3_0" 2 1 1 2 2 '\x00@' 2 '\x04\x00' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	4	BindComplete
B	4	NoData
B	15	CommandComplete	 "INSERT 0 1"
B	5	ReadyForQuery	 I
F	4	Sync
B	5	ReadyForQuery	 I
F	4	Terminate
```
</details>

Maybe something that we could do is to run the extra synchronisation only when people asks for `returning=True`. What do you think?

Unrelated to this (more related to the `returning` feature): now that the default is False, we could make `executemany()` store the results in the state of the cursor even when the queries don't return tuples, in case there is anything interesting they'd want to know there (such as the status message, or anything else we haven't figured out, really, but that the user may access via `cursor.pgresult`).


Apart from these choice, thank you very much for this refactoring: there are definitely improvements in this branch.

dlax:
> Maybe something that we could do is to run the extra synchronisation only when people asks for returning=True. What do you think?

The extra sync is only needed to get `rowcount` right after the `executemany()` call. In fact, when using `returning=True` people would usually call `fetch*()`, which will emit a sync request, so that extra sync would not be needed either.

Independently of `executemany()`, I think that having the `pipeline()` context manager doing a sync at exit makes sense (and only keep the `PQexitPipelineMode()` call for outermost pipeline). So perhaps this boils down to whether we should open a new pipeline at each `executemany()`?

dvarrazzo:
I have set up a few changes in [this branch](https://github.com/psycopg/psycopg/tree/execmany-returning-changes).

I have changed the behaviour of `executemany("INSERT", returning=True)` to really return the result, and to rowcount to be correct only on `returning=True`.

There is also some refactor that I think valuable in the pipeline code.

dvarrazzo:
> Independently of executemany(), I think that having the pipeline() context manager doing a sync at exit makes sense (and only keep the PQexitPipelineMode() call for outermost pipeline). So perhaps this boils down to whether we should open a new pipeline at each executemany()?

I think it does. In my last changes I have removed it, but it's not a bad thing to think.

So yes, maybe we can reinstate it, and change executemany to not be a trivial pipeline in pipeline. I'll give it a try...

dlax:
> I have set up a few changes in [this branch](https://github.com/psycopg/psycopg/tree/execmany-returning-changes).

I've read them; LGTM.

dvarrazzo:
I have restored the sync at the end of the pipeline block, including when nested. Also added tests that the number of syncs is right with a newly introduced `trace` fixture:

```python
def test_pipeline_exit_sync_trace(conn, trace):
    t = trace.trace(conn)
    with conn.pipeline():
        pass
    conn.close()
    assert len([i for i in t if i.type == "Sync"]) == 1


def test_pipeline_nested_sync_trace(conn, trace):
    t = trace.trace(conn)
    with conn.pipeline():
        with conn.pipeline():
            pass
    conn.close()
    assert len([i for i in t if i.type == "Sync"]) == 2
```

However, I have also refactored executemany so that we don't re-enter the pipeline, if present already, in order to keep the number of roundtrips minimal.

Do you have in mind other operations in which the number of roundtrips, or the sequence of be-fe commands, are so important that it's worth testing for it?

I think now this branch is pretty polished... On to the docs :)

dlax:
> Do you have in mind other operations in which the number of roundtrips, or the sequence of be-fe commands, are so important that it's worth testing for it?

No, I have nothing in mind at the moment.

> I think now this branch is pretty polished...

Agreed.

PR #263, opened 2022-04-02
Pipeline mode docs

dvarrazzo:
@dlax happy to have your feedback about these docs. Thank you!

### DIFF
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -290,7 +290,7 @@ The fine prints
 
     The Pipeline mode is an experimental feature.
 
-    Its behaviour, especially around error condtions, hasn't been explored as
+    Its behaviour, especially around error conditions, hasn't been explored as
     much as the normal request-response messages pattern, and its async nature
     makes it inherently more complex.
 
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -13,16 +13,6 @@ advantage of the pipeline mode, a client will wait less for the server, since
 multiple queries/results can be sent/received in a single network roundtrip.
 Pipeline mode can provide a significant performance boost to the application.
 
-The server executes statements, and returns results, in the order the client
-sends them. The server will begin executing the commands in the pipeline
-immediately, not waiting for the end of the pipeline. Note that results are
-buffered on the server side; the server flushes that buffer when a
-*synchronization point* is established. If any statement encounters an error,
-the server aborts the current transaction and does not execute any subsequent
-command in the queue until the next synchronization point; a
-`~errors.PipelineAborted` exception is raised for each such command. Query
-processing resumes after the synchronization point.
-
 Pipeline mode is most useful when the server is distant, i.e., network latency
 (“ping time”) is high, and also when many small operations are being performed
 in rapid succession. There is usually less benefit in using pipelined commands
@@ -32,16 +22,23 @@ would take 30 seconds in network latency alone without pipelining; with
 pipelining it may spend as little as 0.3 s waiting for results from the
 server.
 
-The pipeline mode is available on any currently supported PostgreSQL version
-but, in order to make use of it, the client must use a libpq from PostgreSQL
-14 or higher. You can use `Pipeline.is_supported()` to make sure your client
-has the right library.
+The server executes statements, and returns results, in the order the client
+sends them. The server will begin executing the commands in the pipeline
+immediately, not waiting for the end of the pipeline. Note that results are
+buffered on the server side; the server flushes that buffer when a
+:ref:`synchronization point <pipeline-sync>` is established.
 
 .. seealso:: The `PostgreSQL pipeline mode documentation`__ contains many
     details around when it is most useful to use the pipeline mode and about
     errors management and interaction with transactions.
 
     .. __: https://www.postgresql.org/docs/current/libpq-pipeline-mode.html
+
+
+.. _pipeline-usage:
+
+Pipeline mode usage
+-------------------
 
 Psycopg supports the pipeline mode via the `Connection.pipeline()` method. The
 method is a context manager: at the end of the ``with`` block, the connection
@@ -53,8 +50,40 @@ normal mode, Psycopg will not wait for the server to receive the result of
 each query, which will be received in batches when the server flushes it
 output buffer.
 
-This flush can happen either when a synchronization point is established by
-Psycopg:
+When a flush (or a sync) is performed, all pending results are sent back to
+the cursors which executed them. If a cursor had run more than one query, it
+will receive more than one result; results after the first will be available,
+in their execution order, using `~Cursor.nextset()`.
+
+If any statement encounters an error, the server aborts the current
+transaction and does not execute any subsequent command in the queue until the
+next :ref:`synchronization point <pipeline-sync>`; a `~errors.PipelineAborted`
+exception is raised for each such command. Query processing resumes after the
+synchronization point.
+
+.. warning::
+
+    Certain features are not available in pipeline mode, including:
+
+    - COPY is not supported in pipeline mode by PostgreSQL.
+    - `Cursor.stream()` doesn't make sense in pipeline mode (its job is the
+      opposite of batching!)
+    - `ServerCursor` are currently not implemented in pipeline mode.
+
+.. note::
+
+    Starting from Psycopg 3.1, `~Cursor.executemany()` makes use internally of
+    the pipeline mode; as a consequence there is no need to handle a pipeline
+    block just to call `!executemany()` once.
+
+
+.. _pipeline-sync:
+
+Synchronization points
+----------------------
+
+Flushing query results to the client can happen either when a synchronization
+point is established by Psycopg:
 
 - using the `Pipeline.sync()` method,
 - on `~Connection.rollback()`,
@@ -68,12 +97,12 @@ output buffer is full.
 In contrast with a synchronization point, a flush request (i.e. calling a
 fetch method) will not reset the pipeline error state.
 
-When a flush (or a sync) is performed, all pending results are sent back to
-the cursors which executed them. If a cursor had run more than one query, it
-will receive more than one result; results after the first will be available,
-in their execution order, using `~Cursor.nextset()`.
+
+The fine prints
+---------------
 
 .. warning::
+
     The Pipeline mode is an experimental feature.
 
     Its behaviour, especially around error condtions, hasn't been explored as
@@ -84,14 +113,7 @@ in their execution order, using `~Cursor.nextset()`.
     bugs and shortcomings forcing us to change the current interface or
     behaviour.
 
-.. warning::
-    Certain features are not available in pipeline mode, including:
-
-    - COPY is not supported in pipeline mode by PostgreSQL.
-    - `Cursor.stream()` doesn't make sense in pipeline mode (its job is the
-      opposite of batching!)
-    - `ServerCursor` are currently not implemented in pipeline mode.
-
-.. note::
-    Starting from Psycopg 3.1, `Cursor.executemany()` is optimised to make use
-    of pipeline mode.
+The pipeline mode is available on any currently supported PostgreSQL version,
+but, in order to make use of it, the client must use a libpq from PostgreSQL
+14 or higher. You can use `Pipeline.is_supported()` to make sure your client
+has the right library.
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -20,9 +20,8 @@ buffered on the server side; the server flushes that buffer when a
 *synchronization point* is established. If any statement encounters an error,
 the server aborts the current transaction and does not execute any subsequent
 command in the queue until the next synchronization point; a
-`~errors.PipelineAborted` exception is raised for each such command (including
-a `~Connection.rollback()`). Query processing resumes after the
-synchronization point.
+`~errors.PipelineAborted` exception is raised for each such command. Query
+processing resumes after the synchronization point.
 
 Pipeline mode is most useful when the server is distant, i.e., network latency
 (“ping time”) is high, and also when many small operations are being performed
@@ -57,7 +56,8 @@ output buffer.
 This flush can happen either when a synchronization point is established by
 Psycopg:
 
-- using the `Pipeline.sync()` method, or,
+- using the `Pipeline.sync()` method,
+- on `~Connection.rollback()`,
 - at the end of a `!Pipeline` block;
 
 or using a fetch method such as `Cursor.fetchone()`.
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -51,22 +51,27 @@ resumes the normal operation mode.
 Within the pipeline block, you can use one or more cursors to execute several
 operations, using `~Cursor.execute()` and `~Cursor.executemany()`. Unlike in
 normal mode, Psycopg will not wait for the server to receive the result of
-each query, which will be received in batches when a synchronization point is
-established.
+each query, which will be received in batches when the server flushes it
+output buffer.
 
-Psycopg can establish a synchronization point:
+This flush can happen either when a synchronization point is established by
+Psycopg:
 
-- using the `Pipeline.sync()` method;
+- using the `Pipeline.sync()` method, or,
 - at the end of a `!Pipeline` block;
-- using a fetch method such as `Cursor.fetchone()`.
 
-The server might perform a sync on its own initiative, for instance when the
+or using a fetch method such as `Cursor.fetchone()`.
+
+The server might perform a flush on its own initiative, for instance when the
 output buffer is full.
 
-When a sync is performed, all the pending results are sent back to the cursors
-which executed them. If a cursor had run more than one query, it will receive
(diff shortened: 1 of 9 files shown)


dvarrazzo:
I wonder if we shouldn't declare the feature experimental, and add a disclaimer that we might consider changing the interface or the behaviour of the pipeline mode based on the users' feedback.

docs/advanced/pipeline.rst
```diff
@@ -68,6 +68,14 @@ which executed them. If a cursor had run more than one query, it will receive
 more than one result; results after the first will be available, in their
 execution order, using `~Cursor.nextset()`.
 
+.. warning::
+    Certanin features are not available in pipeline mode, including:
```
dlax, line 72: typo: Certain

docs/advanced/pipeline.rst
```diff
@@ -0,0 +1,73 @@
...
+The pipeline mode is available on any currently supported PostgreSQL version
+but, in order to make use of it, the client must use a libpq from PostgreSQL
+14 or higher. You can use `Pipeline.is_supported()` to make sure your client
+has the right library.
+
+.. seealso:: The `PostgreSQL pipeline mode documentation`__ contains many
+    details around when it is most useful to use the pipeline mode and about
+    errors management and interaction with transactions.
+
+    .. __: https://www.postgresql.org/docs/14/libpq-pipeline-mode.html
```
dlax, line 45: Maybe use `/current/` instead of `14` like elsewhere in our documentation?

docs/advanced/pipeline.rst
```diff
@@ -0,0 +1,73 @@
...
+method is a context manager: at the end of the ``with`` block, the connection
+resumes the normal operation mode.
+
+Within the pipeline block, you can use one or more cursors to execute several
+operations, using `~Cursor.execute()` and `~Cursor.executemany()`. Unlike in
+normal mode, Psycopg will not wait for the server to receive the result of
+each query, which will be received in batches when a synchronization point is
+established.
+
+Psycopg can establish a synchronization points:
```
dlax, line 57: typo: synchronization point*

docs/advanced/pipeline.rst
```diff
@@ -0,0 +1,73 @@
...
+
+- using the `Pipeline.sync()` method;
+- at the end of a `!Pipeline` block;
+- using a fetch method such as `Cursor.fetchone()`.
```
dlax, line 61: It's not quite true that fetch methods trigger a synchronization point; in fact they ask for a flush of server's output buffer. This is different because a synchronization point, in addition to requesting a flush, resets the error state of the pipeline (as explained already earlier).

So maybe we should talk about *flush* here instead of *synchronization point*? (The conditions above apply for flush.)

docs/advanced/pipeline.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The server might perform a sync on its own initiative, for instance when the
+query buffer is full.
```
dlax, line 64: Same here, it's a flush not a sync. (And one last below.)

Also, "when the output buffer is full" seems more appropriate.

dlax (commented):
Very nice!

Just spotted a few typos and suggesting a clarification about sync vs flush.

dlax:
> I wonder if we shouldn't declare the feature experimental, and add a disclaimer that we might consider changing the interface or the behaviour of the pipeline mode based on the users' feedback.

I'm okay if you're more comfortable with this approach. Requesting users' feedback is always good, we can start a thread on the ML for this. Perhaps we could ping downstream projects about this as well (e.g. sqlalchemy)?

docs/advanced/pipeline.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Pipeline mode support
+=====================
+
+.. versionadded:: 3.1
+
+The *pipeline mode* allows PostgreSQL client applications to send a query
+without having to read the result of the previously sent query. Taking
+advantage of the pipeline mode, a client will wait less for the server, since
+multiple queries/results can be sent/received in a single network transaction.
```
asqui, line 13: What do we mean by 'network transaction' here? 'network packet'? 'network interaction'?

Might be good to avoid the term 'transaction', since that has a very specific (and different) meaning in databases.

asqui (commented):
I must admit I've been largely ignoring the work on this feature, so I figured now would be a good time to review the docs "with fresh eyes" knowing pretty much nothing about the feature 😄 

Excellent overview docs, as usual, all very clear. Just one minor nitpick on the use of the term "network transaction" from me.

Only other suggestion is that it might be nice to have an example of nested `Pipeline` blocks -- coming to this feature 'cold' I can't quite picture how you're meant to use these.

i.e. if fetching from a cursor automatically causes a flush, then why would you need to establish a nested `Pipeline` block rather than just having one level, with periodic fetches as needed?

Perhaps the only use of nesting is to enable composability of independent functions that make use of pipelines? In that case, maybe rather than an example, a one-line statement to explain this would be more appropriate?

docs/advanced/pipeline.rst
dvarrazzo: Yes, I think the same. I lifted a few of those paragraphs directly from PostgreSQL documentation (first paragraph, [here](https://www.postgresql.org/docs/14/libpq-pipeline-mode.html)), so I'd have assumed it very strictly nitpicked. But it sounds strange to me too.

I would call it network roundtrip.

dvarrazzo:
@dlax 

> I'm okay if you're more comfortable with this approach. Requesting users' feedback is always good, we can start a thread on the ML for this. Perhaps we could ping downstream projects about this as well (e.g. sqlalchemy)?

I fear the semver police...

I would explain that the whole pipeline feature leads to a pretty different way of communicating between backend and frontend. There could be corner cases, especially around errors management, in which the states might get out of sync. Something might be a bug with a straightforward fix, other things may lead to rethink the whole approach more deeply, and we might see the behaviour or maybe even the interface changing.

So pretty much I'd ask people to cut us some slack for a minor version or two, as we gain experience :D

I will write the mailing list about the feature. The SQLAlchemy author is subscribed so we might get his input too.

dlax:
@asqui 
> Only other suggestion is that it might be nice to have an example of nested `Pipeline` blocks -- coming to this feature 'cold' I can't quite picture how you're meant to use these.
> 
> i.e. if fetching from a cursor automatically causes a flush, then why would you need to establish a nested `Pipeline` block rather than just having one level, with periodic fetches as needed?

That's the difference between flush and sync. A sync triggers a flush *and* resets the error state. So if one only requests flushes and any error happen during the pipeline then any statement emitted after that error will be ignored. Typically, one might want to emit a sync after each transaction commit/rollback, so if the connection is operating in pipeline mode as a whole, opening inner pipeline for transaction would make sense (it's not really opening new pipeline, just syncing the main one).

> Perhaps the only use of nesting is to enable composability of independent functions that make use of pipelines? In that case, maybe rather than an example, a one-line statement to explain this would be more appropriate?

That's another case, yes: one opens a pipeline while possibly within another pipeline that's controlled from outer scope.

docs/advanced/pipeline.rst
dlax: I've drafted commit https://github.com/dlax/psycopg3/commit/fa703b5b (`pipeline-docs` branch on my repo) as an attempt to improve this part, if that helps.

dvarrazzo:
Thank you both for the review. I have fixed the obvious typo so far, thank you.

Two non-minor thing I think should add:

- as @asqui suggests, examples would be useuful
- some comment about implicit transactions. Pipeline mode is less autocommit than it looks, for instance:

```python
with psycopg.connect(autocommit=True) as conn:
    try:
        with conn.pipeline() as p:
            conn.execute("create table done()")
            conn.execute("error")
    except Exception as e:
        print(e)
    conn.execute("select * from done")  # this fails
```

unlike in normal mode, the table is not created. Statements are wrapped into an implicit transaction and the Sync act as commit/rollback.

I'm doing some experiments with small pipelines and looking at the message flow to wrap my head around flush and sync. Just read @dlax changes in https://github.com/dlax/psycopg3/commit/fa703b5b and they look good, I'm going to integrate them. What I can see from these tests is:

dvarrazzo:
I assume that Flush is part of the Sync. Sync does Flush + reset of the internal transaction or of the pipeline aborted state (which I understand to be a client-side state, not a server-side one).

Exiting the block or calling `sync()` calls sync. `fetch*()` is the only way we can send a naked Flush, ISTM. Is this correct, and what you expect, @dlax?

Two reference pages in postgres docs:
- [protocol flow](https://www.postgresql.org/docs/current/protocol-flow.html)
- [pipeline mode](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html)

My tests:

This is the baseline behaviour: a sequence of non-pipeline statements. I force prepare to give them a trace more similar to the pipeline mode, by forcing the advanced query protocol.

```python
import psycopg
with psycopg.connect(autocommit=True, prepare_threshold=0) as conn:
    f = open("tmp.pipe", "w")
    conn.pgconn.trace(f.fileno())
    conn.pgconn.set_trace_flags(psycopg.pq.Trace.SUPPRESS_TIMESTAMPS)
    conn.execute("select %s", [1])
    conn.execute("select %s", [2])
```
resutls in:
```
F	27	Parse	 "_pg3_0" "select $1" 1 21
F	4	Sync
B	4	ParseComplete
B	5	ReadyForQuery	 I
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x01' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '1'
B	5	ReadyForQuery	 I
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x02' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '2'
B	5	ReadyForQuery	 I
F	4	Terminate
```
It can be seen that every `F` streak (frontend sending messages) is terminated by a `Sync`, which is the way the client tells the server that it's its turn to talk now. The server closes the implicit transactions there.

----

Two statements in pipeline mode go:

```python
# boilerplate statements removed from here on
with psycopg.connect(autocommit=True, prepare_threshold=0) as conn:
    with conn.pipeline() as p:
        conn.execute("select %s", [1])
        conn.execute("select %s", [2])
```

```
F	27	Parse	 "_pg3_0" "select $1" 1 21
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x01' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x02' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '1'
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '2'
B	5	ReadyForQuery	 I
F	4	Terminate
```

there is only one sync, and act as a flush too: the messages are sent to the frontend.

---

Sending an explicit flush, as well as exiting an internal pipeline block, results in the same: a flush being sent.

```python
with psycopg.connect(autocommit=True, prepare_threshold=0) as conn:
    with conn.pipeline() as p:
        with conn.pipeline() as p:
            conn.execute("select %s", [1])
        conn.execute("select %s", [2])
    # or, alternatively,
    with conn.pipeline() as p:
        conn.execute("select %s", [1])
        p.sync()
        conn.execute("select %s", [2])

```
```
F	27	Parse	 "_pg3_0" "select $1" 1 21
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x01' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '1'
B	5	ReadyForQuery	 I
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x02' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '2'
B	5	ReadyForQuery	 I
F	4	Terminate
```

Note for @asqui: the reason to allow nested pipeline is to make easier to compose functions without the knowledge of where they are called from. A function receiving a connection can start a pipeline and this would work the same way either if the connection was in pipeline mode already or not. Nestable blocks is something we introduced only lately: the idea came from a first implementation of executemany which was pretty much brutally calling `with pipeline` internally. Later on, looking at the flow, we realised that it wasn't the most optimal implementation (because each executemany would have issued a probably unnecessary sync), however the nestable pipeline block stayed. I still think it makes sense to have them, but I'm open to hear opposite opinions.

---

Fetching seems the only way to ask for a Flush, which I understand doesn't set a sync point.

```python
with psycopg.connect(autocommit=True, prepare_threshold=0) as conn:
    with conn.pipeline() as p:
        cur = conn.execute("select %s", [1])
        cur.fetchone()
        cur = conn.execute("select %s", [2])
```

```
F	27	Parse	 "_pg3_0" "select $1" 1 21
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x01' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Flush
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '1'
F	28	Bind	 "" "_pg3_0" 1 1 1 2 '\x00\x02' 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	BindComplete
B	33	RowDescription	 1 "?column?" 0 0 21 2 -1 0
B	11	DataRow	 1 1 '2'
B	5	ReadyForQuery	 I
F	4	Terminate
```

@dlax is this exactly what you expect to happen or is there anything expected for you here?

dlax:
Yes, that's exactly how it's supposed to work.

* Sync also does a flush of server's output buffer (it also flushes the query from client by the way, similarly to `PQflush()`)
* to fetch, we need the server's output buffer to be flushed (otherwise `PQgetResult` would block) but we don't need a Sync
* in contrast with Sync, I don't see a use case for a flush outside fetch, hence it's fine not to have an API for this (in case one wonders why there is none); so yes: "fetch*() is the only way we can send a naked Flush"
* about implicit transaction and Sync, I think this is part of the normal [protocol flow](https://www.postgresql.org/docs/current/protocol-flow.html):
  > At completion of each series of extended-query messages, the frontend should issue a Sync message. This parameterless message causes the backend to close the current transaction if it's not inside a BEGIN/COMMIT transaction block (“close” meaning to commit if no error, or roll back if error). Then a ReadyForQuery response is issued. The purpose of Sync is to provide a resynchronization point for error recovery. When an error is detected while processing any extended-query message, the backend issues ErrorResponse, then reads and discards messages until a Sync is reached, then issues ReadyForQuery and returns to normal message processing.

> Two non-minor thing I think we should add:
>     * as @asqui suggests, examples would be useful
>     * some comment about implicit transactions.

Agreed on both points. While examples are often better that abstract explanations, the second is a behavior change to normal mode so it's worth a comment.

dvarrazzo:
> Yes, that's exactly how it's supposed to work.

Thank you for confirming, @dlax! It seems straightforward in terms of relation between the actions on the interface and the communications with the server :slightly_smiling_face:

dlax:
While playing more on my side, I'd suggest https://github.com/psycopg/psycopg/pull/269 about the `Sync` thing...

asqui:
> Note for @asqui: the reason to allow nested pipeline is to make easier to compose functions without the knowledge of where they are called from. A function receiving a connection can start a pipeline and this would work the same way either if the connection was in pipeline mode already or not.

+100 for this design principle. Something that was a key feature of the transaction design too.

docs/advanced/pipeline.rst
```diff
@@ -73,6 +73,17 @@ the cursors which executed them. If a cursor had run more than one query, it
 will receive more than one result; results after the first will be available,
 in their execution order, using `~Cursor.nextset()`.
 
+.. warning::
+    The Pipeline mode is an experimental feature.
+
+    Its behaviour, especially around error condtions, hasn't been explored as
```
dlax, line 121: typo: condtions

dlax (commented):
Looks good; just one typo and suggestion #295.

dvarrazzo:
I have merged the documentation to master in order to publish it. It is now [available here](https://www.psycopg.org/psycopg3/docs/advanced/pipeline.html), which makes reviewing it as a whole easier.

Still open to improvements if you have any feedback, thank you very much!

@dlax super happy we have finished this work! :partying_face: Thank you very much!!!

dlax:
The examples you added are great; thank you.

Very glad we finished this!!

Issue #101, opened 2021-10-10
Create a `ClientBindingCursor`?


dvarrazzo:
Client-side statement can be merged using `sql.SQL()`. However we could create a cursor doing that automatically.

It would be handy in porting psycopg2 code more easily maybe, even if less efficient.

It could be made a mixin to add to `ClientCursor` and `AsyncClientCursor`, replacing the `PostgresQuery` with a `PostgresClientQuery` doing a different transformation.

sureshdsk:
@dvarrazzo looks like something I can do. but I don't understand the use case clearly since I'm not familiar with sql.SQL. can you give me some examples of how the existing merging works with manual steps and how the expected automatic way?

dvarrazzo:
Hi @sureshdsk,

Psycopg 2 composes the queries client-side: if you run `execute("insert into table values (%s)", ["hello"])` it will create a string on the client, which is valid sql (e.g. `insert into table values ('hello')`) and send it to the server.

Psycopg 3 uses a different mechanism and passes the query and argument separated, using a different protocol: it transform the query to postgres parameters (`insert into table values ($1)`) and leaves the server do the binding.

This has advantages in terms of performance, memory usage, flexibility. However it only works with certain statements (SELECT, data manipulation statements) but not with others (data-definition statements and others). So you cannot run anymore `cur.execute("SET TimeZone to %s")` or `alter table foo set default %s`.

The main workaround is to use SQL composition with the objects in the `psycopg.sql` module. This however requires the strings to be translated because it uses a third way of specifying placeholders (e.g. `alter table foo set default {}`).

The issue is explained in the docs: https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#server-side-binding

So, the idea would be to have a cursor which, given a `cur.execute("alter table foo set default %s", [42])`, internally would transform the statement and call `sql.SQL("alter talbe foo set default {}").format(sql.Literal(42))` and then send the query already merged to the server (using `PGconn.exec_()` instead of `PGconn.exec_params()`).

However:

I opened this ticket to keep a note of this, because it's doable, but nobody has asked for it yet. My idea is that if people start pressuring that composing queries client-side explicitly is difficult or is a chore, we could introduce it. But at the moment no such pressure has materialised yet: it might be because the library is not very used yet, or it might be that people just understand the issue and use the workaround proposed.

[...]

PR #264, opened 2022-04-03
Typed literal

### LINKED ISSUE #205: Add `sql.TypedLiteral`
In order to make client-side queries more similar to the server-side ones, we can introduce a `sql.TypedLiteral` object, which, on `as_string()`, would:

- use the transformer to find the object oid
- use the TypesRegistry to find the name of the type
- add a cast to the object representation (e.g. `'2021-01-01'::date`).

Note that this might require to create many more Transformer. A worthwile optimisation (which could be useful for any recursive type) could be to not create a transformer when a transformer is passed as a context, but use always the same instance.

dvarrazzo:
Close #205 

Does the heavy lifting to implement #101


dlax (approved):
I read all commits, and all look good to me (though I'm not familiar with that part of the code base, nor with the issue at stake).

PR #265, opened 2022-04-03
Add ConnectionInfo.pipeline_status

dvarrazzo:
@dlax It occurred to me that we don't have a public way to check for the pipeline status. There is `Connection.pgconn.pipeline_status` but it sort of require "internal libpq knowledge". Ok with this?

Will flesh more docs (even just a "see also: pipeline mode") to the docs branch.


dlax (approved):
Yes, totally useful (I found myself needing this yesterday while playing more with the feature).
Suggesting #268 as a follow-up.

PR #267, opened 2022-04-04
Fix pipeline rollback

dvarrazzo:
A work in progress. Rolling back within a pipeline should probably take care of the pipeline status too.

@dlax, what do you think?

### DIFF
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -531,6 +531,11 @@ class BaseConnection(Generic[Row]):
             raise e.ProgrammingError(
                 "rollback() cannot be used during a two-phase transaction"
             )
+
+        # Get out of a "pipeline aborted" state
+        if self._pipeline and self.pgconn.pipeline_status == pq.PipelineStatus.ABORTED:
+            yield from self._pipeline._sync_gen()
+
         if self.pgconn.transaction_status == TransactionStatus.IDLE:
             return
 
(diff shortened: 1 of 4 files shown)


dlax:
> Rolling back within a pipeline should probably take care of the pipeline status too.

Not sure; it might be premature to handle this on our side. I mean: the `ROLLBACK` not being accounted for in an aborted pipeline is normal. Perhaps, as a start, we could document that that calling `sync()` before commit or rollback is wise? (Postgres documentation mentions that "Pipelines should be scoped to logical units of work, usually (but not necessarily) one transaction per pipeline.")

dlax:
Just to clarify my point:
- the current implementation, relying on an explicit `with conn.pipeline()` statement from the user is somewhat _low-level_ (less than plain libpq, but still),
- while we probably managed to get a it working fine in a number of situations we discovered along the way, this requires some kind of "expert" knowledge (typically, the: "it might be wise to `sync` the pipeline before commit or rollback"),
- if we want to get further on the way to make it work more seamlessly (i.e. hiding the complexity), I think we'd benefit from designing the overall behavior: for instance, having an automatic pipeline mode, or some option at connection and transaction contexts like `with psycopg.connect(pipeline=True):` or `with conn.transaction(pipeline=True):`... (I'm willing to investigate in this direction.)

At this point, listing examples and edge cases as considered in #263 would really help, I think.

(So, about this change proposed here, I don't know for now.)

dlax:
I've thinking about this more and suggest https://github.com/psycopg/psycopg/pull/271 as an alternative. The idea is to use nested pipeline contexts to handle transaction state rather that (as done here) use pipeline internal logic in transaction code. Not totally sure about it, still.

PR #268, opened 2022-04-04
doc: add PipelineStatus to 'pq' page

dlax:


dvarrazzo:
Ah, I thought I was adding it to the docs, but I missed it. Thank you

PR #269, opened 2022-04-04
pipeline: various improvements on sync, flush and fetch steps 

dlax:
(see commit messages for details; part of this would make sense with #271 as well)

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -86,10 +86,13 @@ class BasePipeline:
 
     def _sync_gen(self) -> PQGen[None]:
         self._enqueue_sync()
+        yield from self._communicate_gen()
+
+    def _exit_gen(self) -> PQGen[None]:
         try:
             # Send any pending commands (e.g. COMMIT or Sync);
             # while processing results, we might get errors...
-            yield from self._communicate_gen()
+            yield from self._sync_gen()
         finally:
             # then fetch all remaining results but without forcing
             # flush since we emitted a sync just before.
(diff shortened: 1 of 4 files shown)


dvarrazzo:
I don't see tests here. Is this something addressing the problems found in #267?

Juggling a few things right now, I'll be back on this later

psycopg/psycopg/_pipeline.py
```diff
@@ -196,7 +197,7 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> None:
         try:
-            self.sync()
+            self.sync(True)
```
dvarrazzo, line 200: I would make this a kwarg-only argument, as `_sync_gen()` is,  and force user to use `fetch=True`.

dlax:
> I don't see tests here. Is this something addressing the problems found in #267?

No, it's not related.

Rather it's a matter of avoiding surprises from user's perspective: "eh, I just wanted to sync, what's all this server->client network traffic?" (But the definition of "surprise" is subjective of course.)

> I don't see tests here.

I'd try, but I'm afraid it's going to be tricky to write reliable tests for this very change because, when we call `Pipeline.sync()`, we then call the `pipeline_communicate()` generator which might in turn fetch some results *if* the server has decided to flush its output buffer.

(In fact, this change reverts back to the original implementation I submitted where calling `sync()` did not involved explicit results fetching.)

psycopg/psycopg/_pipeline.py
dlax: I'm not sure there is a use case for an explicit fetch so I'd rather rework the commit to not have the parameter in `sync()`. Do you see a use case on your side?

dvarrazzo:
No worries about the tests. I was just wondering if this refactoring was something addressing the test failed that I introduced to show the rollback problem in #267

psycopg/psycopg/_pipeline.py
dvarrazzo: No use case in mind. Just looking at that naked `True` in the `sync()`, which doesn't give any explanation, whereas the internal `_sync_gen()` is self explanatory with a kwargs-only argument :slightly_smiling_face:

dlax:
Second commit is another (micro?) optimization on `executemany(..., returning=True)`.

dvarrazzo:
Rebased on master, will merge as tests pass.

PR #271, opened 2022-04-06
feat: wrap transaction in pipelines if the connection has one

dlax:
(Alternative solution to #267; test introduced here taken from there.)

When the connection is in pipeline mode, we wrap each transaction block
with two pipelines: the inner one ensures that we'd get a Sync at the
end of transaction *before* the commit or rollback statement, thus
ensuring that the latter gets run even if the pipeline got in aborted
state during the transaction; the outer one ensures (also emitting a
sync) that the connection is restored into a usable state even when the
transaction rolled back.

Introduce nullcontext and Abstract(Async)ContextManager compat
definitions.

### DIFF
--- a/psycopg/psycopg/_compat.py
+++ b/psycopg/psycopg/_compat.py
@@ -30,16 +30,45 @@ else:
 if sys.version_info >= (3, 9):
     from zoneinfo import ZoneInfo
     from collections import Counter, deque as Deque
+    from contextlib import AbstractAsyncContextManager, AbstractContextManager
 else:
     from backports.zoneinfo import ZoneInfo
-    from typing import Counter, Deque
+    from typing import (
+        Counter,
+        Deque,
+        AsyncContextManager as AbstractAsyncContextManager,
+        ContextManager as AbstractContextManager,
+    )
 
 if sys.version_info >= (3, 10):
+    from contextlib import nullcontext
     from typing import TypeAlias, TypeGuard
 else:
+    from contextlib import nullcontext as _nullcontext
     from typing_extensions import TypeAlias, TypeGuard
 
+    if sys.version_info >= (3, 9):
+
+        class nullcontext(_nullcontext[T]):
+            async def __aenter__(self) -> T:
+                return self.enter_result
+
+            async def __aexit__(self, *excinfo: Any) -> None:
+                pass
+
+    else:
+
+        class nullcontext(_nullcontext):
+            async def __aenter__(self) -> Any:
+                return self.enter_result
+
+            async def __aexit__(self, *excinfo: Any) -> None:
+                pass
+
+
 __all__ = [
+    "AbstractAsyncContextManager",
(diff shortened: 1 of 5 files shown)


dvarrazzo:
I am trying to use your solution to solve the problem highlighted in #267. I have simplified the implementation a bit, dropping the abstract context manager. As there is already an `if` there:

```python
        tx = Transaction(self, savepoint_name, force_rollback)
        if self._pipeline:
            with self.pipeline(), tx, self.pipeline():
                yield tx
        else:
            with tx:
                yield tx
```

However, `test_rollback_explicit()` doesn't pass with your implementation either: I'll look into it.

dvarrazzo:
The `test_rollback_explicit()` was easy to fix on #267: we were bailing out too early and this seems enough to fix it:

```diff
diff --git a/psycopg/psycopg/connection.py b/psycopg/psycopg/connection.py
index 587cef22..fc92c051 100644
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -531,13 +531,14 @@ class BaseConnection(Generic[Row]):
             raise e.ProgrammingError(
                 "rollback() cannot be used during a two-phase transaction"
             )
-        if self.pgconn.transaction_status == TransactionStatus.IDLE:
-            return
 
         # Get out of a "pipeline aborted" state
         if self._pipeline and self.pgconn.pipeline_status == pq.PipelineStatus.ABORTED:
             yield from self._pipeline._sync_gen()
 
+        if self.pgconn.transaction_status == TransactionStatus.IDLE:
+            return
+
         yield from self._exec_command(b"ROLLBACK")
         self._prepared.clear()
         for cmd in self._prepared.get_maintenance_commands():
```

I have rebased your branch on master, now that we have merged #269, and for some reason `tests/test_pipeline_async.py::test_transaction` started to fail randomly, some 20% of the times, with "psycopg.errors.NoActiveSqlTransaction: ROLLBACK TO SAVEPOINT can only be used in transaction blocks" in `AsyncPipeline.__aexit__()`. It doesn't seem that it was doing it originally, so maybe it's not compatible with some of these commits:

- 785b1b73 feat: only Sync nested pipelines at exit
- e49aa7a7 fix: only request a pipeline flush in returning executemany()
- ec32a8a9 fix: do not fetch in Pipeline.sync()

I assume the patch above would fix the same tests in this branch, however I don't know what to do with this random failure and it seems that #267 is behaving better.

dvarrazzo:
I have merged #267. I am closing this, but I'm still open to consider if this is a better approach, once all tests pass. Thank you very much!

dlax:
The approach from 060118f8f731559c45823ae3030a657ded5385a6 that you merged looks good to me. It's simpler than this suggestion here, but, now comparing, the latter was probably *too* conservative (namely, there does not seem to be a need for syncing before `commit()`, only before `rollback()` is needed).
Thanks for moving things forward!!

Issue #273, opened 2022-04-12
Support for custom enums and enum arrays


ertaquo:
Cheers!

I found that psycopg doesn't support custom enum types, so wrote [little support function](https://gist.github.com/ertaquo/91619a9c021fad7d7b0b6219ebce9f93) to generate loaders for them. Maybe it can be useful for someone?

ertaquo:
Done, but please verify the PR - maybe i did something wrong?

dvarrazzo:
Thank you, that's amazing! Will review soon

dvarrazzo:
Overtaken by #274.

PR #274, opened 2022-04-13
feat: add enum support

ertaquo:
Postgres enum support (casting from/to Python `enum.Enum` types).

Based on #273


psycopg/psycopg/types/enum.py
```diff
@@ -0,0 +1,95 @@
...
+    loader = type(name, (base,), attribs)
+    adapters.register_loader(info.oid, loader)
+
+    base = EnumBinaryLoader
+    name = f"{info.name.title()}{base.__name__}"
+    attribs = {"python_type": info.python_type}
+    loader = type(name, (base,), attribs)
+    adapters.register_loader(info.oid, loader)
+
+
```
dvarrazzo, line 92: We need dumpers dumping to the right oid.
Untested:
```python
attribs = {"oid": info.oid}

name = f"{info.name.title()}Dumper"
dumper = type(name, (EnumDumper,), attribs)
adapters.register_dumper(info.python_type, dumper)

name = f"{info.name.title()}BinaryDumper"
dumper = type(name, (EnumBinaryDumper,), attribs)
adapters.register_dumper(info.python_type, dumper)
```

tests/types/test_enum.py
```diff
@@ -0,0 +1,125 @@
+from enum import Enum
+
+import pytest
+from psycopg.types.enum import EnumInfo, register_enum
+
+from psycopg import pq
+
+
+class _TestEnum(str, Enum):
```
dvarrazzo, line 10: Why the underscore?

tests/types/test_enum.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        drop type if exists testenum cascade;
+
+        create type testenum as enum('ONE', 'TWO', 'THREE');
+        """
+    )
+    return EnumInfo.fetch(svcconn, "testenum")
+
+
+def test_fetch_info(conn, testenum):
```
dvarrazzo, line 40: There is no test to fetch with an async connection. Could you please add one?

dvarrazzo (changes requested):
Hello! Thank you very much, this looks good.

Main shortcomings to address:

- we need a dumper able to roundtrip. If we pass an enum as query param, we get back the same enum (at `is` level, in the enum case).
- we should support non-string enums too.

I have pushed a commit with some failing tests to your branch.

psycopg/psycopg/types/enum.py
```diff
@@ -0,0 +1,95 @@
...
+def register_default_adapters(context: AdaptContext) -> None:
```
dvarrazzo, line 93: Looks like the generic dumpers are ok. My fear was about creating an incompatibility with people dumping int enums as postgres integers. But there is a test in `test_numeric.py` proving that this use is not broken.

However, in order to work ok, the oid of the generic dumpers should be 0, or you get an error such:
```
psycopg.errors.DatatypeMismatch: column "val" is of type testenum but expression is of type text
```
added a failing test to prove it.

dvarrazzo (changes requested):

dvarrazzo:
Please also add an entry to `docs/news.rst` for psycopg 3.1 :)

docs/basic/pgtypes.rst
```diff
@@ -365,3 +365,73 @@ connection or cursor), other connections and cursors will be unaffected::
     ...     "coordinates":[-48.23456,20.12345]}')
     ... """).fetchone()[0]
     '0101000020E61000009279E40F061E48C0F2B0506B9A1F3440'
+
+.. _adapt-enum:
+
+Enum types casting
```
dvarrazzo, line 371: "Type casting" is more a psycopg2 term. Please use "Enum adaptation"

psycopg/psycopg/types/enum.py
```diff
@@ -0,0 +1,95 @@
...
+    """Register the adapters to load and dump a enum type.
+
+    :param info: The object with the information about the enum to register.
+    :param python_type: Python enum type matching to the Postgres one. If `!None`,
+        the type will be generated and put into info.python_type.
+    :param context: The context where to register the adapters. If `!None`,
+        register it globally.
+
+    .. note::
+        Only string enums are supported.
```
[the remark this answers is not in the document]
ertaquo: As far as I know, Postgres enum labels can be string only.
I don't know how to manage enums of another types. Maybe add optional parameter to use Python enum keys instead of values?
```sql
CREATE TYPE some_enum AS ENUM ('ONE', 'TWO');
```
```python
class SomeEnum(int, Enum):
  ONE = 1
  TWO = 2
...
register_enum(info, SomeEnum, use_keys_as_labels=True, context=conn)
```

docs/basic/pgtypes.rst
```diff
@@ -365,3 +365,73 @@ connection or cursor), other connections and cursors will be unaffected::
     ...     "coordinates":[-48.23456,20.12345]}')
     ... """).fetchone()[0]
     '0101000020E61000009279E40F061E48C0F2B0506B9A1F3440'
+
```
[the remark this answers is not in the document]
ertaquo: `docs/basic/params.rst`? Not a `docs/basic/adapt.rst`?

tests/types/test_enum.py
ertaquo: pytest shows error if class name starts with `Test`:
```
E   pytest.PytestCollectionWarning: cannot collect test class 'TestEnum' because it has a __new__ constructor (from: tests/types/test_enum.py)
```
Maybe rename it (`StrTestEnum`)?

dvarrazzo: Ah you are right, Test is collected. Yes, something like StrTestEnum, MyEnum, MyStrEnum would do.

psycopg/psycopg/types/enum.py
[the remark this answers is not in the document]
dvarrazzo: Yes, I have been thinking all the time during my review that adaptation was based on the names. Only towards the end I figured out you were relying on the values.

There is also the case that keys and values might be different for str enums in python, oelr that the mapping python-postgres is not 1:1 perfect.

What I had in mind is that maybe we could accept a mapping enum -> label name to customize the mapping when the postgres label doesn't match the python enum name.

```python
class LowEnum(str, Enum):
    FOO = "VAL1"
    BAR = "VAL2"

# create type as enum ('FOO', 'VAL3')
register_enum(... labels={LowEnum.foo: "VAL3"})
```

He adapters could use, by default, a 1:1 mapping based on names. The `labels` param might override the mapping only wnrre needed.

What do you think?

docs/basic/pgtypes.rst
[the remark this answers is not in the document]
dvarrazzo: Yes, sorry, `basic/adapt` :)

https://www.psycopg.org/psycopg3/docs/basic/adapt.html

psycopg/psycopg/types/enum.py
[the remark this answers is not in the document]
ertaquo: Existing serializers (like `json.dumps`) uses enum values, not keys.
`labels` parameter looks not really clean, as for me. Enum labels in Postgres and in Python should match - otherwise it looks like architecture problem.
I'd insist on `use_keys_as_labels` parameter, it should be more clean and clear and easy to use. Just compare:
```sql
CREATE TYPE some_enum AS ENUM ('ONE', 'TWO');
```
```python
class SomeEnum(Enum):
  ONE = 1
  TWO = 2
...
register_enum(info, SomeEnum, labels={SomeEnum.ONE: "ONE", SomeEnum.TWO: "TWO"}, context=conn)
# vs
register_enum(info, SomeEnum, use_keys_as_labels=True, context=conn)
```

docs/basic/pgtypes.rst
```diff
@@ -365,3 +365,73 @@ connection or cursor), other connections and cursors will be unaffected::
...
+   documentation for the generic usage, especially the
+   `~psycopg.types.TypeInfo.fetch()` method.
+
+   .. attribute:: enum_labels
+
+       Contains labels available in the PostgreSQL enum type.
+
+   .. attribute:: python_type
+
+       After `register_composite()` is called, it will contain the python type
```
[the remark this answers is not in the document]
ertaquo: Fixed

psycopg/psycopg/_typeinfo.py
```diff
@@ -285,6 +285,42 @@ def _get_info_query(
...
+    def __init__(
+        self,
+        name: str,
+        oid: int,
+        array_oid: int,
+        enum_labels: Sequence[str],
+    ):
+        super().__init__(name, oid, array_oid)
+        self.enum_labels = enum_labels
+        # Will be set by register() if the `python_type` is a type
```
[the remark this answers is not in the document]
ertaquo: Fixed, plus added better type hint for `python_type`.

psycopg/psycopg/types/enum.py
```diff
@@ -0,0 +1,95 @@
...
+    format = Format.TEXT
+    python_type: Type[E]
+
+    def __init__(self, oid: int, context: Optional[AdaptContext] = None):
+        super().__init__(oid, context)
+
+    def load(self, data: Buffer) -> E:
+        if isinstance(data, memoryview):
+            data = bytes(data)
+        return self.python_type(data.decode())
```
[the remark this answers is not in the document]
ertaquo: Copy-pasted some code from the `TextLoader`, except `ascii` encoding handling - obviously I need to use strings only. Can you check it please?

psycopg/psycopg/types/enum.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class EnumBinaryLoader(EnumLoader[E]):
+    format = Format.BINARY
+
+
+class EnumDumper(StrDumper):
```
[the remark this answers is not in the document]
ertaquo: Can I use `StrDumper` as base class, but set custom `oid` in attributes?

psycopg/psycopg/types/enum.py
ertaquo: Fixed.

tests/types/test_enum.py
ertaquo: Rewritten tests in more universal way, to use different enums, labels, formats and encodings. Can you check it please?

docs/basic/pgtypes.rst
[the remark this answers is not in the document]
ertaquo: Done.

docs/basic/pgtypes.rst
ertaquo: Fixed.

psycopg/psycopg/types/enum.py
ertaquo: What if just remove the generic dumpers? They are kinda useless, because AFAIK `StrDumperUnknown` is being used for all Python enums by default.

tests/types/test_enum.py
ertaquo: Added general test, I guess it should be enough? Can you check it please?

psycopg/psycopg/types/enum.py
ertaquo: Also generic dumpers breaks tests.
Some tests generates queries based on existing dumpers. There was no `make_Enum` method in Faker, but if I add it:
```python
    def make_Enum(self, spec, length=5):
        return Enum(
            f"FakeEnum{randrange(0, 1 << 90)}",
            {f"K{i}": f"V{i}" for i in range(1, randrange(1, length))},
        )
```
It still raises error:
```
psycopg.ProgrammingError: cannot adapt type 'EnumMeta' using placeholder '%t' (format: TEXT)
```
Excuse me, but I just don't know how to fix it.

psycopg/psycopg/types/enum.py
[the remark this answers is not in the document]
dvarrazzo: That would only work for str enum and might break if the connection is sql_ascii (because it would return bytes, not str, for reasons explained [here](https://www.psycopg.org/psycopg3/docs/basic/adapt.html#strings-adaptation)).

psycopg/psycopg/types/enum.py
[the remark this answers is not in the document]
dvarrazzo: > Enum labels in Postgres and in Python should match - otherwise it looks like architecture problem.

I totally agree with you, from the  point of view of an application developer. However, psycopg is a driver and its main objective is to try to be complete in the features it supports. That's why, if this feature is about "enum adaptation", I feel it should cover every kind of enum and to make possible to create a flexible mapping to make Python and Postgres talk to each other the way the user desires.

That's why I think we should cover other types of enums too, especially int, because Python makes it possible to define them.

I see the inconsistency you say, regards using enum in json. That comes probably from the fact that `MyEnum(str, Enum)` is more a `str` than an `Enum` (`str` comes first in `MyEnum.__mro__`, so, if `MyEnum` doesn't have a dumper registered, the `str` dumper takes effect. For the same reason, `MyIntEnum(int, Enum)` gets dumped successfully (as an integer).

Dumping using keys works with every type of enum (str, int, pure enums): it seems a much more complete solution. ISTM that it's the str enum to be special, because it happens that its values are also valid postgres labels. So, if anything, maybe we could add a parameter `use_values_as_labels=True`, which could be applied only when the values are strings.

Note that, in the example you make above:

```python
register_enum(info, SomeEnum, labels={SomeEnum.ONE: "ONE", SomeEnum.TWO: "TWO"}, context=conn)
```

in this case, `labels` wouldn't be needed at all. I would use it only if there are cases in which the Python key doesn't match the Postgres label. So I'd see that this should work out of the box:

```python
# CREATE TYPE some_enum AS ENUM ('ONE', 'TWO');

class SomeEnum(Enum):
  ONE = 1
  TWO = 2

register_enum(info, SomeEnum, context=conn)
```

and only for keys that don't match you could override the mapping:

```python
class MoreEnum(Enum):
  ONE = 1
  MORE = 99

register_enum(info, SomeEnum, labels={MoreEnum.MORE: "TWO"} context=conn)
```

so it still seems that if the user has a straightforward mapping, then their `register_enum()` is simple; otherwise, the more messy it is, the more complex the mapping.

Maybe it's worth to think more about the scope of the feature; however, covering pure and int enums, seems a necessary thing to habe.

psycopg/psycopg/types/enum.py
dvarrazzo: I'll look into that, but probably we will have to exclude enums from these random tests, because having them would also require to generate a type on the postgres side.

tests/types/test_enum.py
dvarrazzo: It looks very good, thank you

psycopg/psycopg/types/enum.py
dvarrazzo: I made a fix to disable random tests of enums.

psycopg/psycopg/types/enum.py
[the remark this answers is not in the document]
ertaquo: How about little compromise?
What about to use some priority:
 * custom labels first (`labels` parameter);
 * enum values (for string-based or mixed enums);
 * enum keys.
If would work for most cases. Even if there will be some weird enum like:
```python
class WeirdEnum(Enum):
  ONE = "TWO"
  TWO = "THREE"
```
It will be possible to deal with it using custom labels (`labels={item: item.name for item in WeirdEnum}`).
Or even better variant:
```python
class EnumLabels(Enum):
    AUTO = 0  # try values first, then keys
    KEYS = 1 # use keys as labels
    VALUES = 2 # use values as labels

def register_enum(
    info: EnumInfo,
    python_type: Optional[Type[E]] = None,
    labels: Union[EnumLabels, Dict[E, str]] = EnumLabels.AUTO,
    context: Optional[AdaptContext] = None,
) -> None
```
Or
```python
class EnumLabels(Enum):
    AUTO = 0  # try values first, then keys
    KEYS = 1 # use keys as labels
    VALUES = 2 # use values as labels
    CUSTOM_ONLY = 3 # use custom labels only

def register_enum(
    info: EnumInfo,
    python_type: Optional[Type[E]] = None,
    labels: EnumLabels = EnumLabels.AUTO,
    custom_labels: Optional[Dict[E, str]] = None,
    context: Optional[AdaptContext] = None,
) -> None
```

psycopg/psycopg/types/enum.py
ertaquo: Thank you!

dvarrazzo:
I have made some interface and cleanup and some changes in the direction of using keys, in [our enum branch](https://github.com/psycopg/psycopg/tree/enum). I still want to verify if everything makes sense this way.

Note that in the previous branch there was an inconsistency: if you didn't provide an enum to adapt, one would have been created from the postgres enum, but it would have been a pure enum, not a string one.

I want to make more tests especially regarding the default adaptation behaviour. Please let me know if what done here makes sense.

dvarrazzo:
I have added a mapping feature too. I think, with this, enum mapping is feature-complete. [The docs](https://github.com/psycopg/psycopg/blob/enum/docs/basic/adapt.rst#enum-adaptation) should describe well how the labels mapping works.

A review is welcome, otherwise I think I'm happy with this feature.

ertaquo:
Looks great! Thank you very much!

dvarrazzo:
@ertaquo, no, let's not close this merge from here anyway. I have force-pushed to your branch.

dvarrazzo:
Merging the branch as it is. Further changes may be made in master or in followup branches.

Thank you very much for the proposal and the first implementation, @ertaquo!

PR #287, opened 2022-05-01
Re-export from psycopg._cmodule

hauntsaninja:
I'm fixing a bug in mypy's `--no-implicit-reexport` logic in
https://github.com/python/mypy/pull/12704 and mypy-primer flagged this


dvarrazzo:
Thank you!

Issue #296, opened 2022-05-09
Commit result managed too late in pipeline mode?


dvarrazzo:
Hi @dlax,

Looking at [the example](https://www.psycopg.org/psycopg3/docs/advanced/pipeline.html#synchronization-points) added yesterday on the pipeline docs page:

```python
with psycopg.connect(autocommit=True) as conn:
    with conn.pipeline() as p, conn.cursor() as cur:
        cur.execute("INSERT INTO mytable (data) VALUES (%s)", ["one"])
        cur.execute("INSERT INTO no_such_table (data) VALUES (%s)", ["two"])
        conn.execute("INSERT INTO mytable (data) VALUES (%s)", ["three"])
        p.sync()
        cur.execute("INSERT INTO mytable (data) VALUES (%s)", ["four"])
```

there is something fishy here. I think Python shouldn't have reached `four` at all, if not with a try ... except - especially after `sync()` is called. While this might be tolerable, and the error deferred to the end of the block, although with ample disclaimers, this might be too flaky for a COMMIT.

> [from #271] there does not seem to be a need for syncing before commit(), only before rollback() is needed

I am not sure about this after all. Not so much syncing before commit, but after commit it might be useful. From the pipeline mode docs:

> The client must not assume that work is committed when it sends a COMMIT — only when the corresponding result is received to confirm the commit is complete.

Looking at our behaviour instead:

```python
# A table to cause an error on commit
# create table selfref (x serial primary key, y int references selfref (x) deferrable initially deferred);
>>> conn = psycopg.connect()
>>> pg = conn.pipeline()
>>> p = pg.__enter__()
>>> conn.execute("insert into selfref (y) values (-1)")
>>> conn.commit()  # no error here
>>> p.sync()  # no error here either!
>>> conn.execute("select 1")
Traceback (most recent call last)
ForeignKeyViolation: insert or update on table "selfref" violates foreign key constraint "selfref_y_fkey"
DETAIL:  Key (y)=(-1) is not present in table "selfref".
```

I think that it would be reasonable to expect, on `conn.commit()` (or at exit of a `transaction()` block), that, if there is an error invalidating the transaction, it is raised there.

As far as I can see, we can raise an error on commit pretty much only by doing what `Pipeline.__exit__()` does (the outermost one, not even the nested ones):

```python
# ... as above
>>> conn.commit()
>>> p.sync()
>>> conn.wait(p._fetch_gen(flush=False))
Traceback (most recent call last)
[...]

PR #297, opened 2022-05-09
Sync the client on sync() and commit()

dvarrazzo:
As highlighted in #296, we should make sure that on `sync()` and on `commit()` the state of the Python connection reflects what's in the server, in particular we should throw an exception if there is any error.

This MR fixes the behaviour of these functions as well as the nested transaction.

The tests also highlight a difference between pipeline and non-pipeline mode: if commit fails in pipeline mode, the transaction is left INERROR instead of going IDLE.

### DIFF
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -237,7 +237,7 @@ Flushing query results to the client can happen either when a synchronization
 point is established by Psycopg:
 
 - using the `Pipeline.sync()` method;
-- on `Connection.rollback()`;
+- on `Connection.commit()` or `~Connection.rollback()`;
 - at the end of a `!Pipeline` block;
 - using a fetch method such as `Cursor.fetchone()` (which only flushes the
   query but doesn't issue a Sync and doesn't reset a pipeline state error).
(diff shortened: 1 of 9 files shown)


dvarrazzo:
@dlax I managed to clean up all the `_sync_gen()` usage from the `transaction.py` module. Most of it is replaced by the extra `self._pipeline.sync()` in the `transaction()` method, and by the `with pipeline()` you added there.

I changed the pipeline exit code by fetching the result at any level, not only the outermost, effectively reverting 785b1b73e97eb461ed9dca4830684c7e7cb9e3ba. This is requested to pass the nested transactions test (hardened in ea03fd166e0af4c570ee22b89cd0383832fe6aff). If not doing so, instead of using a nested pipeline in `transaction()`:

```python
with tx, self.pipeline():
    yield tx
```

we should have needed the likes of:

```python
with tx:
    try:
        yield tx
    finally:
        self._pipeline.sync()
```

...pretty ugly. it doesn't seem that any feature, or measurable optimization, was relying on the only-fetch-on-outer-pipeline behaviour.

Shall we close this branch here?

dlax:
> it doesn't seem that any feature, or measurable optimization, was relying on the only-fetch-on-outer-pipeline behaviour.

Agreed; this was introduced when there was no reason to do otherwise, but now there is one, and the transaction behaviour that's here in place indeed seems more important.

This branch looks good to me; thanks!

dvarrazzo:
Perfect, thank you very much!

PR #298, opened 2022-05-09
fix: wrap transaction in pipelines if the connection has one

dlax:
When entering a transaction on a connection in pipeline mode, we open an
inner pipeline to ensure that a Sync is emitted at the end of
transaction thus restoring the connection in its expected state (i.e.
the same as in non-pipeline mode).


dvarrazzo:
Much better than the direction the thing was taking, thank you.

PR #299, opened 2022-05-10
fix: wrap transaction in an outer pipeline if the connection has one

dlax:
Follow-up #297, almost resurrecting #271 (but without `nullcontext()`):
* Now running (almost) all transaction tests with and without the connection in pipeline mode;
* Fixing connection state at exit of transaction state;
* See the TODO in tests, about `Connection._check_intrans()`.

### DIFF
--- a/tests/test_transaction.py
+++ b/tests/test_transaction.py
@@ -7,6 +7,11 @@ import pytest
 from psycopg import Connection, ProgrammingError, Rollback
 
 
+@pytest.fixture
+def conn(conn, pipeline):
+    return conn
+
+
 @pytest.fixture(autouse=True)
 def create_test_table(svcconn):
     """Creates a table called 'test_table' for use in tests."""
@@ -70,10 +75,12 @@ class ExpectedException(Exception):
     pass
 
 
-def test_basic(conn):
+def test_basic(conn, pipeline):
     """Basic use of transaction() to BEGIN and COMMIT a transaction."""
     assert not in_transaction(conn)
     with conn.transaction():
+        if pipeline:
+            pipeline.sync()
         assert in_transaction(conn)
     assert not in_transaction(conn)
 
@@ -103,11 +110,13 @@ def test_cant_reenter(conn):
             pass
 
 
-def test_begins_on_enter(conn):
+def test_begins_on_enter(conn, pipeline):
     """Transaction does not begin until __enter__() is called."""
     tx = conn.transaction()
     assert not in_transaction(conn)
     with tx:
+        if pipeline:
+            pipeline.sync()
         assert in_transaction(conn)
     assert not in_transaction(conn)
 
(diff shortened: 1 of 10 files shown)


dlax:
https://github.com/dlax/psycopg3/commit/dd638575fe4aff22ac3ae9db2f1e73ad00743f1e would resolve the TODO about `_check_intrans()`; I can pull it here if it sounds appropriate.

dvarrazzo:
`_check_intrans()` seems only used for user-accessed functions to set the connection state. It doesn't seem there is any performance problem there. I don't see a strong reason to avoid to do so.

tests/test_transaction.py
```diff
@@ -7,6 +7,11 @@
 from psycopg import Connection, ProgrammingError, Rollback
 
 
+@pytest.fixture
+def conn(conn, pipeline):
+    return conn
```
dlax, line 12: This effectively duplicates almost all tests in the module. That a simple and exhaustive approach, but perhaps it's too much from developer point of view? If so, we can instead just pull the `pipeline` fixture in test cases where it matters (which is to be determined).

dvarrazzo: It seems a good idea for me. If large part of the tests make sense to be in both versions we can keep it this way. If let's say half of them or more don't make sense then maybe we can introduce the pipeline fixture where we need.

If we have effectively duplicated some of the tests in the test_pipeline files, we can drop them from there.

dlax:
I don't get why tests on `str(Transaction)` fail and cannot reproduce:
```
AssertionError: assert '[INTRANS]' in '<psycopg.AsyncTransaction (active) [INTRANS, pipeline=ON] (host=127.0.0.1 database=postgres) at 0x7f6c2f3befd0>'
E            +  where '<psycopg.AsyncTransaction (active) [INTRANS, pipeline=ON] (host=127.0.0.1 database=postgres) at 0x7f6c2f3befd0>' = str(<psycopg.AsyncTransaction (active) [INTRANS, pipeline=ON] (host=127.0.0.1 database=postgres) at 0x7f6c2f3befd0>)
```
Where would this `pipeline=ON` in transaction status value come from?

dvarrazzo:
I added it a few hours ago on master: 66bf06794c3b77dcb65101a3300e940127e4c63a

dlax:
> I added it a few hours ago on master: 66bf06794c3b77dcb65101a3300e940127e4c63a

BUT?! This branch here is not based on that commit. So it seems that 
github runs tests on a merge, which is good, but surprising.

Thanks for the hint, I'll rebase then.

dvarrazzo:
Pretty surprising, yes. Probably because the job comes from a MR from your clone of the repository. Indeed the checkout action shows that it's testing 7f56ef1a710fb06bb6f72a9af61d19f67a2d31c1 which seems to be a merge commit created automatically.

PR #302, opened 2022-05-13
Add client-side cursor

### LINKED ISSUE #101: Create a `ClientBindingCursor`?

dvarrazzo:
The heavy lifting for this changeset is actually done in #264. That changeset adds cast to literal values (e.g. casting `date(2000, 1, 1)` to `'2000-01-01'::date` rather than `'2000-01-01'`.

Once that building block is in place, a client-side cursor is little more than a cursor which uses a different `PostgresQuery`, which puts together the query with the values adapted to literals.

As a bonus of having introduced `PostgresClientQuery`, we can now very simply add parameters to `copy()`.

Plenty of tests, documentation, and also added `cursor.mogrify()` which I'm sure a lot of people may want to use, although there are better ways to do the same.

Close #101

### DIFF
--- a/docs/advanced/cursors.rst
+++ b/docs/advanced/cursors.rst
@@ -39,6 +39,58 @@ reasonably small result sets.
 
 
 .. index::
+    double: Cursor; Client-binding
+
+.. _client-side-binding-cursors:
+
+Client-side-binding cursors
+---------------------------
+
+.. versionadded:: 3.1
+
+The previously described :ref:`client-side cursors <client-side-cursors>` send
+the query and the parameters separately to the server. This is the most
+efficient way to process parametrised queries and allows to build several
+features and optimizations. However, not all types of queries can be bound
+server-side; in particular no Data Definition Language query can. See
+:ref:`server-side-binding` for the description of these problems.
+
+The `ClientCursor` (and its `AsyncClientCursor` async counterpart) merge the
+query on the client and send the query and the parameters merged together to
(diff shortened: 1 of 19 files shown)


dlax (approved):
LGTM; nice work!

dlax:
> also added cursor.mogrify() which I'm sure a lot of people may want to use, although there are better ways to do the same.

Are these better ways documented somewhere?

dvarrazzo:
> Are these better ways documented somewhere?

I refer to the whole `psycopg.sql` module, which allows to parametrise also table/fields names, and generic parts of SQL.

dvarrazzo:
> LGTM; nice work!

Thank you very much for reviewing it, @dlax. I've left it simmering a couple of days and I don't see any reason to delay merging to master further.

With this, I'd say 3.1 is ready to release :tada: I will cut a beta, write to the ML to ask for general comments, wait for some feedback, and then release a stable version.

dlax:
> With this, I'd say 3.1 is ready to release tada I will cut a beta, write to the ML to ask for general comments, wait for some feedback, and then release a stable version.

Great!

PR #306, opened 2022-05-19
tweaks about Row type variable

dlax:


psycopg/psycopg/cursor.py
```diff
@@ -781,7 +781,8 @@ def stream(
             first = True
             while self._conn.wait(self._stream_fetchone_gen(first)):
                 # We know that, if we got a result, it has a single row.
-                rec: Row = self._tx.load_row(0, self._make_row)  # type: ignore
+                rec = self._tx.load_row(0, self._make_row)
+                assert rec is not None
```
dvarrazzo, line 785: Actually I dropped this assert investigating in the stream() slowness (https://github.com/psycopg/psycopg/commit/562e8c46bbe5d2c93d4e4ac2203c7255d363c708). It wasn't caused by this assert of course, but I'm still mildly upset that Mypy checks affect runtime. I forgot to update the async side too.

What do you think? Two asserts or two type-ignore?

dvarrazzo:
Thank you for this improvement. Is this change safe to be backported to the 3.0 branch?

psycopg/psycopg/cursor.py
dlax: Ah, didn't noticed that; `# type: ignore` are fine I think, changed accordingly.

dlax:
> Is this change safe to be backported to the 3.0 branch?

Yes, it should be; the commit applies fine to maint-3.0.

dvarrazzo:
Thank you, that's neat :)

PR #307, opened 2022-05-19
fix: sync pipeline after tpc_prepare()

dlax:
This is to get the NotSupported raised directly when calling
tpc_prepare() as expected in test_tpc_disabled().

(Other TPC tests pass when enabling pipeline mode overall.)


dvarrazzo:
Thank you! :+1:

Issue #309, opened 2022-05-25
Transactions in pipeline mode issue more than one BEGIN


dvarrazzo:
WIth this test modified to read notices, it becomes visible:

https://github.com/psycopg/psycopg/blob/8712aecd4c8726d0ba44f1af13d5367e912a5bce/tests/test_pipeline.py#L404-L422

the trace shows:

```
2022-05-25 02:19:19.705316	F	4	Sync
2022-05-25 02:19:19.705436	B	5	ReadyForQuery	 I
2022-05-25 02:19:19.718020	F	13	Parse	 "" "BEGIN" 0
2022-05-25 02:19:19.718039	F	12	Bind	 "" "" 0 0 0
2022-05-25 02:19:19.718042	F	6	Describe	 P ""
2022-05-25 02:19:19.718045	F	9	Execute	 "" 0
2022-05-25 02:19:19.718047	F	6	Close	 P ""
2022-05-25 02:19:19.718085	F	13	Parse	 "" "BEGIN" 0
2022-05-25 02:19:19.718089	F	12	Bind	 "" "" 0 0 0
2022-05-25 02:19:19.718091	F	6	Describe	 P ""
2022-05-25 02:19:19.718093	F	9	Execute	 "" 0
2022-05-25 02:19:19.718096	F	6	Close	 P ""
2022-05-25 02:19:19.718112	F	4	Sync
2022-05-25 02:19:19.718284	B	4	ParseComplete
2022-05-25 02:19:19.718291	B	4	BindComplete
2022-05-25 02:19:19.718293	B	4	NoData
2022-05-25 02:19:19.718296	B	10	CommandComplete	 "BEGIN"
2022-05-25 02:19:19.718312	B	4	CloseComplete
2022-05-25 02:19:19.718315	B	4	ParseComplete
2022-05-25 02:19:19.718317	B	4	BindComplete
2022-05-25 02:19:19.718319	B	4	NoData
2022-05-25 02:19:19.718344	B	111	NoticeResponse	 S "WARNING" V "WARNING" C "25001" M "there is already a transaction in progress" F "xact.c" L "3687" R "BeginTransactionBlock" \x00
2022-05-25 02:19:19.718350	B	10	CommandComplete	 "BEGIN"
2022-05-25 02:19:19.718361	B	4	CloseComplete
2022-05-25 02:19:19.718364	B	5	ReadyForQuery	 T
[...]

PR #310, opened 2022-05-25
fix: sync nested pipeline with pending commands upon enter

dlax:
When entering a nested pipeline and the outer one has pending commands,
we now sync the pipeline. This is probably less surprising at it makes
the implicit transaction from a nested pipeline isolated from the outer
one.

Fix https://github.com/psycopg/psycopg/issues/309.

---
second commit cherry-picked from 9ca0991f892d6ede9e6f5cd6327f9149170ac56a

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -77,9 +77,11 @@ class BasePipeline:
             BasePipeline._is_supported = pq_version >= 140000
         return BasePipeline._is_supported
 
-    def _enter(self) -> None:
+    def _enter_gen(self) -> PQGen[None]:
         if self.level == 0:
             self.pgconn.enter_pipeline_mode()
+        elif self.command_queue:
+            yield from self._sync_gen()
         self.level += 1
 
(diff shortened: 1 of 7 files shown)


psycopg/psycopg/_pipeline.py
```diff
@@ -77,9 +77,11 @@ def is_supported() -> bool:
             BasePipeline._is_supported = pq_version >= 140000
         return BasePipeline._is_supported
 
-    def _enter(self) -> None:
+    def _enter_gen(self) -> PQGen[None]:
         if self.level == 0:
             self.pgconn.enter_pipeline_mode()
+        elif self.command_queue:
+            yield from self._sync_gen()
```
dvarrazzo, line 84: I assume that, with this hangeset, we can drop this sync? https://github.com/psycopg/psycopg/blob/4e7cf85e94edef99e9135aa3603e5bf6d6f642c2/psycopg/psycopg/connection.py#L906

dlax: Indeed! Done.

PR #332, opened 2022-07-11
Numpy types support

vertefra:
dumpers for numpy int, bool and uint


tests/types/test_numpy.py
```diff
@@ -0,0 +1,131 @@
+import pytest
+from psycopg.adapt import PyFormat
+from psycopg.types.numpy import register_numpy
+
+pytest.importorskip("numpy")
```
dvarrazzo, line 5: There should be a workflow in CI to run these tests too, but not the main one because numpy shouldn't be a mandatory dependency. See dns/postgis tests:

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/.github/workflows/tests.yml#L37

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/.github/workflows/tests.yml#L94-L98

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/tests/types/test_shapely.py#L13

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/psycopg/tox.ini#L31-L33

tests/types/test_numpy.py
```diff
@@ -0,0 +1,131 @@
...
+
+# np Numeric types
+#
+# np.int8 (-128, 127) -> smallint (alias int2)
+# np.int16 (-32768, 32767) -> smallint
+# np.int32 (-2147483648, 2147483647) -> integer (alias int4)
+# np.int64 (-9223372036854775808, 9223372036854775807) -> bigint (alias int8)
+# np.uint8 (0 - 255)
+
+register_numpy()
```
dvarrazzo, line 17: The adapters shouldn't be registered globally: this pollutes the other tests.

Please register numpy types in the tests, on the connection only, or use a fixture to return a numpy-registered connection. See for instance:

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/tests/types/test_shapely.py#L58-L69

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,202 @@
...
+
+class NPUintBinaryDumper(NPUintDumper):
+
+    _int_2_dumper = NPInt2BinaryDumper(Int2)
+    _int_4_dumper = NPInt4BinaryDumper(Int4)
+    _int_8_dumper = NPInt8BinaryDumper(Int8)
+    _int_numeric_dumper = NPIntNumericBinaryDumper(IntNumeric)
+
+
+def register_numpy(context: AdaptContext | None = None) -> None:
```
dvarrazzo, line 179: The  `AdaptContext | None` syntax is not supported on some of the supported Python versions (>= 3.7)

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,202 @@
...
+    adapters = context.adapters if context else postgres.adapters
+
+    adapters.register_dumper(np.int8, NPInt2Dumper)
+    adapters.register_dumper(np.int16, NPInt2Dumper)
+    adapters.register_dumper(np.int32, NPInt4Dumper)
+    adapters.register_dumper(np.int64, NPInt8Dumper)
+    adapters.register_dumper(np.bool8, NPBooleanDumper)
+    adapters.register_dumper(np.uint, NPUintDumper)
```
dvarrazzo, line 187: This is not necessary:

```python
>>> numpy.uint is numpy.uint64
True
```

tests/types/test_numpy.py
```diff
@@ -0,0 +1,131 @@
...
+
+import numpy as np  # noqa: E402
+
+# np Numeric types
+#
+# np.int8 (-128, 127) -> smallint (alias int2)
+# np.int16 (-32768, 32767) -> smallint
+# np.int32 (-2147483648, 2147483647) -> integer (alias int4)
+# np.int64 (-9223372036854775808, 9223372036854775807) -> bigint (alias int8)
+# np.uint8 (0 - 255)
```
dvarrazzo, line 15: This table is not complete: either we fill it in altogether or we drop it (in case it was just a note for you as you were looking into the types).

tests/types/test_numpy.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@pytest.mark.parametrize(
+    "val, expr",
+    [
+        (np.int32(0), "'0'::int4"),
+        (np.int32(-2147483648), "'-2147483648'::int4"),
+        (np.int32(2147483647), "'2147483647'::int4"),
+        # int32 is an alias for intc
+        (np.intc(-2147483648), "'-2147483648'::int4"),
```
dvarrazzo, line 60: There are several aliases in numpy: I would rather have a test to double-check that the types are what expected (e.g. `assert np.intc is np.int32`) and write proper tests on a single alias of the types

tests/types/test_numpy.py
```diff
@@ -0,0 +1,131 @@
...
+        (np.intc(2147483647), "'2147483647'::int4"),
+    ],
+)
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_dump_numpy_to_int4(conn, val, expr, fmt_in):
+
+    assert isinstance(val, np.int32)
+
+    cur = conn.cursor()
+    cur.execute(f"select {expr} = %{fmt_in.value}", (val,))
```
dvarrazzo, line 70: I would also have tests to verify that the postgres types match what expected, e.g. using pg_typeof()

https://github.com/psycopg/psycopg/blob/85c965089ae9cd7234dcced357b6e9b4a0637b6d/tests/types/test_numeric.py#L598

for instance checking that `cur.execute(f"select pg_typeof({expr}) = pg_typeof(%{fmt_in.value})", (val,))` is true.

dvarrazzo (changes requested):
I have noticed that you can add [numpy support to mypy](https://numpy.org/devdocs/reference/typing.html#examples), but I'm not sure it's worth, as we should add numpy as a dev dependency. If we decide we don't, mypy errors must be silenced.

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,202 @@
...
+        return pack_int8(int(obj))
+
+
+class NPIntNumericBinaryDumper(IntNumericDumper):
+
+    format = Format.BINARY
+
+    def dump(self, obj: np.uint) -> Buffer:
+
+        # dump_int_to_numeric_binary
```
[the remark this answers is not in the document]
vertefra: I had to write an adapter in order to implement `bit_length()` method used in this function that numpy doesn't have. I don't know if you like the solution.

tests/types/test_numpy.py
vertefra: Numpy workflow is up and running and passes the test. I still trying to figure it out how to skip a test when numpy is not installed. Looking at `test_shapely`,  my assumption was that `pytest.importskyp` was responsible for skipping the test in case the import failed. But I can't reproduce the same behavior in `test_numpy`. Will figure it out though

tests/types/test_numpy.py
```diff
@@ -0,0 +1,376 @@
...
+        (
+            np.uint(4_294_967_295),
+            "'4294967295'::int8",
+        ),
+    ],
+)
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_dump_numpy_uint(numpy_conn, val, expr, fmt_in):
+
+    assert isinstance(val, np.uint)
```
dvarrazzo, line 367: I think that the problem you are having with the imports is caused by using np objects in the fixtures. You should use pure Python numbers.

```suggestion
        (
            4_294_967_295,
            "'4294967295'::int8",
        ),
    ],
)
@pytest.mark.parametrize("fmt_in", PyFormat)
def test_dump_numpy_uint(numpy_conn, val, expr, fmt_in):

    val = np.uint(val)
```

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+
+# Copyright (C) 2022 The Psycopg Team
+
+import operator
+from typing import Any, Optional, cast
+
+from psycopg.types.numeric import (
+    dump_int_to_numeric_binary,
+)
```
dvarrazzo, line 12:
```suggestion
from psycopg.types.numeric import dump_int_to_numeric_binary
```

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,153 @@
...
+class NPIntBinaryNumericDumper(NPIntNumericDumper):
+
+    format = Format.BINARY
+
+    def dump(self, obj: Any) -> Buffer:
+
+        np_adapted = NPUInt64Adapter(obj)
+
+        # Casting to int mainly to silence mypy complaints
+        return dump_int_to_numeric_binary(cast(int, np_adapted))
```
dvarrazzo, line 129: I'd rather not waste opcodes in the dumpers, they are a pretty hot path.

```suggestion
        return dump_int_to_numeric_binary(int, np_adapted)  # type: ignore
```

psycopg/psycopg/types/numpy.py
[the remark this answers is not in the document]
dvarrazzo: Probably the previous solution was better in that case. Using the adapter is complex and surely hurt performance.

Is there a reson to not implement just:

```python

class NPIntBinaryNumericDumper(NPIntNumericDumper):

    format = Format.BINARY

    def dump(self, obj: Any) -> Buffer:
        return dump_int_to_numeric_binary(int(obj))
```

dvarrazzo (changes requested):

tests/types/test_numpy.py
```diff
@@ -0,0 +1,376 @@
...
+        ),
+        (
+            np.short(0),
+            "'0'::int2",
+        ),
+        (
+            np.short(45),
+            "'45'::int2",
+        ),
+        # short overflow
```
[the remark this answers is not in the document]
vertefra: I was very confused by this behavior as well, and not sure what I had to test in terms of overflow. I read that this “silent wrap around” is not guaranteed for the future, so I added that test to make sure that the behavior will remain consistent.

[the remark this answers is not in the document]
dvarrazzo: Even if the behaviour changes, I don't think there's much we can do. The test will fail, but not for psycopg behaviour or misbehaviour. A short int will surely not contain a value > 32767; the only behaviour I see changing from their part is that the constructor may overflow instead of clipping the value to max.

psycopg/psycopg/types/numpy.py
```diff
@@ -0,0 +1,186 @@
...
+    oid = postgres.types["float8"].oid
+
+    _special = {
+        b"inf": b"'Infinity'::float8",
+        b"-inf": b"'-Infinity'::float8",
+        b"nan": b"'NaN'::float8",
+    }
+
+
+class NPFloat4Dumper(NPFloatDumper):
```
dvarrazzo, line 79: Nit: the general convention in the project is to give dumpers a name similar to the python object they dump, and to loaders a name similar to the Postgres type they load.

Am I too pedantic if I ask to call the dumpers with a name similar to the numpy type name?

```suggestion
class NPFloat32Dumper(NPFloatDumper):
```

tests/types/test_numpy.py
```diff
@@ -0,0 +1,379 @@
...
+    cur.execute(f"select {expr} = %{fmt_in.value}", (bool(val),))
+    assert cur.fetchone()[0] is True
+
+
+@pytest.mark.parametrize(
+    "val, expr",
+    [(0, "'0'::int2"), (255, "'255'::int2")],
+)
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_dump_numpy_ubyte_(numpy_conn, val, expr, fmt_in):
```
dvarrazzo, line 162: Please call the tests after the canonical types, not the aliases. You can also avoid `numpy` in the name, as it's clear from the test module.

```suggestion
def test_dump_uint8(numpy_conn, val, expr, fmt_in):
```

tests/types/test_numpy.py
```diff
@@ -0,0 +1,379 @@
...
+)
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_dump_numpy_float64(numpy_conn, val, expr, fmt_in):
+
+    val = np.float64(val)
+
+    cur = numpy_conn.cursor()
+
+    cur.execute(f"select pg_typeof({expr}) = pg_typeof(%{fmt_in.value})", (val,))
+    cur.fetchone()[0] is True
```
dvarrazzo, line 371: Missing assert here and in other places
```suggestion
    assert cur.fetchone()[0] is True
```

tests/types/test_numpy.py
```diff
@@ -0,0 +1,379 @@
...
+
+    cur.execute(
+        f"select abs(({expr}::float4 - %{fmt_in.value}) / {expr}::float4)", (val,)
+    )
+    res = cur.fetchone()[0]
+
+    assert res <= 1e-6
```
dvarrazzo, line 353: This test doesn't convince me.

We are testing the dumper, so what we want to be sure is that in the leg from the numpy.XXX type to postgres we don't change the data we are dumping. As such I don't think it's a good idea to do arithmetic, and it wouldn't even be correct to select the value in Python and test the result, because we would test the loading part too. So, the test should be in postgres itself.

I expect us to be able to write test that verify that we don't add approximation to floating point values, so we should check that `a = b` in postgres, which is not what you usually do with floats (you test that |a - b| < Ɛ), but here we really want to be bit-precise, if we can.

For the same reason, I'd express the value in the fixtures as string instead of numbers, so we never pass through the Python float type and its approximations.

How about the following? I also added a few other values in the samples, which are obtained by printing the repr of the numpy values (`repr(numpy.float32(numpy.e))`)


```suggestion
@pytest.mark.parametrize(
    "val",
    [
        "256e6",  # a large value
        "256e-6",  # a small value
        "2.7182817",  # e
        "3.1415927",  # pi
    ],
)
@pytest.mark.parametrize("fmt_in", PyFormat)
def test_dump_numpy_float32(numpy_conn, val, fmt_in):

    val = np.float32(val)

    cur = numpy_conn.cursor()

    cur.execute(f"select pg_typeof({val}::float4) = pg_typeof(%{fmt_in.value})", (val,))
    assert cur.fetchone()[0] is True

    cur.execute(f"select {val}::float4 = %{fmt_in.value}", (val,))
    assert cur.fetchone()[0] is True
```

dvarrazzo (changes requested):

dvarrazzo:
Test pass clean finally :slightly_smiling_face:

psycopg/psycopg/postgres.py
```diff
@@ -108,18 +108,22 @@
...
     net.register_default_adapters(context)
     none.register_default_adapters(context)
-    numeric.register_default_adapters(context)
     range.register_default_adapters(context)
     string.register_default_adapters(context)
     uuid.register_default_adapters(context)
+
+    # Make sure to register dumpers to the standard types last, in order to
+    # pick them by default when looking up by oid.
+    numpy.register_default_adapters(context)
```
dlax, line 127: Strange comment: numpy's types are not "standard types", no?

psycopg/psycopg/types/numeric.py
```diff
@@ -43,10 +52,6 @@ def dump(self, obj: Any) -> Buffer:
 
         return str(obj).encode()
```
dlax, line 53:
```suggestion
        return super().dump(obj)
```

dlax (commented):
Changes look good to me; just a few nits.

That's a really great addition!

psycopg/psycopg/postgres.py
dvarrazzo: Maybe this comment needs rewording then 🙂

What I meant is that there was [this failure](https://github.com/psycopg/psycopg/actions/runs/3716378652/jobs/6302634045#step:9:599). This was caused by the fact that both Python's Decimal and numpy's uint64 have a dumper registered with the Postgres' numeric oid. But the former covers the entire numeric domain, the latter just the integers. If the latter is registered later then, if we look up for dumpers by oids, we will pick it even in order to dump a Decimal, [we will cast it to int](https://github.com/vertefra/psycopg/blob/0807d176e1c01c5fd2a31363070b1c8b1c59f355/psycopg/psycopg/types/numpy.py#L67), dropping the decimal part.

The situation happens whenever there is more than one dumper registered which can produce the same oid. So far, IIRC, it only happened for [the dbapi-compliant Binary wrapper](https://github.com/psycopg/psycopg/blob/14bfb96c4ad53a0c426406a587bea5d45fd605e4/psycopg/psycopg/dbapi20.py#L110-L112), whose dumper is chosen in preference of the basic binary dumper, because it can cover both types of objects.

dvarrazzo:
State of this brach to complete, methinks:

- [ ] add test to verify that lists of these types are handled correctly (although probably this is already covered by the random tests).
- [ ] maybe: verify that bumpy arrays are handled correctly by array dumpers. This would be inefficient, but would start covering the array domain. See #336 to deal with them properly.
- [ ] test with C dumpers too. At the moment this is cumbersome because of the way the CI workflow is organised, especially because tox gets in the way. Easy to add after finishing #462

psycopg/psycopg/types/numeric.py
dvarrazzo: TBH I really don't like the need of an isinstance per int dump.

Now that we have Enum dumpers, I would like to try if it's possible to associate a pure int dumper to ints (one that just assumes that str(obj) returns the number representation) and associate a dumper converting to int only to enums.

Volatus:
When could this possibly be finalized? @dvarrazzo

dvarrazzo:
Finally merged to master! Thank you very much!

PR #352, opened 2022-08-17
test: use anyio instead of pytest-asyncio

dlax:
Following comments from https://github.com/psycopg/psycopg/pull/146#issuecomment-1217852061, this is in preparation for #29 via #146.

* * *

This is in preparation for adding support for other async libraries,
through anyio. AnyIO pytest plugin is used in replacement for
pytest-asyncio:

- either the pytest.mark.asyncio is replaced by pytest.mark.anyio, or,
- we rely on the 'anyio_backend' fixture that is pulled in 'aconn_cls'
  fixture (and hence 'aconn') providing automatic detection for test
  functions using it.

The 'anyio_backend' fixture is parametrized to only use asyncio and
selects the event loop policy we need on Windows platform as previously
done in pytest_sessionstart(), but only for Python version 3.8 or
higher.

This fixture is defined in main conftest.py, as well as in
pool/conftest.py since we'll change the former to support more async
backend while keeping the later asyncio-only for now.

Some tests use asyncio explicitly so they would not run with another
backend; accordingly, we mark them with
@pytest.mark.parametrize('anyio_backend', ['asyncio']).

Function test_concurrency_async.py::test_ctrl_c is no longer 'async'
because its code does not directly use asyncio (it's done through a
subprocess); but the 'async def' was needed before in order for
pytest-asyncio to run it since the test module had a global
pytest.mark.asyncio (and we were using the "auto" mode).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,9 @@
 import sys
 import asyncio
 import selectors
-from typing import List
+from typing import Any, Dict, List
+
+import pytest
 
 pytest_plugins = (
     "tests.fix_db",
@@ -64,17 +66,21 @@ def pytest_sessionstart(session):
         raise session.Failed
     cache.set("segfault", True)
 
-    # Configure the async loop.
-    loop = session.config.getoption("--loop")
-    if loop == "uvloop":
-        import uvloop
 
-        uvloop.install()
-    else:
-        assert loop == "default"
+asyncio_options: Dict[str, Any] = {}
+if sys.platform == "win32" and sys.version_info >= (3, 8):
+    asyncio_options["policy"] = asyncio.WindowsSelectorEventLoopPolicy()
 
-    if sys.platform == "win32":
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+
+@pytest.fixture(
+    params=[pytest.param(("asyncio", asyncio_options.copy()), id="asyncio")],
+    scope="session",
+)
+def anyio_backend(request):
+    backend, options = request.param
+    if request.config.option.loop == "uvloop":
+        options["use_uvloop"] = True
+    return backend, options
 
 
 allow_fail_messages: List[str] = []
--- a/tests/fix_db.py
+++ b/tests/fix_db.py
@@ -239,7 +239,7 @@ def conn_cls(session_dsn):
 
 
 @pytest.fixture(scope="session")
-def aconn_cls(session_dsn):
+def aconn_cls(session_dsn, anyio_backend):
     cls = psycopg.AsyncConnection
     if crdb_version:
         from psycopg.crdb import AsyncCrdbConnection
--- a/tests/pool/conftest.py
+++ b/tests/pool/conftest.py
@@ -1,0 +1,14 @@
+import pytest
+
+from ..conftest import asyncio_options
+
+
+@pytest.fixture(
+    params=[pytest.param(("asyncio", asyncio_options.copy()), id="asyncio")],
+    scope="session",
+)
+def anyio_backend(request):
+    backend, options = request.param
+    if request.config.option.loop == "uvloop":
+        options["use_uvloop"] = True
+    return backend, options
--- a/tests/test_concurrency_async.py
+++ b/tests/test_concurrency_async.py
@@ -12,8 +12,6 @@ import psycopg
 from psycopg import errors as e
 from psycopg._compat import create_task
 
-pytestmark = pytest.mark.asyncio
-
 
 @pytest.mark.slow
 async def test_commit_concurrency(aconn):
@@ -194,7 +192,7 @@ async def test_identify_closure(aconn_cls, dsn):
     sys.platform == "win32", reason="don't know how to Ctrl-C on Windows"
 )
 @pytest.mark.crdb_skip("cancel")
-async def test_ctrl_c(dsn):
+def test_ctrl_c(dsn):
     script = f"""\
 import signal
 import asyncio
--- a/psycopg/setup.cfg
+++ b/psycopg/setup.cfg
@@ -65,10 +65,10 @@ binary =
 pool =
     psycopg-pool
 test =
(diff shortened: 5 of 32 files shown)


tests/fix_db.py
```diff
@@ -211,8 +211,8 @@ def conn_cls(session_dsn):
     return cls
 
 
-@pytest.fixture(scope="session")
-def aconn_cls(session_dsn):
+@pytest.fixture
+def aconn_cls(dsn, anyio_backend):
```
dvarrazzo, line 215: It seems that this fixture can be defined `scope=function` (the default) but `conn_cls` cannot, it must be `session`. I'd prefer to keep the symmetry: conceptually these object don't change across tests in the same session. Can you define `anyio_backend` as session-scoped or there are problems doing so?

tests/conftest.py
```diff
@@ -65,18 +65,19 @@ def pytest_report_header(config):
...
-    if loop == "uvloop":
-        import uvloop
-
-        uvloop.install()
-    else:
-        assert loop == "default"
-
-    if sys.platform == "win32":
-        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+asyncio_options = {}  # type: ignore[var-annotated]
```
dvarrazzo, line 68: This should do too, with the right imports:

```suggestion
asyncio_options: Dict[str, Any] = {}
```

dvarrazzo: Uhm... but mostly, why do we need a global `asyncio_options` module-level dictionary at all here and don't we do everything in the fixture itself?

tests/test_concurrency_async.py
```diff
@@ -45,7 +44,8 @@ async def runner():
 
 
 @pytest.mark.slow
-async def test_concurrent_execution(aconn_cls, dsn):
+@pytest.mark.parametrize("anyio_backend", ["asyncio"])
+async def test_concurrent_execution(aconn_cls, dsn, anyio_backend):
```
dvarrazzo, line 48: I don't get exactly what is the reason for specifying this parametrize here. I understand that `anyio_backend` is a parametrized fixture that yields exactly one backend right now, right?

I understand that your plan is to eventually change the fixture and make it yield `['asyncio', 'trio']`, but that a test like this cannot run as it is on trio because it uses explicitly `asyncio.gather()`. That might be fine, but I think that pre-emptively skipping this test on trio might hide failures. I would rather not specify a `@parametrize` for this test, look at everything failing on `trio`, and only then write a matching test `test_concurrent_execution_trio`... Although that wouldn't scale in the direction of having different backends.

What do you think will be supported in the future?

Also, alternative implementation, to make tests more readable (although it would add the same shortcoming of hiding tests when a new backend is introduced): how about dropping the module-level `import asyncio`, and having a fixture called `asyncio`, which would yield the asyncio module if we are testing asyncio and would skip the tests if we are testing trio? Same for other backends.

```suggestion
async def test_concurrent_execution(aconn_cls, dsn, asyncio):
```

tests/pool/conftest.py
```diff
@@ -0,0 +1,13 @@
+import pytest
```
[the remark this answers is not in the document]
dlax: Yes, pytest handles this as expected: the `anyio_backend` fixture defined here will override the one defined in outer `conftest.py`. See https://docs.pytest.org/en/latest/how-to/fixtures.html#override-a-fixture-on-a-folder-conftest-level

In fact, that's temporary as this file would go away when we'll port pool code to use anyio.

tests/conftest.py
dlax: The global `asyncio_options` is imported from elsewhere (from `tests/pool/conftest.py` in this PR, and from `test_waiting.py`, at least temporarily in   https://github.com/psycopg/psycopg/pull/146/commits/89bdf08bfe660a1a345d582583fc7b01606e1f01#diff-ec162367f2b494baea5f25f3d81908ca9edbbae2b6f129fd2b3bb3c89d626a20).

tests/test_concurrency_async.py
dlax: Indeed, this `mark` is somehow a left-over which could be dropped. Namely, I initially did not port those tests (the ones using `asyncio` explicitly) and thus needed this `mark`; but ultimately, I decided to port them in https://github.com/psycopg/psycopg/pull/146/commits/f43c709e3e90ebe1701e0c564f19512fd648f192#diff-b3b8cfbbb6a3720571b3ceceb3dc998d151667f62e5738a225bd0234cbde51ff (done through `if anyio_backend_name == "asyncio":`, though we might arguably find a nicer way).

tests/fix_db.py
dlax: I'll give it a try; thanks for the review!

dlax: Seems to work with session-scope.

dhirschfeld:
Looks good barring what appears to be a network issue:

image

saolof:
This was just marked as ready for review? I skimmed through it and it looked good to me, though someone more deeply familiar with the actual psycopg codebase may be better qualified to review this.

dvarrazzo:
Merged, thank you! I'll try to cherry-pick this on maint 3.1 too, otherwise it will be easy to cause conflicts.

dvarrazzo:
This change seems to have caused some instability in the tests on windows:

- https://github.com/psycopg/psycopg/actions/runs/4096339028/jobs/7063905786
- https://github.com/psycopg/psycopg/actions/runs/4091876834/attempts/1

we should figure out what is going on there... :\

dlax:
> we should figure out what is going on there... :\

I think this is because there used to be an unconditional `asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())` which was replaced by the `asyncio_options` dict in the assumption that no asyncio loop would be instantiated outside the anyio fixture, which is wrong in `tests/pool/test_pool_async_noasyncio.py`.

I'll work on a fix; thank you for letting me know.

PR #373, opened 2022-09-14
document PGconn encrypt_password() and tracing methods

dlax:


dvarrazzo:
Amazing, thank you :)

PR #376, opened 2022-09-15
fix: also sync nested pipeline when a transaction is active

### LINKED ISSUE #374: transaction() broken with implicit pipeline transaction
When a pipeline is entered and does not start with an explicit transaction (i.e. a `BEGIN` statement), the server opens an "implicit transaction". But this transaction cannot hold a `SAVEPOINT`. So on our side, after this implicit transaction got started, the connection is in `IDLE` state, so that when entering a `transaction()` block, we try to insert a `SAVEPOINT` and get:
```python
>>> conn = psycopg.connect(autocommit=True)
>>> with conn.pipeline():
...   conn.execute("select 'implicit'")
...   with conn.transaction():
...     conn.execute("select 'tx'")
...   conn.execute("select 'after'")
... 
<psycopg.Cursor [no result] [ACTIVE, pipeline=ON] (database=denis) at 0x7f7203a3ee80>
error ignored terminating <psycopg.Pipeline [IDLE, pipeline=ON] (database=denis) at 0x7f72038bbfa0>: ROLLBACK TO SAVEPOINT can only be used in transaction blocks
  File "<stdin>", line 3, in <module>
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/home/denis/src/psycopg3/psycopg/psycopg/connection.py", line 905, in transaction
    with self.pipeline(), tx, self.pipeline():
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/home/denis/src/psycopg3/psycopg/psycopg/connection.py", line 938, in pipeline
    with pipeline:
  File "/home/denis/src/psycopg3/psycopg/psycopg/_pipeline.py", line 210, in __enter__
    self._conn.wait(self._enter_gen())
  File "/home/denis/src/psycopg3/psycopg/psycopg/connection.py", line 954, in wait
    return waiting.wait(gen, self.pgconn.socket, timeout=timeout)
  File "/home/denis/src/psycopg3/psycopg/psycopg/waiting.py", line 240, in wait_epoll
    s = gen.send(ready)
  File "/home/denis/src/psycopg3/psycopg/psycopg/_pipeline.py", line 96, in _enter_gen
    yield from self._sync_gen()
  File "/home/denis/src/psycopg3/psycopg/psycopg/_pipeline.py", line 107, in _sync_gen
    yield from self._fetch_gen(flush=False)
  File "/home/denis/src/psycopg3/psycopg/psycopg/_pipeline.py", line 156, in _fetch_gen
    self._process_results(queued, results)
  File "/home/denis/src/psycopg3/psycopg/psycopg/_pipeline.py", line 171, in _process_results
    raise e.error_from_result(result, encoding=pgconn_encoding(self.pgconn))
psycopg.errors.NoActiveSqlTransaction: SAVEPOINT can only be used in transaction blocks
```

I'll investigate and hopefully get a fix.

dlax:
Upon enter, transaction() checks the in-transaction status to determine if a BEGIN (when IDLE) or a SAVEPOINT (otherwise) should be inserted. However, in pipeline mode, if the pipeline is in "implicit transaction" (i.e. it has no active BEGIN) and statements have been sent, the in-transaction status might be ACTIVE if those statements have not yet completed (typically, when results have not been fetched).

By issuing a sync() before entering a nested pipeline (which will happen when entering transaction(), if already within a pipeline), we force completion of those statements, and thus get a predictable in-transaction status before entering the transaction() block.

Closes #374.

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -31,6 +31,8 @@ FATAL_ERROR = pq.ExecStatus.FATAL_ERROR
 PIPELINE_ABORTED = pq.ExecStatus.PIPELINE_ABORTED
 BAD = pq.ConnStatus.BAD
 
+ACTIVE = pq.TransactionStatus.ACTIVE
+
 logger = logging.getLogger("psycopg")
 
 
@@ -92,7 +94,13 @@ class BasePipeline:
             )
         if self.level == 0:
             self.pgconn.enter_pipeline_mode()
-        elif self.command_queue:
+        elif self.command_queue or self.pgconn.transaction_status == ACTIVE:
+            # Nested pipeline case.
+            #  Transaction might be ACTIVE when the pipeline uses an "implicit
+            #  transaction", typically in autocommit mode. But when entering a
+            #  Psycopg transaction(), we expect the IDLE state. By sync()-ing,
+            #  we make sure all previous commands are completed and the
+            #  transaction gets back to IDLE.
             yield from self._sync_gen()
         self.level += 1
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -16,6 +16,8 @@ Psycopg 3.1.2 (unreleased)
 - Fix handling of certain invalid time zones causing problems on Windows
   (:ticket:`#371`).
 - Fix segfault occurring when a loader fails initialization (:ticket:`#372`).
+- Fix invalid SAVEPOINT issued when entering `Connection.transaction()` within
+  a pipeline using an implicit transaction (:ticket:`#374`).
 
 
 Current release
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -266,6 +266,23 @@ def test_errors_raised_on_nested_transaction_exit(conn):
     assert cur2.fetchone() == (2,)
 
 
+def test_implicit_transaction(conn):
+    conn.autocommit = True
+    with conn.pipeline():
+        assert conn.pgconn.transaction_status == pq.TransactionStatus.IDLE
+        conn.execute("select 'before'")
+        # Transaction is ACTIVE because previous command is not completed
+        # since we have not fetched its results.
+        assert conn.pgconn.transaction_status == pq.TransactionStatus.ACTIVE
+        # Upon entering the nested pipeline through "with transaction():", a
+        # sync() is emitted to restore the transaction state to IDLE, as
+        # expected to emit a BEGIN.
+        with conn.transaction():
+            conn.execute("select 'tx'")
+        cur = conn.execute("select 'after'")
+    assert cur.fetchone() == ("after",)
+
+
 @pytest.mark.crdb_skip("deferrable")
 def test_error_on_commit(conn):
     conn.execute(
--- a/tests/test_pipeline_async.py
+++ b/tests/test_pipeline_async.py
@@ -267,6 +267,23 @@ async def test_errors_raised_on_nested_transaction_exit(aconn):
     assert await cur2.fetchone() == (2,)
 
 
+async def test_implicit_transaction(aconn):
+    await aconn.set_autocommit(True)
+    async with aconn.pipeline():
+        assert aconn.pgconn.transaction_status == pq.TransactionStatus.IDLE
+        await aconn.execute("select 'before'")
+        # Transaction is ACTIVE because previous command is not completed
+        # since we have not fetched its results.
+        assert aconn.pgconn.transaction_status == pq.TransactionStatus.ACTIVE
+        # Upon entering the nested pipeline through "with transaction():", a
+        # sync() is emitted to restore the transaction state to IDLE, as
+        # expected to emit a BEGIN.
+        async with aconn.transaction():
+            await aconn.execute("select 'tx'")
+        cur = await aconn.execute("select 'after'")
+    assert await cur.fetchone() == ("after",)
+
+
 @pytest.mark.crdb_skip("deferrable")
 async def test_error_on_commit(aconn):
     await aconn.execute(


psycopg/psycopg/_pipeline.py
```diff
@@ -92,7 +94,7 @@ def _enter_gen(self) -> PQGen[None]:
             )
         if self.level == 0:
             self.pgconn.enter_pipeline_mode()
-        elif self.command_queue:
+        elif self.command_queue or self.pgconn.transaction_status == ACTIVE:
```
dvarrazzo, line 97: Can you add a comment here to remember why we need to check for active?

dvarrazzo (changes requested):
Do I understand correctly that, with this change, the tx enter finds the connection INTRANS and hence executes (successfully) a SAVEPOINT?

asqui:
I don't understand how this is working.

Given the description of the problem, I'd intuitively expect the fix to be some change to handle this edge-case in `Transaction.__enter__()`.

Instead, your change is to `Pipeline.__enter__()` and even then, to a branch that only seems to deal with nested pipeline enters.

I don't see where your test case invokes the changed code. What am I missing?

dlax:
> Do I understand correctly that, with this change, the tx enter finds the connection INTRANS and hence executes a SAVEPOINT?

No, the connection is in `IDLE`, and thus `transaction()` emits a `BEGIN`.
See the pq-trace (in particular the `B	5	ReadyForQuery	 I`, just before the `BEGIN`):
```
================ tests.test_pipeline::test_implicit_transaction ================
F	23	Parse	 "" "select 'before'" 0
F	14	Bind	 "" "" 0 0 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" NNNN 0 NNNN 65535 -1 0
B	16	DataRow	 1 6 'before'
B	5	ReadyForQuery	 I
F	13	Parse	 "" "BEGIN" 0
F	14	Bind	 "" "" 0 0 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	4	NoData
B	10	CommandComplete	 "BEGIN"
B	5	ReadyForQuery	 T
F	19	Parse	 "" "select 'tx'" 0
F	14	Bind	 "" "" 0 0 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" NNNN 0 NNNN 65535 -1 0
B	12	DataRow	 1 2 'tx'
B	5	ReadyForQuery	 T
F	14	Parse	 "" "COMMIT" 0
F	14	Bind	 "" "" 0 0 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	4	NoData
B	11	CommandComplete	 "COMMIT"
B	5	ReadyForQuery	 I
F	22	Parse	 "" "select 'after'" 0
F	14	Bind	 "" "" 0 0 1 0
F	6	Describe	 P ""
F	9	Execute	 "" 0
F	4	Sync
B	4	ParseComplete
B	4	BindComplete
B	33	RowDescription	 1 "?column?" NNNN 0 NNNN 65535 -1 0
B	15	DataRow	 1 5 'after'
B	5	ReadyForQuery	 I
```

dlax:
@asqui:
> Given the description of the problem, I'd intuitively expect the fix to be some change to handle this edge-case in `Transaction.__enter__()`.

I could indeed have been there. But there is, in fact, already code handling transaction-specific code in `Pipeline.__enter__()` so it felt more natural to keep everything there. (For completeness, I should mention that code used to be on transaction side, and got moved in commit 459ddafc0cafaa73519c4dcd519147c0f0cd5a81 and others.)

> Instead, your change is to `Pipeline.__enter__()` and even then, to a branch that only seems to deal with nested pipeline enters.

Nested pipelines are essentially (only?) useful for `transaction()`. It's a concept from Psycopg, this does not exist in the libpq: it consists in emitting a sync() message in the (outer and only real) pipeline when entering the "nested" one. This is in order to make sure transaction() behavior in Psycopg is consistent from user's perspective: for instance, making sure a `ROLLBACK` gets executed (on the server) even if there's an error in transaction block.

For example, within a plain lipq pipeline:
```sql
BEGIN;  # global transaction
INSERT INTO s VALUES ('abc');
BEGIN;  # inner transaction
INSERT INTO no_such_table VALUES ('x');
ROLLBACK;  # NOT executed because of FATAL error from previous statement
INSERT INTO s VALUES ('xyz');
COMMIT;

SELECT * from s;
-> []
```
but in Psycopg (thanks to nested pipelines, which works under the hood out of user's control):
```python
with conn.pipeline():
   with conn.transaction():                                                  
          conn.execute("INSERT INTO s VALUES (%s)", ("abc",))
          try:
              with conn.transaction():
                  conn.execute("INSERT INTO no_such_table VALUES (%s)", ("x",))
          except errors.UndefinedTable:
              pass
          conn.execute("INSERT INTO s VALUES (%s)", ("xyz",))

    conn.execute("SELECT * FROM s").fetchall()
    # -> [('abc',), ('xyz',)]
```
(we get the same behaviour with or without the pipeline)

> I don't see where your test case invokes the changed code. What am I missing?

The test case demonstrates that, in autocommit mode within a pipeline, the transaction gets in `ACTIVE` state (it'd be `IDLE` in non-pipeline mode, as expected by `Transaction._push_savepoint()`), and it checks that a `with transaction():` can still be used in this case. See how it fails in issue description #374.
I will add comments to make this clearer.

dvarrazzo:
I had a though similar to @asqui's one: "surely there's a 'if state == IDLE' in tx opening that should become 'if state in (IDLE, ACTIVE)'". But your fix makes totally sense in bringing back the connection into a state that we manage already.

dvarrazzo (approved):

asqui:
Nice, I am clearly way out of touch in my understanding of how pipelines and transactions are interacting. :-)

Thanks for your detailed explanation.

PR #388, opened 2022-09-25
Better _as_python_identifier()

Jolbas:
Let more valid identifiers through _as_python_identifier(). Also 2.5x faster if no change is needed


tests/test_rows.py
```diff
@@ -56,6 +56,12 @@ def test_namedtuple_row(conn):
     assert r2.number == 1
     assert not cur.nextset()
     assert type(r1) is not type(r2)
+    
+    cur.execute('select 1 as üåäö, 1 as _, 1 as "123"')
+    (r3,) = cur.fetchall()
+    assert 'üåäö' in r3._fields
+    assert 'f_' in r3._fields
+    assert 'f123' in r3._fields
```
dlax, line 64: Black wants you to use double-quotes (`"`) here:
```suggestion
    assert "üåäö" in r3._fields
    assert "f_" in r3._fields
    assert "f123" in r3._fields
```

tests/test_rows.py
```diff
@@ -56,6 +56,12 @@ def test_namedtuple_row(conn):
     assert r2.number == 1
     assert not cur.nextset()
     assert type(r1) is not type(r2)
+
```
dlax, line 59: Also, this line has trailing whitespaces:
```suggestion

```

Jolbas: Thanks. I will install black before next PR.

dlax (approved):

Issue #162, opened 2021-11-23
Provide pre-compiled wheels for Mac OS 12 (Monterey)


RootLUG:
Hello,
as the name title suggest, this issue is about providing precompiled wheels for the newest MacOS Monterey (12.0).
The main `psycopg` package installs fine (via poetry or pip! but the provider (psycopg-binary) can't be installed. Pip currently doesn't find the matching wheel/distribution for the latest MacOS, I would guess the OS selector does not match the existing uploaded wheels. 
Tested on py3.9 with poetry & pip, Mac OS 12.0.1; I've checked also wheels on pypi and it appears this is the case for all other python version for which the psycopg-binary is released as well.

dvarrazzo:
Does cibuildwheel enable to create and test such packages on GitHub Actions?

RootLUG:
I don't believe so. In CI I see it builds and test the wheel for Mac OS py3.9 which matches my environment but the wheel name (pip filters on filename) that is build contains also the macos version (psycopg_binary-3.0.4-cp39-cp39-macosx_10_9_x86_64.whl).

Here are some relevant logs I found in buildci:
```
Build options:
  platform: 'macos'
  architectures: {<Architecture.universal2: 'universal2'>, <Architecture.x86_64: 'x86_64'>, <Architecture.arm64: 'arm64'>}
  build_selector: BuildSelector(build_config='cp39-macosx_x86_64')

...

Building cp39-macosx_x86_64 wheel
CPython 3.9 macOS x86_64

...

Building wheels for collected packages: psycopg-binary
    Building wheel for psycopg-binary (PEP 517): started
    Building wheel for psycopg-binary (PEP 517): finished with status 'done'
    Created wheel for psycopg-binary: filename=psycopg_binary-3.0.4-cp39-cp39-macosx_10_9_x86_64.whl size=527183 sha256=46ab30feba31e3b9ac300af5f91fdf53a497d4ee074044e1651435890a5cf1ba
    Stored in directory: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-ephem-wheel-cache-5uwk8le0/wheels/29/20/88/f5f60833d067ccc96553500a229cd797858bab78af94536410
  Successfully built psycopg-binary

```

[...]

PR #396, opened 2022-09-30
Remove warning about missing binary package for M1

magnuswatn:
Saw this warning, and I think it's outdated now, after #162, so I thought I'd just remove it.

### DIFF
--- a/docs/basic/install.rst
+++ b/docs/basic/install.rst
@@ -57,10 +57,6 @@ to update it beforehand.
 The above package should work in most situations. It **will not work** in
 some cases though.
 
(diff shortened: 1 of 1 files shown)


pauloxnet (approved):

dvarrazzo:
You are right, thank you very much for the contribution :)

asqui:
On a related note: The psycopg2 issue template still has "Before opening this ticket, please confirm that..." [...] "I am not trying to install psycopg2-binary on an Apple M1, for which binary packages are not supported." which I think is no longer relevant?

PR #402, opened 2022-10-05
refactor: handle exit exception in BasePipeline._exit()

dlax:
We remove repeated code in `__(a)exit__()` method of `(Async)Pipeline` by passing the exception value to `_exit()` method of `BasePipeline`.


dvarrazzo:
it makes sense. Thank you very much!

PR #409, opened 2022-10-10
make Cursor.stream() work in pipeline mode

### LINKED ISSUE #406: single-row and pipeline modes
Currently, using `Cursor.stream()` in pipeline mode is disabled, but [libpq documentation](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html#LIBPQ-PIPELINE-RESULTS) states that it's possible to set single-row mode when in pipeline mode.

Just to mention that I am working on this at the moment. I can confirm that in pure `psycopg.pq`, it works; but things are more complex with the psycopg API obviously...

dlax:
This adds support for the pipeline for `Cursor.stream()`, aka single-row mode.

Due to how `.stream()` is implemented, it somehow _interrupts_ the pipeline as it forces results fetch alongside query send so that's a bit suboptimal. But at least, people using the pipeline mode would not need to exit/re-enter it to use `.stream()`. 

I think there is a problem in the libpq when using single-row mode and pipeline mode together; this has been [reported](https://www.postgresql.org/message-id/flat/01af18c5-dacc-a8c8-07ee-aecc7650c3e8%40dalibo.com) (waiting for feedback as the time of this writing), and is worked around on our side.

Closes #406


dvarrazzo:
Hi @dlax,

are we sure that the work done here is valuable? It seems to me that streaming and batching go in two different directions, and there are other features that cannot be supported in pipeline mode (e.g. COPY).

Is there a concrete gain in adding this extra complexity and branching in the codebase?

dlax:
Hi @dvarrazzo,

That's a good question and I understand the concern about possible extra maintenance burden.

I do remember that we stated that pipeline and streaming modes were not supposed to be used together in the initial implementation. But then I saw that there was code handling this and documentation in the libpq ,so I thought it could be worth.

Currently, it's a limitation we impose on our side (or an oversight), in contrast with `COPY`.

In fact, I'm not sure these two modes go in "different directions"; namely the pipeline mode helps when network latency matters, and single-row mode would help when, e.g., memory consumption on the client matters. I would suppose someone might be interested in sending a batch of queries, but receiving the results of some of them row-by-row.

Now, as mentioned above, the implementation proposed here does not decouple query sending from results fetch in `.stream()`, so this use case cannot be realized as is (unless the streaming query is the last one in the pipeline). But that's a limitation I'd like to remove in the future (if we decide to go further with this idea).

I also remember you raised the "silly" idea that we could make the connection always in pipelining, and I was thinking about the blocker for this. Maybe that's theoretical, maybe not?

dlax:
It's probably worth doing some refactoring before proceeding with this feature so as to limit branching in the code base. 

Marking as *draft* accordingly now; will come back later to this.

PR #410, opened 2022-10-10
fix: Pass encoding to bytes() when argument is a string

danipozo:
This line:
https://github.com/psycopg/psycopg/blob/b8027b2481b49b87c4b1344d4b098e6f04531ea0/psycopg/psycopg/pq/pq_ctypes.py#L965
raises a `TypeError` when `data` is a `str` object, as the `bytes` builtin must be passed the `encoding` parameter if the first parameter is a string. Example:
```python
>>> import psycopg
>>> from psycopg.pq import Escaping
>>> db = psycopg.connect("dbname=eolica user=postgres host=localhost")
>>> Escaping(db.pgconn).escape_identifier('hello')
  File "<stdin>", line 1, in <module>
  File "/opt/pyenv/versions/3.10.6/lib/python3.10/site-packages/psycopg/pq/pq_ctypes.py", line 960, in escape_identifier
    data = bytes(data)
TypeError: string argument without an encoding
```
I think that this should fix the error, but I haven't been able to test it.

### DIFF
--- a/psycopg/psycopg/pq/pq_ctypes.py
+++ b/psycopg/psycopg/pq/pq_ctypes.py
@@ -18,7 +18,7 @@ from ctypes import addressof, c_char_p, c_int, c_size_t, c_ulong, c_void_p, py_o
 from typing import Any, Callable, List, Optional, Sequence, Tuple
 from typing import cast as t_cast, TYPE_CHECKING
 
-from .. import errors as e
+from .. import errors as e, _encodings
 from . import _pq_ctypes as impl
 from .misc import PGnotify, ConninfoOption, PGresAttDesc
 from .misc import error_message, connection_summary
@@ -961,7 +961,9 @@ class Escaping:
 
         self.conn._ensure_pgconn()
         # TODO: might be done without copy (however C does that)
-        if not isinstance(data, bytes):
+        if isinstance(data, str):
+            data = bytes(data, encoding=_encodings.pgconn_encoding(self.conn))
+        elif not isinstance(data, bytes):
             data = bytes(data)
         out = impl.PQescapeLiteral(self.conn._pgconn_ptr, data, len(data))
         if not out:


dlax:
`escape_identifier` is expecting a `bytes` value, not a `str` but the [documentation](https://www.psycopg.org/psycopg3/docs/api/sql.html) indeed has an example with a string. Maybe the doc needs updating instead?

danipozo:
Yes, I was confused by that example, but that said, wouldn't it be more natural for `escape_identifier` to accept a string? Is there a reason why a SQL identifier should be represented by a `bytes` object (maybe they are required to be ASCII-only)?

dvarrazzo:
I looked at the aptxh and it seems that this function was actually trying to encode objects to bytes, but doing it wrong. Mmm... I think this should be accepted then, but it needs tests. 

It's worth looking at other similar functions and make sure they behave consistently

danipozo:
> I looked at the aptxh

aptxh?

> Mmm... I think this should be accepted then, but it needs tests.

Okey, I can add some tests passing a string.

> It's worth looking at other similar functions and make sure they behave consistently

Do you see this as being done in this same PR or separately? If the former, do you have some pointers where to look at?

dvarrazzo:
> > I looked at the aptxh
> 
> aptxh?

Ehmmm... maybe it was "patch" :smile: 


> > Mmm... I think this should be accepted then, but it needs tests.
> 
> Okey, I can add some tests passing a string.
> 

I've looked into it, but I think that the object is good as it is: everything in the `psycopg.pq` layer only operates at bytes level, and sometimes, like in this case, at memoryview level, to avoid unnecessary data copies.

The principle is that the `psycopg` layer will instead work in strings.

> [@dlax] escape_identifier is expecting a bytes value, not a str but the [documentation](https://www.psycopg.org/psycopg3/docs/api/sql.html) indeed has an example with a string. Maybe the doc needs updating instead?

Yes, that example is wrong. The problem is that the correction is *very* ugly:

```python
>>> print("INSERT INTO %s VALUES (%%s, %%s)" % (bytes(Escaping(conn.pgconn).escape_identifier(table_name.encode()))).decode())
INSERT INTO "foo" VALUES (%s, %s)
```

eww...

So, this is not a bug, but it is a missing feature. We probably need a high level interface to perform the job of `psycopg.pq.Escaping` "in userland", aka in strings converted in the connection encoding when available, utf-8 as fallback.

I have open a conversation at #420. Your input is welcome. Thank you very much!

dvarrazzo:
Yes. I'm also considering that those Python functions are probably not used in tight loop (where it matters in C adapters, probably the libpq function is used), so at least we can restrict them to return bytes to Python, not memoryview.

Opened #422

PR #430, opened 2022-11-03
Fix problems with pool created before the async loop is started

### LINKED ISSUE #219: bug: PoolTimeout on pool with available connections
I'm running a global AsyncConnectionPool that I use in various classes in my project, which seems to be working fine. However, after running the pool for a few hours, it stops giving connections despite having many available ones.

`PoolTimeout: couldn't get a connection after 30.0 sec`

This seems to happen at random after keeping the pool active for about an hour.

Here's the code I use to initialize the pool
```py
global pool


async def start_pool():
    global pool
    pool = AsyncConnectionPool(conninfo=f"""
            host=localhost
            port=5432
            dbname=database
            user=username
            password=password
        """)
    await pool.resize(32, 512)  # Pool size set to large number for testing purposes
```
And here's an example of how I'm using it 
```py
    async def get_stats(self):
        sql = """SELECT * FROM stats WHERE user_id=%s"""
        sqldata = (self.user_id,)
        async with self.pool.connection() as conn:
            async with conn.cursor(row_factory=dict_row) as cur:
                await cur.execute(sql, sqldata)
                result = await cur.fetchone()
        return result
```
Here's the values recieved from `pool.get_stats()` when the issues start to happen:
```js
{'requests_num': 74, 
'requests_queued': 9,
'connections_num': 9, 
'connections_ms': 130, 
'requests_wait_ms': 240030, 
'usage_ms': 1249, 
'requests_errors': 8, 
'pool_min': 32, 
'pool_max': 512, 
'pool_size': 33, 
'pool_available': 0, 
'requests_waiting': 8}
```

Note that I increased the `max_size` to a pretty large number just to see if it was the size causing the issues, but clearly it is not. Either I missed something in the documentation, or something is wrong under the hood.

dvarrazzo:
Close #219.

@w4tsn are you able to test this branch before merging it? Thank you!

### DIFF
--- a/docs/news_pool.rst
+++ b/docs/news_pool.rst
@@ -7,6 +7,16 @@
 ``psycopg_pool`` release notes
 ==============================
 
+Future releases
+---------------
+
+psycopg_pool 3.1.4 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix async pool exhausting connections, happening if the pool is created
+  before the event loop is started (:ticket:`#219`).
+
+
 Current release
 ---------------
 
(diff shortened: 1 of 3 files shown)


w4tsn:
LGTM. Works as expected :+1:

PR #441, opened 2022-11-21
Build parameters in composables

### LINKED ISSUE #339: Build parameters in Composables
`SQL` and `Composable` are great for composing queries. Coming from SQLAlchemy, I made an incorrect assumption that it would allow me to build up the query, including parameters at the same time. It doesn't, but I'm not understanding any technical limitation that would make that a bad idea. Am I missing a good reason to avoid that approach, or some technical limits that make that more difficult than it appears to me?

Here's a rough example, ignorant of the details of what would need to happen, of the kind of thing I'm talking about:

```python
filters = {'a': 1, 'b': 2}
query = (
    sql.SQL("SELECT * FROM {tbl} WHERE {filters}")
    .format(
        tbl=sql.Identifier('mytable'),
        filters=sql.SQL(' AND ').join(
            sql.SQL("{col} = %s", (value,)).format(col=sql.Identifier(row))
            for row, value in filters.items()
        )
    )
)
cursor.execute(query)  # OR
q, params = query.compile()
```

There's sure to be some complexity regarding when to use positional and keyword `%` parameters. And more complexity with composing the parameters, especially if they are positional parameters and we need to figure out where the `%` parameters are compared to the `{}` type formatting as they are composed. This seems like a really nice way to build dynamic queries, at least from my admittedly limited perspective.

ryanhiebert:
I've been sitting on this work for too long. It's not polished, I haven't updated it in a while, but I need to get it out 
for feedback. I hope that it doesn't prove too insulting, and I intend to edit it based on feedback. I _think_ that I was 
getting in my own head, afraid of producing something low-quality, but I don't want it to let it stop me from submitting at 
all.

It did take a little bit to figure out what compromises I wanted to be able to make with it, and how to work through them. 
I think for the most part this is all completely backward compatible. There was one place where I wished I could break 
backward compatibility, the second argument to `Placeholder`. I would prefer that the `value` be the 2nd positional 
parameter, but that would be backward incompatible because `format` is currently allowed to be a positional argument. All 
told, I'd prefer it to be a keyword argument instead. I created a `Param` class that acts just like `Placeholder`, but 
substituting my preferred constructor API so that you can see what it looks like.

Dictionaries and Sequence parameter accumulation works differently from each other, but I think sanely in both cases. I 
don't allow for conflicts, because its a guessing game what that should mean. I do allow for overrides when running the query.

I haven't attempted to write any docs just yet, and I do recognize that is important.

Fix #339

### DIFF
--- a/psycopg/psycopg/_queries.py
+++ b/psycopg/psycopg/_queries.py
@@ -280,7 +280,7 @@ def _validate_and_reorder_params(
 
 
 _re_placeholder = re.compile(
-    rb"""(?x)x
+    rb"""(?x)
         %                       # a literal %
         (?:
             (?:


dvarrazzo:
Hello! Thank you for your contributions

Just a heads up: for a few days I won't be able to review this MR as I'm stuck deeply in a couple of other projects. I'll look into it as soon as I'm relieved of some load.

Thank you very much!

ryanhiebert:
Thank you so much for setting my expectations on your scheduling, it makes me feel that my time and work is valued, even if this doesn’t prove to be the right direction for the project.

dvarrazzo:
Thank you very much for this PR proposal, @ryanhiebert. As anticipated, we will close it here because the feature can be now achieved with [nested template strings](https://www.psycopg.org/psycopg3/docs/basic/tstrings.html#example-nested-templates).

ryanhiebert:
Thank you!

PR #455, opened 2022-12-12
Run tests aaginst CockroachDB v22.2

rafiss:
v22.2 was released recently, so we can test it.

This also removes testing for v21.2, since it is no longer supported.


.github/workflows/tests.yml
```diff
@@ -266,8 +266,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - {crdb: "latest-v22.2", python: "3.10", impl: "c", libpq: "newest"}
```
dvarrazzo, line 269: Let's spice up the crdb tests by using different psycopg implementations (c and Python), different libpq and Python versions

```suggestion
          - {crdb: "latest-v22.2", python: "3.11", impl: "python", libpq: ""}
```

dvarrazzo:
@rafiss A couple of tests fail with CRDB 22.2. Has something changed around cancel? https://github.com/psycopg/psycopg/actions/runs/3680084773/jobs/6225285824#step:8:291

rafiss:
Running from a CLI (without psycopg), yes it seems like something has changed unintentionally.

This fails:
```
root@localhost:26257/defaultdb> cancel session (select session_id from [show session_id]);

ERROR: rpc error: code = Internal desc = An internal server error has occurred. Please check your CockroachDB logs for more details.
```

But this works as expected:
```
root@localhost:26257/defaultdb> show session_id;

             session_id
------------------------------------
  17310e97c6319f200000000000000001
(1 row)

root@localhost:26257/defaultdb> cancel session '17310e97c6319f200000000000000001';

warning: connection lost!
```

rafiss:
I filed https://github.com/cockroachdb/cockroach/issues/93739. In the meantime, I think I can just refactor this test.

dvarrazzo (approved):
Works for me! I have re-introduced a mix of python and libpq versions in the tests, which was pasto-ed away in your commit.

Rebased on master, will merge as tests pass.

Thank you very much!

PR #462, opened 2022-12-17
Remove tox from CI 

dvarrazzo:
I have realised that tox gets in the way of writing an efficient and succinct ci.

Dropping it would allow to have a single test step. Using a simple bash wrapper to retry the tests would be much simpler than using the corresponding tox idiom. The features offered by tox (env isolation) is of no use for local development (virtualenv is fine) and ci execution (runners are isolated).


.github/workflows/tests.yml
```diff
@@ -19,32 +19,37 @@ jobs:
 
   linux:  # {{{
     runs-on: ubuntu-latest
+    if: true
```
dlax, line 22: What is this (and other similar below) for?

psycopg/.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
```
dlax, line 1: Is this file really needed? It seems redundant with the one at root level.

.github/workflows/lint.yml
```diff
@@ -19,17 +19,27 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
-      - name: Install tox
-        run: pip install "tox < 4"
+
+      - name: install packages to tests
+        run: sudo pip install ./psycopg[dev,test] codespell
```
dlax, line 24: It'd be nice to also drop the `sudo pip install` in this file.

dlax (commented):
I'm a bit sad about this because I, as a casual contributor (in general, not so casual for this project) I find it very comfortable to find tox as an entry point for setting up a development environment. I also like the pattern of keeping the CI configuration simple by relying on tox.

Just to be clear, I don't run `tox` either on projects I work on regularly (I just run `pytest` or `mypy` as you do, and other linters are in `$EDITOR`); this applies to psycopg for me.

But if I put myself in *me-as-a-casual-contributor*'s place, I would be annoyed to have to decipher those large YAML files to know how to get a working dev env.

Again, for me, in this project, this does not matter much so I can live with this change. Just wanted to get another perspective.

dvarrazzo:
Thank you for your feedback, @dlax! Of course you are much more than an occasional contributor. Trying to step in the shoes of one such contributor, I think the best favour we may make them is to provide a simple README to start from.

I don't expect people to dig into the `.github` files, I don't think that's the right entry point. However, tox wasn't particularly representative about how to run tests. The test stanza was:

```
    -python -bb -m pytest {posargs}
    -python -bb -m pytest --lf --lfnf=none --no-collect-ok --randomly-seed=last {posargs}
    python -bb -m pytest --lf --lfnf=none --no-collect-ok --randomly-seed=last {posargs}
```

which has _a lot_ to explain, whereas, on desktop, all you need to do for testing is more similar to

```
pip install ./psycopg[test]  # once
pytest
```

the isolated tox environment was causing problems I could never solve, such as why multiprocess tests were failing using the C extension, and it's fundamentally redundant given the throwaway nature of current ci runners.

.github/workflows/tests.yml
dvarrazzo: It is because, when I want to run tests only a subset of the platforms (for instance debugging a windows problem), I wish to disable the other platform running. It can be done adding `if: false` to the other platforms... but only after I make the regular mistake of adding `when: false` and being scolded by github :smile: 

So it's a reminder about the syntax to use to disable a branch.

psycopg/.flake8
dvarrazzo: I have to double-check if they are still both needed. Until recently (where this config was in the `tox.ini` files), the top-level flake8 file was taken into account when running `flake8` from the root of the project, to check all the code, but, when the vim plugin checks the syntax (using `flake8 /absolute/path/to/file.py` I assume), I understand that flake8 stops at the first `setup.py` or similar file it finds, which in most projects identify the root of the project, but not in this "monorepo".

I hate the duplication too. I'll check if I can drop it.

.github/workflows/lint.yml
dvarrazzo: Yes, done. It was needed because I forgot the
```
with:
    python-version: something
```
in the `setup-python` action.

psycopg/.flake8
dvarrazzo: If the top file is dropped, flake8 gets lost in the `.venv` directories. It doesn't have the smarts to check in the `.gitignore`.

```
$ flake8  # finishes quickly
$ rm .flake8 
$ flake8  # several seconds of inactivity
^C... stopped while processing files

$ flake8 -v
flake8.checker            MainProcess     46 INFO     Making checkers
^Cflake8.checker            MainProcess  28457 WARNING  Flake8 was interrupted by the user
... stopped while processing files
```
so that's kinda needed. Trying to drop the directory ones, vim doesn't pick up on the config:

image

I use vim with [ALE](https://github.com/dense-analysis/ale). Using `:ALEInfo` it shows that the command line executed is:
``` 
['/bin/bash', '-c', 'cd ''/home/piro/dev/psycopg3/psycopg'' && ''/home/piro/dev/psycopg3/.venv/bin/flake8''
--format=default --stdin-display-name ''/home/piro/dev/psycopg3/psycopg/psycopg/errors.py'' - < ''/tmp/vOedQyL/4248/errors.py''']
```
the problem with this is the `cd` to the `psycopg` sub-directory of the project instead of the root:

```
(.venv) piro@sherekhan:~/dev/psycopg3$ flake8 /home/piro/dev/psycopg3/psycopg/psycopg/errors.py
# finds no problem

(.venv) piro@sherekhan:~/dev/psycopg3$ cd psycopg
/home/piro/dev/psycopg3/psycopg
(.venv) piro@sherekhan:~/dev/psycopg3/psycopg$ flake8 /home/piro/dev/psycopg3/psycopg/psycopg/errors.py 
# finds the problem
```

I'll check if I can configure ALE better...

dvarrazzo: Ale, as I understood, takes the `psycopg` directory as the root as it looks [one of these files](https://github.com/dense-analysis/ale/blob/8bcbb2995822c4062d00cb41df713fd76092f04f/doc/ale-python.txt#L24).  There seem to be a way to customize it [using `ale_root`](https://github.com/dense-analysis/ale/blob/8bcbb2995822c4062d00cb41df713fd76092f04f/doc/ale.txt#L1953) but I'm not so much an expert of vim so it will take some time to figure out how to write a function e.g. to use `.git` as the project root.

dvarrazzo: Asked in https://github.com/dense-analysis/ale/discussions/4391

dlax: Indeed I also have trouble within vim when the inner file is dropped. That's for the investigation.
But maybe the file could be a symlink?

dlax:
> Trying to step in the shoes of one such contributor, I think the best favour we may make them is to provide a simple README to start from.

Agreed.


> However, tox wasn't particularly representative about how to run tests. The test stanza was:

Oh; I didn't realize that. Indeed, not very helpful.


> the isolated tox environment was causing problems I could never solve, such as why multiprocess tests were failing using the C extension, and it's fundamentally redundant given the throwaway nature of current ci runners.
>
Ok. I understand. So if the tool is not helpful, that's another argument 
for dropping it's usage.


Thank you for these clarifications, let's move on then!

PR #463, opened 2022-12-18
refactor: returning condition directly

apinanyogaratnam:


dvarrazzo:
Very impolite. Don't do that anymore.

Issue #448, opened 2022-12-02
Performance degradation under concurrency


dvarrazzo:
@dvarrazzo  Something bad is happening with concurrency. Increasing the number of concurrent connections makes psycopg performance worse (!). I am talking about both psycopg and psycopg_async.

Here is the modified [script](https://gist.github.com/pwtail/5f056752fed2760bc78d3fea64b6f854) I used.
In the case of sync driver adding concurrency means multiple threads, in the async case - multiple coroutines.

Both psycopg2 and asyncpg behave as expected so using concurrency roughly doubles their performance

_Originally posted by @pwtail in https://github.com/psycopg/psycopg/discussions/411#discussioncomment-4288249_

dvarrazzo:
I have added a `--concurrency` option to the `bench-411.py` script and can verify that concurrent operation seem indeed serialized.

dvarrazzo:
Update: I have been profiling psycopg under concurrency and compare it whith psycopg2. Dumping from my diary:

```
    strace -fTr python tests/scripts/bench-411.py -c1 -n 10000 --no-drop --no-create psycopg > strace-c1 2>&1
    strace -fTr python tests/scripts/bench-411.py -c4 -n 10000 --no-drop --no-create psycopg > strace-c4 2>&1
```

Looking the non-concurrent strace, it is a loop of:

```
     0.000030 sendto(3, "B\0\0\0\34\0_pg3_0\0\0\1\0\1\0\1\0\0\0\2#o\0\1\0\0D\0\0"..., 51, MSG_NOSIGNAL, NULL, 0) = 51 <0.000006>
     0.000018 poll([{fd=3, events=POLLIN|POLLERR}], 1, 100) = 1 ([{fd=3, revents=POLLIN}]) <0.000004>
     0.000014 recvfrom(3, "2\0\0\0\4T\0\0\0\264\0\10id\0\0\2=\273\0\1\0\0\0\27\0\4\377\377\377\377\0"..., 16384, 0, NULL, NULL) = 282 <0.000003>
```

In psycopg2, most often, the interlocked loops don't bother the kernel to grab
the GIL:

```
    [pid T1]      0.000014 sendto(3, "Q\0\0\0\244\nSELECT customer.id, custom"..., 165, MSG_NOSIGNAL, NULL, 0) = 165 <0.000015>
    [pid T1]      0.000028 poll([{fd=3, events=POLLIN|POLLERR}], 1, -1 <unfinished ...>
[...]

Issue #415, opened 2022-10-16
Consider dropping the use of generators/wait


dvarrazzo:
I think that paycopg3 is taking a measurable performance hit with the internal use of generators and wait functions.

This architecture allows to have most code in common between the sync and the async version of the objects but I think that such simplification is paid in term of performance.

I think it's worth as an experiment to refactor a code path to be "linear" and to measure if there is a performance improvement.

asqui:
Hmm interesting. I thought this was a clever way to ensure not too much code duplication between sync and async implementations.

I wonder if there is another standard approach to solving this without the performance impediment?

dvarrazzo:
Yes, it worked well in that regards, but tracing the code it seems that an unnecessary amount of time is spent in that machinery. Every interaction with I/O requires setting up and triggering StopIteration exceptions and in a tight loop I think we feel that (see #411). This might have become faster in Python 3.10 (I should test it, I'm still working with Python 3.8 day-to-day) but I think that, in that version, setting up the exception was optimized, not triggering it.

In terms of avoiding code duplication, I'm considering that we might autogenerate the sync objects from the async ones...

dvarrazzo:
I started doing a quick and incomplete test in the [no-generators](https://github.com/psycopg/psycopg/tree/no-generators) branch.

From a quick test, dropping generators and passing a wait function down to the users doesn't make a big performance difference.

dlax:
@dvarrazzo, which "benchmark" case are you using?

With the one from #411 (removing some superfluous `rollback()` and `connect()`), I see no obvious difference between python 3.9 and 3.10.

dvarrazzo:
@dlax I have dropped it in in the form I was running it as `tests/scripts/bench-411.py`

dvarrazzo:
@dlax note that the test in this branch is half in C (it uses the psycopg.pq.__impl__ == 'c' wrapper) but it uses the Python generators.

What I'd do for a more robust test would be to create new modules, instead of hijacking `generators.py` (and `generators.pyx`) to implement similar functions, but with non-generator interface (i.e. taking a wait function as input) and run some tests to compare the C path.

My plan is to work on this in the next days, but please let me know if you do something, so that we don't duplicate our work :)

dlax:
@dvarrazzo, thanks. I think I'd like to get back to more in-depth benchmarks first because things are not so clear to me yet... I'll let you know in any case!

dlax:
[...]

PR #469, opened 2022-12-28
High level generators

dvarrazzo:
This MR is a refactoring of the generators module, with the idea of having generators doing more high-level work. For instance a public generator would take care of a whole execute-fetch cycle (similar to e.g. what `PQexecParams()` does) whereas, until 3.1, generators take care separately of flushing and receiving data.

The idea is that, using this organization, we can, at a later stage, release the GIL for the whole duration of the send/fetch cycle and avoid the GIL thrashing that might be causing #448. This will also need to actually drop the use of generators and use a blocking function to wait, so we will have to build #415 on top of this in order to really address #448.

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -11,11 +11,11 @@ from typing_extensions import TypeAlias
 
 from . import pq
 from . import errors as e
+from . import generators
 from .abc import PipelineCommand, PQGen
 from ._compat import Deque
 from ._encodings import pgconn_encoding
 from ._preparing import Key, Prepare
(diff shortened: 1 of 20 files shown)


dvarrazzo:
This branch is now pretty stale as it hasn't been worked on for more than one year.

The aim for this work was to make some gain in concurrency situation (#448). However, I have measured that any gain would be dwarfed by network latency even in fast situations. Therefore the interest for this refactoring is now little.

Closing the MR and moving the branch to my personal clone at https://github.com/dvarrazzo/psycopg/tree/high-level-generators in case me, or someone else, may want to revive this work.

PR #470, opened 2022-12-28
No generators

dvarrazzo:
Initial attempt to refactor the generators + wait consumer approach into process functions taking a wait function to block.

There is some speedup to gain from this approach because of less use of generators/exceptions. But generators also require to reacquire the GIL at every yield, so the idea, unless I'm terribly mistaken, is that not using them, when we can, should allow to take longer dives into C code without resurfacing back to Python (and re-acquire the GIL), addressing the concurrency problem exposed in #448.


dvarrazzo:
Note: the code so far doesn't address the async code. With this refactoring we wouldn't be able to use the current structure with a base sync/async class and just differentiating the interface.

If the reorganisation delivers the speed improvement I would like to see, the idea would be to write only the async part of the code and use code generation to build the sync counterpart (basically dropping the async/await keywords and generating the small differences needed some other way).

dvarrazzo:
This branch has gone stale. In the meantime, changes in master happened, and sync code is generated from the async code. This potentially enables other types of refactoring, for instance dropping the shared code from a base class.

Branch moved to https://github.com/dvarrazzo/psycopg/tree/no-generators in case there is some interest. Removing it from the public repo to avoid everyone cloning the repository to get a copy of it.

PR #474, opened 2023-01-03
Use to_regtype instead of ::regtype in type info queries

### LINKED ISSUE #473: Use `to_regtype($1)` instead of `$1::regtype` in TypeInfo.fetch query
Using `$1::regtype` will raise an error if the type does not exist. An example is if the hstore is not installed in the db.

ref https://github.com/sqlalchemy/sqlalchemy/issues/8631#issuecomment-1369037460 and https://github.com/sqlalchemy/sqlalchemy/issues/8631#issuecomment-1369184601

cc @om26er

note: it seems `to_regtype` was added in 9.4 (at least the documentation of 9.3 does not mention it)

CaselIT:
Fixes #473

I haven't modified the test, but it seems that the case of not existing type was already tested. Let me know if you thing additional tests are needed

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,15 @@
 ``psycopg`` release notes
 =========================
 
+Future releases
+---------------
+
+Psycopg 3.1.8 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Don't pollute server logs when types looked for by `TypeInfo.fetch()` 
+  are not found (:ticket:`#473`).
+
 Current release
 ---------------
 
--- a/psycopg/psycopg/_typeinfo.py
+++ b/psycopg/psycopg/_typeinfo.py
@@ -12,13 +12,14 @@ from typing import Sequence, Tuple, Type, TypeVar, Union, TYPE_CHECKING
 from typing_extensions import TypeAlias
 
 from . import errors as e
-from .abc import AdaptContext
+from .abc import AdaptContext, Query
 from .rows import dict_row
 
+
 if TYPE_CHECKING:
-    from .connection import Connection
+    from .connection import BaseConnection, Connection
     from .connection_async import AsyncConnection
-    from .sql import Identifier
+    from .sql import Identifier, SQL
 
 T = TypeVar("T", bound="TypeInfo")
 RegistryKey: TypeAlias = Union[str, int, Tuple[type, int]]
@@ -71,7 +72,7 @@ class TypeInfo:
     @classmethod
     def fetch(
         cls: Type[T],
-        conn: "Union[Connection[Any], AsyncConnection[Any]]",
+        conn: Union["Connection[Any]", "AsyncConnection[Any]"],
         name: Union[str, "Identifier"],
     ) -> Any:
         """Query a system catalog to read information about a type."""
@@ -146,17 +147,36 @@ class TypeInfo:
             register_array(self, context)
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
-        return """\
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
+        from .sql import SQL
+
+        return SQL(
+            """\
 SELECT
     typname AS name, oid, typarray AS array_oid,
     oid::regtype::text AS regtype, typdelim AS delimiter
 FROM pg_type t
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = {regtype}
 ORDER BY t.oid
 """
+        ).format(regtype=cls._to_regtype(conn))
+
+    @classmethod
+    def _has_to_regtype_function(cls, conn: "BaseConnection[Any]") -> bool:
+        # introduced in PostgreSQL 9.4 and CockroachDB 22.2
+        info = conn.info
+        return info.vendor == "PostgreSQL" or (
+            info.vendor == "CockroachDB" and info.server_version >= 220200
+        )
+
+    @classmethod
+    def _to_regtype(cls, conn: "BaseConnection[Any]") -> "SQL":
+        from .sql import SQL
+
+        if cls._has_to_regtype_function(conn):
+            return SQL("to_regtype(%(name)s)")
+        else:
+            return SQL("%(name)s::regtype")
 
     def _added(self, registry: "TypesRegistry") -> None:
         """Method called by the `!registry` when the object is added there."""
@@ -181,16 +201,15 @@ class RangeInfo(TypeInfo):
         self.subtype_oid = subtype_oid
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
+        # CockroachDB does not support range so no need to use _to_regtype
         return """\
 SELECT t.typname AS name, t.oid AS oid, t.typarray AS array_oid,
     t.oid::regtype::text AS regtype,
     r.rngsubtype AS subtype_oid
 FROM pg_type t
 JOIN pg_range r ON t.oid = r.rngtypid
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = to_regtype(%(name)s)
 """
 
     def _added(self, registry: "TypesRegistry") -> None:
@@ -218,20 +237,19 @@ class MultirangeInfo(TypeInfo):
         self.subtype_oid = subtype_oid
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
         if conn.info.server_version < 140000:
             raise e.NotSupportedError(
                 "multirange types are only available from PostgreSQL 14"
             )
+        # CockroachDB does not support multirange so no need to use _to_regtype
         return """\
 SELECT t.typname AS name, t.oid AS oid, t.typarray AS array_oid,
     t.oid::regtype::text AS regtype,
     r.rngtypid AS range_oid, r.rngsubtype AS subtype_oid
 FROM pg_type t
 JOIN pg_range r ON t.oid = r.rngmultitypid
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = to_regtype(%(name)s)
 """
 
     def _added(self, registry: "TypesRegistry") -> None:
@@ -262,9 +280,8 @@ class CompositeInfo(TypeInfo):
         self.python_type: Optional[type] = None
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
+        # CockroachDB does not support composite so no need to use _to_regtype
         return """\
 SELECT
     t.typname AS name, t.oid AS oid, t.typarray AS array_oid,
@@ -281,14 +298,14 @@ LEFT JOIN (
         SELECT a.attrelid, a.attname, a.atttypid
         FROM pg_attribute a
         JOIN pg_type t ON t.typrelid = a.attrelid
-        WHERE t.oid = %(name)s::regtype
+        WHERE t.oid = to_regtype(%(name)s)
         AND a.attnum > 0
         AND NOT a.attisdropped
         ORDER BY a.attnum
     ) x
     GROUP BY attrelid
 ) a ON a.attrelid = t.typrelid
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = to_regtype(%(name)s)
 """
 
 
@@ -310,10 +327,11 @@ class EnumInfo(TypeInfo):
         self.enum: Optional[Type[Enum]] = None
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
-        return """\
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
+        from .sql import SQL
+
+        return SQL(
+            """\
 SELECT name, oid, array_oid, array_agg(label) AS labels
 FROM (
     SELECT
@@ -322,11 +340,12 @@ FROM (
     FROM pg_type t
     LEFT JOIN  pg_enum e
     ON e.enumtypid = t.oid
-    WHERE t.oid = %(name)s::regtype
+    WHERE t.oid = {regtype}
     ORDER BY e.enumsortorder
 ) x
 GROUP BY name, oid, array_oid
 """
+        ).format(regtype=cls._to_regtype(conn))
 
 
 class TypesRegistry:
--- a/tests/test_typeinfo.py
+++ b/tests/test_typeinfo.py
@@ -4,6 +4,10 @@ import psycopg
 from psycopg import sql
 from psycopg.pq import TransactionStatus
 from psycopg.types import TypeInfo
+from psycopg.types.composite import CompositeInfo
+from psycopg.types.enum import EnumInfo
+from psycopg.types.multirange import MultirangeInfo
+from psycopg.types.range import RangeInfo
 
 
 @pytest.mark.parametrize("name", ["text", sql.Identifier("text")])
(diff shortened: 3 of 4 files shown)


psycopg/psycopg/_typeinfo.py
```diff
@@ -154,7 +154,7 @@ def _get_info_query(
     typname AS name, oid, typarray AS array_oid,
     oid::regtype::text AS regtype, typdelim AS delimiter
 FROM pg_type t
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = to_regclass(%(name)s)
```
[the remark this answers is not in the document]
dvarrazzo: Together with these changes, you want to drop the `try... except UndefinedObject` where the query is called, because the function, unlike the cast, is not supposed to raise it.

dvarrazzo (changes requested):
The goal of this change is to avoid an internal error, which is currently handled correctly by psycopg, but which causes noise in the server logs, which we want to avoid.

I think you should add the test that no exception is raised in the `transaction()` context. For instance, this change to the tests fail on master:

```diff
diff --git a/tests/test_typeinfo.py b/tests/test_typeinfo.py
index d0e57e66..3090b644 100644
--- a/tests/test_typeinfo.py
+++ b/tests/test_typeinfo.py
@@ -46,7 +46,15 @@ async def test_fetch_async(aconn, name, status):
 
 @pytest.mark.parametrize("name", ["nosuch", sql.Identifier("nosuch")])
 @pytest.mark.parametrize("status", ["IDLE", "INTRANS"])
-def test_fetch_not_found(conn, name, status):
+def test_fetch_not_found(conn, name, status, monkeypatch):
+    exit_orig = psycopg.Transaction.__exit__
+
+    def exit(self, exc_type, exc_val, exc_tb):
+        assert exc_val is None
+        return exit_orig(self, exc_type, exc_val, exc_tb)
+
+    monkeypatch.setattr(psycopg.Transaction, "__exit__", exit)
+
     status = getattr(TransactionStatus, status)
     if status == TransactionStatus.INTRANS:
         conn.execute("select 1")
```

and should pass in your branch instead.

Can you also please add an entry to `docs/news.rst`? Like:

```
Psycopg 3.1.8 (unreleased)
^^^^^^^^^^^^^^^^^^^^^^^^^^

- Don't pollute server logs when types looked for by `TypeInfo.fetch()` 
  are not found (:ticket:`#473`).
```

CaselIT:
I've updated with the suggested changed. The test change are maybe not needed, since removing the try-except it would now raise in the test with the `::regtype` version, but this is probably clearer

docs/news.rst
```diff
@@ -7,6 +7,15 @@
 ``psycopg`` release notes
 =========================
 
+Future releases
```
dvarrazzo, line 10: Thank you for the care! :+1:

CaselIT:
Mmm, a pipeline fails with `unknown function: to_regtype()` not sure what crdb is testing. Is it an old pg version (the to_regtype function was added in pg9.4, at least according to the docs)

I guess we can do a change the query dynamically depending on the db but it complicates things a bit

CaselIT:
An alternative is using the typename instead of the oid to filter. I don't remember if the name must be unique or not (I guess so), but there's probably not index on it so the query is likely worse

docs/news.rst
CaselIT: I looked at the file history since I did not know what was the format when there are unreleased changes

dvarrazzo:
yes, it looks like we need to generate that query dynamically. We can do that by checking `conn.info.vendor == "CockroachDB" and conn.server_version < 220200`, in which case we need to return the old-style query. We can change that function, because it's internal and declare it returning an `abc.Query` (which can be a string or a `sql.SQL` type of object).

Note that crdb doesn't support ranges and composite, so we would need to make the query dynamic only in the base `TypeInfo` object and in `EnumInfo`

Of course, the try-except needs to be reinstated in order to handle these case.

dvarrazzo:
The problem with using the typename is that it doesn't take the search path into account. We should only look at the types in schemas mentioned in the search path, in the order they are mentioned there, and we should consider the case in which a schema-specified name is passed to `fetch()`, while also consider quoting: `TypeInfo.fetch("foo.bar") is the type `bar` in the schema `foo`, whereas `TypeInfo.fetch('"foo.bar"') is the type `foo.bar` somewhere in the search path.

The idea is that `TypeInfo.fetch()` takes an argument the way Postgres understands it, and if we wanted to use the catalog and the type names to reproduce it it would be a complex job, which I don't think it's worth only to save users a spurious line in the logs.

CaselIT:
I was thinking of looking in the function catalog instead (pg_proc). The price would be a second query (at least once per connection)

Regarding using the name, I agree, having to consider schema and search path makes things complex

dvarrazzo:
> I was thinking of looking in the function catalog instead (pg_proc). The price would be a second query (at least once per connection)

I don't see the reason to do so. We can check by database vendor/version whether that function is supported. It looks like it was added in CRDB 22.2, and it's missing in 22.1.

https://www.cockroachlabs.com/docs/stable/functions-and-operators.html

CaselIT:
> > I was thinking of looking in the function catalog instead (pg_proc). The price would be a second query (at least once per connection)
> 
> I don't see the reason to do so. We can check by database vendor/version whether that function is supported. It looks like it was added in CRDB 22.2, and it's missing in 22.1.
> 
> https://www.cockroachlabs.com/docs/stable/functions-and-operators.html

Well there are 100 and one different dbs that claim to be more or less pg wire compatible, so in sqlalchemy we sometime resort on detecting the feature instead of checking the version. Still, for this we can probably use the version. If needed it can be changed later.

Do you want to also support pg<9.4 that also doesn't have the function (at least according to the docs)?

dvarrazzo:
No, psycopg supports only PostgreSQL >= 10 and CRDB, per agreement with CockroachLabs. We don't support other postgres-like databases, unless they sponsor the project.

If you prefer, you can express the test as "if vendor == postgres or (vendor == crdb and version >= 22.2)" to use `to_regtype()`, so that, as a fallback, the more common `::regtype` expression is used. Should more vendors appear, we can set up a system of capabilities (`has_to_regtype_function`), but for now it would be overkilling.

CaselIT:
Should be updated. I've also made the not found test complete (hopefully that does not break anythin)

I've placed private function in the _typeinfo module to detect the feature. Connection does not seem to have any similar function and I'm not sure if you have preferences about this.
Let me know if you want me to change their location or make them part of the `TypeInfo` class

psycopg/psycopg/_typeinfo.py
```diff
@@ -263,9 +275,8 @@ def __init__(
         self.python_type: Optional[type] = None
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
+    def _get_info_query(cls, conn: _AnyConn) -> str:
+        # CockroachDB does not support range so no need to use _to_regtype
```
CaselIT, line 279: ok this is should say composite. Will change after review

psycopg/psycopg/_typeinfo.py
```diff
@@ -219,20 +232,19 @@ def __init__(
...
             )
+        # CockroachDB does not support range so no need to use _to_regtype
         return """\
 SELECT t.typname AS name, t.oid AS oid, t.typarray AS array_oid,
     t.oid::regtype::text AS regtype,
     r.rngtypid AS range_oid, r.rngsubtype AS subtype_oid
 FROM pg_type t
 JOIN pg_range r ON t.oid = r.rngmultitypid
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = to_regtype(%(name)s)
```
dvarrazzo, line 247: I think this should be something of the likes:
```python
    @classmethod
    def _get_info_query(cls, conn: _AnyConn) -> Query:
        return sql.SQL("""`
SELECT ...
WHERE t.oid = {regtype}
""".format(regtype=cls._to_regtype(conn)))

    @classmethod
    def _to_regtype(cls, conn) -> sql.Composible:
        if ...
            return sql.SQL("to_regtype(%(name)s)")
        else:
            return sql.SQL("(%(name)s::regtype)")
```
this way we respect the `LiteralString` signature of `Query` and `sql.SQL` (Mypy doesn't check it yet, but we try to respect it already).

tests/test_typeinfo.py
```diff
@@ -44,29 +48,69 @@ async def test_fetch_async(aconn, name, status):
     assert info.array_oid == psycopg.adapters.types["text"].array_oid
 
 
-@pytest.mark.parametrize("name", ["nosuch", sql.Identifier("nosuch")])
-@pytest.mark.parametrize("status", ["IDLE", "INTRANS"])
-def test_fetch_not_found(conn, name, status):
+_name = pytest.mark.parametrize("name", ["nosuch", sql.Identifier("nosuch")])
+_status = pytest.mark.parametrize("status", ["IDLE", "INTRANS"])
+_has_to_ragtype = pytest.mark.parametrize("has_to_regtype", [True, False])
```
dvarrazzo, line 53: typo: `_has_to_ragtype`

psycopg/psycopg/_typeinfo.py
```diff
@@ -22,6 +22,7 @@
 
 T = TypeVar("T", bound="TypeInfo")
 RegistryKey: TypeAlias = Union[str, int, Tuple[type, int]]
+_AnyConn = Union["Connection[Any]", "AsyncConnection[Any]"]
```
dvarrazzo, line 25:
```suggestion
_AnyConn: TypeAlias = Union["Connection[Any]", "AsyncConnection[Any]"]
```

with its import

psycopg/psycopg/_typeinfo.py
```diff
@@ -163,6 +162,21 @@ def _added(self, registry: "TypesRegistry") -> None:
         pass
 
 
+def _has_to_regtype_function(conn: _AnyConn) -> bool:
```
dvarrazzo, line 165: I would prefer to have these as TypeInfo methods.

tests/test_typeinfo.py
dvarrazzo, line 53: I wouldn't add this parametrization: it's not an user input. What I would do is maybe to move the test for no exception in `Transaction.__exit__` separate from the others, and to skip it on databases that are known to not pass that test.

You can easily achieve it by using:
```diff
diff --git a/tests/fix_crdb.py b/tests/fix_crdb.py
index ad85698e..32ec457b 100644
--- a/tests/fix_crdb.py
+++ b/tests/fix_crdb.py
@@ -122,6 +122,7 @@ _crdb_reasons = {
     "server-side cursor": 41412,
     "severity_nonlocalized": 81794,
     "stored procedure": 1751,
+    "to_regtype": None,
 }
 
 _crdb_reason_version = {
@@ -130,4 +131,5 @@ _crdb_reason_version = {
     "cancel": "skip < 22",
     "server-side cursor": "skip < 22.1.3",
     "severity_nonlocalized": "skip < 22.1.3",
+    "to_regtype": "skip < 22.2",
 }
```
and decorate the test with `@pytest.mark.crdb_skip("to_regtype")`

dvarrazzo (changes requested):

psycopg/psycopg/_typeinfo.py
CaselIT: Is this needed by some type checkers?

I've updated but I don't think mypy complains about it.

CaselIT:
I've updated. Let's see if the ci is happy.

tests/test_typeinfo.py
```diff
@@ -44,34 +48,165 @@ async def test_fetch_async(aconn, name, status):
     assert info.array_oid == psycopg.adapters.types["text"].array_oid
 
 
-@pytest.mark.parametrize("name", ["nosuch", sql.Identifier("nosuch")])
-@pytest.mark.parametrize("status", ["IDLE", "INTRANS"])
-def test_fetch_not_found(conn, name, status):
+_name = pytest.mark.parametrize("name", ["nosuch", sql.Identifier("nosuch")])
+_status = pytest.mark.parametrize("status", ["IDLE", "INTRANS"])
+_has_to_regtype = pytest.mark.parametrize("has_to_regtype", [True, False])
```
CaselIT, line 53: forgot this

tests/test_typeinfo.py
```diff
@@ -44,34 +48,164 @@ async def test_fetch_async(aconn, name, status):
...
+@_status
+@pytest.mark.crdb_skip("to_regtype")
+def test_fetch_not_found_EnumInfo_to_regtype(conn, name, status, monkeypatch):
+    _fetch_not_found(conn, name, status, monkeypatch, True, EnumInfo)
+
+
+@_name
+@_status
+def test_fetch_not_found_EnumInfo_cast(conn, name, status, monkeypatch):
+    _fetch_not_found(conn, name, status, monkeypatch, False, EnumInfo)
```
dlax, line 122: Would it be possible to add a new
```python
@pytest.mark.parametrize(                                                             
    "info_cls, has_to_regtype",
    [
        pytest.param(TypeInfo, True, marks=pytest.mark.crdb_skip("to_regtype")),
        pytest.param(TypeInfo, False),
        pytest.param(RangeInfo, True, marks=pytest.mark.crdb_skip("range")),
        pytest.param(MultirangeInfo, True, marks=pytest.mark.crdb_skip("range")),
        pytest.param(CompositeInfo, True, marks=pytest.mark.crdb_skip("composite")),
        pytest.param(EnumInfo, True, marks=pytest.mark.crdb_skip("to_regtype")),
        pytest.param(EnumInfo, False),
    ],
)
```
to the previous `def test_fetch_not_found()` instead of those new test functions?

psycopg/psycopg/_typeinfo.py
```diff
@@ -146,17 +148,36 @@ def register(self, context: Optional[AdaptContext] = None) -> None:
...
     oid::regtype::text AS regtype, typdelim AS delimiter
 FROM pg_type t
-WHERE t.oid = %(name)s::regtype
+WHERE t.oid = {regtype}
 ORDER BY t.oid
 """
+        ).format(regtype=cls._to_regtype(conn))
+
+    @classmethod
+    def _has_to_regtype_function(cls, conn: _AnyConn) -> bool:
```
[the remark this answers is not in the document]
dvarrazzo: I asked for these functions to be methods. I prefer them to still have the class available. I'm not of the school that "if you don't need self be a classmethod, if you don't need the class be a staticmethod". I don't see why to paint yourself in a corner with an interface :)

psycopg/psycopg/_typeinfo.py
```diff
@@ -15,13 +15,15 @@
...
+
 if TYPE_CHECKING:
     from .connection import Connection
     from .connection_async import AsyncConnection
-    from .sql import Identifier
+    from .sql import Composed, Identifier, SQL
 
 T = TypeVar("T", bound="TypeInfo")
 RegistryKey: TypeAlias = Union[str, int, Tuple[type, int]]
+_AnyConn: TypeAlias = Union["Connection[Any]", "AsyncConnection[Any]"]
```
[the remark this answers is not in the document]
dvarrazzo: This signature was already used here, @CaselIT only made a declaration for it. I seem to remember that there is a reason to use `Union[sync, async]` - I can't remember which one, but likely something related to something missing on BaseConnection.

Maybe `Union["Connection[Any]", "AsyncConnection[Any]"]` is important enough to deserve a place in `psycopg.abc`. I've been thinking to add it as `AnyConnection`, but there's ambiguity to what the `Any` refers to (the type, or the asyncness?)

tests/test_typeinfo.py
CaselIT: I completely forgot about marks in params. That's what I wanted. Will look at the pr, thanks

CaselIT:
Thanks for the review, will try to update later today

psycopg/psycopg/_typeinfo.py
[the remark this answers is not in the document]
dlax: I agree the already-there `Union` is needed because `BaseConnection` is not enough (namely, in `fetch()`, the `transaction()` method is used). But in newly added methods, only `.info` is used, so `BaseConnection` would be enough. Not a big deal though.

psycopg/psycopg/_typeinfo.py
```diff
@@ -146,17 +148,36 @@ def register(self, context: Optional[AdaptContext] = None) -> None:
             register_array(self, context)
 
     @classmethod
-    def _get_info_query(
-        cls, conn: "Union[Connection[Any], AsyncConnection[Any]]"
-    ) -> str:
-        return """\
+    def _get_info_query(cls, conn: _AnyConn) -> Union[str, "Composed"]:
```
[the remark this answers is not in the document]
dvarrazzo: Probably even better `Composible`, as this could be no problem just a const `sql.SQL`.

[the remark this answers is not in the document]
CaselIT: I think the union here is needed otherwise mypy should complain if a subclass returns a string instead of a `Composible`

psycopg/psycopg/_typeinfo.py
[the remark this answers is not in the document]
CaselIT: I can move it to abc, but it's probably better to do this in a separate PR, since there are probably more cleanup opportunities.

I'm gonna restore the union, and use base connection in the other places so that's easier to cleanup later

psycopg/psycopg/_typeinfo.py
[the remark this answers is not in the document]
CaselIT: nope, Composable does not work
> Argument 1 to "execute" of "Cursor" has incompatible type "Union[str, Composable]"; expected "Union[str, bytes, SQL, Composed]"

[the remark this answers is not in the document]
dvarrazzo: In this case probably abc.Query is the right return type.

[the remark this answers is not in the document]
CaselIT: will change to that

CaselIT:
I've updated. Sorry for the numerous commits, let my know if you prefer if I squash them into a single one

CaselIT:
the fail in the last commit does not seem related to this change. Can you restart the job to see if it passes? (I don't have the option)

tests/test_typeinfo.py
```diff
@@ -44,29 +48,75 @@ async def test_fetch_async(aconn, name, status):
...
+
+    if has_to_regtype:
+
+        def exit(self, exc_type, exc_val, exc_tb):
+            assert exc_val is None
+            return exit_orig(self, exc_type, exc_val, exc_tb)
+
+        monkeypatch.setattr(psycopg.Transaction, "__exit__", exit)
+    else:
+        monkeypatch.setattr(TypeInfo, "_has_to_regtype_function", lambda _: False)
```
dvarrazzo, line 85: I don't understand the need of this branch. Could you please clarify?

CaselIT: the original idea was to test the case of no support for `to_regtype` even for server that do support it.
I guess it could be removed is an old version of cockroach db is always tested (at least in the pipeline)

dvarrazzo:
@CaselIT 

> the fail in the last commit does not seem related to this change. Can you restart the job to see if it passes? (I don't have the option)

Test is re-running. Sorry about that: I don't know why that test failure is resilient to repeating too.

> I've updated. Sorry for the numerous commits, let my know if you prefer if I squash them into a single one

Yes, squashing either everything together or just the typing attempts would be appreciated, thank you :)

tests/test_typeinfo.py
dvarrazzo: Yes, my plan is for CI to run on all the databases/versions supported, so I don't think that testing the monkeypatched version of the adapter is useful. You might even want to drop the `has_to_regtype` param and use `TypeInfo._has_to_regtype_function()` to decide whether to test `__exit__` or not.

CaselIT: ok, let's do that

CaselIT:
Updated and squashed. The failed test on crdb seems unrelated (seen that fail multiple times btw)

dvarrazzo:
Merged, thank you very much!

> The failed test on crdb seems unrelated (seen that fail multiple times btw)

I have pushed a change that should make it more robust.

CaselIT:
thanks for the feedbacks! until next time!

PR #475, opened 2023-01-06
Refactor typeinfo 1

dvarrazzo:
A little refactoring of what just done by @CaselIT, touching on a couple of topic emerged during review. This MR includes two different things, in the two commits, but I hope it's simple enough to review.

Another more complex refactoring is on the way, whose aim is to break the import loop that requires this, and other modules, to use lazy imports, and I'd like to make that MR more self-contained as @dlax has pointed out :)

This MR:

- removes the use of the Union[Connection, AsyncConnection] type, which was only used internally,
- removes the assumption that Postgres > 10 and CRDB will only have ranges after it already acquired `to_regtype()`: use the `cls._to_regtype()` uniformly.

@CaselIT, @dlax What do you think?

### DIFF
--- a/psycopg/psycopg/_typeinfo.py
+++ b/psycopg/psycopg/_typeinfo.py
@@ -15,7 +15,6 @@ from . import errors as e
 from .abc import AdaptContext, Query
 from .rows import dict_row
 
-
 if TYPE_CHECKING:
     from .connection import BaseConnection, Connection
     from .connection_async import AsyncConnection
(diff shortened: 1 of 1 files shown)


psycopg/psycopg/_typeinfo.py
```diff
@@ -163,14 +162,21 @@ def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
...
+        if info.vendor == "PostgreSQL":
+            return info.server_version >= 90400
+        elif info.vendor == "CockroachDB":
+            return info.server_version >= 220200
+        else:
+            return False
 
     @classmethod
     def _to_regtype(cls, conn: "BaseConnection[Any]") -> "SQL":
+        # `to_regtype()` returns the type oid or NULL, unless the :: operator,
```
CaselIT, line 176: `unless` here seems out of place. maybe while sounds better? (not a native speaker, so feel free to ignore)

CaselIT (approved):

psycopg/psycopg/_typeinfo.py
dvarrazzo: Right, it should have been "unlike" 🙂

dlax (approved):

PR #476, opened 2023-01-06
Refactor typeinfo

dvarrazzo:
This MR remove some ugliness in the internal psycopg imports:

- The `TypeInfo` subclasses are now defined in the appropriate module, the ones where they are exported from, e.g. `psycopg.types.range.RangeInfo`
- The import loop between `adapt`, `sql`, `_typeinfo` module is finally broken, so that we don't need so many lazy imports.

The main obstacle to this refactoring was the use of the `psycopg.types` registry in order to define the `oid` class attribute of dumper classes: this required the registry to be populate early at import time. Now, the registry is populated later: still at import time, but only in `psycopg/__init__.py`, by the `register_default_types()` function, and no module expect it to be ready at import time.

To break the import-time dependency on the registry, I've added a `psycopg._oids` module containing just the OID constants. The module is mostly auto-generated, as other OID mappings are.

This MR is based on #475.


dvarrazzo:
@CaselIT, @dlax your opinion is appreciated, as you have seen this area of code lately, but no problem if you would rather not look at it :smile:

psycopg/psycopg/types/enum.py
```diff
@@ -21,6 +24,43 @@
...
+        labels: Sequence[str],
+    ):
+        super().__init__(name, oid, array_oid)
+        self.labels = labels
+        # Will be set by register_enum()
+        self.enum: Optional[Type[Enum]] = None
+
+    @classmethod
+    def _get_info_query(cls, conn: "BaseConnection[Any]") -> Query:
+        from ..sql import SQL
```
CaselIT, line 44: maybe you can remove the lazy load here too?

CaselIT (commented):
Looked quickly at the type info changes, and they make sense to me.
Also the use of constants instead of string keys from a dict make sense to me

psycopg/psycopg/types/enum.py
dvarrazzo: Most likely. Will tweak.

psycopg/psycopg/_transform.py
```diff
@@ -348,3 +350,11 @@ def get_loader(self, oid: int, format: pq.Format) -> "Loader":
                 raise e.InterfaceError("unknown oid loader not found")
         loader = self._loaders[format][oid] = loader_cls(oid, self)
         return loader
+
+
+def _transformer_cls() -> Type["abc.Transformer"]:
```
dlax, line 355: Does it need to be implemented as a function or could it be like it was in `adapt` before, just exposing the `Transformer` class directly? If so, maybe other references to `Transformer` could use the definition from here, instead of from `adapt` to remove the indirection?

dlax: Ah, I see. It'll need the class to be renamed as `_Transformer` or so. And `adapt.Transformer` is part of the public API it seems.

dlax (approved):

psycopg/psycopg/_transform.py
dvarrazzo: Yes, the function is mostly to placate Mypy and flake8 angry about the class redefinition.

However it would be easy to give the class a different name.

```diff
diff --git a/psycopg/psycopg/_transform.py b/psycopg/psycopg/_transform.py
index e75e5c1e..43e39e48 100644
--- a/psycopg/psycopg/_transform.py
+++ b/psycopg/psycopg/_transform.py
@@ -31,7 +31,7 @@ TEXT = pq.Format.TEXT
 PY_TEXT = PyFormat.TEXT
 
 
-class Transformer(AdaptContext):
+class PyTransformer(AdaptContext):
     """
     An object that can adapt efficiently between Python and PostgreSQL.
 
@@ -94,13 +94,13 @@ class Transformer(AdaptContext):
         self._encoding = ""
 
     @classmethod
-    def from_context(cls, context: Optional[AdaptContext]) -> "Transformer":
+    def from_context(cls, context: Optional[AdaptContext]) -> "PyTransformer":
         """
         Return a Transformer from an AdaptContext.
 
         If the context is a Transformer instance, just return it.
         """
-        if isinstance(context, Transformer):
+        if isinstance(context, PyTransformer):
             return context
         else:
             return cls(context)
@@ -352,9 +352,9 @@ class Transformer(AdaptContext):
         return loader
 
 
-def _transformer_cls() -> Type["abc.Transformer"]:
-    # Return the fast object if available
-    if _psycopg:
-        return _psycopg.Transformer
-    else:
-        return Transformer
+Transformer: "Type[abc.Transformer]"
+
+if _psycopg:
+    Transformer = _psycopg.Transformer
+else:
+    Transformer = PyTransformer
diff --git a/psycopg/psycopg/adapt.py b/psycopg/psycopg/adapt.py
index 03e911c9..964c6a55 100644
--- a/psycopg/psycopg/adapt.py
+++ b/psycopg/psycopg/adapt.py
@@ -10,7 +10,7 @@ from typing import Any, Optional, TYPE_CHECKING
 from . import pq, abc
 from . import _adapters_map
 from ._enums import PyFormat as PyFormat
-from ._transform import _transformer_cls
+from ._transform import Transformer as Transformer  # exported here
 
 if TYPE_CHECKING:
     from .connection import BaseConnection
@@ -20,8 +20,6 @@ Buffer = abc.Buffer
 
 ORD_BS = ord("\\")
 
-Transformer = _transformer_cls()
-
 
 class Dumper(abc.Dumper, ABC):
     """
diff --git a/psycopg/psycopg/sql.py b/psycopg/psycopg/sql.py
index ae9695d5..df1f487d 100644
--- a/psycopg/psycopg/sql.py
+++ b/psycopg/psycopg/sql.py
@@ -14,10 +14,7 @@ from .abc import AdaptContext
 from ._enums import PyFormat
 from ._compat import LiteralString
 from ._encodings import conn_encoding
-from ._transform import _transformer_cls
-
-# Don't import Transformer from .adapt to avoid a circular import.
-Transformer = _transformer_cls()
+from ._transform import Transformer
 
 
 def quote(obj: Any, context: Optional[AdaptContext] = None) -> str:
```

I am not sure which solution is less confusing though.

CaselIT: I prefer the alternative name approach, or alternatively a `_py_transformer.py` module that contains only the python fallback code, so that's only executed if the cython version is not installed (we use a similar approach in sqlalchemy, example `engine/_py_row.py`)

dvarrazzo:
By closing the first MR, this got closed automatically by github, and it doesn't seem there is a way to change it and set the target branch to master. Will merge it anyway.

psycopg/psycopg/_transform.py
dvarrazzo: I like the idea to create `_py` modules. Will give it a try.

CaselIT: The objective in sqlalchemy was to have the py fallback code be testable in all cases (even when the cython version is installed) but avoid executing it when importing the library if it's not needed. Using a _py module and conditional import has worked well so far

dlax: > However it would be easy to give the class a different name.

Indeed, it looks reasonable as well.

About the `_py` modules, why not; but we'll need to do it for all of them or none, otherwise, this will be even more confusing, no? Could be a `_py(thon)` subpackage also.

PR #477, opened 2023-01-07
feat: allow to disable pipeline mode from the environment

dlax:
This can be useful in development when testing code paths with automatic pipeline mode, such as executemany(), while still using a recent libpq.

### DIFF
--- a/psycopg/psycopg/_pipeline.py
+++ b/psycopg/psycopg/_pipeline.py
@@ -5,6 +5,7 @@ commands pipeline management
 # Copyright (C) 2021 The Psycopg Team
 
 import logging
+import os
 from types import TracebackType
 from typing import Any, List, Optional, Union, Tuple, Type, TypeVar, TYPE_CHECKING
 from typing_extensions import TypeAlias
@@ -35,6 +36,7 @@ BAD = pq.ConnStatus.BAD
 ACTIVE = pq.TransactionStatus.ACTIVE
 
 logger = logging.getLogger("psycopg")
+disabled = "PSYCOPG_NO_PIPELINE" in os.environ
 
 
 class BasePipeline:
@@ -62,6 +64,8 @@ class BasePipeline:
     @classmethod
     def is_supported(cls) -> bool:
         """Return `!True` if the psycopg libpq wrapper supports pipeline mode."""
+        if disabled:
+            return False
         if BasePipeline._is_supported is None:
             BasePipeline._is_supported = not cls._not_supported_reason()
         return BasePipeline._is_supported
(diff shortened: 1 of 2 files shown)


dvarrazzo:
What is the reason to disable the pipeline mode? Are there cases in which executemany fails in pipeline mode?

Shouldn't it rather change the `pipeline()` decorator to make it no-op? If used, at a glance, it would change the `executemany()` behaviour, but would make `pipeline()` fail like it was used on libpq < 14

If we introduce this change (I'm not opposed, I just would like to understand better the use case), we should also document the feature.

dlax:
It's for being able to cover code paths that are normally not entered if `Pipeline.is_supported()`, such as `BaseCursor._executemany_gen_no_pipeline()`, when running the test suite with libpq >= 14. So it's essentially a convenience for  developers[^1] and I'm not sure it's worth being documented.

That being said, this could also be implemented as a pytest option, e.g. `--no-pipeline`, and monkeypatch; maybe even more convenient.

[^1]: otherwise, I guess one needs two versions of libpq available and tweak `LD_LIBRARY_PATH` or even rebuild; unless I'm missing something?

dlax:
But indeed, this breaks some tests calling `Connection.pipeline()`...

dvarrazzo:
In our test suite, there is an option to install the oldest libpq available in the Debian repository, so we test the non-pipeline code that way.

...at least that was the original idea: looks like [the oldest libpq available now is 14](https://github.com/psycopg/psycopg/actions/runs/3862182187/jobs/6583603324#step:5:48)... There are other code paths that depend on the libpq version so we are not actually testing them :\ I'd be interested in testing with something actually older.

If we want to allow people to test their software in non-pipeline code, that makes sense, but I assume we should document it somehow.

dlax:
In `tools/build/ci_install_libpq.sh`, we could use apt.postgresql.org repository for the `oldest` case as well but with a specific version as advised [here](https://wiki.postgresql.org/wiki/Apt/FAQ#I_want_libpq5_for_version_X.2C_but_there_is_only_version_Y_in_the_repository); trying this in https://github.com/psycopg/psycopg/pull/478

dvarrazzo:
Cool! 

About this MR, if the goal is just about our test suite, not for clients, I think we can just use pytest monkeypatch support. Or am I missing something?

dlax:
Yes; the pytest option seems more appropriate.

But since it doesn't seem so easy to make it work with other things like "pytestmark", I'll work back on it.

dvarrazzo:
I have merged #478. Thank you very much!

Can we close this instead? I think our tests now cover all supported code paths.

dlax:
Yes; closing as the implementation here is not very good.

PR #478, opened 2023-01-07
ci: install the "oldest" libpq from pgdg repository

dlax:


dvarrazzo:
Looks good: it's currently [installing libpq 10](https://github.com/psycopg/psycopg/actions/runs/3863612370/jobs/6585903060#step:5:56).

tools/build/ci_install_libpq.sh
```diff
@@ -18,9 +18,16 @@ case "$libpq" in
...
 
     oldest)
+        curl -sL https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+            | gpg --dearmor \
+            | sudo tee /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg > /dev/null
+
+        rel=$(lsb_release -c -s)
+        echo "deb http://apt.postgresql.org/pub/repos/apt ${rel}-pgdg main 10" \
+            | sudo tee -a /etc/apt/sources.list.d/pgdg.list > /dev/null
+        sudo apt-get -qq update
```
dvarrazzo, line 28: This block of line, from `curl` to `apt-get update`, can be made common between the oldest and the newest branch.

dvarrazzo (changes requested):

tools/build/ci_install_libpq.sh
dlax: Not quite, because the "echo" line is different. But I've factor out common bits still. Also simplified key retrieval a bit.

tools/build/ci_install_libpq.sh
```diff
@@ -9,6 +9,13 @@ set -euo pipefail
 set -x
 
 libpq=${1:-}
+rel=$(lsb_release -c -s)
+
+setup_repo() (
+        curl -sL https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+            | sudo tee /etc/apt/trusted.gpg.d/apt.postgresql.org.asc > /dev/null
+        sudo apt-get -qq update
+)
```
dvarrazzo, line 18: We can include the source line too here, passing the version as a parameter.

Also, this looks a strange function definition. Shouldn't it use curly brackets (`function NAME { ... }` or `NAME () { ... }`) The body is also over-indented.

As you have simplified the key retrieval, now we don't need `curl | sudo tee` anymore: we can just use `sudo curl`.

```suggestion
setup_repo () {
    version=${1:-}
    sudo curl -sL -o /etc/apt/trusted.gpg.d/apt.postgresql.org.asc \
        https://www.postgresql.org/media/keys/ACCC4CF8.asc
    echo "deb http://apt.postgresql.org/pub/repos/apt ${rel}-pgdg main ${version}" \
        | sudo tee -a /etc/apt/sources.list.d/pgdg.list > /dev/null
    sudo apt-get -qq update
}
```

It should be possible to call it using `setup_repo 10` in the `oldest` branch and leave it as `setup_repo` in `newest`.

tools/build/ci_install_libpq.sh
```diff
@@ -18,23 +25,21 @@ case "$libpq" in
...
 
     oldest)
+        echo "deb http://apt.postgresql.org/pub/repos/apt ${rel}-pgdg main 10" \
+            | sudo tee -a /etc/apt/sources.list.d/pgdg.list > /dev/null
+        setup_repo
+
         pqver=$(apt-cache show libpq5 | grep ^Version: | tail -1 | awk '{print $2}')
-        sudo apt-get -qq -y --allow-downgrades install \
-            "libpq-dev=${pqver}" "libpq5=${pqver}"
+        sudo apt-get -qq -y --allow-downgrades install "libpq-dev=${pqver}" "libpq5=${pqver}"
```
dvarrazzo, line 33: Observation: there are `sudo` everywhere. We might as well drop all of them and call this script using `sudo` in the CI workflow?

dvarrazzo (changes requested):

tools/build/ci_install_libpq.sh
dvarrazzo: If we dropped sudo, this would become just:

```suggestion
setup_repo () {
    version=${1:-}
    curl -sL -o /etc/apt/trusted.gpg.d/apt.postgresql.org.asc \
        https://www.postgresql.org/media/keys/ACCC4CF8.asc
    echo "deb http://apt.postgresql.org/pub/repos/apt ${rel}-pgdg main ${version}" \
        >> /etc/apt/sources.list.d/pgdg.list
    apt-get -qq update
}
```

It should be possible to call it using `setup_repo 10` in the `oldest` branch and leave it as `setup_repo` in `newest`.

PR #479, opened 2023-01-08
rowcount in executemany()

dlax:
I'm trying to improve the behavior around "rowcount" in `executemany()` as follows:
* when not "returning", `rowcount` is set to the number of changes (e.g. `INSERT`s) that resulted from `executemany()`
* otherwise, `rowcount` is set to the number of rows of the first result set so that it matches with the number of items then returned by `.fetchall()`

This stems from the [inline comment nobody has read](https://github.com/psycopg/psycopg/blob/3.1.7/psycopg/psycopg/cursor.py#L276), and me having tried to understand it.

It *seems* more correct to me, but maybe I'm missing something?

(commit messages have more details)


dvarrazzo:
What I think is that the common expectation is that `executemany("INSERT ...")` and `executemany("INSERT ... RETURNING ...")` return the same value. As a consequence, when we return from the function, we set the sum of all the results' rowcounts as the cursor rowcount. 

This breaks at the moment of calling `nextset()`, if someone is really interested in the content of the results, yes. However I don't think people really read `cursor.rowcount` while iterating on `nextset()`. If any, the most correct result would be maybe to keep the same accumulated `rowcount` even after switching result using `nextset()`.

Reading your comment, I understand that using `returning=False` the previous behaviour is maintained; with `returning=True` the rowcount matches the current result... I think this makes sense, yes.

Is there, in the tests, a test that the value is correct for a command returning `TUPLES_OK` with `returning=False`? I seem to remember that the number of rows affected must be read from different attributes (`ntuples` vs. `command_tuples`?): I'd like to make sure we handle both the cases.

We also need to update the documentation (for both `executemany` and `rowcount`, to explain how tuples are counted) and add a news entry.

Thank you very much!

dlax:
> Is there, in the tests, a test that the value is correct for a command returning TUPLES_OK with returning=False?  seem to remember that the number of rows affected must be read from different attributes (ntuples vs. command_tuples?): I'd like to make sure we handle both the cases.

I think [test_cursor.py::test_executemany_returning_discard()](https://github.com/dlax/psycopg3/blob/4454581d1327e6994ca49ef694a3cf1e56215dde/tests/test_cursor.py#L319) does this; it's using a `RETURNING` clause with `executemany(..., returning=False)`. With `RETURNING`, an `INSERT`/`UPDATE`/`DELETE` will return values based on each row actually inserted/updated/deleted so `ntuples` and `command_tuples` are the same (I can't think of a case where that wouldn't be true). Still, on our side, since we have `returning=False`, using `command_tuples` unconditionally seems valid to me (since we discard query results).

dlax:
> We also need to update the documentation (for both executemany and rowcount, to explain how tuples are counted) and add a news entry.

Of course; done now.

PR #484, opened 2023-01-16
Fixed TypeInfo for ClientCursor.

apollo13:
TypeInfo had a hardcoded binary=True which doesn't work to well for ClientCursor.

Without this change we'd get:
```
  File "/home/florian/sources/django.git/test.py", line 5, in <module>
    print(TypeInfo.fetch(conn, "geometry"))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/florian/.local/share/virtualenvs/django/lib/python3.11/site-packages/psycopg/_typeinfo.py", line 93, in fetch
    cur.execute(cls._get_info_query(conn), {"name": name})
  File "/home/florian/.local/share/virtualenvs/django/lib/python3.11/site-packages/psycopg/cursor.py", line 728, in execute
    raise ex.with_traceback(None)
psycopg.NotSupportedError: client-side cursors don't support binary results
```

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -17,6 +17,7 @@ Psycopg 3.1.8 (unreleased)
   are not found (:ticket:`#473`).
 - Set `Cursor.rowcount` to the number of rows of each result set from
   `~Cursor.executemany()` when called with `!returning=True` (:ticket:`#479`).
+- Fix `TypeInfo.fetch()` when used with `ClientCursor` (:ticket:`#484`).
 
 Current release
 ---------------
--- a/psycopg/psycopg/_typeinfo.py
+++ b/psycopg/psycopg/_typeinfo.py
@@ -93,7 +93,7 @@ class TypeInfo:
         # or intrans)
         try:
             with conn.transaction():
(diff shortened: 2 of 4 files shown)


apollo13:
Hi @dvarrazzo, I hope this PR is correct -- I couldn't find any reference to why `binary=True` was used and whether it was needed at some point.

Assuming the tests pass and this PR is merge-able, when do you think you could cut a release with it? I am asking because I'd like to figure out if we should patch around this in Django or can just raise the minimum required version of psycopg in time for Django 4.2.

dvarrazzo:
Hey hello! Thank you very much for the MR!

The binary was there because all the types involved support binary loaders so it was ok to do that, and going in a general direction of trying to use more binary paths. This predates the introduction of the client cursors, and the combination was missing from the test suite.

Thank you very much for adding tests and release notes too!

I can merge and make a new release pretty soon, maybe even tonight.

dvarrazzo:
@pauloxnet @apollo13 Psycopg 3.1.8 released!

apollo13:
Wow, much appreciated. Thank you for the rapid turnaround!

Issue #488, opened 2023-01-17
Add integration tests with SQLAlchemy


dvarrazzo:
See #487 

We all love (but hate (but love)) semver, but it would be nice to check for accidental breakages.

Testing would slow down. Maybe we should add an "integration tests" workflow to run before release and daily on cron?

@zzzeek, @CaselIT would you like to help me with this? Either providing a MR or suggesting a "script" procedure to run a SQLAlchemy+psycopg test, and which versions to test?

Thank you very much!

CaselIT:
Hi Daniele,

I can look at creating an integration pypeline test for sqlalchemy, that maybe could be extended also to other 3rd party dialects.

How about making it run on commits on the master branch and optionally manually? That would not impact normal PRs but it would catch issues fairly quickly once a new feature is in marter

dvarrazzo:
@CaselIT can you explain better: do you see this pipeline running in the SA repo or in the PG repo?

CaselIT:
I meant in PG repo

There are already tox test in the pipeline that runs using psycopg master in sqlalchemy

zzzeek:
right but I hardly ever look at our "master" runner because it breaks so often for things that aren't our issue.  Though in this case, it did pick up this problem; it runs nightly.    but if psycopg had this in CI in response to new changes then it would be more of an alert that  a particular change is breaking.

CaselIT:
I'm creating something

dvarrazzo:
> I meant in PG repo

That's ok, we are talking about the same idea then :D I was a bit confused.

> right but I hardly ever look at our "master" runner because it breaks so often for things that aren't our issue.

On our side, I would run integration tests on branches master and maint-3.1 of psycopg 3 (and analogous for psycopg2 if easy - it should).

Considering that the goal of this test is to avoid that changes in PG accidentally break SA (and similarly other 3rd party integrations), I think it would be appropriate to run SA tests using the psycopg dialect on the currently supported versions of SA. Maybe 1.4.x and 2.0.x? Looking at your timeline on pypi it seems 1.3.x is not maintained anymore.

Is it worth to test Alembic too or you don't think that a breakage in Alembic would be missed by a test in SA? Should we test it, which versions?

I think that it would be already valuable to test the packages released on PyPI rather than the development branches, the reason being that, if a breakage happens in psycopg maint branch vs. a SA released version, is our fault; if it happens on an SA maint branch against psycopg it's yous :yum: 

[...]

PR #490, opened 2023-01-17
Add sqlalchemy to the pipeline

CaselIT:
Fixes: #488

NOTE: this is currently broken when using the file from pypi because of https://github.com/sqlalchemy/sqlalchemy/issues/9106

Since this pipeline does not run on PR, you can use https://github.com/CaselIT/psycopg/actions/workflows/3rd-party-tests.yml to see it running

### DIFF
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -1,0 +1,107 @@
+name: Test on 3rd party libraries
+
+on:
+  push:
+    branches:
+      - "master"
+      - "sqlalchemy_pipeline"
+    paths-ignore:
+      - "docs/*"
+      - "tools/*"
+  workflow_dispatch:
+
+concurrency:
+  # Cancel older requests of the same workflow in the same branch.
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  sqlalchemy:
+    name: sqlalchemy-${{ matrix.sqlalchemy_label }},psycopg-${{ matrix.impl }},py-${{ matrix.python-version }}
+    # linux should be enough to test if everything works.
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - "3.11"
+          - "3.7"
+        sqlalchemy_label:
+          # what version of sqlalchemy to download is defined in the "include" section below,
+          # in the variable pip_sqlalchemy
+          - git_main
+          - release
+        impl:
+          - c
+          - python
+        include:
+          - sqlalchemy_label: git_main
+            pip_sqlalchemy: git+https://github.com/sqlalchemy/sqlalchemy.git#egg=SQLAlchemy
+          - sqlalchemy_label: release
+            # TODO: remove pre once v2 is stable
+            pip_sqlalchemy: --pre sqlalchemy>=2a
+
+    env:
+      PSYCOPG_IMPL: ${{ matrix.impl }}
+      DEPS: ./psycopg pytest pytest-xdist
+
+    services:
+      pg:
+        image: postgres:15
+        env:
+          POSTGRES_PASSWORD: password
+          POSTGRES_DB: test
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup PG
+        env:
+          PGPASSWORD: password
+        run: |
+          psql -AXqte "host=127.0.0.1 dbname=test user=postgres" << HERE
+          SELECT version();
+          CREATE SCHEMA test_schema;
+          CREATE SCHEMA test_schema_2;
+          CREATE EXTENSION hstore;
+          HERE
+
+      - name: Include psycopg-c to the packages to install
+        if: ${{ matrix.impl == 'c' }}
+        run: |
+          echo "DEPS=$DEPS ./psycopg_c" >> $GITHUB_ENV
+
+      - name: Install pycopg packages
+        run: pip install $DEPS
+
+      - name: Setup and install sqlalchemy
+        run: |
+          pip download --no-deps --no-binary :all: ${{ matrix.pip_sqlalchemy }}
+          mkdir sa_home
+          case $(file --brief --mime-type SQLAlchemy*) in
+            application/gzip)
+              tar -C sa_home -xzf SQLAlchemy*
+              ;;
+            application/zip)
+              unzip -d sa_home -q SQLAlchemy*
+              ;;
+            *)
+              echo "Unexpected format for $(file --mime-type SQLAlchemy*)" >&2
+              exit 1
+              ;;
+          esac
+          mv sa_home/$( ls sa_home ) sa_home/sa
+          cd sa_home/sa
+          pip install .
+
+      - name: Run sqlalchemy tests
+        env:
+          URL: postgresql+psycopg://postgres:password@127.0.0.1/test
+        working-directory: sa_home/sa
+        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -46,13 +46,19 @@ jobs:
       DEPS: ./psycopg pytest pytest-xdist
 
     services:
-      pg:
+      postgresql:
         image: postgres:15
         env:
           POSTGRES_PASSWORD: password
           POSTGRES_DB: test
         ports:
           - 5432:5432
+        # Wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -109,4 +109,4 @@ jobs:
         env:
           URL: postgresql+psycopg://postgres:password@127.0.0.1/test
         working-directory: sa_home/sa
-        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst
+        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst --color=yes
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -1,4 +1,4 @@
-name: Test on 3rd party libraries
+name: 3rd party tests
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+          - sqlalchemy
+        impl:
+          - c
+          - python
+
+    env:
+      PSYCOPG_IMPL: ${{ matrix.impl }}
+      DEPS: ./psycopg[test] ./psycopg_pool pytest pytest-xdist
+
+    # could not translate host name "pg" to address: Temporary failure in name resolution
```
CaselIT, line 38: remove comment

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,89 @@
+name: Test on 3rd party libraries
+
+on:
+  push:
+    branches:
+      - "master"
+      - "sqlalchemy_pipeline"
```
CaselIT, line 7: this branch can be removed (or not)

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+      - "tools/*"
+  workflow_dispatch:
+
+concurrency:
+  # Cancel older requests of the same workflow in the same branch.
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  sqlalchemy:
```
CaselIT, line 19: probably need a custom name, since the default ones are not clear at all

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+    # linux should be enough to test if everything works.
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - "3.11"
```
dvarrazzo, line 26: Only this one? Maybe the newest and the oldest supported (3.7 for pg3).

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+        sqlalchemy:
+          - git+https://github.com/sqlalchemy/sqlalchemy.git#egg=SQLAlchemy
+          - sqlalchemy
```
dvarrazzo, line 29: This, at the moment, means 1.4.x and will become 2.0.x once the latter is out of beta, right? Maybe it's better to be more explicit?

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+            extract="unzip -d sa_home -q"
+          fi
+          $extract $( ls SQLAlchemy* )
+          cd sa_home/$( ls sa_home )
+          pip install .
+
+      - name: Run sqlalchemy tests
+        run: bash -c "cd sa_home/$( ls sa_home ) &&
+          pytest -n 2 -q --dburi postgresql+psycopg://postgres:password@127.0.0.1/test --backend-only --dropfirst
+          "
```
dvarrazzo, line 88: Multiline would be more readable:

```suggestion
        run: |
          cd sa_home/$( ls sa_home )
          pytest -n 2 -q --backend-only --dropfirst \
             --dburi postgresql+psycopg://postgres:password@127.0.0.1/test

```

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup PG
+        env:
+          PGPASSWORD: password
+        run: |
+          run_sql="psql -h 127.0.0.1 -d test -U postgres -c "
+          $run_sql 'SELECT version();'
+          $run_sql 'CREATE SCHEMA test_schema; CREATE SCHEMA test_schema_2;CREATE EXTENSION hstore;'
```
dvarrazzo, line 60: They can be run all together in the same command. We can use -e to echo the commands run. I also prefer to specify a connection string instead of the loose parameters.

```suggestion
        run: |
          psql -AXqte "host=127.0.0.1 dbname=test user=postgres" << HERE
          SELECT version();
          CREATE SCHEMA test_schema;
          CREATE SCHEMA test_schema_2;
          CREATE EXTENSION hstore;
          HERE
```

This would produce an output like:

```
SELECT version();
PostgreSQL 15.1 (Ubuntu 15.1-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0, 64-bit
CREATE SCHEMA test_schema;
CREATE SCHEMA test_schema_2;
CREATE EXTENSION hstore;
```

dvarrazzo (changes requested):

dvarrazzo:
I seem to understand that there isn't a way to run this pipeline unless merged to master, or something like that? I have had this problem whenever I have added a new pipeline.

.github/workflows/3rd-party-tests.yml
CaselIT: The pip command is using `--pre` so it's installing v2 already. but we can add `>=2`

.github/workflows/3rd-party-tests.yml
CaselIT: Makes sense

.github/workflows/3rd-party-tests.yml
CaselIT: I was experimenting, since GitHub doesn't show the output of the test until they are done for some reason.

I wanted to try another alternative to see if it works, if it doesn't I'll use what you suggested

CaselIT:
> I seem to understand that there isn't a way to run this pipeline unless merged to master, or something like that? I have had this problem whenever I have added a new pipeline.

Yes, since otherwise anyone could create a pr adding a workflow that does whatever without the owner control.
You can look at the link I added above to see some runs

.github/workflows/3rd-party-tests.yml
CaselIT: thanks, that looks better!

CaselIT:
I've updated.

On sqlalchemy side we were thinking about releasing an rc3. Once that's live this should pass all jobs (it's currently correctly failing on the version taken from pypi that's rc2, and passing on main)

CaselIT:
Sqlalchemy rc3 was released yesterday, so all jobs should pass now

.github/workflows/3rd-party-tests.yml
dvarrazzo: I don't see this branch in your repo. Should it be removed?

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+              echo "PIP_SQLALCHEMY=git+https://github.com/sqlalchemy/sqlalchemy.git#egg=SQLAlchemy" >> $GITHUB_ENV
+              ;;
+            release)
+              echo "PIP_SQLALCHEMY=sqlalchemy>=2a" >> $GITHUB_ENV
+              ;;
+            *)
+              echo "Unknown value for sqlalchemy: ${{ matrix.sqlalchemy }}"
+              exit 1
+              ;;
+          esac
```
dvarrazzo, line 63: I would prefer to have these entries directly in the matrix and used later at `pip download`, without this intermediate step.

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+        sqlalchemy:
+          - git_main
+          - release
+        impl:
+          - c
+          - python
+
+    env:
+      PSYCOPG_IMPL: ${{ matrix.impl }}
+      DEPS: ./psycopg[test] ./psycopg_pool pytest pytest-xdist
```
dvarrazzo, line 38: I think we don't need the `[test]` of `psycopg` here, and maybe we don't need `psycopg_pool` either, unless you use it?

dvarrazzo (changes requested):

.github/workflows/3rd-party-tests.yml
dvarrazzo, line 38: Out of curiosity, why don't you have a `[test]` extra yourself, specifying `pytest-xdits`? You can then `pip install .[test]` in the `sa` dir to install library and test dependencies.

.github/workflows/3rd-party-tests.yml
CaselIT: it's definetly there, I'm using it to run the pipeline in my fork https://github.com/caselit/psycopg/tree/sqlalchemy_pipeline

My comment was just to mention that I can remove before the merge

.github/workflows/3rd-party-tests.yml
CaselIT: I did this for the name of the job. I'm not sure if I can otherwise create something meaningful. Maybe GitHub supports json objects as variable values? I'll do somr tries

.github/workflows/3rd-party-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+        run: |
+          pip download --pre --no-deps --no-binary :all: $PIP_SQLALCHEMY
+          mkdir sa_home
+          if [[ ${{ matrix.sqlalchemy }} = "release" ]]
+          then
+            extract="tar -C sa_home -xzf"
+          else
+            extract="unzip -d sa_home -q"
+          fi
+          $extract $( ls SQLAlchemy* )
```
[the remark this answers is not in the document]
CaselIT: I use the sdist to also run the tests, so in any case it needs to be extracted

.github/workflows/3rd-party-tests.yml
CaselIT: I honestly did not look if sqlalchemy had one.

Maybe we can just use tox here? I'll take a look at what target there are

dvarrazzo: It doesn't seem it has one, no.

TBH Tox has caused me so many problems that doing without it would be preferable for me.

.github/workflows/3rd-party-tests.yml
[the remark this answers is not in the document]
dvarrazzo: > I use the sdist to also run the tests, so in any case it needs to be extracted

ah sure, that makes sense.

.github/workflows/3rd-party-tests.yml
CaselIT: looking at the defined env there is one that could maybe be used, that's the github one, but running pytest should be easier

.github/workflows/3rd-party-tests.yml
CaselIT: I may have found a reasonable solution to have both a label and the variable defined in the matrix section

CaselIT:
sorry forgot to mention yesterday. I've updated

dvarrazzo:
Merged, thank you very much!

CaselIT:
I did not know that github automatically named the matrix jobs for you. It even does a better job that what I had before
image

PR #499, opened 2023-01-28
refactor(pool): make constructor configuration parameters type-safe

### LINKED ISSUE #493: `ConnectionPool`'s kwargs parameter is not ideal for typing and auto-completion
The forwarding kwargs in `ConnectionPool`'s constructor means that all of the forwarded arguments are not type checked and can't be auto-completed/discovered in code editors.

This is the kwargs:

https://github.com/psycopg/psycopg/blob/f058590f2d2ffa122cff2aa81c2646dcfba4b4ff/psycopg_pool/psycopg_pool/pool.py#L30-L40

and these are the "missing" parameters:

https://github.com/psycopg/psycopg/blob/f058590f2d2ffa122cff2aa81c2646dcfba4b4ff/psycopg_pool/psycopg_pool/base.py#L40-L56

This also holds for the other pool types.

Possible solutions I can think of are:

- Eliminate the kwargs; duplicate all of the parameters.
- Wait for [PEP-692](https://peps.python.org/pep-0692/). This will fix the type safety, but probably not the code editor part.
- Replace the params with a configuration type. This is a breaking change, unless moved to a factory method.

bluetech:
Previously, arguments to ConnectionPool and friends would be forwarded using `**kwargs` to `BasePool`. This is however not type-safe, and prevents code editors from auto-completing the parameters.

Drop the `**kwargs`, duplicate the parameters instead.

Fixes #493.


dvarrazzo:
This is... awful :D but it is what it is :) It will be extra chore to maintain, but I accept this is better for type completion, type safety, and it generates the complete signature in the documentation.

Will merge and release in psycopg-pool 3.1.6.

PR #504, opened 2023-02-01
Fix TypeInfo.fetch() returning errors when passed a `sql_ascii` connection

### LINKED ISSUE #503: TypeInfo has different types depending on postgres local settings
Under different locales, for example `C`, the return types of the query in `psycopg._typeinfo.TypeInfo._fetch(_async)`are bytes instead of strings which means that the [example](https://www.psycopg.org/psycopg3/docs/basic/pgtypes.html#hstore-adaptation) from the psycopg docs for hstore does not work. `psycopg.types.array.register_array` expects the attributes of `TypeInfo` to be strings and attempts to encode them. Normally I use `en_US.UTF-8` but I neglected a parameter to `initdb` so I discovered this. I've since changed my configuration but the problem here remains.

I'd be happy to submit a PR for this if it is an actual bug and not intended behavior, but it might take me a bit - this is a busy week for me.

dvarrazzo:
Close #503.

@rafmagns-skepa-dreag could you please check if it solves the problem you have found?

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,16 @@
 ``psycopg`` release notes
 =========================
 
+Future releases
+---------------
+
+Psycopg 3.1.9 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix `TypeInfo.fetch()` using a connection in `!sql_ascii` encoding
+  (:ticket:`#503`).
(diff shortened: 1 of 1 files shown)


rafmagns-skepa-dreag:
yes. I just tested this code and it resolved the issue

PR #505, opened 2023-02-03
Fix broken link to pip docs

Harry-Lees:
The previous [link](https://pip.pypa.io/en/stable/reference/pip_install/#install-editable) to pip.pypa was broken. This PR changes to the newer page on editable installs.


dvarrazzo:
Thank you very much! Too bad that it looks like the new page doesn't contain info about unbreak editable mode, but hey. Just gave the new page a cursory look though.

PR #523, opened 2023-03-08
doc: fix pg2pyenc() docstring

dlax:


dvarrazzo:
Oops, thank you very much :grimacing: 

Will merge in the next few days: I have to fix the build pipeline and currently pressed by other jobs.

dlax:
By the way, I was wondering if functions from this `_encodings` module could be exposed as part of the public API? Namely, I'd need to use the encoding of a database (not necessarily the one I'm connected to, but, for instance, as reported by `pg_database.encoding`) and this function would be quite useful.

dvarrazzo:
I understand. Would you like to propose an interface?

dvarrazzo:
Cherry-picked, thank you

PR #531, opened 2023-03-23
tests: skip test_sql::test_invalid_name if not a super-user

dlax:
Creating a base type, as done in that test, requires being a superuser.

It's annoying to me while running the test suite as the user I'm using on some machines is not a super-user.


dvarrazzo:
Merged, thank you!

If we need this check more, we can either add a marker, or use `svcconn` for this test, and skip all tests using `svcconn` if it's not `is_superuser`.

PR #533, opened 2023-03-28
chore: fix declaration of PQsetTraceFlags in pyi

dlax:
Running 'python -m psycopg.pq._pq_ctypes' currently produces a diff because _PQsetTraceFlags is not present in .pyi but PQsetTraceFlags is.


dvarrazzo:
Thank you! :slightly_smiling_face:

PR #540, opened 2023-04-04
row factories and connection pools 

dlax:
Connection pools currently produces `Connection[Any]` type and there does not seem to be a way to tell the type system about the row type of connections in the pool. I've been thinking a bit about this and drafted a proposal for discussion. ~Best start is to read the documentation change, does this sound good from a user perspective?~

### DIFF
--- a/psycopg_pool/psycopg_pool/null_pool.py
+++ b/psycopg_pool/psycopg_pool/null_pool.py
@@ -6,10 +6,11 @@ Psycopg null connection pools
 
 import logging
 import threading
-from typing import Any, Callable, Dict, Optional, Tuple, Type
+from typing import Any, Callable, Dict, Optional, Tuple, Type, overload
 
 from psycopg import Connection
 from psycopg.pq import TransactionStatus
+from psycopg.rows import Row, RowFactory
 
 from .pool import ConnectionPool, AddConnection, ConnectFailedCB
 from .errors import PoolTimeout, TooManyRequests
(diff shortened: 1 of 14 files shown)


docs/advanced/pool.rst
```diff
@@ -284,6 +284,45 @@ working connections, as soon as they are available.
...
+psycopg.rows.class_row(MyRow)}`` to ``ConnectionPool``, but the type parameter
+for ``ConnectionPool`` remains needed.
+
+Alternatively, the `!connection_class` constructor may be used:
+::
+
+    class MyConnection(psycopg.Connection[MyRow]):
+        def __init__(self, *args):
+            super().__init__(*args)
+            self.row_factory = psycopg.rows.class_row(MyRow)
```
dlax, line 317: Somehow unrelated: writing this made me realize that this not very convenient; I'd have appreciated if the following worked:
```python
class MyConnection(psycopg.Connection[MyRow]):
    row_factory = psycopg.rows.class_row(MyRow)
```
and, at a first glance, I think it can be made working with a bit of work.
Thoughts?

dvarrazzo: If I understand correctly, this could be a mistake, of the type I've already made, for instance for `Dumper.oid`, which is a class attribute, but can be overridden by instances. It is valid python, but Mypy doesn't like it very much. It cannot be annotated as a ClassVar and mypyc is confused as of whether it's a class or instance attribute. So I would like to refactor it away in psycopg 3.2 (if I can do it without losing backward compatibility).

docs/advanced/pool.rst
```diff
@@ -284,6 +284,45 @@ working connections, as soon as they are available.
...
+        row = conn.execute(...).fetchone()
+        # row has type MyRow (or None)
+        # conn has type Connection[MyRow]
+
+Here it is needed to set ``ConnectionPool``'s type parameter to ``MyRow`` as
+the `!configure` callback is not enough for type inference.
+
+The same can be achieved by passing ``kwargs={"row_factory":
+psycopg.rows.class_row(MyRow)}`` to ``ConnectionPool``, but the type parameter
+for ``ConnectionPool`` remains needed.
```
dvarrazzo, line 309: How about we also add a `row_factory` parameter to the pool constructor, and we assign the type parameter from there?

dvarrazzo: We could make an error to specify row_factory both in the parameters and in the `kwargs` parameter.

dlax: Indeed, this would be simpler. Thank you.

psycopg_pool/psycopg_pool/null_pool.py
```diff
@@ -40,7 +41,55 @@ def _maybe_grow_pool(self) -> None:
...
+        max_idle: float = ...,
+        reconnect_timeout: float = ...,
+        reconnect_failed: Optional[ConnectFailedCB] = ...,
+        num_workers: int = ...,
+    ):
+        ...
+
+    @overload
+    def __init__(
+        self: "NullConnectionPool[Any]",
```
dvarrazzo, line 71: This actually returns a `NullConnectionPool[TupleRow]`. In my merged branch I have used this definition.

dvarrazzo:
Superseded by #559. I have cherry-picked the changeset to fix the __enter__ return value, which can be released in 3.1.x as it's a backward-compatible change.

PR #559, opened 2023-05-09
make pool generic on connection type

dvarrazzo:
This allows to have the connection returned by connection() type-annotated.

Draft to play around #558, but possibly to merge with #540 in order to specify just a connection row without subclassing.

### DIFF
--- a/psycopg_pool/psycopg_pool/null_pool_async.py
+++ b/psycopg_pool/psycopg_pool/null_pool_async.py
@@ -6,28 +6,74 @@ psycopg asynchronous null connection pool
 
 import asyncio
 import logging
-from typing import Any, Awaitable, Callable, Dict, Optional, Type
+from typing import Any, Awaitable, Callable, cast, Dict, Optional, overload, Type
 
 from psycopg import AsyncConnection
 from psycopg.pq import TransactionStatus
+from psycopg.rows import TupleRow
 
 from .errors import PoolTimeout, TooManyRequests
 from ._compat import ConnectionTimeout
 from .null_pool import _BaseNullConnectionPool
-from .pool_async import AsyncConnectionPool, AddConnection, AsyncConnectFailedCB
+from .pool_async import AsyncConnectionPool, ACT, AddConnection, AsyncConnectFailedCB
 
 logger = logging.getLogger("psycopg.pool")
 
 
-class AsyncNullConnectionPool(_BaseNullConnectionPool, AsyncConnectionPool):
+class AsyncNullConnectionPool(_BaseNullConnectionPool, AsyncConnectionPool[ACT]):
+    @overload
+    def __init__(
+        self: "AsyncNullConnectionPool[AsyncConnection[TupleRow]]",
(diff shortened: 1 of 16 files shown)


psycopg_pool/psycopg_pool/null_pool_async.py
```diff
@@ -6,28 +6,74 @@
...
 
 
-class AsyncNullConnectionPool(_BaseNullConnectionPool, AsyncConnectionPool):
+class AsyncNullConnectionPool(_BaseNullConnectionPool, AsyncConnectionPool[ACT]):
+    @overload
+    def __init__(
+        self: "AsyncNullConnectionPool[AsyncConnection[TupleRow]]",
+        conninfo: str = "",
+        *,
+        open: bool = True,
```
dvarrazzo, line 29: wrong default: should be `...`.

dvarrazzo:
@dlax, @guacs,

I think this is not bad: I have merged this attempt to make the pool generic on the connection together with your row_factory branch in #540, and it seems to work well in most cases.

The pools are now:

- `Pool()` -> `Pool[Connection[TupleRow], TupleRow]`
- `Pool(row_factory=DictRow)` -> `Pool[Connection[DictRow], DictRow]`
- `Pool(connection_class=MyConn)` -> `Pool[MyConn[TupleRow], TupleRow]`
- `Pool(connection_class=MyConn, row_factory=DictRow)` -> `Pool[MyConn[DictRow], DictRow]`

Type annotation for connections and results follows as expected.

The only sub-optimal behaviour if is someone derives a non-parametric connection class which never returns tuples, for instance:

```python
    class MyConnDict(psycopg.Connection[DictRow]):
        def __init__(self, *args: Any, **kwargs: Any):
            kwargs["row_factory"] = dict_row
            super().__init__(*args, **kwargs)
```

in this case, `Pool(connection_class=MyConnDict)` would be annotated as `Pool[MyConnDict, TupleRow]`, with the annotation indicating that it would return tuples, but actually returning dictionaries. Passing `row_factory` to the pool would do nothing. If someone had this connection, they could still use a redundant declaration: `Pool(connection_class=MyConnDict, row_factory=dict_row)` and align the annotation with the reality.

I think this is the only case in which there would be needed two levels of generic and can't be done in Python, but it seems that all the other cases work well with Python types model.

What do you think?

dlax:
I think this is very reasonable. The edge case you mention seems acceptable to me.

I'm not sure if you want me to change something in #540 or not, let me know in any case.

dvarrazzo:
#540 looks good as it is. There was only that comment of yesterday, about the default returned row to be `TupleRow` rather than `Any`, but we can leave the MR as it is. I will merge it here, and them merge this to master.

Need to add docs for the feature to finish this.

dvarrazzo:
@dlax heads up: I'll try to review the state of parametrising the pools again this weekend. In the past days I was finding some strange interaction between the row type and the connection class parameters but now I'm struggling to reproduce the behaviour I wasn't liking.

Issue #565, opened 2023-05-15
On connection failure, PGconn is leaked until finalization


jchampio:
Hello,

The `connect` generator appears to have no cleanup step (i.e. a call to `PQfinish`) if the generator's `PQconnectPoll` calls fail. This results in resources leaking (or hanging open, in the case of the underlying socket) until the finalizer is called at process exit.

I'm using `psycopg[c]`, if that matters, but from inspection it looks like the implementation in `generators.py` might have the same problem.

Here's a quick and dirty program to illustrate. The "server" causes a connection failure by writing unexpected data to the socket, then waits for the client to close the connection. With the old `psycopg2.connect`, the script correctly exits after raising `OperationalError`. With `psycopg.connect`, the script hangs.

```py
#! /usr/bin/env python3

import socket
import threading

from psycopg import connect
#from psycopg2 import connect  # uncomment to fix the program


class Server(threading.Thread):
    """
    Accepts exactly one Postgres connection, reads a startup packet, writes
    garbage to the socket, and then tries to read more data.
    """
    def __init__(self):
        super().__init__()

        # Listen on localhost.
        self._listener = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self._listener.bind(("127.0.0.1", 0))
        self._listener.listen(1)

        _, self.port = self._listener.getsockname()

    def run(self):
        sock, _ = self._listener.accept()
        with sock:
            # Read the startup packet.
            data = sock.recv(4096)
            print("startup: ", data)

            # Send some garbage to the client to cause a connection failure.
            sock.send(b"garbage data")

            # Client should disconnect at this point; the recv() call should
            # immediately return with no data. With psycopg 3.1, this call
            # hangs.
            data = sock.recv(4096)
            print("read: ", data)


server = Server()
server.start()

connect(host="127.0.0.1", port=server.port)  # raises OperationalError

# Python will join the server thread at this point. With psycopg2, the program
# exits; with psycopg 3.1, the program hangs.
```

dvarrazzo:
Hello,

the PQfinish is called [on PGconn deallocation](https://github.com/psycopg/psycopg/blob/8ec349089b60ff1ea52700b81431183e0f3c9867/psycopg_c/psycopg_c/pq/pgconn.pyx#L46-L50). As such, it is invoked usually as soon as the `conn` object gets out of scope.

[...]

PR #566, opened 2023-05-16
tests: check that OperationalError raised by connect() holds a pgconn

dlax:
Sort of a non-regression test for the part of commit 9220293dc023b757f2a57702c14b1460ff8f25b0 concerning generators. It used roughly the same logic as tests/pq/test_pgconn.py::test_used_password() to determine if the test connection needs a password and gets skipped otherwise.

* * *

It's related to #565, to illustrate that we need to keep the `pgconn` attribute bound to `OperationalError` raised by `generators.connect()`.

### DIFF
--- a/tests/test_generators.py
+++ b/tests/test_generators.py
@@ -7,6 +7,35 @@ import pytest
 import psycopg
 from psycopg import waiting
 from psycopg import pq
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Closing this as merged in #570. Thank you very much!

PR #609, opened 2023-07-21
fix: don't clobber a Python exception on COPY FROM with QueryCanceled

### LINKED ISSUE #593: Getting the underlying exception in `copy()`
### Problem

When an exception is raised from the user python code during `copy()`, the traceback looks like this:

```
    <snipped>
requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

The above exception was the direct cause of the following exception:

  File "venv/lib/python3.11/site-packages/psycopg/cursor.py", line 900, in copy
    yield copy
  File "myfile.py", line 86, in myfunc
    raise MyException() from e
MyException

  File "myotherfile.py", line 1251, in myotherfunc
    with cursor.copy(sql.SQL('''
  File "/usr/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
  File "venv/lib/python3.11/site-packages/psycopg/cursor.py", line 902, in copy
    raise ex.with_traceback(None)
psycopg.errors.QueryCanceled: COPY from stdin failed: error from Python: MyException - 
```

If I want to inspect the underlying exception, I currently need to check `__context__`. But `__context__` should be reserved for unintended bugs in error handling code. This makes it hard to distinguish real underlying errors from exceptions unintentionally raised in psycopg's exception handling code.

### Suggested solution

Change to explicit [chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining), that is `raise QueryCanceled() from underlying_exc` instead of `raise` without `from` in the exception handling context.

This will change the traceback message to say "The above exception was the direct cause of the following exception". Also will change `__context__` to None and make `__cause__` have the exception instead.

There are linters that can help with detecting this more holistically, for example flake8-bugbear [B904](https://github.com/PyCQA/flake8-bugbear#opinionated-warnings), though I didn't check if it will detect this case.

dvarrazzo:
We trigger the server to raise the QueryCanceled; however, the original exception has more information (the traceback). We can consider the server exception just a notification that cancellation worked as expected.

This is a mild change in behaviour, as the fixed tests state. However, raising QueryCanceled is not explicitly documented and not part of a strict interface, so we can probably change the exception raised without needing to wait for psycopg 4.

Close #593

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -10,8 +10,8 @@
 Future releases
 ---------------
 
-Psycopg 3.1.10
-^^^^^^^^^^^^^^
+Psycopg 3.1.10 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Fix prepared statement cache validation when exiting pipeline mode (or
   `~Cursor.executemany()`) in case an error occurred within the pipeline
@@ -20,6 +20,9 @@ Psycopg 3.1.10
   `OperationalError` in case of connection failure. `Error.pgconn` is now a
   shallow copy of the real libpq connection, and the latter is closed before
   the exception propagates (:ticket:`#565`).
+- Don't clobber a Python exception raised during COPY FROM with the resulting
+  `!QueryCanceled` raised as a consequence (:ticket:`#593`).
+
 
 Current release
 ---------------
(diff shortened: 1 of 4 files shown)


dvarrazzo:
I have refactored the MR to move the query canceled management closer to where it is raised and to fix async copy too.

dlax (approved):

Issue #604, opened 2023-07-13
Weird pipeline issue where results are not available


JelteF:
To reproduce:

```python
import sys
import random
import psycopg
from psycopg import sql

max_pipeline_length = 10
max_prepared_stmt = 100
n_iterations = 100
size_of_param = 4096
for pipeline_length in range(max_pipeline_length):
    with psycopg.connect("port=5432 client_encoding=UTF8", autocommit=True) as conn:
        with conn.pipeline() as p, conn.cursor() as cur:
            for i in range(pipeline_length):
                stmt_id = random.randint(1, max_prepared_stmt)
                cur.execute(
                    sql.SQL("SELECT %s, %s as {}").format(
                        sql.Identifier(f"s{stmt_id}")
                    ),
                    params=(i, str(i).zfill(size_of_param)),
                    prepare=True,
                )
            # Uncommenting this line causes the issue to go away
            # p.sync()
            for i in range(pipeline_length):
                result = cur.fetchall()
                if result != [(i, str(i).zfill(size_of_param))]:
                    print(pipeline_length, i, result)
                    sys.exit(1)

                cur.nextset()
```

A few things that this prints on my machine:
```
3 1 []
```
```
[...]

PR #613, opened 2023-08-01
Pipeline nextset warning

dvarrazzo:
As described in #604, we want to discourage people to use multiple `execute()` in a pipeline to accumulate results in the same cursor.

This MR raises a warning if  `nextset()` is used after `execute()` in pipeline mode. This will become an error in 3.2 (and in a followup MR).

### DIFF
--- a/psycopg/psycopg/cursor.py
+++ b/psycopg/psycopg/cursor.py
@@ -9,6 +9,7 @@ from types import TracebackType
 from typing import Any, Generic, Iterable, Iterator, List
 from typing import Optional, NoReturn, Sequence, Tuple, Type, TypeVar
 from typing import overload, TYPE_CHECKING
+from warnings import warn
 from contextlib import contextmanager
 
 from . import pq
@@ -156,6 +157,18 @@ class BaseCursor(Generic[ConnectionType, Row]):
         Return `!True` if a new result is available, which will be the one
         methods `!fetch*()` will operate on.
         """
+        # Raise a warning if people is calling nextset() in pipeline mode
+        # after a sequence of execute() in pipeline mode. Pipeline accumulating
+        # execute() results in the cursor is an unintended difference w.r.t.
+        # non-pipeline mode.
+        if self._execmany_returning is None and self._conn._pipeline:
+            warn(
+                "using nextset() in pipeline mode for several execute() is"
+                " deprecated and will be dropped in 3.2; please use different"
+                " cursors to receive more than one result",
+                DeprecationWarning,
+            )
+
         if self._iresult < len(self._results) - 1:
             self._select_current_result(self._iresult + 1)
             return True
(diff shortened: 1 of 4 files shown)


dvarrazzo:
@dlax could you please review this? The idea would be to merge this branch for 3.1.10 in the next few days.

dlax (approved):
LGTM.

dvarrazzo:
Thank you @dlax! Will merge soon.

psycopg/psycopg/cursor.py
```diff
@@ -156,6 +157,18 @@ def nextset(self) -> Optional[bool]:
         Return `!True` if a new result is available, which will be the one
         methods `!fetch*()` will operate on.
         """
+        # Raise a warning if people is calling nextset() in pipeline mode
+        # after a sequence of execute() in pipeline mode. Pipeline accumulating
+        # execute() results in the cursor is an unintended difference w.r.t.
+        # non-pipeline mode.
+        if self._execmany_returning is None and self._conn._pipeline:
```
dlax, line 164: I think checking `self._conn._pipeline` is not enough because one might have exited the pipeline before calling `.nextset()`, e.g.:
~~~python
    with conn.cursor() as cur:                                                                                                          
        with conn.pipeline() as p:                                                                                                      
            cur.execute("select 1")                                                                                                     
            cur.execute("select 2")                                                                                                     
            assert len(p.result_queue) == 2                                                                                             
        (r,) = cur.fetchone()                                                                                                           
        assert r == 2  # expected after clobbering                                                                                                   
        assert cur.nextset() is None  # no warning since we're out of the pipeline         
~~~

dvarrazzo: ...I'd say if they do that they deserve an unexpected behaviour :grimacing: 

I know that a cursor can be created independently before or after the pipeline. However, it seems reasonable to require (maybe to enforce?) that if an `execute()` is performed before/inside/after a pipeline, the matching `fetch*()` should happen in the same area. I expect the unexpected, otherwise.

Is this something you would consider enforcing? An effective and efficient way to do so would be to `_reset()` the cursor upon entering and exiting the pipeline.

dlax: I agree: calling `.fetch*()` within the same scope as respective `.execute()` is probably the most sensible things to do. So the implementation here is probably enough.

> Is this something you would consider enforcing?

Hm, no. `.fetch*()` invocations after exiting the pipeline work fine[^1] as results of pipeline's `execute()`s are retrieved at pipeline exit and bound to their cursor (so that those `.fetch*()` do not actully do I/O but just return).  So I'm not sure there is any value in changing/breaking this: that would rather be a burden on our side.

[^1]: We have tests for this: https://github.com/psycopg/psycopg/blob/915593ad6f3577f67ba73d0105fec0e88940ae8e/tests/test_pipeline.py#L131

PR #614, opened 2023-08-01
Pipeline nextset error and clobbering behaviour

dvarrazzo:
As described in #604, we want to replace the accumulation behaviour of cursors in pipeline mode with the normal clobbering behaviour.

This MR, at this moment, only replaces the `nextset()` warning introduced in #613 with an error.

The clobbering behaviour is not implemented yet. @dlax, is it something you could contribute?

### DIFF
--- a/docs/advanced/pipeline.rst
+++ b/docs/advanced/pipeline.rst
@@ -185,26 +185,8 @@ several operations, using `Connection.execute()`, `Cursor.execute()` and
 
 Unlike in normal mode, Psycopg will not wait for the server to receive the
 result of each query; the client will receive results in batches when the
-server flushes it output buffer.
-
-When a flush (or a sync) is performed, all pending results are sent back to
-the cursors which executed them. If a cursor had run more than one query, it
-will receive more than one result; results after the first will be available,
-in their execution order, using `~Cursor.nextset()`:
-
-.. code:: python
-
-    >>> with conn.pipeline():
-    ...     with conn.cursor() as cur:
-    ...        cur.execute("INSERT INTO mytable (data) VALUES (%s) RETURNING *", ["hello"])
-    ...        cur.execute("INSERT INTO mytable (data) VALUES (%s) RETURNING *", ["world"])
-    ...        while True:
-    ...            print(cur.fetchall())
-    ...            if not cur.nextset():
-    ...                break
-
-    [(1, 'hello')]
-    [(2, 'world')]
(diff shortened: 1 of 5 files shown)


dlax:
> The clobbering behaviour is not implemented yet. @dlax, is it something you could contribute?

Yes, that's something I'm planning to work on[^1].

About this MR, do we intent to keep the error when calling `.nextset()` *after* the "clobbering behaviour" gets implemented?

[^1]: hopefully this week, but not sure; late August otherwise.

dvarrazzo:
> Yes, that's something I'm planning to work on

Amazing, thank you! No hurry, it will be a lot of work before 3.2 is wrapped up. Enjoy your summer!


> About this MR, do we intent to keep the error when calling `.nextset()` _after_ the "clobbering behaviour" gets implemented?

I think it is a good idea because, if someone misses the warning in 3.1.10 will not be able to avoid missing the error in 3.2. If we drop the error people might just run 3 query and fail to spot the fact that they have lost the result of the first two results.

However, I have my share of doubts about keeping the error. If there are good reason to drop it, stronger than my reason to keep it, I'm open to change the behaviour as currently tested.

dlax:
One reason to drop the error is that, with the "clobbering behaviour" in place, #619, now `.nextset()` would always return `None`. This is a different behaviour from before:
~~~python
    with conn.pipeline():
        cur.execute("select 1")
        cur.execute("select 2")

        assert cur.fetchall() == [(1,)]
        assert cur.nextset()
        assert cur.fetchall() == [(2,)]
        assert cur.nextset() is None
~~~
now (with the clobbering, and dropping the error in `.nextset()`):
~~~python
    with conn.pipeline():
        cur.execute("select 1")
        cur.execute("select 2")

        assert cur.fetchall() == [(2,)]
        assert cur.nextset() is None
~~~
*if* we expect users to check the return value of `.nextset()` (reasonable expectation IMO), it seems to me that they would notice the behaviour change.

dvarrazzo:
Yes, I thought too that what we are really stopping the user from doing is to call a `lambda: None` and that error is also a difference in behaviour of pipeline vs. non-pipeline...

Sounds good: I will drop the error, we will only leave the warning in 3.1.x from 3.1.10 onwards.

Thank you very much, @dlax!

dvarrazzo:
Error dropped, rebased on master. Once tests pass I'll merge. Thank you very much for the collaboration!

PR #649, opened 2023-09-25
Fix memory leak with repeated `register_*()` call

### LINKED ISSUE #647: Memory leak with psycopg-c (but not pure python)
I found a memory leak when using psycopg-c (not present in pure python version) when using with Django and Celery. I reported the issue to Django [here](https://code.djangoproject.com/ticket/34870) but it is not clear exactly which project is at fault.

Here is a [reproduction repo](https://gitlab.com/bufke/psycopg-memory-leak).

Django/Celery seems to call `hstore.register_hstore(ti, connection.connection)` on every celery task instead of just once. That might be a bug in one of those projects - but I would think psycopg should not memory leak when doing such. And the problem goes away with psycopg (sans c).

I will continue to try to create a more minimal reproduction set up. Please let me know if you have ideas to things to try out to help debug/fix the issue.

dvarrazzo:
The leak is caused by subclasses of C extension types not being GC'd. However this MR introduces caching of all the dynamically generated types.

The cache keys are usually oid, type names, etc, so they are not expected to change repeatedly and, unless connecting to an unbound number of servers, each one with different type oids, the caches themselves are not expected to cause a leak.

Close #647

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,13 @@
 ``psycopg`` release notes
 =========================
 
+Psycopg 3.1.12 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix memory leak when `~register_*()` functions are called repeatedly
+  (:ticket:`#647`).
+
+
 Current release
 ---------------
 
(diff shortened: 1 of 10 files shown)


dvarrazzo:
merging via command line.

Issue #584, opened 2023-06-06
psycopg_pool does not check connection lifetime before giving a connection to a calling code


Tiendil:
How to reproduce:

- Configure `idle_session_timeout` in PostgreSQL, `max_lifetime` & `min_size` in psycopg_pool, `max_lifetime < idle_session_timeout`, `min_size=2`.
- Wait for an `idle_session_timeout` seconds doing nothing.
- Try to get a connection from the pool and execute an SQL query.

Result:

- `psycopg.OperationalError: consuming input failed: server closed the connection unexpectedly`

Expected result:

- Pool detects that existing connections lifetime is expired and requests a new connection(s) from the DB.
- => SQL query executed succesfully.

Temporary solution:

- Periodically execute the `check` method.

Potential cause of the problem:

- Database closes a connection, but psycopg_pool does not close it => next SQL query fails because the connection is closed by the DB.
- psycopg_pool checks connection `_expire_at` attribute only [when connection is returned to pool](https://github.com/psycopg/psycopg/blob/f09eb30f6f7bcf702fb4e42a1a9349f9a9d34b01/psycopg_pool/psycopg_pool/pool_async.py#L533), but not before give it to a code.
- No queries —> no connections returned to a pool —> no connections checked for a lifetime.

Such behaviour complicates the development and life of not-so-popular projects in production. Also, it is not obvious for new developers.

Is it by design, or is it a bug?

Tiendil:
Thanks for the fast answer!

And sorry for the wrong bug report.

dvarrazzo:
No worreis, it's not the first time someone expects that `idle_session_timeout` is kept in consideration by the pool. I think it does need a suggestion in the docs about how to disable it.

dvarrazzo:
Added docs here: https://www.psycopg.org/psycopg3/docs/advanced/pool.html#pool-and-idle-session-timeout-setting

PR #658, opened 2023-10-06
Add check callback to ensure getconn() returns a working connection

### LINKED ISSUE #247: Add disconnection detection to pool
There is a [recipe to detect disconnections](https://www.psycopg.org/psycopg3/docs/advanced/async.html#detecting-disconnections) in the docs. It uses selectors and doesn't use polling, so it's pretty efficient.

There is some [advice about using it with the pool](https://www.psycopg.org/psycopg3/docs/advanced/pool.html#connection-quality) too. However it's up to the client to wrap up something to make use of it, and it wastes a connection entirely.

However we can make the extra step and add a disconnection detection feature? The pool is in a privileged position because it may sneakily use a connection internally, as long as it's well coordinated with external users. For instance, the recipe proposes:

```python
sel.register(conn, selectors.EVENT_READ)
while True:
    if not sel.select(timeout=60.0):
        continue  # No FD activity detected in one minute

    # Activity detected. Is the connection still ok?
    try:
        conn.execute("SELECT 1")
    except psycopg.OperationalError:
        await pool.check()
```

if we did this from within the pool, we could use a connection from within `pool._pool`, without taking it out of the available set. If activity is detected (so we are about to enter the `try` block), we may check `connection._pool`, which is populated only when the connection has been assigned to the client and taken out of the pool: in this case we would avoid to execute the `try` and start using a different connection instead.

Problems ahead:

- There are race conditions to take care of. We might want to check `conn._pool` holding the pool lock
- we should try to use the probing connection as last in the pool, to avoid to start a new selector every `min_size` requests (the pool is a queue: connection are taken with a popleft and returned with an append, because using it as a stack makes hard to implement `max_lifetime`). `pool._pool` is a deque, so inserting in penultimate position rather than append should be possible and efficient.
- if the probing connection is left out of the carousel then `max_lifetime` should be checked in a different way. So maybe it's worth to use a different probing connection every time, e.g. taking the most recent one and move it to the back to the deque, with a rotation...

### LINKED ISSUE #656: Add on-connect callback to check connections as obtained from the pool
A lot of people have the expectation that the pool will only serve valid connection, which I'm starting to see less and less the case in several scenarios, such as using `idle_session_timeout`, with database vendors recycling connections, using haproxy as a front in HA solutions...

In #247 we were thinking about using a probe connection to check for disconnection, but that would only work for network disconnections dropping all the connections in the pool, not for connections lost singularly, e.g. for `idle_session_timeout`.

If the client is willing to pay a little price, we can add a callback to call on getconn: if the callback fails (e.g. to execute a `SELECT 1`) then we discard the connection and ask for a new one, until a working one is obtained, or timeout hits.

The whole loop should honor the getconn timeout: if the timeout is 30s then 30s is a global deadline to obtain a working connection, not per attempt.

dvarrazzo:
The MR introduces a `check` callback which is invoked on `getconn()` and verifies that the connection works correctly.

It is the answer to all the people expecting the pool to work correctly with servers setting `idle_session_timeout` (#584) or whatever other middleware may close an idle connection. This solution is democratic because:

- people who don't want the feature don't pay any price;
- people who want the feature can enable it easily (setting `check=ConnectionPool.check_connection` on init, at its simplest);
- the price is paid by the client requesting it, not by the server or by other clients.

There is no polling for connections working, which might be more impactful on the server, and no complex and race-y worker as proposed in #247. Furthermore, #247 would have only detected a network loss (the loss of all the connections to the server), not the loss of a single connection.

Close #656
Close #247

### DIFF
--- a/docs/advanced/pool.rst
+++ b/docs/advanced/pool.rst
@@ -108,7 +108,11 @@ When a client asks for a connection (typically entering a
 
 - if no connection is available, the client is put in a queue, and will be
   served a connection once one becomes available (because returned by another
-  client or because a new one is created).
+  client or because a new one is created);
+
+- if a `!check` callback was provided, it is called on the connection before
+  passing the connection to the client. If the check fails, a new connection
+  will be obtained.
 
 When a client has finished to use the connection (typically at the end of the
 context stared by `~ConnectionPool.connection()`):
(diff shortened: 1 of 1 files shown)


dvarrazzo:
The MR requires docs, it is otherwise complete feature-wise.

PR #659, opened 2023-10-07
Warn if the user relies on the pool being open implicitly.

dvarrazzo:
The pool consructor's `open` parameter and `open()` method were introduced in psycopg_pool 3.1 to allow delaying pool work after import time. This is especially useful for async code because the pool doesn't work correctly if it's open before the event loop is started.

The default for the parameter is `True` for compatibility with psycopg_pool 3.0 In the future (psycopg_pool 4.0), the default for the open parameter will become `False`.

With this MR, to be released with psycopg_pool 3.2, a warning is raised if the pool is created expecting an implicit open=True. This means that the warning is:

- not raised if `ConnectionPool(dsn, open=True)` is specified (explicit open, the user knows that it's the right moment);
- not raised if `ConnectionPool(dsn, open=False)` is specified (the user knows when it will be the right moment to open the pool);
- not raised if the pool context is used:
```python
with ConnectionPool(dsn) as p:
    with p.connection():
        pass
```
- but only raised if the pool is expected to having been open on init:
```python
p = ConnectionPool(dsn)
with p.connection():
    pass
```

### DIFF
--- a/docs/api/pool.rst
+++ b/docs/api/pool.rst
@@ -159,10 +159,18 @@ The `!ConnectionPool` class
         The class is generic and `!connection_class` provides types type
         variable. See :ref:`pool-generic`.
 
-   .. note:: In a future version, the default value for the `!open` parameter
-        might be changed to `!False`. If you rely on this behaviour (e.g. if
-        you don't use the pool as a context manager) you might want to specify
-        this parameter explicitly.
+   .. warning::
+
+        At the moment, the default value for the `!open` parameter is `!True`;
+        In a future version, the default will be changed to `!False`.
+
+        If you expect the pool to be open on creation even if you don't use
+        the pool as context manager, you should specify the parameter
+        `!open=True` explicitly.
+
+        Starting from psycopg_pool 3.2, a warning is raised if the pool is
+        used with the expectation of being implicitly opened in the
+        constructor and `!open` is not specified.
 
    .. automethod:: connection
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
@dlax I remember that, with trio, just `AsyncConnectionPool(open=True)` was a problem too, and, even in asyncio, it forces us to jump through some hoops in order to start async code from the sync constructor.

Maybe, for the async pools, we want to warn even if `open=True` is open explicitly, and ask the use to use an explicit `await pool.open()` or `async with AsyncPool(): ...`?

dvarrazzo:
@dlax Perfect, thank you for the feedback.

So my proposal is what shown [in the tests here](https://github.com/psycopg/psycopg/pull/659/files#diff-45ac6fb648b8b38d8c7897497e72ebdcb2287b758c598ee3e0ca62001a7be42c): the sync pool may be open on init, but it must be explicit; the async pool will require an explicit open() call or to use the context manager.

There will be a warning starting from psycopg_pool 3.2; from 4.0 the default open parameter in the constructor will become False and it will be an error to use True for the async pool.

psycopg_pool/psycopg_pool/null_pool_async.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         self: AsyncNullConnectionPool[AsyncConnection[TupleRow]],
         conninfo: str = "",
         *,
-        open: bool = ...,
+        open: bool | None = ...,
```
asqui, line 32: I'm not reviewing this PR but just a couple of random drive-by typing questions:
1. Is the type annotation for `self` needed? (I didn't think it generally was, since the type checker knows what type the object is?)
2. What does this ellipsis assignment syntax mean? I've not come across this before.

dlax: Annotation for `self` is not needed in general when used in a normal method.

But here, this is an `@overload` of the `__init__()` method in which the "constructed" object, `self`, will vary depending on other arguments to the constructor: the type of `self` depends on the value of `connection_class` argument. So by annotating `self` along with `connection_class` in each overload definition, we declare this type mapping.

The ellipsis for parameter (`= ...`) in `@overload` definitions is used when the parameter has a default value in order to avoid repeating it in each definitions whereas it's already declared in the non-overloaded definition (the last one): https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

asqui: Thanks for the clarifications @dlax!
(I should have looked at more context than what was shown in the diff, to notice that this was an `@overload` definition 🤦 )

dvarrazzo: Thank you for your answer, @dlax, you beat me to it.

In effect, as `__init__` doesn't return anything, the annotation on `self` is used in place as an annotation on the return value.

PR #660, opened 2023-10-13
improvements to tools/async_to_sync.py

dlax:


tools/async_to_sync.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python
 """Convert async code in the project to sync code.
 
-Note: the version of Python used to run this script affects the output. Please
-use the `async-to-sync.sh` wrapper to use a version consistent with CI checks.
+Note: the version of Python used to run this script affects the output.
```
dvarrazzo, line 4: Oh, thank you for fixing it. I added the build-in-a-container feature as an external script, but I figured out it could have been easy replaced by Python and ended up not committing that version.

tools/async_to_sync.py
```diff
@@ -564,9 +577,9 @@ def parse_cmdline() -> Namespace:
         "--all", action="store_true", help="process all the files of the project"
     )
     parser.add_argument(
-        "--docker",
-        action="store_true",
-        help=f"run in a docker image with Python {PYVER}",
+        "--container",
```
dvarrazzo, line 580: How do you feel about having two mutually exclusive options, `--docker|--podman`?

The reason would be mostly a simpler command line invocation.

If we configure argparse with action="store_const", target="container" the rest of the code would stay largely the same. The command line stripping which would be made simpler, becoming just the likes of `args.remove(f"--{opt.container}")`. The main motivation I have in mind is just command line usage anyway.

tools/async_to_sync.py
```diff
@@ -103,12 +102,8 @@ def check(outputs: list[str]) -> int:
         return 1
 
     # Check that all the files to convert are included in the --all list
-    cmdline = ["git", "ls-files", "**.py"]
-    git_pys = sp.check_output(cmdline, cwd=str(PROJECT_DIR)).decode().split()
-
-    cmdline = ["grep", "-l", f"auto-generated by '{SCRIPT_NAME}'"]
-    cmdline += git_pys
-    maybe_conv = sp.check_output(cmdline, cwd=str(PROJECT_DIR)).decode().split()
+    cmdline = ["git", "grep", "-l", f"auto-generated by '{SCRIPT_NAME}'", "**.py"]
```
dvarrazzo, line 105: Wow, git is a posix-compatible operating system... :smile:

tools/async_to_sync.py
dvarrazzo: We can actually follow up on this comment by printing a warning if the Python version used is different from the expected PYVER.

No need to do it in this MR: I can do it in the main async-to-sync branch after merging this MR to avoid conflicts. Unless you jump on it and do it before me.

dvarrazzo (commented):
Let me know what you prefer about the command line options, and then I'll merge this. Thank you very much for your help!

dvarrazzo:
Edit: Never mind about the version check: I've already done it my side.

tools/async_to_sync.py
dvarrazzo: Edit: Never mind about the version check: I've already done it my side.

tools/async_to_sync.py
dlax: I'm fine with this approach too; even more if it requires less code.

dlax: Done now.

dvarrazzo:
Perfect, thank you very much!

PR #663, opened 2023-10-18
chore: use AnyIO 4.0+ in tests dependencies

### LINKED ISSUE #662: Resolve problems with anyio 4
Anyio 4 has incompatibilities so, when it was released, tests were broken. Because we were still maintaining Python 3.7 and Anyio 4 doesn't support it, I pinned the version we depend on to < 4 (e847f3ca65f23ce0f95eb553edfc04431a0c18ba).

Now we dropped Python 3.7 support in the master branch (02dfd084fc3afa74e6caff42055066f13c0ba69d) so we can remove the upper boundary. However I think tests still need to be fixed, because they use a feature that got refactored.

I don't know if it's feasible to depend on any anyio version or if the only way to support 4 is to drop 3, which is a second choice I'd be ok with, if necessary.

dlax:
Fix #662

### DIFF
--- a/psycopg/setup.cfg
+++ b/psycopg/setup.cfg
@@ -64,8 +64,7 @@ binary =
 pool =
     psycopg-pool
 test =
-    # TODO: move to anyio >= 4.0 once support for Python 3.7 is dropped
-    anyio >= 3.6.2, < 4.0
+    anyio >= 4.0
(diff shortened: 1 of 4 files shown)


dvarrazzo:
Thank you very much for the chore, @dlax!

PR #673, opened 2023-10-24
add `timeout` and `stop_after` parameters to `Connection.notifies`

### LINKED ISSUE #340: AsyncConnection.notifies() gernerator blocks AsyncConnection.execute()
I have a use case where a single `AsyncConnection` needs to listen on different channels. Which to listen on becomes known in runtime, e.g. connected clients determine what is listened and unlistened on. Unfortunately, when using the `AsynConnection.notifies()` generator, further `AsyncConnection.execute()` calls will hang. 

```
async def test_notifies():
    con = await psycopg.AsyncConnection.connect(PSYCOPG_DSN, autocommit=True)
    await con.set_autocommit(True)

    async def notifier():
        i = 0
        while True: 
            await con.execute(f'notify foo, \'{i}\'')
            print(f"notified {i}")
            await asyncio.sleep(1)
            i += 1
            if i==10:
                break
        await con.close()
        
    async def receiver():
        print("receiver: LISTENING ")
        cur = con.cursor()
        await cur.execute("listen foo")
        gen = con.notifies()
        async for n in gen:
            print("receiver: notified ")
            print(n)

    workers = [notifier(), receiver()]
    await asyncio.gather(*workers)
    
asyncio.run(test_notifies())
``` 

Output of the program:
```
receiver: LISTENING 
notified 0
```
After that the  program hangs.

dvarrazzo:
This is a proposal to add `timeout` and `stop_after` parameters to `Connection.notifies()`, in order to address the problems discussed e.g. in #340, #672.

This interface allows to either wait up to a certain timeout or up to a certain number of notifications received. In case notifications are received in a single batch, their number might be higher than `stop_after`, for the last batch received.

To add:

- documentation
- Maybe some test throwing Ctrl-c and other forms of cancelling

Close #340


dvarrazzo:
There is a bug in various wait functions: in `wait_poll()` but not in `wait_c()`: if no signal is received the function loops forever:

```python
def wait_poll(gen: PQGen[RV], fileno: int, timeout: Optional[float] = None) -> RV:
...
            while not fileevs:
                fileevs = poll.poll(timeout)
```

The condition is triggered by `pytest tests/test_notify.py::test_no_notify_timeout`

Issue #674, opened 2023-10-25
Psycopg doesn't support multiple hosts in the connection string


yancyxu:
I have a PostgreSQL server setup with one primary and two secondary servers. The primary server is responsible for read-write operations, while the secondary servers are read-only. If the primary server goes down, one of the secondary servers will be promoted to primary to continue operations. Currently, I am using psycopg_pool to connect to this PostgreSQL cluster for read-write operations, so it can only connect to the primary database.

The issue I am encountering is as follows:
When connecting to the PostgreSQL cluster using psycopg, if my primary node (192.168.1.11) goes down and one of the remaining secondary nodes (192.168.1.3) is promoted to primary, my program throws an error: psycopg_pool.PoolTimeout: couldn't get a connection after 30.0 sec. The configuration target_session_attrs=read-write doesn’t seem to perform automatic failover by connecting to the next available primary node. Even restarting the program doesn’t help. The issue is resolved only when I change the connection string in my program to:
`pool=psycopg_pool.ConnectionPool(postgresql://postgres:postgres@192.168.1.3:5431,192.168.1.2:5431,192.168.1.1:5431/postgres?target_session_attrs=read-write),`
placing the new primary node at the beginning of the DSN, allowing a successful connection.

My questions are:

1. Does psycopg provide functionality for simple automatic failover?
2. By default, does psycopg only connect to the first IP address in the DSN, disregarding the target_session_attrs=read-write setting?

dvarrazzo:
Hello,

there is nothing that shouldn't work.. except #602 where a similar issue has been reported.

Yesterday I noticed an issue that might be related, so I will investigate further into why `target_session_attrs` seem to cause problems.

dvarrazzo:
Work to solve the problem started in #675

dvarrazzo:
Fixed by #675

dvarrazzo:
Psycopg 3.1.13 released with this bug fixed. Testing is welcome. Cheers!

Issue #602, opened 2023-07-12
Regression for support of multiple hosts with `target_session_attrs` comparing to psycopg2


moonrail:
Hello altogether,

first off thank you for work on v3 and psycopg2 before that.

There may be a regression of functionality in v3 in how connections are opened.

Basically it seems that psycopg2 can handle unreachable hosts correctly while v3 cannot. Maybe due to SSL negotiation:
```python
  File "/tmp/test_psycopg/venv/lib64/python3.11/site-packages/psycopg/connection.py", line 728, in connect
    raise ex.with_traceback(None)
psycopg.OperationalError: connection failed: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
could not send SSL negotiation packet: Success
```

From my testing the above behaviour is the same if the IP is not reachable or the port is down.

This is problematic when using the multi-host feature with `target_session_attrs` of PostgreSQL:
- https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNSTRING
- https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS

With this feature a DSN could be e.g.:
`postgresql://username:password@host1,host2,host3/dbname?target_session_attrs=read-write`

With psycopg2 a network downtime of host1 (port down, IP not reachable via TCP) would lead to host2 to be tried for a read-write session and after that host3, and from the beginning. This is useful when having automatic failovers in a PostgreSQL-Cluster.

With v3 and all hosts reachable the feature works correctly and the first host with desired `target_session_attrs` is selected.

### Steps to reproduce

This "ssl" error is easy to reproduce:
1. setup a postgresql host, reachable via TCP
2. try to connect to it:
```python
import psycopg

conn = psycopg.connect(f"postgresql://username:password@host1")
print(conn.info.host)
print(conn.cursor().execute('SELECT 1;'))
```
3. now stop postgresql on the host
4. retry with the same script

For testing with multi host a cluster is not required, just two postgresql hosts with the same credentials:
1. setup 2 postgresql hosts, reachable via TCP
2. create the same user on both
3. try to connect to it:
```python
import psycopg

conn = psycopg.connect(f"postgresql://username:password@host1,host2")
print(conn.info.host)
print(conn.cursor().execute('SELECT 1;'))
```
4. now stop postgresql on host1
5. retry with the same script

The connection should be made to host2, but it already fails with host1.

[...]

PR #675, opened 2023-10-26
Fix handling of multiple hosts

dvarrazzo:
When using the async connection path, the libpq doesn't try connection attempts to multiple hosts. Which is kinda natural, as it doesn't enforce the timeout, so it wouldn't know when to proceed.

This MR splits a conninfo containing more than one host into a sequence of attempt, each one to be tried with the connection timeout requested, or with a finite default (set to an arbitrary 2m 10s because that's what my Ubuntu times out with).

The mechanism of splitting a conninfo into leads itself to a simpler implementation of the async DNS lookup. In particular, previously we would have looked up all the host before attempting a connection; now we look up only the host as we try to connect to.

This MR might close:

- #602
- #674 

To do:

- [x] Add tests for the connection attempts
- [x] Implement [attempts randomization](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-LOAD-BALANCE-HOSTS)
- [x] Move `conninfo.resolve_hostaddr_async` to `conninfo._dsn`: the first version is no more used internally; the latter is exposed as deprecated function but we won't drop it right now.

### DIFF
--- a/docs/api/dns.rst
+++ b/docs/api/dns.rst
@@ -92,12 +92,6 @@ server before performing a connection.
     .. warning::
         This is an experimental method.
 
-    .. versionchanged:: 3.1
-        Unlike the sync counterpart, perform non-blocking address
-        resolution and populate the ``hostaddr`` connection parameter,
-        unless the user has provided one themselves. See
-        `resolve_hostaddr_async()` for details.
-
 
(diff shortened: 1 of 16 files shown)


marselsultanov:
Also this MR might close:
https://github.com/psycopg/psycopg/issues/580
:)

PR #680, opened 2023-11-16
Add test for client cursor with list param

webjunkie:
Coming from psycopg2, I'm trying to update queries that use lists with `IN`. I've already read the docs and try to use `ANY` now, but without success. I'm using the client cursor and what basically happens is that it seems to just make a string out of any list I give it and slap that into the `ANY`.

As I didn't see it covered yet, I added a test case. It fails and hopefully illustrates the issue, given what I'm doing is on the right track.

### DIFF
--- a/tests/test_cursor_client_async.py
+++ b/tests/test_cursor_client_async.py
@@ -131,6 +131,7 @@ async def test_leak(aconn_cls, dsn, faker, fetch, row_factory):
         ("select %%, %s", (["a"],), "select %, 'a'"),
         ("select %%, %(foo)s", ({"foo": "x"},), "select %, 'x'"),
         ("select %%s, %(foo)s", ({"foo": "x"},), "select %s, 'x'"),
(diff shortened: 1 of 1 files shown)


dvarrazzo:
The test shows clearly that if you are passing the Python list `["x", "y"]` it generates the valid Postgres array representation `'{x,y}'` and not what you say it does, "a string representation of the list", nor what you wrote in the test (`{'x','y'}`, which is at best a Python representation of a set).

I understand you have problems with using ANY: what is your problem?

webjunkie:
Thanks for the prompt feedback. I now realize my mistake in interpreting the behavior here.

I have revised the test case to align with the correct usage as per your guidance. While the initial misunderstanding was on my end, maybe this test could still be a valuable addition to ensure clarity for others transitioning from psycopg2 to psycopg3, especially regarding list handling.

I understand if you still feel the test is unnecessary. Thank you for your time and patience.

dvarrazzo:
@webjunkie most definitely, if you have suggestions about how to improve the docs, they are welcome.

Issue #681, opened 2023-11-20
Test issues with 3.1.13


mweinelt:
Hi!

After updating from 3.1.12 to 3.1.13 I see a few tests failing for our package in nixpkgs, and I think they're related to #675.

```
psycopg> ============================= test session starts ==============================
psycopg> platform linux -- Python 3.11.6, pytest-7.4.3, pluggy-1.3.0 -- /nix/store/qp5zys77biz7imbk6yy85q5pdv7qk84j-python3-3.11.6/bin/python3.11
psycopg> cachedir: /build
psycopg> Using --randomly-seed=1210518465
psycopg> default selector: EpollSelector
psycopg> Server version: PostgreSQL 15.5 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 12.3.0, 64-bit
psycopg> libpq wrapper implementation: c
psycopg> libpq used: 150005
psycopg> libpq compiled: 150005
psycopg> rootdir: /build/source
psycopg> configfile: pyproject.toml
psycopg> testpaths: tests
psycopg> plugins: anyio-4.0.0, randomly-3.13.0
psycopg> collected 4724 items / 78 deselected / 4646 selected       

[...]

psycopg> =========================== short test summary info ============================
psycopg> FAILED tests/test_module.py::test_connect[args3-kwargs3-user=foo] - AssertionError: assert {'user': 'foo', 'host': '/build/run/postgresql'} == ...
psycopg> FAILED tests/test_module.py::test_connect[args0-kwargs0-] - AssertionError: assert {'host': '/build/run/postgresql'} == {}
psycopg> FAILED tests/test_module.py::test_connect[args1-kwargs1-dbname=foo user=bar] - AssertionError: assert {'user': 'bar', 'dbname': 'foo', 'host': '/build/run...
psycopg> FAILED tests/test_module.py::test_connect[args2-kwargs2-port=15432] - AssertionError: assert {'host': '/build/run/postgresql', 'port': '15432'} =...
psycopg> FAILED tests/test_connection.py::test_connect_args[args4-kwargs4-dbname=qux user=joe port=5433] - AssertionError: assert {'user': 'joe', 'dbname': 'qux', 'host': '/build/run...
psycopg> FAILED tests/test_connection.py::test_connect_args[args1-kwargs1-] - AssertionError: assert {'host': '/build/run/postgresql'} == {}
[...]

PR #682, opened 2023-11-21
fix: flush PG environment variables in test_connect{,_args}

mweinelt:
These tests were previously affected by environment variables, that can be used to configure a postgresql instance.

Closes: #681

### DIFF
--- a/tests/test_module.py
+++ b/tests/test_module.py
@@ -15,7 +15,7 @@ from ._test_connection import drop_default_args_from_conninfo
         ((), {"user": "foo", "dbname": None}, "user=foo"),
     ],
 )
-def test_connect(monkeypatch, dsn, args, kwargs, want):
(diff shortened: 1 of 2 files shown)


mweinelt:
Looks like the windows runners do rely on at least PGUSER and PGPASSWORD environment variables.

dvarrazzo:
My proposal is in 75c89c3edf14cbae630db10f2b1e8d9bf9834e48: adding a `dns_env` fixture behaving like `dns` but embedding in the `PG*` env vars values too.

Tested in #683 and it worked, so I've pushed it here too (after dropping your temporary commit).

If tests pass here, and if you can please verify that tests pass in your environment too, these two MR can be merged.

Thank you very much!

mweinelt:
Tests on here also run fine for us!

dvarrazzo:
Cool! Merging both branches! Cheers!

PR #683, opened 2023-11-21
[maint-3.1] fix: flush PG environment variables in test_connect{,_args}

mweinelt:
These tests were previously affected by environment variables, that can be used to configure a postgresql instance.

Closes: #681


mweinelt:
Tested and works for us! Thank you.

PR #686, opened 2023-11-28
Add support for PyPy

ngnpope:
Supersedes #62.

This seemed to be enough to get everything passing on PyPy 3.10.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,10 @@ jobs:
           # Test with minimum dependencies versions
           - {impl: c, python: "3.8", ext: min, postgres: "postgres:15"}
 
+          # Test with PyPy.
+          - {impl: python, python: "pypy3.9", postgres: "postgres:13"}
+          - {impl: python, python: "pypy3.10", postgres: "postgres:14"}
+
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
@@ -100,6 +104,11 @@ jobs:
         run: |
           echo "DEPS=$DEPS numpy" >> $GITHUB_ENV
           echo "MARKERS=$MARKERS numpy" >> $GITHUB_ENV
+
+      - name: Exclude certain tests from pypy
+        if: ${{ startsWith(matrix.python, 'pypy') }}
+        run: |
+          echo "NOT_MARKERS=$NOT_MARKERS timing" >> $GITHUB_ENV
 
       - name: Configure to use the oldest dependencies
         if: ${{ matrix.ext == 'min' }}
--- a/psycopg/psycopg/pq/pq_ctypes.py
+++ b/psycopg/psycopg/pq/pq_ctypes.py
@@ -47,7 +47,10 @@ def version() -> int:
 
 @impl.PQnoticeReceiver  # type: ignore
 def notice_receiver(arg: c_void_p, result_ptr: impl.PGresult_struct) -> None:
-    pgconn = cast(arg, POINTER(py_object)).contents.value()
+    pgconn = cast(arg, POINTER(py_object)).contents.value
+    if callable(pgconn):  # Not a weak reference on PyPy.
+        pgconn = pgconn()
+
     if not (pgconn and pgconn.notice_handler):
         return
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,8 @@
+import gc
 import sys
 import asyncio
 import selectors
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Tuple
 
 import pytest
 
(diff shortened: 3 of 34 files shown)


.github/workflows/tests.yml
```diff
@@ -46,6 +46,10 @@ jobs:
           # Test with minimum dependencies versions
           - {impl: c, python: "3.8", ext: min, postgres: "postgres:15"}
 
+          # Test with PyPy.
+          - {impl: python, python: "pypy3.9", postgres: "postgres:13"}
+          - {impl: python, python: "pypy3.10", postgres: "postgres:14"}
```
ngnpope, line 51: I didn't try locally with 3.9, but hoping it should work.

Have used the same versions of PostgreSQL as for the CPython ones above.

psycopg/psycopg/pq/pq_ctypes.py
```diff
@@ -47,7 +47,12 @@ def version() -> int:
 
 @impl.PQnoticeReceiver  # type: ignore
 def notice_receiver(arg: c_void_p, result_ptr: impl.PGresult_struct) -> None:
-    pgconn = cast(arg, POINTER(py_object)).contents.value()
+    pgconn = cast(arg, POINTER(py_object)).contents.value
+
+    if sys.implementation.name != "pypy" or callable(pgconn):
+        # On PyPy this doesn't seem to be a weak reference.
+        pgconn = pgconn()
```
ngnpope, line 54: This was the main thing that was causing me issues when trying to get this up and running with Django.

tests/conftest.py
```diff
@@ -68,6 +68,12 @@ def pytest_sessionstart(session):
     cache.set("segfault", True)
 
 
+def pytest_runtest_setup(item):
+    if sys.implementation.name == "pypy":
+        if any(mark.name == "timing" for mark in item.iter_markers()):
+            pytest.skip("timing tests are unstable")
```
ngnpope, line 74: I skipped all of the timing tests. Initially I tried to come up with a different tolerance, but it was unstable.

tests/fix_pq.py
```diff
@@ -53,7 +53,11 @@ def libpq():
         # Not available when testing the binary package
         libname = find_libpq_full_path()
         assert libname, "libpq libname not found"
-        return ctypes.pydll.LoadLibrary(libname)
+        if sys.implementation.name == "pypy" and not hasattr(ctypes, "pydll"):
+            # https://foss.heptapod.net/pypy/pypy/-/issues/3496
+            return ctypes.cdll.LoadLibrary(libname)
+        else:
+            return ctypes.pydll.LoadLibrary(libname)
```
ngnpope, line 60: Python's documentation states:

> Thus, this is only useful to call Python C api functions directly.

See https://docs.python.org/3/library/ctypes.html#ctypes.PyDLL

Given this is calling into libpq, should this just be using CDLL everywhere?

tests/test_copy.py
```diff
@@ -673,6 +674,9 @@ def test_connection_writer(conn, format, buffer):
 
 
 @pytest.mark.slow
+@pytest.mark.skipif(
+    sys.implementation.name == "pypy", reason="depends on refcount semantics"
+)
```
ngnpope, line 679: I've skipped all of the leak tests which compare reference counts.

psycopg/psycopg/pq/pq_ctypes.py
dvarrazzo: Not a small difference :\

Can we just test `callable()`?

tests/test_copy.py
dvarrazzo: I think you can add a `pytest.skip()` in the `gc_count()` function so that you get all the current and future refcount tests skipped on pypy

tests/conftest.py
dvarrazzo: The test marks are meant to control the test suite from the outside. You can add `timing` to the `NOT_MARKERS` env var.

```yaml
      - name: Exclude certain tests from pypy
        if: ${{ matrix.ext == 'pypy' }}  # figure out the check
        run: |
          echo "NOT_MARKERS=$NOT_MARKERS timing" >> $GITHUB_ENV
```

tests/fix_pq.py
dvarrazzo: Looking at 69fbbec4337494550d76de9837717534e6f70e9b and to the fact that [in the main adapter we use `ctypes.cdll`](https://github.com/psycopg/psycopg/blob/607425c28ba6644163dc9b876f31ac2f480ec81f/psycopg/psycopg/pq/_pq_ctypes.py#L21), I think we can use `cdll` everywhere.

tests/test_copy.py
dvarrazzo: ...Although this would end up skipping the test only after it has run, which is a waste of time.

Can you create a `gc` fixture, to return an object with methods `count()` and `collect()`, but we can just skip the test on fixture creation on pypy.

dvarrazzo:
If we add support for pypy, do we need to add [trove classifiers to the metadata](https://github.com/psycopg/psycopg/blob/607425c28ba6644163dc9b876f31ac2f480ec81f/psycopg/setup.cfg#L28-L44)? [To the pool too](https://github.com/psycopg/psycopg/blob/607425c28ba6644163dc9b876f31ac2f480ec81f/psycopg_pool/setup.cfg#L22-L38)?

tests/test_copy.py
ngnpope: There are some tests that will pass if we add `gc_collect()` after `del`, and a few other cases too that already use `gc_collect()`. If we move the function into the object returned by the fixture object we make skip tests that would happily work. Do you mind this? (Or we can also keep `gc_collect()` around for these few cases?)

dvarrazzo: If tests pass reliably using `gc_collect()` maybe let's keep that function around. Thank you for the care :slightly_smiling_face:

ngnpope: I'm exposing `gc_collect` as an additional fixture. This will allow changing the fixture from `gc_collect` to `gc` and the call from `gc_collect()` to `gc.collect()` if any cases do come up flaky.

ngnpope:
I've pushed up two fixup commits, the first with all the general bits and the second is focused on the gc stuff.

dvarrazzo:
There was a [surprising error](https://github.com/psycopg/psycopg/actions/runs/7067161708) in psycopg 3.1 on pypy 3.9, which didn't appear in psycopg master nor on pypy 3.10. I think that the problem is just an anyio 3 bug (master uses anyio 4) so I've [skipped the failing tests](https://github.com/psycopg/psycopg/commit/602c875f7156e74f2b923a3fb4c2e9aa476ba6cf).

Psycopg 3.1.14 is on the way!

ngnpope:
Thanks Daniele!

om26er:
Did we get around to doing any performance benchmarks? Would be great to know.

Issue #692, opened 2023-12-03
Flaky tests in at least `test_cursor_async.py`


nialov:
# Problem

While trying to fix `psycopg` for `nixpkgs`, I noticed that some of the tests in the test suite fail depending on if they are run together or individually.

If `pytest` is ran with `-k test_cursor_async` some tests will fail:

~~~
       > SKIPPED [1] tests/test_gevent.py:8: could not import 'gevent': No module named 'gevent'
       > SKIPPED [2] tests/crdb/test_cursor_async.py:18: only for CockroachDB
       > FAILED tests/test_cursor_async.py::test_leak[asyncio-dict_row-many-0-b] - AssertionError: objects leaked: 0, -2
       > FAILED tests/test_cursor_async.py::test_leak[asyncio-dict_row-all-0-b] - AssertionError: objects leaked: 0, -2
       > FAILED tests/test_cursor_async.py::test_leak[asyncio-tuple_row-many-0-b] - AssertionError: objects leaked: -2, 2
       > ========== 3 failed, 139 passed, 3 skipped, 4580 deselected in 52.59s ==========
~~~

If then one of these tests is picked and `pytest` is ran with `-k 'test_leak[asyncio-namedtuple_row-many-0-s]'` the test will not fail.

You can also run some of the failing tests together (`-k 'test_leak[asyncio-namedtuple_row-many-0-s] or test_leak[asyncio-dict_row-many-0-b]'`) and they will pass.

# What I expected

These look like side-effects between the tests which is flaky behavior. Tests should pass regardless of if they are run together or individually.

# Additional info

Full test suite log without any `-k` flags can be found here and with failing tests: <https://gist.github.com/nialov/0d7a35e71ec97d1022cf0564884e3354>. Scroll down to find `pytest` logs.

I found a pull request where similar failures were detected without explanation: <https://github.com/psycopg/psycopg/pull/159#pullrequestreview-811932912>. No idea if related.

After a few runs the whole suite suddenly finished successfully... So there is definitely some flaky behavior afoot.

Test setup:

~~~
python3.11-psycopg> ============================= test session starts ==============================
python3.11-psycopg> platform linux -- Python 3.11.6, pytest-7.4.3, pluggy-1.3.0
python3.11-psycopg> cachedir: /build
python3.11-psycopg> Using --randomly-seed=2252975248
python3.11-psycopg> default selector: EpollSelector
[...]

PR #693, opened 2023-12-04
Mark flaky ref count tests

nialov:
All tests that have the 'gc' fixture are now marked with the refcount mark. The reasoning is that they demonstrate flaky behaviour and disabling them in certain CI is necessary to ensure reliable testing. See #692

Besides the suggested fixes by @dvarrazzo  I moved come code from `fix_proxy.py` to `conftest.py` as I believe now that other kinds of tests are also marked in `pytest_collection_modifyitems` it probably does not belong in a file named `fix_proxy`. Let me know what you think!


dvarrazzo:
Thank you for the pr, @nialov 

I don't like to have too much stuff in the conftest, so I would rather not move the proxy `collection_modifyitems()` there.

Could you please revert the proxy change, and instead create a new `fix_gc.py` module, where to move all the gc-related code: a `collection_modifyitems()` on its own only dealing with the `gc` fixture and a `pytest_configure` registering the `refcount` marker.

Thank you very much!

nialov:
Sure, will do! At the latest at the end of the week.

nialov:
Done and hopefully it is according to your request!

dvarrazzo:
Thank you, what I did is part of what I wanted, but what I meant was to move **everything** gc-related from conftest to the module you just created, including the fixtures that were already there: pretty much this block and all the imports it requires, and clean up the unneeded imports from conftest.py:

https://github.com/psycopg/psycopg/blob/4747cdb2701e7df6128ed43022e3d70fb880343b/tests/conftest.py#L106-L170

nialov:
I hope it is one step closer again :D

dvarrazzo:
That looks good! Merging. I will also rebase it on the branch 3.1 and push there, so you can build the 3.1.x

Issue #703, opened 2023-12-17
Regression in async connection on 3.1.15 when multiple hosts are present


CaselIT:
Hi.

The following script that uses multiple hosts specifying the port only on some of them passes in the sync connection but fails on the async one due to an assertion error
```py
import traceback
import psycopg

urls = [
    {
        "dbname": "test",
        "user": "scott",
        "password": "tiger",
        "host": "localhost,localhost,localhost",
        "port": ",5432,",
    },
    {
        "dbname": "test",
        "user": "scott",
        "password": "tiger",
        "host": "localhost,localhost,localhost",
        "port": "5432,,",
    },
    {
        "dbname": "test",
        "user": "scott",
        "password": "tiger",
        "host": "localhost,localhost,localhost",
        "port": "5432,,5432",
    },
]


def sync():
    for url in urls:
        with psycopg.connect(**url) as conn:
            cur = conn.execute("select 42")
            print(cur.fetchall())


async def async_():
    for url in urls:
        try:
            async with await psycopg.AsyncConnection.connect(**url) as conn:
                cur = await conn.execute("select 42")
                print(await cur.fetchall())
        except Exception:
            traceback.print_exc()


sync()
import asyncio

try:
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
except:
    pass

asyncio.run(async_())
```
output is
```
[(42,)]
[(42,)]
[(42,)]
Traceback (most recent call last):
  File "test3.py", line 39, in async_
    async with await psycopg.AsyncConnection.connect(**url) as conn:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cfede\venvs\sa\Lib\site-packages\psycopg\connection_async.py", line 124, in connect
    attempts = await conninfo_attempts_async(params)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cfede\venvs\sa\Lib\site-packages\psycopg\conninfo.py", line 330, in conninfo_attempts_async
    attempts.extend(await _resolve_hostnames(attempt))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cfede\venvs\sa\Lib\site-packages\psycopg\conninfo.py", line 426, in _resolve_hostnames
    assert port and "," not in port  # assume a libpq default and no multi
[...]

PR #705, opened 2023-12-18
Add async driver to sqlalchemy pipeline

CaselIT:
Also run psycopg async in the sqlalchemy tests. This is partially connected to #703

I'm trying the pipeline at https://github.com/CaselIT/psycopg/actions/runs/7253511346
It should fail due to #703

### DIFF
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -110,7 +110,7 @@ jobs:
         env:
           URL: postgresql+psycopg://postgres:password@127.0.0.1/test
         working-directory: sa_home/sa
-        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst --color=yes
+        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst --color=yes --dbdriver psycopg_async
 
   django:
     # linux should be enough to test if everything works.
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -30,6 +30,13 @@ Psycopg 3.2 (unreleased)
 .. __: https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types
 
 
+Psycopg 3.1.16 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix empty ports handling in async multiple connection attempts
+  (:ticket:`#703`).
+
+
 Current release
 ---------------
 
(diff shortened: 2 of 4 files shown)


CaselIT:
>  FAILED test/dialect/postgresql/test_dialect.py::BackendDialectTest_postgresql+psycopg_async_15_5::test_multiple_host_real_connect[***/DB?host=H&host=H:P&host=H-False] - AssertionError
FAILED test/dialect/postgresql/test_dialect.py::BackendDialectTest_postgresql+psycopg_async_15_5::test_multiple_host_real_connect[***/DB?host=H:P&host=H&host=H-False] - AssertionError
FAILED test/dialect/postgresql/test_dialect.py::BackendDialectTest_postgresql+psycopg_async_15_5::test_multiple_host_real_connect[***/DB?host=H:P&host=H&host=H:P-False] - AssertionError

it's correctly running the async version too

.github/workflows/3rd-party-tests.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         env:
           URL: postgresql+psycopg://postgres:password@127.0.0.1/test
         working-directory: sa_home/sa
-        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst --color=yes
```
dvarrazzo, line 113: Does this run only async instead of sync? Don't we want to run both sync and async?

dvarrazzo (commented):
I [run successfully](https://github.com/psycopg/psycopg/actions/runs/7255387291) your changeset against [the commit](https://github.com/psycopg/psycopg/commit/d038fecd9efda8dcb85f2dfe7c3b6a5bc7b4f52d) that fixes issue #703.

After clarifying my question about sync/async run I will merge this MR too.

Thank you very much for your help!

.github/workflows/3rd-party-tests.yml
CaselIT: No, it runs both. It's not overly clear, but the `dbdriver` option is used to add additional drivers to run with the provided uri.

If it's clearer, both versions could be added as `dbdriver` since the testing framework of sqlalchemy will deduplicate them

dvarrazzo: Ok, thank you for confirming, and indeed the test run shows almost twice the number of tests. :+1:

CaselIT: Yes, it basically creates a copy of every "backend" test for each driver selected

Issue #708, opened 2023-12-29
Typing of AsyncConnectionPool is incompatible with pyright


adriangb:
```python
from psycopg_pool import AsyncConnectionPool

async def main() -> None:
    async with AsyncConnectionPool() as pool:
        pass
```

```
test.py
  test.py:5:16 - error: Object of type "AsyncConnectionPool[AsyncConnection[TupleRow]]" cannot be used with "with" because it does not implement __aenter__ (reportGeneralTypeIssues)
  test.py:5:41 - error: Type of "pool" is unknown (reportUnknownVariableType)
  test.py:5:41 - error: Variable "pool" is not accessed (reportUnusedVariable)
3 errors, 0 warnings, 0 informations 
```

pyright==1.1.343
psycopg==3.1.16
psycopg-binary==3.1.16
psycopg-pool==3.2.0

dvarrazzo:
This is clearly incorrect:

https://github.com/psycopg/psycopg/blob/aa4d15b8ba9a7f0865aa486c804f46adadc440a9/psycopg_pool/psycopg_pool/pool_async.py#L522

Pyright is pywrong here.

adriangb:
I think the issue is the _Self type and the bound with generics on it. I’m happy to report an issue on pyright.

dvarrazzo:
Thank you very much, very appreciated!

adriangb:
No problem! It might be worth trying to use `typing_extensions.Self` to see if that resolves the problem.

adriangb:
My suspicion is that although the error from pyright is confusing what it's trying to say is "this class doesn't implement the AsyncContextManager API as I expect it to" due to the typing of `_Self`.

dvarrazzo:
That might be, however it's a construct that mypy understand correctly, and mypy is the static checker we use for our codebase.

I don't remember if there was some work done in the past to use `typing.Self`. I assume we tried, but we found some shortcoming. A lot of time has passed since (I can't find a relevant branch) and maybe we had problems with some Python version (3.6, 3.7) that we don't need to support anymore.

[...]

Issue #308, opened 2022-05-24
Attempting to define the returned value of connect() as generic


dvarrazzo:
Subclassing a connection doesn't require to redefine the `connect()` method, because it is generic. However we can't explain this to mypy yet, because [the function is defined returning `Connection`](https://github.com/psycopg/psycopg/blob/79c14f47f2068cd568ee8b50df79090f16a3f9c5/psycopg/psycopg/connection.py#L673-L713).

Attempts to use [a generic self](https://mypy.readthedocs.io/en/stable/more_types.html#precise-typing-of-alternative-constructors) fail because it seems that the generic self doesn't play well with a default argument for the generic parameter.

This is a self-contained reproduction of the issue: [2023-12: edited using more modern typing techniques]

```python
from typing import Any, Callable, Generic, Self, Sequence, TypeAlias, TypeVar, overload, cast

Row = TypeVar("Row", covariant=True)
TupleRow: TypeAlias = tuple[Any, ...]
DictRow: TypeAlias = dict[str, Any]
RowFactory: TypeAlias = Callable[[Sequence[Any]], Row]

def tuple_row(values: Sequence[Any]) -> TupleRow:
    ...

def dict_row(values: Sequence[Any]) -> DictRow:
    ...

class Connection(Generic[Row]):
    @overload
    @classmethod
    def connect(cls, *, row_factory: RowFactory[Row]) -> Self:
        ...

    @overload
    @classmethod
    def connect(cls) -> Self:
        ...

    @classmethod
    def connect(cls, *, row_factory: RowFactory[Row] | None = None) -> Self:
        ...

c = Connection.connect()  # mypy reports required annotation
reveal_type(c)  # Revealed as Connection[Any], should be Connection[TupleRow]

ct = Connection.connect(row_factory=tuple_row)
reveal_type(ct)  # Revelaled as Connection[TupleRow]

cd = Connection.connect(row_factory=dict_row)
reveal_type(cd)  # Revelaled as Connection[DictRow]
```

with this definition, `connect()` without `row_factory` argument loses the info that it returns tuple rows.

It would be nice to be able to specify `Self[TupleRow]` as return values, but the typevar cannot be used with an argument.

Another attempt that I wish worked is to use a default argument instead of override: [2023-12: edited using more modern typing techniques]

```python
class ConnectionNoOverload(Generic[Row]):

    @classmethod
    def connect(cls, *, row_factory: RowFactory[Row] = tuple_row) -> Self:
        ...
```

[...]

PR #710, opened 2023-12-30
refactor: use typing.Self

dvarrazzo:
The object seems available for all the supported Python version and should avoid problems with PyRight (see #708).

It is not a solution for #308 because we cannot use `Self[Row]`.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,16 @@
 ``psycopg`` release notes
 =========================
 
+Future releases
+---------------
+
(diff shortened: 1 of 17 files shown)


dvarrazzo:
Reading back https://github.com/python/typing/discussions/1197, the reason we didn't use `typing.Self` was that, at the time, it wasn't supported by Mypy.

adriangb:
This seems to fix the pyright issues!

PR #720, opened 2024-01-16
fix: use long for seconds to prevent IntervalLoader from overflowing

### LINKED ISSUE #719: Overflow in interval loading in C module
It seems there's a bug within psycopg-binary `IntervalLoader` implementation, this however can't be reproduced using the python version of `IntervalLoader`.

To reproduce the issue:

```pip install psycopg```

```test.py
import psycopg

conn = psycopg.connect(user='postgres', password='postgres', dbname='postgres')

with conn.cursor() as cur:
    cur.execute("select ('1993534:40:40.447')::interval")
    row=cur.fetchone()
    print(row)
```

output from the above code: ✅
```(datetime.timedelta(days=83063, seconds=81640, microseconds=447000),)```

now `pip install psycopg-binary` and run the code again, it outputs: ❌
```(datetime.timedelta(days=-16357, seconds=35048, microseconds=447000),)```

It seems regular `int` is overflowing so `long` must be used for seconds,  [here's a rough patch](https://github.com/psycopg/psycopg/pull/720)

enapupe:
The patch is incomplete, it only touches `IntervalLoader` implementation. Other implementations that rely on the same helper function are broken.

Should fix #719

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -34,6 +34,7 @@ Psycopg 3.1.18 (unreleased)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Fix possible deadlock on pipeline exit (:ticket:`685`).
+- Fix overflow loading large intervals in C module (:ticket:`719`).
 
 
 Current release
--- a/psycopg_c/psycopg_c/types/datetime.pyx
+++ b/psycopg_c/psycopg_c/types/datetime.pyx
@@ -391,7 +391,7 @@ cdef class DateLoader(CLoader):
         if length != 10:
             self._error_date(data, "unexpected length")
 
-        cdef int vals[3]
+        cdef long vals[3]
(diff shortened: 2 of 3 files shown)


enapupe:
@dvarrazzo this should be ready to review. In the end, cython's `timedelta_new` implementation would also overflow so I had to convert seconds into days before forwarding it. I tried not changing `vals` from `int` to `long` but couldn't make it work, this is my first time ever playing with cython and I have almost no experience with C.

dvarrazzo:
Changing `vals` is ok, no problem with that.

`timedelta_new` is actually a wrapper for [a Python API function](https://docs.python.org/3/c-api/datetime.html#c.PyDelta_FromDSU) that takes only ints, so making sure that seconds don't overflow is right too.

Now I see that only Windows overflow. Must be a 32/64 bits difference between MSVC and GCC... I can take a look at it.

Issue #723, opened 2024-01-26
Add scalar row factory


embecka:
Please, add <code>scalar_row</code> factory to return a scalar value instead of a tuple.

Usage:
```python
with connection.cursor(row_factory=scalar_row()) as cursor:
	cursor.execute("SELECT username FROM user WHERE id = %(id)s", {"id": 1})
	
	print(cursor.fetchone())  # Prints "username" instead of ("username",)
```

Possible implementation:
```python
def scalar_row(index: int = 0) -> BaseRowFactory[Any]:
	def scalar_row_(cursor: BaseCursor[Any, Any]) -> RowMaker[Any]:
		res = cursor.pgresult
		if not res:
			return no_result
		
		nfields = _get_nfields(res)
		if nfields is None:
			return no_result
		
		if index >= nfields:
			return no_result
		
		def scalar_row__(values: Sequence[Any]) -> Any:
			return values[index]
		
		return scalar_row__
	
	return scalar_row_
```

PR #724, opened 2024-01-26
added `scalar_row` factory

embecka:
Closes psycopg/psycopg#723.

### DIFF
--- a/docs/api/rows.rst
+++ b/docs/api/rows.rst
@@ -14,6 +14,10 @@ Check out :ref:`row-factory-create` for information about how to use these objec
 .. autofunction:: tuple_row
 .. autofunction:: dict_row
 .. autofunction:: namedtuple_row
+.. autofunction:: scalar_row
+
+        .. versionadded:: 3.2
+
 .. autofunction:: class_row
 
     This is not a row factory, but rather a factory of row factories.
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -19,6 +19,7 @@ Psycopg 3.2 (unreleased)
   (:ticket:`340`).
 - Add :ref:`raw-query-cursors` to execute queries using placeholders in
   PostgreSQL format (`$1`, `$2`...) (:ticket:`#560`).
+- Add `~rows.scalar_row` to return scalar values from a query (:ticket:`#723`).
 - Add `~Connection.set_autocommit()` on sync connections, and similar
   transaction control methods available on the async connections.
 - Add support for libpq functions to close prepared statements and portals
--- a/psycopg/psycopg/rows.py
+++ b/psycopg/psycopg/rows.py
@@ -207,6 +207,28 @@ def kwargs_row(func: Callable[..., T]) -> BaseRowFactory[T]:
     return kwargs_row_
 
 
+def scalar_row(cursor: "BaseCursor[Any, Any]") -> "RowMaker[Any]":
+    """
+    Generate a row factory returning the first column
+    as a scalar value.
+    """
+    res = cursor.pgresult
+    if not res:
+        return no_result
+
+    nfields = _get_nfields(res)
+    if nfields is None:
+        return no_result
+
+    if nfields < 1:
+        raise e.ProgrammingError("at least one column expected")
+
+    def scalar_row_(values: Sequence[Any]) -> Any:
+        return values[0]
+
+    return scalar_row_
+
+
 def no_result(values: Sequence[Any]) -> NoReturn:
     """A `RowMaker` that always fail.
 
--- a/tests/test_rows.py
+++ b/tests/test_rows.py
@@ -102,6 +102,16 @@ def test_kwargs_row(conn):
     assert p.age == 42
 
 
+def test_scalar_row(conn):
+    cur = conn.cursor(row_factory=rows.scalar_row)
+    cur.execute("select 1")
+    assert cur.fetchone() == 1
+    cur.execute("select 1, 2")
+    assert cur.fetchone() == 1
+    with pytest.raises(psycopg.ProgrammingError):
+        cur.execute("select")
+
+
 @pytest.mark.parametrize(
     "factory",
     "tuple_row dict_row namedtuple_row class_row args_row kwargs_row".split(),


psycopg/psycopg/rows.py
```diff
@@ -211,6 +211,32 @@ def kwargs_row__(values: Sequence[Any]) -> T:
     return kwargs_row_
 
 
+def scalar_row(index: int = 0) -> BaseRowFactory[Any]:
```
dvarrazzo, line 214: Is there really an use case for the index to be something else than 0?

In my understanding this row factory can be used every time you have a query selecting a single row and you want to get a list with all the result, or you run a query and you want a single number as a result.

Is there really a practical case in which someone may run a query wanting the second column as result, discarding all the rest?

I propose to make this a simple row factory, to use as

```python
conn.cursor(row_factory=scalar_row)
```

and always return the first column.

If you think that the index is something that should be specified:

- the index may be optional, in the sense that the factory could be called both as `row_factory=scalar_row` and `row_factory=scalar_row(0)`. This can be implemented the same way decorators that may or may not take arguments are implemented: at the outermost call you can check whether the function is called with an index or with a cursor and act accordingly.
- If we accept an index, it should also be possible to specify a column name, such as `row_factory=scalar_row("id")`.

psycopg/psycopg/rows.py
```diff
@@ -211,6 +211,32 @@ def kwargs_row__(values: Sequence[Any]) -> T:
...
+        res = cursor.pgresult
+        if not res:
+            return no_result
+    
+        nfields = _get_nfields(res)
+        if nfields is None:
+            return no_result
+
+        if index >= nfields:
+            return no_result
```
dvarrazzo, line 230: `no_result` raises an `InterfaceError("the cursor doesn't have a result")` so it doesn't seem appropriate here. It would be more appropriate to raise a `ProgrammingError` if the user asked for a non-existing column. It would be also expected to handle negative numbers as per normal Python indexing and raise the same exception if the index is < -nfields.

docs/api/rows.rst
```diff
@@ -40,6 +40,7 @@ Check out :ref:`row-factory-create` for information about how to use these objec
 
 .. autofunction:: args_row
 .. autofunction:: kwargs_row
+.. autofunction:: scalar_row
```
dvarrazzo, line 43: This requires a `versionadded`.

psycopg/psycopg/rows.py
```diff
@@ -211,6 +211,32 @@ def kwargs_row__(values: Sequence[Any]) -> T:
     return kwargs_row_
 
 
+def scalar_row(index: int = 0) -> BaseRowFactory[Any]:
+    """Generate a row factory returning a column at `!index` as a scalar value.
```
dvarrazzo, line 215:
```suggestion
    """Generate a row factory returning the column at `!index` as a scalar value.
```

dvarrazzo (commented):
The most important thing to figure out is whether to have an index or not. Please let me know if you see an use case for anything different than 0.

The MR also requires an entry in `docs/news.rst` (the feature would be released in 3.2).

Please note that the failure in lint are caused by black 24.1.0, which has been released about one day ago. I have fixed it on master, so you can rebase your branch on it to solve the problem.

psycopg/psycopg/rows.py
embecka: I don't see a practical case running a query containing more columns than one but returning only a scalar value. Such queries can be easily rewritten to return only one column. In this case, I prefer simplicity over non-practical functionality.

psycopg/psycopg/rows.py
embecka: I will check, if the cursor contains exactly one column, if not, i will raise `ProgrammingError`.

docs/api/rows.rst
embecka: Done.

psycopg/psycopg/rows.py
embecka: My english is not good as I would wish

tests/fix_psycopg.py
```diff
@@ -24,7 +24,7 @@ def global_adapters():
 
 
 @pytest.fixture
-@pytest.mark.crdb_skip("2-phase commit")
+# @pytest.mark.crdb_skip("2-phase commit")
```
embecka, line 27: see: https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

dvarrazzo: This is totally unrelated with this merge request.

dvarrazzo: Also, if it doesn't work, the line should be removed, not commented out. But please make a separate MR for it.

psycopg/psycopg/rows.py
```diff
@@ -207,6 +207,28 @@ def kwargs_row__(values: Sequence[Any]) -> T:
...
+    res = cursor.pgresult
+    if not res:
+        return no_result
+
+    nfields = _get_nfields(res)
+    if nfields is None:
+        return no_result
+
+    if nfields != 1:
+        raise e.ProgrammingError(f"Expected exactly one column, found {nfields}")
```
dvarrazzo, line 224: I wonder if we shouldn't accept queries that return more than one column, and just discard the one after the first.

```suggestion

    if nfields < 1:
        raise e.ProgrammingError(f"at least one column expected")
```

If you are wondering if it's possible to write a query that returns zero columns, it is: just `SELECT`.

What do you think?

Regardless of this choice, the error message should start with lowercase (if it start with uppercase, it is a complete sentence, and should end with a full-stop. However it's easier to compose a message if it's lowercase and has no final full-stop).

tests/test_rows.py
```diff
@@ -102,6 +102,12 @@ def test_kwargs_row(conn):
     assert p.age == 42
 
 
+def test_scalar_row(conn):
+    cur = conn.cursor(row_factory=rows.scalar_row)
+    cur.execute("select 1")
+    assert cur.fetchone() == 1
```
dvarrazzo, line 108: The test should include a case to show that the error is raised (with a query returning 0 columns, and with a query returning >1 column if we think that's the way to go).

docs/news.rst
```diff
@@ -28,6 +28,8 @@ Psycopg 3.2 (unreleased)
   in non-pipeline mode and not totally reliable (:ticket:`#604`).
   The `Cursor` now only preserves the results set of the last
   `~Cursor.execute()`, consistently with non-pipeline mode.
+- Add `~rows.scalar_row()` factory to generate `rows.RowMaker` returning the scalar value
+  taken from the first and only column (:ticket:`#724`).
```
dvarrazzo, line 32: I propose the following wording:

```suggestion
- Add `~rows.scalar_row` to return scalar values from a query (:ticket:`#723`).
```

Also I would move this point to line 22, because in this changelist there are first new features, then bugfixes, and roughly in order from the most generally useful to the least, and this feature is pretty useful to several people, more than for instance the features described at line 22.

dvarrazzo (changes requested):

tests/fix_psycopg.py
dvarrazzo: I have seen now that this change has broken the tests. I'll push a commit to master to fix them.

dvarrazzo: Master is now fixed, this changeset can be dropped.

psycopg/psycopg/rows.py
embecka: I don't have a strict opinion on this matter. I am OK with your suggestion.

dvarrazzo:
FYI, master is currently broken, I am investigating why.

embecka:
I will wait until master is fixed

dvarrazzo (approved):
Everything looks ok, thank you very much!

I have rebased and squashed your commits; will merge to master as soon as tests pass.

PR #725, opened 2024-01-27
fix: add `libc.so` fallback for musl systems to the ctypes impl

mgorny:
Add a fallback to `libc.so` library name to fix loading the ctypes implementation on musl systems.  On musl, `find_library("c")` does not work (the problem has been reported to CPython in 2014, and has not been resolved yet), causing the module to fail on `assert libcname`. Instead, add a fallback to using `libc.so` and let ctypes raise an exception if such a library does not exist.

### DIFF
--- a/psycopg/psycopg/pq/_pq_ctypes.py
+++ b/psycopg/psycopg/pq/_pq_ctypes.py
@@ -29,7 +29,10 @@ FILE_ptr = POINTER(FILE)
 
 if sys.platform == "linux":
     libcname = ctypes.util.find_library("c")
-    assert libcname
+    if not libcname:
+        # Likely this is a system using musl libc, see the following bug:
+        # https://github.com/python/cpython/issues/65821
+        libcname = "libc.so"
     libc = ctypes.cdll.LoadLibrary(libcname)
 
     fdopen = libc.fdopen
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -37,6 +37,10 @@ Psycopg 3.1.18 (unreleased)
 
 - Fix possible deadlock on pipeline exit (:ticket:`685`).
 - Fix overflow loading large intervals in C module (:ticket:`719`).
+- Fix compatibility with musl libc distributions affected by `CPython issue
+  #65821`__ (:ticket:`#725`).
+
+.. __: https://github.com/python/cpython/issues/65821
 
 
 Current release


thesamesam:
For completeness, the Gentoo bug is https://bugs.gentoo.org/923003 and the upstream CPython bug is https://github.com/python/cpython/issues/65821.

dvarrazzo:
Hello,

does it mean that the pure Python version of psycopg just doesn't work on musl? I would like to see it tested in CI.

The change you provided has broken the tests. Are you able to provide something working?

psycopg/psycopg/pq/_pq_ctypes.py
```diff
@@ -28,8 +28,8 @@ class FILE(Structure):
 FILE_ptr = POINTER(FILE)
 
 if sys.platform == "linux":
-    libcname = ctypes.util.find_library("c")
-    assert libcname
+    # find_library("c") does not work on musl, fall back to libc.so instead
+    libcname = ctypes.util.find_library("c") or "libc.so"
```
dvarrazzo, line 32: Because the `libcname` is the normal and expected path (so much that there was an assert to enforce that), I would rather make it clear why there are cases where that's not the case.

```suggestion
    libcname = ctypes.util.find_library("c")
    if not cname:
        # Likely this is musl linux, see bugs ...
        # [maybe even add a check to verify we are on musl]
        libcname = "libc.so"
```

dvarrazzo (changes requested):

mgorny:
> does it mean that the pure Python version of psycopg just doesn't work on musl?

Yes, or at least on vanilla musl+CPython. It is possible that some distributions are patching musl and/or CPython to make `ctypes.util.find_library("c")` work.

> I would like to see it tested in CI.

I don't think GitHub Actions support any musl-backed setup.

> The change you provided has broken the tests. Are you able to provide something working?

Could you be more specific? Give that `master` is broken right now and how many different workflows are being run, it's hard for me to find the specific job that I broke.

I'll update the code per the suggestion.

psycopg/psycopg/pq/_pq_ctypes.py
mgorny: I'm afraid that detecting musl would be non-trivial since musl upstream is of pretty strong opinion that people shouldn't be checking what libc is used and therefore do not provide any explicit way of checking for musl.

thesamesam:
> > I would like to see it tested in CI.
> 
> I don't think GitHub Actions support any musl-backed setup.
> 

I think you can ask Github Actions to use an arbitrary Docker container, but then some action-stuff like the Python setup may not work, see e.g. https://github.com/Jannik2099/pms-utils/blob/main/.github/workflows/alpine.yml.

But Alpine patches their musl, so this bug wouldn't have shown up there anyway. We'd need to do a Gentoo image.

dvarrazzo:
> Give[n] that `master` is broken right now

Ah, I didn't notice that the word is on fire. Master was working as of yesterday. I'll check what's up with it.

dvarrazzo:
> But Alpine patches their musl, so this bug wouldn't have shown up there anyway.

I understand, I identified musl == alpine but of course they are two different things.

Nevermind then, I don't want to add other element to the test matrix, it's already slow enough as it is.

dvarrazzo:
I have rebased your branch after fixing master, and added a release note entry. As soon as tests pass I will merge your PR and will release 3.1.18 including this fix in the next few days.

Thank you for your contribution!

mgorny:
Thanks a lot!

PR #746, opened 2024-03-04
Improve performance of copy

### LINKED ISSUE #745: Binary copy performance on Mac
Hi, big thanks for this library!

I wanted to report a large difference in binary copy performance between Mac and Linux (as well as between psycopg and Ruby's [pg gem](https://github.com/ged/ruby-pg) on Mac).

Project | Platform | Time
--- | --- | ---
psycopg 3 | Linux | 10 sec
psycopg 3 | Mac | 321 sec
pg gem | Mac | 4 sec

Here's [a repo](https://github.com/ankane/psycopg-repro) to reproduce ([CI results](https://github.com/ankane/psycopg-repro/actions/runs/8133986713)), but also sharing the scripts below.

Python script

```python
import psycopg

conn = psycopg.connect(dbname='repro', autocommit=True)
conn.execute('DROP TABLE IF EXISTS items')
conn.execute('CREATE TABLE items (id bigserial, description text)')

description = 'test' * 100

cur = conn.cursor()
with cur.copy('COPY items (description) FROM STDIN WITH (FORMAT BINARY)') as copy:
    copy.set_types(['text'])

    for i in range(1000000):
        # show progress
        if i % 10000 == 0:
            print('.', end='', flush=True)

        copy.write_row([description])

print()
```

Ruby script

```ruby
require "pg"

conn = PG.connect(dbname: "repro")
conn.exec("DROP TABLE IF EXISTS items")
conn.exec("CREATE TABLE items (id bigserial, description text)")

description = "test" * 100

coder = PG::BinaryEncoder::CopyRow.new
conn.copy_data("COPY items (description) FROM STDIN WITH (FORMAT BINARY)", coder) do
  1000000.times do |i|
    # show progress
    putc "." if i % 10000 == 0

    conn.put_copy_data([description])
  end
end

puts
```

This happens with `psycopg`, `psycopg[binary]`, and `psycopg[c]` (version 3.1.18).

ankane:
Fixes #745 

I kept the flush code the same as `copy_end`, but think it may be better to break if `f != 1` instead of `f == 0` in both places to account for errors, which return -1 ([Postgres docs](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQFLUSH)).

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -53,6 +53,7 @@ Psycopg 3.1.19 (unreleased)
 - Fix excessive stripping of error message prefixes (:ticket:`#752`).
 - Allow to specify the ``connect_timeout`` connection parameter as float
   (:ticket:`#796`).
+- Improve COPY performance on macOS (:ticket:`#745`).
 
 
 Current release
--- a/psycopg/psycopg/_copy.py
+++ b/psycopg/psycopg/_copy.py
@@ -16,7 +16,7 @@ from typing import Any, Iterator, Type, Tuple, Sequence, TYPE_CHECKING
 from . import pq
 from . import errors as e
 from ._compat import Self
-from ._copy_base import BaseCopy, MAX_BUFFER_SIZE, QUEUE_SIZE
+from ._copy_base import BaseCopy, MAX_BUFFER_SIZE, QUEUE_SIZE, PREFER_FLUSH
 from .generators import copy_to, copy_end
 from ._encodings import pgconn_encoding
 from ._acompat import spawn, gather, Queue, Worker
@@ -199,13 +199,15 @@ class LibpqWriter(Writer):
         if len(data) <= MAX_BUFFER_SIZE:
             # Most used path: we don't need to split the buffer in smaller
             # bits, so don't make a copy.
-            self.connection.wait(copy_to(self._pgconn, data))
+            self.connection.wait(copy_to(self._pgconn, data, flush=PREFER_FLUSH))
         else:
             # Copy a buffer too large in chunks to avoid causing a memory
             # error in the libpq, which may cause an infinite loop (#255).
             for i in range(0, len(data), MAX_BUFFER_SIZE):
                 self.connection.wait(
-                    copy_to(self._pgconn, data[i : i + MAX_BUFFER_SIZE])
+                    copy_to(
+                        self._pgconn, data[i : i + MAX_BUFFER_SIZE], flush=PREFER_FLUSH
+                    )
                 )
 
     def finish(self, exc: BaseException | None = None) -> None:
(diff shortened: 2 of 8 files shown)


dvarrazzo:
Errors don't need to be accounted for because `PGconn.flush()` will raise an exception if `PQflush()` returns -1.

https://github.com/psycopg/psycopg/blob/558cbe7ef7ef6a6d1f6c65e368aaaf9f9a10108e/psycopg/psycopg/pq/pq_ctypes.py#L588

dvarrazzo:
Thank you for your MR. I need to review the whole copy pipeline because there were several considerations we made in the past (around buffering, for instance) that apparently don't hold all the water I thought they did and maybe there is code around that that can be removed.

dvarrazzo:
Hello,

on Linux, I struggle to see any difference between using or not using flush. But I do accept that on macOS it might be different.

I have added a minimal framework that can be used to test copy performance. Could you please give it a try and report what sort of time difference you have with or without your patch, on your platform? 

Can you also report if you are testing on a network connection or if your database is on localhost?

Feel free to make any change you think useful to the test script. Thank you very much!

ankane:
Hi @dvarrazzo, I created a [CI workflow](https://github.com/ankane/psycopg-repro2/blob/master/.github/workflows/build.yml) to show the difference (and can be used to test). The database is on localhost.

Mac - [CI logs](https://github.com/ankane/psycopg-repro2/actions/runs/8962145231/job/24610642251)

```sh
tests/scripts/copytest.py --nrecs 1000000
```

- master: `time to copy: 22.962515 sec`
- PR: `time to copy: 1.659219 sec`

Linux - private runner - [CI logs](https://github.com/ankane/psycopg-repro2/actions/runs/8962145231/job/24610642397)

```sh
tests/scripts/copytest.py --colsize 200 --nrecs 1000000
```

- master: `time to copy: 26.573821 sec`
- PR: `time to copy: 17.158544 sec`

Linux - public runner - [CI logs](https://github.com/ankane/psycopg-repro2/actions/runs/8962145231/job/24610774796)

```sh
tests/scripts/copytest.py --colsize 200 --nrecs 1000000
```

- master: `time to copy: 11.504595 sec`
- PR: `time to copy: 12.674765 sec`

For some reason, the PR is faster than master on CI when the repo is private, but slightly slower on CI when the repo is public.

dvarrazzo:
Thank you @ankane. It looks like the result is pretty clear on macOS, but less so on Linux. You also had results where the performance on linux seemed improved in previous version of you message, right?

For me the right target is 3.1.18 because I would like to release a change in 3.1.19 if it's not excessively invasive.

Maybe we can flush at every row for macOS only. I wonder what we should do for Windows. Is it easy to add it to your workflow?

ankane:
That seems reasonable. Another approach that might work for all platforms would be to keep track of the buffer size and only flush once it hits a certain threshold (so it's not doing it every row).

Here are the [CI results](https://github.com/ankane/psycopg-repro2/actions/runs/8973183518/job/24642772251) for Windows:

```sh
python tests/scripts/copytest.py --colsize 100 --nrecs 1000000
```

- w/ patch: 17.495542 sec
- w/o patch: 16.414442 sec

dvarrazzo:
Ok, given these numbers I will merge this MR with flush only activated on macOS. We can tweak it further in the future, but it seems that at the moment we are hitting some quadratic behaviour on macOS and the most important thing is to avoid them.

ankane:
Great, sounds good.

ankane:
Thanks @dvarrazzo!

Issue #534, opened 2023-03-28
non-blocking cancellation


dlax:
Postgres ~~16~~ 17 might include a libpq API for non-blocking cancellation: https://commitfest.postgresql.org/37/3511/

This seems interesting to me, and I wonder if this would help with respect to various issues related to cancellations that occurred some times ago (ctrl+c, etc.).

I have prepared a ~~["pg16/non-blocking-pqcancel" branch](https://github.com/dlax/psycopg3/tree/pg16/non-blocking-pqcancel)~~ ["pg17/non-blocking-pqcancel" branch](https://github.com/dlax/psycopg3/tree/pg17/non-blocking-pqcancel) on my fork to include this (uncommitted) work into Psycopg; the result is that `AsyncConnection.cancel()` becomes `async`, so it's a breaking change.

Obviously, we'd wait until this eventually lands in Postgres ~~16~~ 17. I just wanted to share this work in some way and possibly discuss the idea a bit.

dvarrazzo:
Hey, I just happened to read your message on -hackers, and that's great! I also took a look at your code.

It would be sad to introduce this change only in psycopg 4. It would be better to add an `async AsyncConnection.acancel()` function on the Async connection, and to make `AsyncConnection.cancel()` raise a warning when used. We could release it in a future psycopg 3.x.

Before PG 16 is released (or in case this libpq feature doesn't make the cut), we could actually implement `acancel()` calling `PQcancel()` in a thread executor and release at least this in psycopg 3.2, so that the interface should be in place.

dlax:
> It would be sad to introduce this change only in psycopg 4. It would be better to add an `async AsyncConnection.acancel()` function on the Async connection, and to make `AsyncConnection.cancel()` raise a warning when used. We could release it in a future psycopg 3.x.

Yes, we can do that, sure. But, in fact, maybe we can keep `AsyncConnection.cancel()` implemented as is it now so as to keep something that can be used within a signal handler (the new libpq API is not supposed to be used in that case)?

dvarrazzo:
> maybe we can keep AsyncConnection.cancel() implemented as is it now so as to keep something that can be used within a signal handler

Sure, I didn't think about this use case. We can document that `cancel()` is blocking on `AsyncConnection` and add a `acancel()` method on a future feature release.

dlax:
Patches have been committed in PostgreSQL; so I'm resuming my prior work for integration in Psycopg.

dlax:
[...]

PR #754, opened 2024-03-29
add libpq interface for encrypted and non-blocking cancellation

dlax:
First step for #534


psycopg_c/psycopg_c/pq/libpq.pxd
```diff
@@ -326,9 +339,19 @@ typedef enum {
 #endif
 
 #if PG_VERSION_NUM < 170000
+typedef struct pg_cancel_conn PGcancelConn;
```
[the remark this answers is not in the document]
dlax: Took a little while for me to sort this out, but, yes, it seems to work :)

psycopg/psycopg/pq/_pq_ctypes.py
```diff
@@ -593,6 +598,135 @@ def PQsendClosePortal(pgconn: PGconn_struct, name: str) -> int:
...
+    _PQcancelFinish.restype = None
+
+
+def PQcancelCreate(pgconn: PGconn_struct) -> PGcancelConn_struct:
+    if not _PQcancelCreate:
+        raise NotSupportedError(
+            "PQcancelCreate requires libpq from PostgreSQL 17,"
+            f" {libpq_version} available instead"
+        )
+    return _PQcancelCreate(pgconn)  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
dvarrazzo: Actually, the "raise unsupported" was becoming a nuisance, so in the same MR I've added further refactoring to remove that pattern. WDYT?

tests/pq/test_pgconn.py
```diff
@@ -389,6 +393,100 @@ def test_set_single_row_mode(pgconn):
     pgconn.set_single_row_mode()
 
 
+@contextlib.contextmanager
+def cancellable_query(pgconn: PGconn, monitor_conn: PGconn) -> Iterator[None]:
+    pgconn.send_query_params(b"SELECT pg_sleep($1)", [b"180"])
```
dvarrazzo, line 398: Wouldn't be better to define this helper using `psycopg.Connection`? If you did it this way because you have found most of the tests in the tests/pq suite using only libpq objects, it's because when they were written psycopg wasn't implemented yet :smile:

tests/pq/test_pgconn.py
```diff
@@ -389,6 +393,100 @@ def test_set_single_row_mode(pgconn):
...
+    while pgconn.is_busy():
+        pgconn.consume_input()
+
+
+@pytest.mark.libpq(">= 17")
+def test_cancel_nonblocking(dsn):
+    # mimic test_cancel() from src/test/modules/libpq_pipeline/libpq_pipeline.c
+    def connect() -> PGconn:
+        conn = pq.PGconn.connect(dsn.encode())
+        if conn.status != pq.ConnStatus.OK:
```
dvarrazzo, line 426: assert?

tests/pq/test_pgconn.py
```diff
@@ -389,6 +393,100 @@ def test_set_single_row_mode(pgconn):
...
+    # test PQcancelBlocking
+    with cancellable_query(conn, monitor_conn):
+        cancel_conn = conn.cancel_conn()
+        assert cancel_conn.status == pq.ConnStatus.ALLOCATED
+        cancel_conn.blocking()
+        assert cancel_conn.status == pq.ConnStatus.OK  # type: ignore[comparison-overlap] # noqa: E501
+    cancel_conn.finish()
+    del cancel_conn
+
+    def wait(cancel_conn: PGcancelConn) -> None:
```
dvarrazzo, line 455: Don't we have something similar to this already in the test suite that we can refactor/reuse?

tests/pq/test_pgconn.py
```diff
@@ -389,6 +393,100 @@ def test_set_single_row_mode(pgconn):
...
+        ), f"unexpected cancel connection status: {cancel_conn.error_message}"
+
+    # test PQcancelCreate and then polling with PQcancelPoll
+    with cancellable_query(conn, monitor_conn):
+        cancel_conn = conn.cancel_conn()
+        assert cancel_conn.status == pq.ConnStatus.ALLOCATED
+        cancel_conn.start()
+        assert cancel_conn.status == pq.ConnStatus.STARTED
+        wait(cancel_conn)
+        assert cancel_conn.status == pq.ConnStatus.OK
```
dvarrazzo, line 477: Wouldn't be better to have several tests instead of a long one? cancellable_query might be a local fixture.

dvarrazzo (changes requested):

psycopg/psycopg/pq/_enums.py
```diff
@@ -31,6 +31,7 @@ class ConnStatus(IntEnum):
     GSS_STARTUP = auto()
     CHECK_TARGET = auto()
     CHECK_STANDBY = auto()
+    ALLOCATED = auto()  # Only for cancel connections.
```
dvarrazzo, line 34: Something broke in the docs generation (in the horrible `docs/lib/pg3_docs.py`) a while ago: the "docstring" style constants were published (e.g. [here](https://www.psycopg.org/psycopg3/docs/api/pq.html#psycopg.pq.ConnStatus.OK)).

However, the libpq declares that `ALLOCATED` is visible, therefore I would document it consistently:

```suggestion
    ALLOCATED = auto()
    """Connection to the server hasn't been initiated yet."""
```

The blurb:

https://github.com/psycopg/psycopg/blob/cae87c4cf6223cc278f36759ea8709cc628ac7f6/docs/api/pq.rst?plain=1#L148-L150

might also require adjustment, for instance:

```
    There are other values in this enum, but only `ALLOCATED`, `OK`, `BAD` 
    are seen after a connection has been established. Other statuses might
    only be seen during the connection phase and are considered internal.
    `!ALLOCATED` is only expected to be returned by `PGcancelConn.status`.
```

docs/api/pq.rst
```diff
@@ -149,7 +149,7 @@ Enumerations
     after a connection has been established. Other statuses might only be seen
     during the connection phase and are considered internal.
 
-    .. seealso:: :pq:`PQstatus()` returns this value.
+    .. seealso:: :pq:`PQstatus()` and `PQcancelStatus()` return this value.
```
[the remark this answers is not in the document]
dlax: I did in another commit not pushed here because we'd need to change `libpq_docs_version = "devel"` for the docs to build; if that's okay for you (considering this would affect published docs I think by linking to https://www.postgresql.org/docs/devel/), I can add the commit here.

psycopg/psycopg/pq/_pq_ctypes.py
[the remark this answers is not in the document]
dlax: > How about specifying the return type as in https://github.com/psycopg/psycopg/pull/768?

I have mixed feelings about that; I think the explicit `# type: ignore[no-any-return]` indicate more clearly that there is a limitation somewhere (on our side or in mypy) and that it might go away someday. However I can live with the approach you suggest, especially if it's more consistent with the rest of the code base (which seems true, as evidenced by `git grep no-any-return` vs `git grep "rv: "`).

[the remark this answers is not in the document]
dlax: Hm, reading the next commits in #768 makes my point vanish; so no objection at all :)

tests/pq/test_pgconn.py
dlax: It's a copy-paste from: https://github.com/psycopg/psycopg/blob/cae87c4cf6223cc278f36759ea8709cc628ac7f6/tests/fix_db.py#L192-L194

tests/pq/test_pgconn.py
dlax: Yes we do, in this very file, in `test_connect_async()`, `test_connect_async_bad()` and `test_reset_async()` ! I'll try to factor out a common helper in a follow-up.

psycopg/psycopg/pq/_pq_ctypes.py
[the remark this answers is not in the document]
dlax: I rebased on now-master so the code now looks like #768.

tests/pq/test_pgconn.py
dlax: Fixed in 41e8b3a628805b4862c6d2bbace87c7cbcd72491

tests/pq/test_pgconn.py
dlax: Improved in 41e8b3a628805b4862c6d2bbace87c7cbcd72491.

psycopg/psycopg/pq/_enums.py
dlax: Done as well.

tests/pq/test_pgconn.py
dlax: Done in a preparatory commit 005e9b1bc593fc003b8cdb241b323bc5d55d85be

dvarrazzo (approved):
Looks good! Rebasing and merging, thank you very much!

dvarrazzo:
Merged now :slightly_smiling_face:

Issue #756, opened 2024-03-29
Notifies not generated if consumed in other thread


malthe:
The same connection `conn` is used in two threads:

1. The first thread waits for `conn.notifies()`.
2. The second thread runs a statement that triggers a notification.

You would expect that the first thread would receive the notification, but that's not the case – the notifies are consumed and distributed only to handlers registered using `conn.add_notify_handler`.

When notifies are consumed, the select/polling happening as part of `conn.notifies()` will cause a loop-around, but we won't see any notifications and will just resume waiting. What should happen is that any notification received on the connection must be distributed to open `notifies` generators.

A `notifies` generator should be considered open when `conn.notifies()` returns (iterating should not be required). When iteration completes (due to a `timeout` or `stop_after` setting), the generator can be considered closed.

dvarrazzo:
> You would expect that the first thread would receive the notification, but that's not the case – the notifies are consumed and distributed only to handlers registered using conn.add_notify_handler.

I don't think that yours is the right expectation. Most likely the notifications are returned together with the return message to the second thread, therefore the first one doesn't get to see them. I would call it a race condition, except that the first thread is likely bound to predictably lose the race.

I would say don't do that.

malthe:
@dvarrazzo since I had a draft changeset ready to make this change, I decided to push it for review – that said, obviously feel free to close this pull request given that you've already stated your view on the current behavior.

I think the proposed behavior is better (more correct, less surprising), but it can be seen in different ways. I'm incorporating your library into https://github.com/malthe/pq and was hoping that I could test things using just a single connection.

dvarrazzo:
Thank you for the MR! Happy to review It and open to change my mind if it
makes sense 🙂

PR #757, opened 2024-03-30
Ensure that notifies include those received during concurrent fetch and pipeline processing

malthe:
This fixes issue #756.

This change addresses a race condition with the concurrent use of the `notifies` generator with fetch and/or pipeline communications in other threads.

The proposed behavior should be less surprising since it streamlines how notifies are distributed – now always through the notify handlers.

When would you rely on this behavior? The issue came up writing test code where a single connection was used to both generate and receive notifications, motivated by the thread-safe nature of the connection object. I think the new behavior aligns better with this promise.

### DIFF
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -11,6 +11,7 @@ from time import monotonic
 from types import TracebackType
 from typing import Any, AsyncGenerator, AsyncIterator, List, Optional
 from typing import Type, Union, cast, overload, TYPE_CHECKING
+from collections import deque
 from contextlib import asynccontextmanager
 
 from . import pq
@@ -25,7 +26,6 @@ from ._compat import Self
 from .conninfo import make_conninfo, conninfo_to_dict
 from .conninfo import conninfo_attempts_async, timeout_from_conninfo
 from ._pipeline import AsyncPipeline
-from ._encodings import pgconn_encoding
 from .generators import notifies
 from .transaction import AsyncTransaction
 from .cursor_async import AsyncCursor
@@ -343,33 +343,35 @@ class AsyncConnection(BaseConnection[Row]):
 
         nreceived = 0
 
-        while True:
-            # Collect notifications. Also get the connection encoding if any
-            # notification is received to makes sure that they are consistent.
-            try:
-                async with self.lock:
-                    ns = await self.wait(notifies(self.pgconn), interval=interval)
-                    if ns:
-                        enc = pgconn_encoding(self.pgconn)
-            except e._NO_TRACEBACK as ex:
-                raise ex.with_traceback(None)
+        ns: deque[Notify] = deque()
+        self.add_notify_handler(ns.append)
+        try:
+            while True:
+                # Collect notifications. Also get the connection encoding if any
+                # notification is received to makes sure that they are consistent.
+                try:
+                    async with self.lock:
+                        await self.wait(notifies(self.pgconn), interval=interval)
+                except e._NO_TRACEBACK as ex:
+                    raise ex.with_traceback(None)
 
-            # Emit the notifications received.
-            for pgn in ns:
-                n = Notify(pgn.relname.decode(enc), pgn.extra.decode(enc), pgn.be_pid)
-                yield n
-                nreceived += 1
+                # Emit the notifications received.
+                while ns:
+                    yield ns.popleft()
+                    nreceived += 1
 
-            # Stop if we have received enough notifications.
-            if stop_after is not None and nreceived >= stop_after:
-                break
-
-            # Check the deadline after the loop to ensure that timeout=0
-            # polls at least once.
-            if deadline:
-                interval = min(_WAIT_INTERVAL, deadline - monotonic())
-                if interval < 0.0:
+                # Stop if we have received enough notifications.
+                if stop_after is not None and nreceived >= stop_after:
                     break
+
+                # Check the deadline after the loop to ensure that timeout=0
+                # polls at least once.
+                if deadline:
+                    timeout = min(_WAIT_INTERVAL, deadline - monotonic())
+                    if timeout < 0.0:
+                        break
+        finally:
+            self.remove_notify_handler(ns.append)
 
     @asynccontextmanager
     async def pipeline(self) -> AsyncIterator[AsyncPipeline]:
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -347,11 +347,10 @@ class AsyncConnection(BaseConnection[Row]):
         self.add_notify_handler(ns.append)
         try:
             while True:
-                # Collect notifications. Also get the connection encoding if any
-                # notification is received to makes sure that they are consistent.
+                # Wait for notifications, collecting and processing them using
+                # thread-safe operations.
                 try:
-                    async with self.lock:
-                        await self.wait(notifies(self.pgconn), interval=interval)
+                    await self.wait(notifies(self.pgconn), interval=interval)
                 except e._NO_TRACEBACK as ex:
                     raise ex.with_traceback(None)
 
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -7,6 +7,7 @@ Psycopg connection object (async version)
 from __future__ import annotations
 
 import logging
+from socket import close, dup
 from time import monotonic
 from types import TracebackType
 from typing import Any, AsyncGenerator, AsyncIterator, List, Optional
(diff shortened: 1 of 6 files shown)


psycopg/psycopg/connection_async.py
```diff
@@ -319,34 +319,36 @@ async def notifies(
...
-            # polls at least once.
-            if deadline:
-                timeout = min(_WAIT_INTERVAL, deadline - monotonic())
-                if timeout < 0.0:
+        ns: deque[Notify] = deque()
+        self.add_notify_handler(ns.append)
+        try:
+            while True:
+                # Collect notifications. Also get the connection encoding if any
+                # notification is received to makes sure that they are consistent.
```
dvarrazzo, line 327: This comment is now stale: we don't get the encoding anymore here.

It would be also worth mentioning the reason why we want to generate notification in this slightly roundabout way, receiving them from handler rather than from as return value.

```suggestion
                # Collect notifications. Also get the connection encoding if any
                # notification is received to makes sure that they are consistent.
                # Collect notifications. Note that we don't use the return
                # value of the generator but we rely on the handler in order
                # to collect notifications that might have been generated and
                # consumed by another concurrent thread.
```

psycopg/psycopg/generators.py
```diff
@@ -251,28 +253,21 @@ def _pipeline_communicate(
...
-    # Consume notifies
-    while True:
-        n = pgconn.notifies()
-        if not n:
-            break
-        if pgconn.notify_handler:
-            pgconn.notify_handler(n)
+    # Consume notifies. The lock ensures that notifies
+    # are consumed and handled atomically.
+    with lock:
```
dvarrazzo, line 258: This is absolutely not ok. This lock is application-global and not even asyncio-friendly.

Why is a lock needed here anyway?

psycopg/psycopg/generators.py
```diff
@@ -251,28 +253,21 @@ def _pipeline_communicate(
...
+        while True:
+            n = pgconn.notifies()
+            if not n:
+                break
+            if pgconn.notify_handler:
+                pgconn.notify_handler(n)
 
 
-def notifies(pgconn: PGconn) -> PQGen[List[pq.PGnotify]]:
+def notifies(pgconn: PGconn) -> PQGen[None]:
```
dvarrazzo, line 267: Even if the main user of this function doesn't make use of this return value anymore, I don't see a reason to change this function's signature.

tests/test_notify_async.py
```diff
@@ -71,14 +71,16 @@ async def receiver():
         gen = aconn.notifies()
         async for n in gen:
             ns.append((n, time()))
-            if len(ns) >= 2:
-                await gen.aclose()
+            if len(ns) == 2:
+                await aconn.execute("notify foo, '3'")
+            if len(ns) == 3:
+                break
```
dvarrazzo, line 77: I don't get this test. It doesn't seem to test the feature you introduced.

Can you demonstrate in a test that notifications generated in a concurrent thread or task are actually collected by the receiving generator?

dvarrazzo (changes requested):
I have marked some changes I would introduce. However, I'm really not convinced about this changeset and how much we can promise about using a connection while listening for notifications at the same time.

The changes introduced in 3.2 to the notifies generator are already designed to allow to use a connection for some query run as well as to consume notification (see #340), although with client coordination rather than relying on thread preemption. While the generator is waiting, we actually want to use the connection in an exclusive way: the intent is clearly stated by the `self.lock` context. Using a connection in exclusive way during a query is a consistent pattern of usage of psycopg: you can use two cursor concurrently but the querying process can only be sequential and is serialized by `self.lock`. You cannot run a query while running `copy()`, etc.

The reason you can sneak a concurrent query while `notifies()` is executed is only because the polling from signal; if that interruption wasn't there, your concurrent `execute()` would have to wait until `timeout`, or would be stuck forever if `timeout=None`. So the way your feature is working seems more or less an implementation accident. If any there should be a test to prove that you can run a query timely while a `notifies()` with a long timeout is running. However, because this happens as an implementation accident, I don't really want to commit to this feature.

Another side effect is that running any concurrent `execute()` incurs a penalty because it will have to wait an average of `_WAIT_INTERVAL` / 2 in order for the query to wait for the lock in `notifies()` to be temporarily released.

So all in all, this seems a promise we can hardly keep, unless incurring in a performance degradation on the connection used that probably users wouldn't be willing to pay.

psycopg/psycopg/generators.py
malthe: The lock is there to satisfy that in `notifier`, if `wait(...)` completes then either notification processing has happened either:

1) in the `notifier` thread
2) in another thread

But not partially in one thread, partially in another – because this could mean that the `notifier` thread never actually gets to yield notifications but just goes back to sleep, waiting.

That said, you're absolute right that this global lock isn't workable. There's also the `with self.lock` around `wait(...)` that I'm a little unsure of – why is this one really necessary?

I don't consider it to be an implementation accident in the sense that I think `notifier` should work just like `add_notify_handler` which is as a kind of observable: as soon as the database sends the notification, we can pick it up either by registering a handler, or using `notifier` (or both).

But taking a step back, I think we could probably fix the code so that the implementation is solid, but at the same time, we should only take the time to do that if we can agree that it's the right behavior, worth committing to. I think we can avoid the lock that causes the delay you mention so that should probably not tip the boat.

malthe: Just a quick note to say that `pgconn.consume_input` which calls into `PQconsumeInput` does not block:

```c
int
PQconsumeInput(PGconn *conn)
...
  /*
  * Load more data, if available. We do this no matter what state we are
  * in, since we are probably getting called because the application wants
  * to get rid of a read-select condition. Note that we will NOT block
  * waiting for more input.
  */
  if (pqReadData(conn) < 0)
    return 0;
```
(Reformatted to fit constraints here.)

That is, we can have two threads suspended waiting for the same socket to become read-ready and it's okay that both proceed to `pgconn.consume_input` in the sense that this won't cause any unexpected blocking.

malthe: I think perhaps we should consider a `pgconn.consume_notifies() -> bool` instead – which then atomically (using a lock specific to this particular `pgconn`):

1. Retrieve a `PQnotifies`
2. Calls the `notify_handler` (if defined)

(Alternatively, the function could also do the `while True` loop, pulling out all notifications. In this case, the method doesn't have a return value).

The reason why it would seem we can't use the connection lock is that the `_consume_notifies` call is sort of intermingled between other blocks, since notifies are transmitted before results (apparently).

malthe: I have reworked the implementation now, having better understood the requirements.

In `connection.py` I have added a _read lock_ which works to synchronize the program flow after a select on a file descriptor.

If the notifies thread holds the read lock, then an active query that resumes after a select that wants `READY_R` must wait to acquire this read lock – which is released as soon as the notifies thread has consumed the input and processed notifies.

Conversely, if the active query is able to acquire the read lock, the notifies thread simply waits for the query to complete before continuing its loop.

In the common case, where a connection is used exclusively by a single thread, there is virtually no cost: an uncontended lock is practically free to acquire and release.

As an implementation note for the asynchronous case, I did have to `socket.dup` the file descriptor inside `notifies()` in order for the asynchronous event loop to correctly fire (to achieve the same behavior as `select.select` where every blocked thread wakes up).

psycopg/psycopg/generators.py
malthe: There is no return value now – so _mypy_ would complain if we provided anything else.

tests/test_notify_async.py
malthe: I have improved and expanded this test case now (also adding an additional case to test the behavior during a long-running query).

psycopg/psycopg/connection_async.py
malthe: Resolving this comment because it's been supplanted by more recent changes.

malthe:
@dvarrazzo having reworked this substantially, if you have some time for a second look it would be great.

There's some more detail in the comment I [posted above](https://github.com/psycopg/psycopg/pull/757#discussion_r1559774785).

dvarrazzo:
I Will review It again. Thank you very much!

dvarrazzo:
I am sorry but I really don't feel comfortable to accept this changeset.

The use case you are proposing to cover is marginal: using the same connection both to generate notifications and to send command. This is not what the connection is designed for, and the need to using `dup()` on the file descriptor is a demonstration that the design doesn't integrate smoothly.

The addition of a second lock is also something I'd be extremely wary of. I haven't measured the happy path and I should take face-value your assertion that it has negligible performance impact. But two locks are a higher risk of deadlock and I have no experience of how the two would behave in uncommon conditions.

These are extremely deep architectural changes and affect every single query executed just to allow the feature of using the connection while listening for notifications. I think that the trade-off is not a net gain: it increases the maintenance complexity and introduce possible failure modes whereas the same problem can be resolved easily in a client program. Or just by using two connections.

malthe:
Just a couple of notes here:

The need to `socket.dup` in the asynchronous case is due to the design of the [asyncio event loop](https://docs.python.org/3/library/asyncio-eventloop.html) which allows just a single callback per file descriptor (registered using the `add_reader` or `add_writer` methods).

It's not ideal, but it's also not really a problem: setting up a notifier loop should not happen in a tight loop and even so, the `socket.dup` call is cheap.

Are we adding a lot of complexity? Perhaps, but the concept of the read lock is pretty intelligible I think: you can read only if you've got the read lock, while you can make a query if you have the connection lock.

I don't see how a would-be deadlock could materialize, but it's hard to make such guarantees of course besides pointing to the following:
```python
...
    def wait(self, ...
        acquired = self._read_lock.acquire(False)
        if not acquired:
            gen = read_guard(self._read_lock, gen, True)

```
In the happy path, the read lock will always be acquired (with trivial cost) and so there's no `read_guard` involved. This only comes into play if a notifier has already been established first.

That's an argument to say that this proposed functionality does not affect users that aren't using both modes concurrently.

PR #760, opened 2024-04-01
fix: lock the connection during a 'notifies()' call

dvarrazzo:
With the previous implementation, it was possible to sneak an execute() while the generator is consumed. This gives the false impression that it's possible to use the connection while listening (see #756), which is false for reason better explored in #757.

Therefore, lock the connection while listening to notifications. If someone wants to mix commands with listening on the same connection, they should do it collaboratively with an adequately short notifies() timeout.


dlax (approved):
Changes, and the reasoning behind, look good to me.

PR #763, opened 2024-04-03
encrypted and non-blocking cancellation

### LINKED ISSUE #534: non-blocking cancellation

dlax:
~~based on #754 (first commit); implements the proposal from https://github.com/psycopg/psycopg/issues/534#issuecomment-2034747801~~

fix #534

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -35,6 +35,10 @@ Psycopg 3.2 (unreleased)
   in non-pipeline mode and not totally reliable (:ticket:`#604`).
   The `Cursor` now only preserves the results set of the last
   `~Cursor.execute()`, consistently with non-pipeline mode.
+- Add `~Connection.cancel_safe()` for encrypted and non-blocking cancellation
+  (:ticket:`#754`).
+- Possibly use non-blocking cancellation upon `KeyboardInterrupt`
+  (:ticket:`#754`).
 
 .. __: https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -36,11 +36,8 @@ Psycopg 3.2 (unreleased)
   The `Cursor` now only preserves the results set of the last
   `~Cursor.execute()`, consistently with non-pipeline mode.
 - Add `~Connection.cancel_safe()` for encrypted and non-blocking cancellation
-  (:ticket:`#754`).
-- Possibly use non-blocking cancellation upon `KeyboardInterrupt`
-  (:ticket:`#754`).
-- Possibly use non-blocking cancellation upon `Copy` termination
-  (:ticket:`#754`).
+  If possible, use such method internally upon `KeyboardInterrupt` and `Copy`
+  termination (:ticket:`#754`).
 
 .. __: https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -39,6 +39,8 @@ Psycopg 3.2 (unreleased)
   (:ticket:`#754`).
 - Possibly use non-blocking cancellation upon `KeyboardInterrupt`
   (:ticket:`#754`).
+- Possibly use non-blocking cancellation upon `Copy` termination
+  (:ticket:`#754`).
 
 .. __: https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types
 
--- a/psycopg/psycopg/_connection_base.py
+++ b/psycopg/psycopg/_connection_base.py
@@ -293,21 +293,29 @@ class BaseConnection(Generic[Row]):
 
     def cancel(self) -> None:
         """Cancel the current operation on the connection."""
-        # No-op if the connection is closed
+        if self._should_cancel():
+            self._try_cancel(self.pgconn)
+
+    def _should_cancel(self) -> bool:
+        """Check whether the current command should actually be cancelled when
+        invoking cancel*().
+        """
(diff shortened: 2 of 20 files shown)


psycopg/psycopg/_connection_base.py
```diff
@@ -313,6 +321,15 @@ def _try_cancel(cls, pgconn: "PGconn") -> None:
         else:
             c.cancel()
 
+    def _cancel_gen(self) -> PQGenConn[None]:
+        try:
+            cancel_conn = self.pgconn.cancel_conn()
+        except e.OperationalError as ex:  # if the connection is closed
+            logger.warning("couldn't create a cancel connection: %s", ex)
```
dvarrazzo, line 328: Why to catch an exception, and assume that it was caused by the connection closed? We can just check if the connection is close if we want to verify that. It actually seems that we do it already, in virtue of the fact that all the calls to this method are guarded by `_should_cancel()`.

I don't see a strong reason to catch exceptions here. I don't think you want to catch *any* exception, because you don't catch them in the two statements below. If someone, or ourselves in certain case, want to just try to send a cancel as a best-effort thing but ignore failure I think they should catch the exception at that level.

dvarrazzo: Looked more around and I understand that this pattern was largely copied from `_try_cancel()` which received a similar guard in b3e0be9869257cfa9602608caee4ebe96148f63f.

I think that my refactoring in that comment was somewhat misguided: the call of `_try_cancel()` from `cancel()` is already guarded by a check for connection closed (and the comment is wrong, I assume I meant signal handler, not callback handler). So I don't think I meant to make that change for the benefit of `close()`, rather I think I wanted to make the other call to `_try_cancel()` (the one in the interruption of `wait()`) a best-effort thing.

How it happens that after a Ctrl-C the connection is closed is not entirely clear to me... But I can't remember any better.

To try and recap, I don't think that `cancel()` or `cancel_safe()` should have a built-in policy of doing cancelling as as a best-effort thing. I think that these methods should just raise an exception if they fail, and in specific places we might want to ignore that exception raised and to just log a warning. So, in the code affected by b3e0be9869257cfa9602608caee4ebe96148f63f, it would be in `wait()`, but not in `cancel()`, and I haven't looked on a wider range (I haven't checked if in that moment `cancel()` was being used in other operations).

What do you think of reviewing the calls to ensure:

- `cancel()` to bail out early for closed connection, raise whatever exception
- `cancel_safe()` to do the same
- `wait()` to guard for exceptions on `close()`, log and proceed
- other uses to review.

I think that the guideline should be that if a cancelling fails when we are handling an exception, we should log the episode and continue, especially when the original exception wanted to be re-raised, rather than clobbering it with the cancelling error.

dvarrazzo:
I have pushed a docs commit. With that changeset now the docs look like:

image

Does it look ok for you?

psycopg/psycopg/_connection_base.py
dlax: What you suggest above is not entirely clear to me; I would agree that `cancel()` and `cancel_safe()` should behave similarly w.r.t. failures to create the underlying `PQcancel` or `PQcancelConn` object, which is why I included the exception handling here, indeed. Similarly, if something needs to be changed, it must be done consistently on both sides; but can we do this in another PR?

I can remove the `except e.OperationalError` right now though. Let me know.

docs/news.rst
```diff
@@ -33,6 +33,12 @@ Psycopg 3.2 (unreleased)
   in non-pipeline mode and not totally reliable (:ticket:`#604`).
   The `Cursor` now only preserves the results set of the last
   `~Cursor.execute()`, consistently with non-pipeline mode.
+- Add `~Connection.cancel_safe()` for encrypted and non-blocking cancellation
+  (:ticket:`#754`).
+- Possibly use non-blocking cancellation upon `KeyboardInterrupt`
+  (:ticket:`#754`).
+- Possibly use non-blocking cancellation upon `Copy` termination
+  (:ticket:`#754`).
```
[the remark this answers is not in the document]
dlax: Fixed this, along with the signature of `to_thread()` below.

dlax:
> Does it look ok for you?

LGTM.

psycopg/psycopg/_connection_base.py
dvarrazzo: Opened #778 to look into that.

dvarrazzo:
I have opened #778 and #779 to address the points open: reviewing the exception swallowing behaviour (to be backported to 3.1) and introducing a timeout. This way we can merge this state now.

Rebasing and merging. Thank you very much for the amazing work, @dlax!

dlax:
> I have opened #778 and #779 to address the points open: reviewing the exception swallowing behaviour (to be backported to 3.1) and introducing a timeout. This way we can merge this state now.

Sounds good; thank you!

PR #764, opened 2024-04-03
disable flake8's line length check

dlax:
This lets black solely handle the line length when formatting the code. In particular, this allows for longer string literals as black would produce when generating code from tools/async_to_sync.py (even if the original string is broken and concatenated), e.g. in https://github.com/psycopg/psycopg/actions/runs/8540753061/job/23398501922?pr=763

Suggesting this but maybe there is another solution to the issue?


dvarrazzo:
Hello Denis,

I found the issue too using async_to_sync. My way to solve it was to use the `+` operator to split large strings. See here for example:

https://github.com/psycopg/psycopg/blob/cae87c4cf6223cc278f36759ea8709cc628ac7f6/psycopg_pool/psycopg_pool/pool_async.py#L126-L132

I think that for non time-sensitive parts of the code this is an acceptable solution and it effectively generates narrow code rather than a long string of which we ignore the length.

https://github.com/psycopg/psycopg/blob/cae87c4cf6223cc278f36759ea8709cc628ac7f6/psycopg_pool/psycopg_pool/pool.py#L117-L123

What do you think?

dlax:
> What do you think?

In general, I find that having to manually tweak string literals is tedious and I usually give up on this in most projects I work on by just ignoring E501 or setting flake8's `max-line-length` to `120` or so (while keeping black's limit). Leaving this to developer's (or reviewer's) judgement seems reasonable to me.

I can live with the workaround you suggest, though; but I will probably forget about it next time I'll stubble upon this with the generated code.

dvarrazzo:
Yes, having to use redundant string concatenation operators is not pretty. However, looking at the current flake8 configuration, we can use per-file ignores. AFAICS there was only one file where this was necessary, so I don't expect these config files to be a big chore to maintain.

So, what do you think about #767 as an alternative for this changeset? If merged, you can add a per-file ignore for the autogenerated file in `.flake8` and `psycopg/.flake8` to ignore your long warnings.

dlax:
I think I will also probably forget about this per-file rules the next time I'll stubble upon this.

dlax:
Sorry to insist, but, turning the problem around, what's the value of the `max-line-length` rule by flake8 considering that black already does "the job"? Maybe I'm missing something, and if so I'd be happy to learn about.

dvarrazzo:
> Sorry to insist, but, turning the problem around, what's the value of the max-line-length rule by flake8 considering that black already does "the job"? Maybe I'm missing something, and if so I'd be happy to learn about.

Black won't reformat strings or comments if they are too long. With the current setup there is a warning:

image

Disabling the flake8 warning will only remove the warning, but will leave the line too large to be practical.

image

If the question becomes "why do you say that large lines are unpractical, haven't you bought a 132 column printer yet?", some of my answers are:

- for side-by-side diff in a reasonable size screen, for instance [this MR](https://github.com/psycopg/psycopg/pull/767/files#diff-57b26a96183c435163abf36561884639eb2df77a0ce05da68f469787d5f02df6). I'm ok with large lines in these files just because they are autogenerated, otherwise I really don't like these types of diffs.

image

- I heard that people buy large monitors but they hang them the wrong way... I prefer this setup, with an 80 cols editor and a bunch of shells.

image

dlax:
Please don't misunderstand me: I'm not advocating for having very long lines and I am definitely in favour of line's width not exceeded ~100 columns.

However, as I mentioned before, I think the way to break or not-break string literals or comments can be left to developer's or reviewer's judgement because no automated tool does the job properly as far as I know (more precisely, black and flake8 together do get in my way in that case). Typically, it's often better to break lines logically, e.g. at end of sentences.

Apart from the example discussed here, another common pain point for me are '`#type: ignore[bla bla bla]` comments, which, if the line is already close to 88 columns, need to be complemented with a '`# noqa: E501`.

Another thing that might be worth having is a `.editorconfig` (https://editorconfig.org/) with:
```
[*.{py,pyx,pxd}]
max_line_length = 78
```
so as to make it a soft rule.

Just trying to share my developer experience, for what it's worth. :)

dvarrazzo:
I find myself working in several projects with different and incompatible styles. I wasn't familiar with EditorConfig, I can take a look at that, thank you for sharing :slightly_smiling_face: 

In this project I am trying to use a standard Python setup, hence I haven't changed Black standard of 88 columns rules, and I'm happy that the linter, flake8, warns about lines too long. I think in the past there was a 79 cols limit ([yes there was](https://github.com/psycopg/psycopg/commit/aecc520b010bc58f86c90ef9c39bc86a3343aa73#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)) and that used to get in the way. The bullet was bitten in 212ad14a7fd60c770d5016586713b7ab7342f18c.

The Python literal concatenation works most of the times to allow to express long strings within the chosen number of columns in hand-edited code. The only thing that gets in the way is autogenerated code. I think it's a pretty uncommon occurrence, so uncommon that, when I introduced async-to-sync, I had forgotten that I used the per-file exclude in the past for another autogenerated section of code (the error classes).

So yes, maybe the third time in 4 years that we stubble around autogenerated code we will have forgotten again about the per-file excludes flake8 feature :smile: but because autogenerated code is pretty uncommon, I prefer to solve that type of problem when we have it rather than disabling line length linting altogether. If feels like throwing away the baby with the bathwater :slightly_smiling_face: Hence my preference for #767

PR #766, opened 2024-04-04
Connection generator refactoring

dvarrazzo:
This MR is a refactoring of the connection and the wait methods.

The connection timeout is now handled in the `connect()` generator rather than in the wait_conn function. The former is more appropriate because it contains all the "policies" to connect, the other is only a mechanism to push the wheels.

Note that the meaning of `timeout` parameter had diverged between `wait()` and `wait_conn()`: in the former the timeout is mostly a poll interval, in the latter it was signifying the connection timeout. Now the meaning is the same. For this reason, in the second commit, the parameter is renamed `interval`.

Note that these changesets were already proposed as part of the `notifies()` method refactoring in 557137dd; however test started failing in CI and the changeset was reverted in 557137dd. Now I found some time to investigate why it was failing and it turned out to be [this erroneous definition](https://github.com/psycopg/psycopg/commit/07da60bedb4aa398100ece1c193f592c750d94f0#diff-4c78d1652a0d0f6518b1452103c4da9c49b1042add006e492fe8b0eb7637cf31R35) of the deadline as an integer. Sanity restored.


dvarrazzo:
@dlax ok to merge these changes? They will affect your `Connection.cancel()` MR.

dlax (approved):

PR #770, opened 2024-04-05
chore: regenerate _pq_ctypes.pyi

dlax:
Following commit a7f280227923ccb5de8928f7335e092531eedea1.


dvarrazzo:
Right, let's merge this. However that stub generation is pretty brutal... with 4 years of experience, now I would do that job parsing the AST, so that the lines protected by `if libpq_version` can be parsed anyway and they don't need to be listed as special cases.

PR #603, opened 2023-07-12
Add Close wrappers to PGconn

JelteF:
This adds wrappers around the Close portal/prepared statement functions that got added in libpq version 17: https://github.com/postgres/postgres/commit/28b5726561841556dc3e00ffe26b01a8107ee654

Related to #589, but this doesn't start using these new functions yet to close prepared statements.

PR #773, opened 2024-04-06
Make prepared statements compatible with PgBouncer

### LINKED ISSUE #589: Using Close protocol message instead of DEALLOCATE to help prepared statements in PgBouncer
PgBouncer does currently not parse queries at all, only protocol messages. But psycopg closes its statements using DEALOCATE sql call:

https://github.com/psycopg/psycopg/blob/63803637e5b20547807bcd632ce2487d871df44e/psycopg/psycopg/_preparing.py#L119

What do you think of about, instead of sending DEALLOCATE text query here, to instead send a Close protocol message? Because that would make it work automatically with the work-in-progress PgBouncer prepared statements patch:  https://github.com/pgbouncer/pgbouncer/pull/845

P.S. The DEALLOCATE ALL usage is similarly problematic, but maybe PgBouncer can make an exception for supporting that since it would be extremely easy to parse. And also it should not be necessary to use as I explained a few minutes ago on the PG mailing list thread that triggered this DEALLOCATE ALL usage (according to git blame): https://www.postgresql.org/message-id/flat/CAGECzQSUPTjfb7BYLg5c22tN4SSuLTRv9-6TpuZTpJ0q8hqp2w%40mail.gmail.com#b9b91ddf3d37160a3e8771e69b27f0ac

dvarrazzo:
This MR introduces compatibility with PgBouncer by using protocol level messages whose libpq interface was added in PostgreSQL 17 and wrapped by psycopg in #603.

PgBouncer supports close messages [from 1.21](https://www.pgbouncer.org/2023/10/pgbouncer-1-21-0) and DEALLOCATE ALL [from 1.22](https://www.pgbouncer.org/2024/01/pgbouncer-1-22-0), therefore we consider the latter as the minimum supported version for the feature.

Close #589


docs/advanced/prepare.rst
```diff
@@ -50,8 +50,30 @@ Statement preparation can be controlled in several ways:
...
+
+.. __: https://www.pgbouncer.org/
+.. __: https://www.pgbouncer.org/2024/01/pgbouncer-1-22-0
+
+.. hint::
+
+    If libpq 17 is not available on your client, but PgBouncer is 1.22 or
+    higher, you can still use Psycopg *as long as you disable deallocation*.
+
+    You can do so by setting `Connection.prepared_max` to `!sys.maxsize`.
```
JelteF, line 79: How about we make `None` mean no maximum (and thus no deallocation)? That feels more natural to me than setting the maximum to some very high number.

docs/advanced/prepare.rst
```diff
@@ -50,8 +50,30 @@ Statement preparation can be controlled in several ways:
...
+
+.. _pgbouncer:
+
+Using prepared statements with PgBouncer
+----------------------------------------
+
+Starting from 3.2, Psycopg supports prepared statements when using the
+PgBouncer__ middleware, using the following caveats:
+
+- PgBouncer version must be at least version `1.22`__.
```
JelteF, line 68:
```suggestion
- PgBouncer version must be at least version `1.22`__.
- PgBouncer its max_prepared_statements configuration should set to a positive number.
```

docs/advanced/prepare.rst
dvarrazzo: That makes sense

docs/advanced/prepare.rst
dvarrazzo: Will reword it a bit, but thank you for the info

dlax (approved):

docs/advanced/prepare.rst
dvarrazzo: Done

PR #774, opened 2024-04-08
document libpq 17 features

dlax:
We bump `libpq_docs_version = "17"` in docs configuration (first commit), as the `REL_17_STABLE` branch is now available in PostgreSQL git repository. And then document features of libpq 17: `PGCancelConn` and `change_password()`.

#534


dvarrazzo:
Looks good, thank you for the docs refresh!

Issue #778, opened 2024-04-09
Review error swallowing in cancel methods


dvarrazzo:
It seems that the `cancel()` method swallows exception in more cases than it should be, and the `cancel_safe()` has adopted the same pattern. ISTM that we are doing it too eagerly.

_See thread and comment by @dlax in https://github.com/psycopg/psycopg/pull/763#discussion_r1557068227_

PR #780, opened 2024-04-09
feat: add a timeout parameter to Connection.cancel_safe()

### LINKED ISSUE #779: Add timeout (with a sensible default) to cancel_safe
As suggested in https://github.com/psycopg/psycopg/issues/534#issuecomment-2037987893 I think that we should have a timeout on cancelling, because with all async methods the libpq expects the client to to that.

@dlax suggested to leave it to our client (https://github.com/psycopg/psycopg/issues/534#issuecomment-2042757578) but I think it's our responsibility to do so, because most time people won't even know that they need a timeout until when they will misconfigure their firewall and their program will get stuck (hence the default) and because the asyncio facilities are not available in sync code.

Opening this ticket to allow #763 to proceed and add the timeout after merging.

dlax:
Fix #779


psycopg/psycopg/connection.py
```diff
@@ -261,7 +261,7 @@ def rollback(self) -> None:
         with self.lock:
             self.wait(self._rollback_gen())
 
-    def cancel_safe(self) -> None:
+    def cancel_safe(self, *, timeout: float = 0.0) -> None:
```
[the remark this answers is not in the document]
dlax: Done.

dvarrazzo:
Hi @dlax, I'm playing a bit with this changeset.

One thing I see is that it lacks tests. I am trying to whip one together by:

- create a proxy to the database
- connect to the proxy
- stop the proxy
- open a deaf port to the proxy
- try cancelling with a timeout and see if it's respected

Part of the testing infrastructure is in place, but not all of it: there isn't a facility to opening a "deaf port" on a previously listening port. I am adding it with:

```python
# in class Proxy in fix_proxy.py
    @contextmanager
    def deaf_listen(self):
        with socket.socket(socket.AF_INET) as s:
            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
            s.bind((self.client_host, self.client_port))
            s.listen(0)
            yield s
```

with this in place I could add a test that looks like:

```python
async def test_cancel_safe_timeout(aconn_cls, proxy):
    proxy.start()
    async with await aconn_cls.connect(proxy.client_dsn) as aconn:
        proxy.stop()
        with proxy.deaf_listen():
            t0 = time.time()
            with pytest.raises(e.CancellationTimeout):
                await aconn.cancel_safe(timeout=1)
            elapsed = time.time() - t0
            assert 1.0 < elapsed < 1.5
```

This works, but I've found a few things to review. One is this pattern in BaseConnection:

```python
    def _cancel_gen(self, *, timeout: float) -> PQGenConn[None]:
        try:
            cancel_conn = self.pgconn.cancel_conn()
        except e.OperationalError as ex:  # if the connection is closed
            logger.warning("couldn't create a cancel connection: %s", ex)
        else:
            cancel_conn.start()
            yield from generators.cancel(cancel_conn)
```

the way it's written, if `cancel_conn()` raises a different exception (in my case it was raising NotSupported because I was trying with libpq 16 we end in the else branch and cause other error downstream. This should be fixed in the context of #778.

Another issue we have is that, if we fall back on the legacy challenge, we don't have a way to time out. We can do it in async but I don't think there's an obvious way to timeout a hung PGcancel. So I'd make at least this change, to be on the safe side in async code:

```
@@ -296,9 +296,15 @@ class AsyncConnection(BaseConnection[Row]):
                 await waiting.wait_conn_async(
                     self._cancel_gen(timeout=timeout), interval=_WAIT_INTERVAL
                 )
+            # TODO: check for explicit support when capabilities implemented
             except e.NotSupportedError:
                 if True:  # ASYNC
-                    await to_thread(self.cancel)
+                    try:
+                        await asyncio.wait_for(to_thread(self.cancel), timeout)
+                    except asyncio.TimeoutError:
+                        raise e.CancellationTimeout(
+                            f"couldn't camcel connection after {timeout} s"
+                        )
                 else:
                     self.cancel()
```

and on the sync code... nothing. I guess we will add to the docs that timeout is not supported in the sync case if libpq < 17.

As the comment says, the cancel_safe() code can make use of #782. So I think I would first get to finish that MR, and to add the relevant testing method before continuing here.

dlax:
Indeed, not I was also bothered by not being able to easily add a test. Thank you for looking into it.
So I should wait for #782 to get in before getting back to this PR? (Probably not before Friday in any case.)

dlax:
I've added the test you suggested above; the proxy fixture is really nice! Also wrapped the libpq<17 code path with a `wait_for(..., timeout=)` and added a note about the timeout not being used for the sync connection and old libpq.

dlax:
Not sure about the test failures in CI; should I mark the test as `flakey`?

dvarrazzo:
I think my last commit fixed the issue. I've repeated [this workflow](https://github.com/psycopg/psycopg/actions/runs/8662300672/job/23759044628) twice and no problem was found.

dlax:
LGTM; thank you for fixing that.

dvarrazzo:
It seems that #778 didn't go in the right direction. If we consistently swallow exceptions in `cancel_safe()` then we will also swallow the timeout...

Swallowing exceptions is useful internally, hence the previous `_try_cancel()`, which was introduced to avoid to extend the interface. But now that we are extending it for 3.2, maybe we can add an `ignore_errors` parameter. I'll give it a try, but please let me know what do you think.

dlax:
At a first glance, I'm not really exited by the `ignore_errors`-kind API; to me, it feels like a symptom of a bad design and something fixed when it was too late to not break compatibility (which is not our case). Also, as a user, if I want to ignore errors, I'd rather do this explicitly with a `try: except:` or `contextlib.suppress()`.

But, if #785 was wrong, why not revert it, reopen #778, proceed with this feature here and then think more about #778 as we'll get the whole picture?

In any case, I haven't read the code changes, being afk this weekend, so I can take a closer look next week.

dvarrazzo:
Thank you for your input, I will consider that.

Also, don't mind the state of this branch... it's pretty broken, something gets stuck forever on cancel with timeout and legacy function.

I'll fix it. Enjoy your weekend!

dlax:
I've been looking more closely the recent commits from #785, especially these changes: https://github.com/psycopg/psycopg/commit/46f661dbbc30808d11a71c0ae7bbfc8245b6fd42#diff-32c2d53ec9106be2c604c0d178a4c2808c4a240de359a5454b983c9a1582a8cfR299 and https://github.com/psycopg/psycopg/commit/0bbbe4bf02a391589c65e88601f9474191d786bd#diff-67bdf0925106fb7b713c19d325d4eb43d8f4a27878c9018f706543b456850743R300. I could track the logic down to this commit https://github.com/psycopg/psycopg/commit/24774792abf1d89b037926815daf2da5ceac93bd but it lacks a bit of context :); you also mentioned that commit the other day, in https://github.com/psycopg/psycopg/pull/763#discussion_r1556622532, but I now admit that I don't understand the point you made at that time. 

So in fact, I actually don't understand why we'd suppress `Exception` (while emitting a warning) in case of cancellation failure. As a user, if cancelling my query fails, I think I'd rather be informed clearly and quickly (per the principle of least-surprise somehow) and I'd be fine being told to handle the exception myself, as any other kind of operational errors.  

(Besides, the situation here is partly my fault, as I probably reviewed #785 too quickly assuming it would unblock other things like changeset here.)

dvarrazzo:
I think that the state must be:

- if an user calls `cancel*()` explicitly, they should receive an exception raised.
- if we use `cancel*()` internally, and an exception would clobber a previous exception, we should swallow it.

So, the problem started from 24774792abf1d89b037926815daf2da5ceac93bd where I think it's ok to call `_try_cancel()` on error handling, but it was wrong to call it from `cancel()`.

In the last commit in this branch (1918e634f6ac2654a07880d0378e8775646b86a2) I have tried to reintroduce the same logic, but as you point out maybe the `ignore_errors` argument is not clean design. So, probably the best thing to do is to reintroduce a `_try_cancel()` internal method (this time async too) and call it everywhere but not in the outermost `cancel()`.

A different problem emerged in the last commit is the hang with the C library and the legacy functions.

I will refactor my last commit in this branch (or just drop by and work in a different branch, which might be more polite), reintroducing `_try_cancel()` in lieu of `ignore_errors` and will drop the timeout management on the legacy function. If I manage to introduce it, I'll do it in another branch, otherwise, no big deal.

dvarrazzo:
I've pushed a new commit 826115c9f7a303bd667068c1fd62f17d9a324bf9 [in this branch](https://github.com/psycopg/psycopg/tree/try-cancel-again) which would follow up the current state of this MR. It should fix the two errors currently failing here (caused by the fact that the timeout is swallowed in the `cancel_safe()` public call and, compared, to the previous attempt last week, ignores the exceptions explicitly instead of using an extra argument.

As there are tests failing here, I propose to include that commit in this branch and merge it with it, but first let's check if tests pass.

psycopg/psycopg/connection.py
```diff
@@ -265,29 +265,36 @@ def rollback(self) -> None:
...
-                )
-            except Exception as ex:
-                logger.warning("couldn't try to cancel query: %s", ex)
+            waiting.wait_conn(
+                self._cancel_gen(timeout=timeout), interval=_WAIT_INTERVAL
+            )
         else:
             self.cancel()
 
+    def _try_cancel(self, *, timeout: float = 30.0) -> None:
```
dlax, line 292: I would suggest to remove the default value for `timeout` as callers always pass a (different) value. (The other option would be to set the default value to the one that's actually used, i.e. `5.0`, but I find it clearer to have it set by the caller.)

dlax:
> I think that the state must be:
> 
>   * if an user calls `cancel*()` explicitly, they should receive an exception raised.
>   * if we use `cancel*()` internally, and an exception would clobber a previous exception, we should swallow it.
> 
> 
> So, the problem started from [2477479](https://github.com/psycopg/psycopg/commit/24774792abf1d89b037926815daf2da5ceac93bd) where I think it's ok to call `_try_cancel()` on error handling, but it was wrong to call it from `cancel()`.

Expressed that way, it totally makes sense. Thanks.

dlax (commented):
I'm a bit concerned by the last commit 95bedf9a62c96932bdbdf92988378f680fa15b39 mixing several somewhat unrelated changes: docs, commented code/TODO (esp. this one); would you mind splitting it up?

psycopg/psycopg/connection.py
```diff
@@ -265,29 +265,36 @@ def rollback(self) -> None:
     def cancel_safe(self, *, timeout: float = 30.0) -> None:
         """Cancel the current operation on the connection.
 
+        :param timeout: raise a `~errors.CancellationTimeout` if sending
+            the timeout doesn't succeed within `timeout` seconds.
```
dlax, line 269: I suppose you meant: "if sending the *cancellation* doesn't succeed". But in fact, it's not clear to me why you're talking about "sending the cancellation". How about: "if the cancellation request does not succeed within `timeout` seconds." ?

psycopg/psycopg/connection.py
```diff
@@ -265,29 +265,36 @@ def rollback(self) -> None:
...
+                self._cancel_gen(timeout=timeout), interval=_WAIT_INTERVAL
+            )
         else:
             self.cancel()
 
+    def _try_cancel(self, *, timeout: float = 30.0) -> None:
+        try:
+            self.cancel_safe(timeout=timeout)
+        except Exception as ex:
+            logger.warning("couldn't send query cancel message: %s", ex)
```
dlax, line 296: Same here, not sure about the "send query cancel" meaning. I would suggest `"query cancellation failed: %s"`.

dlax:
> I'm a bit concerned by the last commit [95bedf9](https://github.com/psycopg/psycopg/commit/95bedf9a62c96932bdbdf92988378f680fa15b39) mixing several somewhat unrelated changes: docs, commented code/TODO (esp. this one); would you mind splitting it up?

I have actually done, and also fixed the other nits, in another branch `cancel-safe-timeout-fixups`. I can push this here if that's okay by you?

dvarrazzo:
> I have actually done, and also fixed the other nits, in another branch `cancel-safe-timeout-fixups`. I can push this here if that's okay by you?

Hello! You raise valid points, including splitting the commits. If you have done it already, sure, feel free to force-push here.

psycopg/psycopg/connection.py
dvarrazzo: What I thought is that, as a default timeout, something in the 30-60s region seems ok, because it's pretty much just to avoid to hang forever. That's what end users would call by default.

As the timeout to "get out of troubles" in our connection, I thought I wouldn't want to add a pretty long operation by default (although it's long only if the situation is really bad...)

If you think it's unclear why we propose 30s to the user, but use 5s ourselves, it's probably ok to drop the 5s introduced in the callers and to use the default which I'd leave to 30a.

dlax:
Ok; pushed all commits but the last one ed5e3bbb8168afb20c91b634a39c9e1c5a0ca294 as I'm not sure what it is about (the hang you mentioned earlier at https://github.com/psycopg/psycopg/pull/780#issuecomment-2056498550?).

Releasing the lock for today :)

psycopg/psycopg/connection.py
dlax: Yes, 30s as a default for user-facing code is totally fine. But, IIUC, this timeout here concerns `_try_cancel()` which is not called (or no longer called) by `cancel_safe()`.

dvarrazzo: Oh yeah, you are totally right. We can make 5s a default of _try_cancel().

dvarrazzo:
> Ok; pushed all commits but the last one [ed5e3bb](https://github.com/psycopg/psycopg/commit/ed5e3bbb8168afb20c91b634a39c9e1c5a0ca294) as I'm not sure what it is about (the hang you mentioned earlier at [#780 (comment)](https://github.com/psycopg/psycopg/pull/780#issuecomment-2056498550)?).
> 
> Releasing the lock for today :)

The failing test runs are [these](https://github.com/psycopg/psycopg/actions/runs/8672989229). Adding the timeout to the async task running the legacy `PQcancel()` function hangs forever. It's reproducible locally.

It's a marginal change and I'm happy to leave it unresolved in this MR, hence leaving the comment in case we want to go back on this in a future iteration.

psycopg/psycopg/connection.py
dlax: Fixed in f9c3d202495b34a7516480bfb9cebb93c1775301.

dlax:
> The failing test runs are [these](https://github.com/psycopg/psycopg/actions/runs/8672989229). Adding the timeout to the async task running the legacy `PQcancel()` function hangs forever. It's reproducible locally.

It seems this does not fail with the current state of the branch.

So this is ready from my point of view, if you can review again; thanks.

dvarrazzo:
> It seems this does not fail with the current state of the branch.

No, this branch is ok as it is. Feature merged, thank you very much :)

The problem arises with a changeset such as:

```
diff --git a/psycopg/psycopg/connection_async.py b/psycopg/psycopg/connection_async.py
index b0c8b753..6d7f9ea1 100644
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -304,7 +304,7 @@ class AsyncConnection(BaseConnection[Row]):
             )
         else:
             if True:  # ASYNC
-                await to_thread(self.cancel)
+                await asyncio.wait_for(to_thread(self.cancel), timeout)
             else:
                 self.cancel()
```

It might even be that #792 was the cause and the problem is fixed already. But it's an issue that can be looked into separately from this changeset :slightly_smiling_face:

dlax:
> The problem arises with a changeset such as:

Indeed; I forgot this was reverted. But now, adding it again does not fail on my side. Just opened #798 to see how it goes in CI.

dlax:
> Adding the timeout to the async task running the legacy PQcancel() function hangs forever. It's reproducible locally.

In fact, this, I can reproduce (took time before I could actually understand what you meant :).

But I suspect this because of how `test_cancel_safe_timeout` is implemented with the deaf listen which makes `PQcancel` hang forever. So in #798, I made this test skipped for the sync Connection and libpq < 17; which seems fine to me as the `timeout` parameter is [documented](https://www.psycopg.org/psycopg3/docs/api/connections.html#psycopg.Connection.cancel_safe) to have no effect in this very case.

dlax:
> So in #798, I made this test skipped for the sync Connection and libpq < 17

Correction: it also hangs for the async case (obviously, since it's related to `PQcancel`); still investigating...

PR #782, opened 2024-04-09
Add capabilities objects

### LINKED ISSUE #772: Idea: adding a capabilities fuction?
Several psycopg features depend on the libpq version and require precise knowledge of when a certain feature was added. For instance, skimming the ifs in the libpq: prepared statements, prepared statements compatibility with pgbouncer, PQhostAddr, async cancelling, set trace flag.

For pipelines the ad-hoc solution of `Pipeline.is_supported()`, but new needs are emerging for newer features.

Proposal: adding a `psycopg.capability()` function to check if a certain capability is supported or not. For instance:

```python
def capability(name: str, *, check: bool = False) -> bool | None:
    """
    Verify if a capability is supported by the current libpq library or database function.

    :param name: the name of the capability to check
    :param check: if True, raise `NotSupportedError` if the capability is unsupported or unknown.
    :return: `True` if the capability is supported, `False` if not, `None` if unknown
    """
    ...
```

If `check = True` we may raise an error with a descriptive message for what is required to gain the capability, such as [these](https://github.com/psycopg/psycopg/blob/640d6046be5e8b660d7782bb0d64f8120f7c1108/psycopg/psycopg/_pipeline.py#L77-L87). This could be add at import level to detect early if a capability required by a program is not supported. Otherwise, the result can be checked in an `if`. 

```python
import psycopg
psycopg.capability("pipeline", check=True)

...

if python.capability("pipeline"):
    with conn.pipeline():
       ...
```

The capabilities might be just strings or we might provide a string enum too to represent them. I would reserve ourselves the possibility to add a new capability check in bugfix releases.

```python
class Capability(str, Enum):
    pipeline = "pipeline"
    close_prepared = "close_prepared"
    pgbouncer_prepared = "pgbouncer_prepared"  # maybe? Just a friendly alias for the above.
    hostaddr = "hostaddr"
    async_cancel = "async_cancel"
    trace_flags = "trace_flags"
```

I would like to add this feature in 3.2, because several feature conditional to the libpq version are being introduced.

What do you think? @dlax? @asqui?

dvarrazzo:
The object, exposed as `psycopg.capabilities`, exposes `has_STUFF()` methods to verify if a certain feature is supported.

If `check=True` is specified, raise an exception instead of returning False, with a finely crafted message such as:

> the feature 'Connection.pipeline()' is not available: you are using a psycopg[c] libpq wrapper built with libpq version 9.5.3; the feature requires libpq version at least 14.0
>
> the feature 'Connection.pipeline()' is not available: the client libpq version (imported from system libraries) is 13.4; the feature requires libpq version at least 14.0
>
> the feature 'Connection.pipeline()' is not available: the client libpq version (imported from psycopg[binary] version 3.2.0) is 13.4; the feature requires libpq version at least 14.0

helping to diagnose the problem.

As a side effect, added a `pq.version_pretty()` function to convert the numieric version to the postgres version. Code refactored to use it instead of homegrown `// 10000`, divmod, and similar.

The `Pipeline.is_supported()` method has been reimplemented using `Capabilities.has_pipeline()`.

Close #772

### DIFF
--- a/docs/api/connections.rst
+++ b/docs/api/connections.rst
@@ -287,6 +287,11 @@ The `!Connection` class
 
     .. automethod:: cancel_safe
 
+        .. note::
+
+            You can use the `~Capabilities.has_cancel_safe` capability to check
+            if `!cancel_safe()` will not fall back on the legacy implementation.
+
         .. warning::
 
             This method shouldn't be used as a `~signal.signal` handler.
--- a/docs/api/objects.rst
+++ b/docs/api/objects.rst
@@ -74,8 +74,9 @@ Connection information
 
     .. autoattribute:: hostaddr
 
-        Only available if the libpq used is at least from PostgreSQL 12.
-        Raise `~psycopg.NotSupportedError` otherwise.
+        Only available if the libpq used is from PostgreSQL 12 or newer.
+        Raise `~psycopg.NotSupportedError` otherwise. You can use the
+        `~Capabilities.has_hostaddr` capability to check for support.
 
     .. autoattribute:: port
     .. autoattribute:: dbname
@@ -107,6 +108,58 @@ Connection information
             The `PostgreSQL supported encodings`__.
 
             .. __: https://www.postgresql.org/docs/current/multibyte.html
+
+
+.. _capabilities:
+
+Libpq capabilities information
+------------------------------
+
+.. autoclass:: Capabilities
+
+    An instance of this object is normally exposed by the module as the object
+    `psycopg.capabilities`.
+
+    Every feature check is implemented by an `!has_SOMETHING()` method. All
+    the methods return a boolean value stating if the capability is supported,
+    which can be used by a program to degrade gracefully::
+
+        if psycopg.capabilities.has_pipeline()
+            with conn.pipeline():
+                operations(conn)
+        else:
+            logger.warning("slower")
+            operations(conn)
+
+    If `check` is `!True`, and the capability is not supported, raise a
+    `NotSupportedError` instead of returning `!False`, explaining why the
+    feature is not supported. This allows to make a check at import time,
+    crashing early and with a clear description of the problem.
+
+        >>> import psycopg
+        >>> psycopg.capabilities.has_pipeline(check=True)
+        Traceback (most recent call last):
+          ...
+        psycopg.NotSupportedError: the feature 'Connection.pipeline()' is not available:
+            the client libpq version (imported from system libraries) is 13.4; the
+            feature requires libpq version 14.0 or newer
+
+    .. versionadded:: 3.2
+
+    .. automethod:: has_encrypt_password
+    .. automethod:: has_hostaddr
+    .. automethod:: has_pipeline
+    .. automethod:: has_set_trace_flags
+    .. automethod:: has_cancel_safe
+
+        .. note::
+
+            The `!cancel_safe()` method is implemented anyway, but it will use
+            the legacy :pq:`PQcancel` implementation.
+
+    .. automethod:: has_pgbouncer_prepared
+
+        .. seealso:: :ref:`pgbouncer`
 
 
 The description `Column` object
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -19,6 +19,8 @@ Psycopg 3.2 (unreleased)
   (:ticket:`340`).
 - Add :ref:`raw-query-cursors` to execute queries using placeholders in
   PostgreSQL format (`$1`, `$2`...) (:ticket:`#560`).
+- Add `psycopg.capabilities` object to :ref:`inspect the libpq capabilities
+  <capabilities>` (:ticket:`#772`).
 - Add `~rows.scalar_row` to return scalar values from a query (:ticket:`#723`).
 - Prepared statements are now :ref:`compatible with PgBouncer <pgbouncer>`.
   (:ticket:`#589`).
--- a/psycopg/psycopg/_capabilities.py
+++ b/psycopg/psycopg/_capabilities.py
@@ -11,36 +11,49 @@ from .errors import NotSupportedError
 
 class Capabilities:
     """
-    Check if a feature is supported.
-
-    Every feature check is implemented by a check method `has_SOMETHING`.
-    All the methods return a boolean stating if the capability is supported.
-    If not supported and `check` is True, raise a `NotSupportedError` instead
-    explaining why the feature is not supported.
+    An object to check if a feature is supported by the libpq available on the client.
     """
 
     def has_encrypt_password(self, check: bool = False) -> bool:
-        """Check if the `~PGconn.encrypt_password()` method is implemented."""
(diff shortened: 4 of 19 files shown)


asqui:
Drive-by naming nitpick: Would `raise: bool` be a better name for `check`?

I think `capabilities.has_pipeline(raise=True)` is a bit more descriptive about the behaviour to someone who is unfamiliar with the capabilities API.

asqui:
(PS: I agree that the `has_STUFF()` API is nicer than `has_capability(capability: str)`, and latter gives extra flexibility that I think is not necessary. 👍 )

dvarrazzo:
Hi, nice to see you dropping by

> Drive-by naming nitpick: Would `raise: bool` be a better name for `check`?
>
> I think `capabilities.has_pipeline(raise=True)` is a bit more descriptive about the behaviour to someone who is unfamiliar with the capabilities API.

You might be right. I should "check" what other use we make of similarly named parameter around in the code. Actually, one that I have in mind is the pool which has a `check` method and build parameter... and they mean a different thing.

So, at a glance, raise might be a better name.

psycopg/psycopg/_capabilities.py
```diff
@@ -0,0 +1,89 @@
...
+    explaining why the feature is not supported.
+    """
+
+    def has_encrypt_password(self, check: bool = False) -> bool:
+        """Check if the `~PGconn.encrypt_password()` method is implemented."""
+        return self._has_feature("PGconn.encrypt_password()", 100000, check=check)
+
+    def has_hostaddr(self, check: bool = False) -> bool:
+        """Check if the `~ConnectionInfo.hostaddr` attribute is implemented."""
+        return self._has_feature("Connection.pipeline()", 120000, check=check)
```
dlax: Should be `"ConnectionInfo.hostaddr"`.

docs/api/connections.rst
```diff
@@ -287,6 +287,12 @@ The `!Connection` class
 
     .. automethod:: cancel_safe
 
+        .. note::
+
+            You can use the `~Capabilities.has_cancel_safe` capability to check
+            if `!cancel_safe()` will will not fall back on the legacy
```
dlax, line 293: typo: will will

docs/api/objects.rst
```diff
@@ -109,6 +110,58 @@ Connection information
...
+.. autoclass:: Capabilities
+
+    An instance of this object is normally exposed by the module as the object
+    `psycopg.capabilities`.
+
+    Every feature check is implemented by an `!has_SOMETHING()` method. All
+    the methods return a boolean value stating if the capability is supported,
+    which can be used by a program to degrade gracefully::
+
+        if psycopg.capabilities.has_pipeline(check=True)
```
dlax, line 127: `check=False` (or omitted) in this case, no?

docs/news.rst
```diff
@@ -19,6 +19,8 @@ Psycopg 3.2 (unreleased)
   (:ticket:`340`).
 - Add :ref:`raw-query-cursors` to execute queries using placeholders in
   PostgreSQL format (`$1`, `$2`...) (:ticket:`#560`).
+- Add `psycopg.capabilities` object to :ref:`inspect the libpq capabilities
+  <capabilities>` (:ticket:`#772`).
```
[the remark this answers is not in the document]
dvarrazzo: The `_` syntax is a reST feature and works only in the same file. Because of this limitation, Sphinx added the `ref` role which allows linking to references in different files in the doc.

psycopg/psycopg/_capabilities.py
```diff
@@ -11,36 +11,49 @@
...
+        return self._has_feature("pq.PGconn.encrypt_password()", 100000, check=check)
 
     def has_hostaddr(self, check: bool = False) -> bool:
-        """Check if the `~ConnectionInfo.hostaddr` attribute is implemented."""
-        return self._has_feature("Connection.pipeline()", 120000, check=check)
+        """Check if the `ConnectionInfo.hostaddr` attribute is implemented.
+
+        The feature requires libpq 12.0 and greater.
+        """
+        return self._has_feature("Connection.info.hostaddr", 120000, check=check)
```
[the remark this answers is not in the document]
dvarrazzo: Yes, it was an early pasto.

docs/api/connections.rst
dvarrazzo: Fixed, thank you

docs/api/objects.rst
dvarrazzo: Yes, it's a mistake. Fixed, thank you

asqui:
Ok, I was actually thinking of `raises: bool` which I'd seen in another library, but I think @dlax's citation from `subprocess.run(..., check=True)` is a more compelling argument for calling it `check`, so I withdraw my suggestion 😄

psycopg/psycopg/_pipeline.py
```diff
@@ -63,36 +64,12 @@ def status(self) -> pq.PipelineStatus:
...
-
-        return ""
-
     def _enter_gen(self) -> PQGen[None]:
-        if not self.is_supported():
-            raise e.NotSupportedError(
-                f"pipeline mode not supported: {self._not_supported_reason()}"
-            )
+        if not self._is_supported:
+            capabilities.has_pipeline(check=True)
```
dlax, line 72: IIUC, we'd now always run `capabilities.has_pipeline(check=True)` when entering the pipeline, unless `is_supported()` has been called before to set the `_is_supported` attribute, no?

Also, it might interesting to only call `capabilities.has_pipeline()` once; something like 663e64265cb9aa422aa401e03877ab063f6ef08f.

dvarrazzo: Yes, that's xorrect. The if is a micro optimization to avoid to call `has_pipeline()` every time in the most common path (when we already established that pipelines are supported).

Maybe it's worth nothing, or maybe in general I should have added caching in the Capabilities object, if my gripe is to avoid to call `pq.version()` over and over. 

TBH I've been a bit lazy here and I didn't add caching because I thought it'd make tests bit more cumbersome... But maybe I should do that.

dvarrazzo:
I have added caching on the capabilities instance, and simplified the pipeline objects as a consequence:

https://github.com/psycopg/psycopg/blob/c4dc5d1ce236ded8a31bf25f8fed8791901f372c/psycopg/psycopg/_pipeline.py#L63-L68

If that's ok, can merge this, then use it in:

https://github.com/psycopg/psycopg/blob/7cf456a06a79b08bdbd825913fa1fe998eb71044/psycopg/psycopg/connection.py#L277-L278

and merge #785 too.

dlax:
Out of curiosity, why not using `functools.cache()`?

dvarrazzo:
> Out of curiosity, why not using `functools.cache()`?

It doesn't declare explicitly that it works to wrap a method. For instance I don't know if and how it includes the `self` parameter in the key calculation. Therefore I prefer to use `@cache` only on pure functions, not methods.

dlax (approved):

PR #783, opened 2024-04-10
Replace `deaf_port` fixture with `deaf_listen` method on `proxy` fixture

dvarrazzo:
This allows to replace the listening proxy port with a port that blocks connection and to write more easily cancel timeout tests (as in #780).


tests/fix_proxy.py
```diff
@@ -110,6 +101,17 @@ def stop(self):
         logging.info("proxy stopped")
         self.proc = None
 
+    @contextmanager
+    def deaf_listen(self):
```
dlax, line 105: The method name is not quite self-explanatory to me; would you mind adding a tiny docstring explaining what this does functionally? (Something like what was on the previous `deaf_port()`.)

dlax (approved):

tests/fix_proxy.py
dvarrazzo: Sure, will do, thank you. It will do something like:

> Open the proxy port to listen, but without accepting a connection.
>
> A program trying to connect will block. Useful to test connection timeouts.

Will commit and merge later, today I'm not in front of a computer

PR #785, opened 2024-04-10
Review cancel methods implementations

### LINKED ISSUE #778: Review error swallowing in cancel methods

dvarrazzo:
I have cleaned up both `cancel()` and `cancel_safe()`. Both the methods now consistently warn and swallow exceptions on error, but only in the outermost shell; inner methods are more purely mechanism.

Close #778. In preparation for #779. Can make use of #782.

### DIFF
--- a/psycopg/psycopg/_connection_base.py
+++ b/psycopg/psycopg/_connection_base.py
@@ -320,13 +320,9 @@ class BaseConnection(Generic[Row]):
         return True
 
     def _cancel_gen(self) -> PQGenConn[None]:
-        try:
-            cancel_conn = self.pgconn.cancel_conn()
-        except e.OperationalError as ex:  # if the connection is closed
-            logger.warning("couldn't create a cancel connection: %s", ex)
-        else:
-            cancel_conn.start()
-            yield from generators.cancel(cancel_conn)
+        cancel_conn = self.pgconn.cancel_conn()
+        cancel_conn.start()
+        yield from generators.cancel(cancel_conn)
 
     def add_notice_handler(self, callback: NoticeHandler) -> None:
         """
--- a/psycopg/psycopg/_connection_base.py
+++ b/psycopg/psycopg/_connection_base.py
@@ -302,7 +302,7 @@ class BaseConnection(Generic[Row]):
             c = self.pgconn.get_cancel()
             c.cancel()
         except Exception as ex:
-            logger.warning("couldn't try to cancel query: %s", str(ex))
+            logger.warning("couldn't try to cancel query: %s", ex)
 
     def _should_cancel(self) -> bool:
         """Check whether the current command should actually be cancelled when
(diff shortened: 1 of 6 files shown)


psycopg/psycopg/_connection_base.py
```diff
@@ -293,8 +293,16 @@ def fileno(self) -> int:
...
+        if not self._should_cancel():
+            return
+
+        # Don't fail cancelling (which might happen on connection closing) to
+        # avoid clobbering eventual exceptions with ours, which is less important.
+        try:
+            c = self.pgconn.get_cancel()
+            c.cancel()
+        except Exception as ex:
+            logger.warning("couldn't try to cancel query: %s", str(ex))
```
dlax, line 305:
```suggestion
            logger.warning("couldn't try to cancel query: %s", ex)
```
seems better to me; by avoiding explicit `str()`, which will be done when the log message gets formatted in any case,  we save it if the log level is higher than `WARNING`. (Sure, that's not a critical part; more a nit :)

psycopg/psycopg/connection.py
```diff
@@ -271,11 +271,17 @@ def cancel_safe(self) -> None:
...
+            return
+
+        # TODO: replace with capabilities.has_safe_cancel after merging #782
+        if pq.__build_version__ >= 170000:
             try:
                 waiting.wait_conn(self._cancel_gen(), interval=_WAIT_INTERVAL)
-            except e.NotSupportedError:
-                self.cancel()
+            except Exception as ex:
+                logger.warning("couldn't try to cancel query: %s", str(ex))
```
dlax, line 282:
```suggestion
                logger.warning("couldn't try to cancel query: %s", ex)
```

psycopg/psycopg/connection_async.py
```diff
@@ -288,16 +288,23 @@ async def cancel_safe(self) -> None:
...
                 await waiting.wait_conn_async(
                     self._cancel_gen(), interval=_WAIT_INTERVAL
                 )
-            except e.NotSupportedError:
-                if True:  # ASYNC
-                    await to_thread(self.cancel)
-                else:
-                    self.cancel()
+            except Exception as ex:
+                logger.warning("couldn't try to cancel query: %s", str(ex))
```
dlax, line 301:
```suggestion
                logger.warning("couldn't try to cancel query: %s", ex)
```

dlax (approved):
LGTM; just a small nit.

psycopg/psycopg/connection.py
dvarrazzo: I started to use this pattern since I discovered that Sentry might leak secrets in its default configuration (https://github.com/getsentry/sentry-python/issues/2417#issuecomment-1752447375).

Writing this code actually made me think I'd I shouldn't actually protect every single rerause of exceptions of too broad classes.

dvarrazzo:
I have dropped the `str()` calls because it's not our responsibility to fix Sentry.

PR #787, opened 2024-04-12
Async to sync refactor

dvarrazzo:
- Reimplement rules matching using pattern matching rather than a cascade of negated ifs to bail out early
- Drop the `--all` option as it's a meaningful default and running the script on no input seems an unneeded use case.

@dlax Are you ok with these change? If `--all` hadn't been there earlier, how would you have called `--convert-all/-B` insttead? I'd be ok to rename it to `--all`.


dlax:
> Are you ok with these change?

Yes; fine by me.

> If `--all` hadn't been there earlier, how would you have called `--convert-all/-B` insttead? I'd be ok to rename it to `--all`.

`-B` came from `make`; I'm also fine with using `--all` if you like it better.

dlax (approved):

Issue #790, opened 2024-04-15
Suggestion: Replace SELECT 1 with an empty query


Eyal-Shalev:
The current implementation of `psycopg_pool.pool.ConnectionPool.check_connection` uses the `SELECT 1` statement to check if a connection is valid.

However this goes through the PostgreSQL query planner which means we can make this process a little bit faster by sending an empty query instead.

This can be done by replacing `conn.execute("SELECT 1")` with `conn.execute("--ping")`

References:
- https://github.com/rails/rails/pull/42368
- [jackc/pgx ping implementation (Go)](https://github.com/jackc/pgx/blob/a3d9120636fc263debce595f3ba05bcd9f4ee809/pgconn/pgconn.go#L1815-L1817)
- https://github.com/sqlalchemy/sqlalchemy/issues/8491

dvarrazzo:
Thank you very much, I will look into this idea and your implementation :slightly_smiling_face:

dvarrazzo:
Implemented in #791.

PR #791, opened 2024-04-15
Replace SELECT 1 with an empty query

Eyal-Shalev:
From #790:
> The current implementation of `psycopg_pool.pool.ConnectionPool.check_connection` uses the `SELECT 1` statement to check if a connection is valid.
> 
> However this goes through the PostgreSQL query planner which means we can make this process a little bit faster by sending an empty query instead.
> 
> This can be done by replacing `conn.execute("SELECT 1")` with `conn.execute("--ping")`
> 
> References:
> - https://github.com/rails/rails/pull/42368
> - [jackc/pgx ping implementation (Go)](https://github.com/jackc/pgx/blob/a3d9120636fc263debce595f3ba05bcd9f4ee809/pgconn/pgconn.go#L1815-L1817)
> - https://github.com/sqlalchemy/sqlalchemy/issues/8491

### DIFF
--- a/psycopg_pool/psycopg_pool/pool.py
+++ b/psycopg_pool/psycopg_pool/pool.py
@@ -537,11 +537,11 @@ class ConnectionPool(Generic[CT], BasePool):
         for instance as `!check` callback when a pool is created.
         """
         if conn.autocommit:
-            conn.execute("SELECT 1")
+            conn.execute("--ping")
         else:
             conn.autocommit = True
             try:
-                conn.execute("SELECT 1")
+                conn.execute("--ping")
             finally:
                 conn.autocommit = False
 
--- a/psycopg_pool/psycopg_pool/pool.py
+++ b/psycopg_pool/psycopg_pool/pool.py
@@ -537,11 +537,11 @@ class ConnectionPool(Generic[CT], BasePool):
         for instance as `!check` callback when a pool is created.
         """
         if conn.autocommit:
-            conn.execute("--ping")
+            conn.execute("")
         else:
             conn.autocommit = True
             try:
-                conn.execute("--ping")
+                conn.execute("")
             finally:
                 conn.autocommit = False
 
--- a/docs/news_pool.rst
+++ b/docs/news_pool.rst
@@ -16,6 +16,8 @@ psycopg_pool 3.2.2 (unreleased)
 - Raise a `RuntimeWarning` instead of a `DeprecationWarning` if an async pool
   is open in the constructor.
 - Fix connections possibly left in the pool after closing (:ticket:`#784`).
+- Use an empty query instead of ``SELECT 1`` to check connections
+  (:ticket:`#790`).
 
 
 Current release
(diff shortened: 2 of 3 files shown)


psycopg_pool/psycopg_pool/pool.py
```diff
@@ -530,11 +530,11 @@ def check_connection(conn: CT) -> None:
         for instance as `!check` callback when a pool is created.
         """
         if conn.autocommit:
-            conn.execute("SELECT 1")
+            conn.execute("--ping")
```
dvarrazzo, line 533: Wouldn't it be even better to execute an empty string? Why to give the backend the task of parsing a comment? (`--ping` is not a command: `--` is a comment marker, I find misleading to make it look like a sort of command line option).

AFAICS the result would be the same:

```python
>>> import psycopg
>>> conn = psycopg.connect()
>>> conn.execute("")
<psycopg.Cursor [EMPTY_QUERY] [INTRANS] (database=piro) at 0x7e1de96cb690>

>>> psycopg.connect().execute("select pg_terminate_backend(%s)", [conn.info.backend_pid])

>>> conn.execute("")
OperationalError: consuming input failed: terminating connection due to administrator command
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

asqui: Agree. I read this suggestion when it was first posted, wondered what this magic `--ping` command was that I'd never heard of before, searched around a bit and found nothing, and the penny never dropped that it was a comment.

Executing an empty string seems clearer.

(Aside: I believe `SELECT` is also a valid postgres query, that also returns an empty resultset. Does that involve the query planner too? If not, I marginally prefer this to the empty string query, as it's a little more obvious that it's just an empty SELECT.)

dvarrazzo:
`SELECT` is a strange beast: it's a valid query that returns exactly one tuples with zero fields. Maybe it's like a zero in Codd's theory, but it's quite the special case (there are a few ifs in psycopg code to manage it without barfing :sweat_smile:).

```python
>>> conn.execute("SELECT").fetchall()
[()]
```

An empty query returns a different type of results. It's probably lighter to handle by both server and client, so from the point of view of the intents of the OP it is probably a better choice.

```python
>>> conn.execute("SELECT")
<psycopg.Cursor [TUPLES_OK] [INTRANS] (database=piro) at 0x734a9d8d1d20>

>>> conn.execute("")
<psycopg.Cursor [EMPTY_QUERY] [INTRANS] (database=piro) at 0x734a96f6af00>
```

psycopg_pool/psycopg_pool/pool.py
Eyal-Shalev: Thanks for the review 😊

Truth be told, I copied what the pgx library in go did. 

I'll get to updating the pr with an empty string in the next couple of days.

Eyal-Shalev: @dvarrazzo @asqui thank you both for the review and sorry it took me that long to do this small change 😅 

I've updated the PR to use an empty string instead of the `--ping` comment _marker_ (thanks for the correction)

dvarrazzo: Thank you! No worries, I will merge this MR as soon as I've done some other changes I'm into (mostly the last bugfixes for the next 3.1.x and I've just found another bug in pool, not released, but the maintenance branch will need some attention.

I think this MR is good to merge, thank you very much!

PR #792, opened 2024-04-16
fix(c): fix the goto label in the HAVE_POLL branch of wait_c_impl()

dlax:
This resolves the following warning when building the extension module:

    $ python psycopg_c/setup.py build_ext -i
    [...]
    building 'psycopg_c._psycopg' extension
    x86_64-linux-gnu-gcc -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -I/home/denis/.local/pgsql/include -I/home/denis/src/psycopg3/.venv/include -I/usr/include/python3.11 -c psycopg_c/_psycopg.c -o build/temp.linux-x86_64-cpython-311/psycopg_c/_psycopg.o
    psycopg_c/_psycopg.c: In function ‘wait_c_impl’:
    psycopg_c/_psycopg.c:1649:1: warning: label ‘error’ defined but not used [-Wunused-label]
     1649 |
          | ^

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/waiting.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/waiting.pyx
@@ -84,7 +84,7 @@ retry_eintr:
     }
 
     if (PyErr_CheckSignals()) { goto finally; }
-    if (select_rv < 0) { goto finally; }  /* poll error */
+    if (select_rv < 0) { goto error; }  /* poll error */
 
     rv = 0;  /* success, maybe with timeout */
     if (select_rv >= 0) {


dvarrazzo:
Thank you for looking into this. Actually I think that your change fixes more than just this problem. If we jump to `finally` when `poll()` has returned an error we end up returning -1 without an exception set, and I think Python will take umbrage of that (throwing a RuntimeError maybe).

As all tests passed, this seems the right thing to do. Will backport to 3.1 too.

dvarrazzo:
> As all tests passed, this seems the right thing to do. Will backport to 3.1 too.

No need for that actually. The label was accidentally changed in the 3.2 branch.

PR #793, opened 2024-04-17
add libpq interface for retrieving results in chunks

dlax:
https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=4643a2b265e967cc5f13ffa0c7c6912dbb3466d0

PR #794, opened 2024-04-17
add a 'size' parameter to Cursor.stream()

dlax:
Based on #793

### DIFF
--- a/docs/api/cursors.rst
+++ b/docs/api/cursors.rst
@@ -180,7 +180,14 @@ The `!Cursor` class
         .. __: https://materialize.com/docs/sql/subscribe/
         .. __: https://www.cockroachlabs.com/docs/stable/changefeed-for.html
 
-        The parameters are the same of `execute()`.
+        The parameters are the same of `execute()`, except for `size` which
+        can be used to set results retrieval by chunks instead of row-by-row.
+
+        .. note::
+
+            This `size` parameter is only available from libpq 17, you can use
+            the `~Capabilities.has_stream_chunked` capability to check if this
+            is supported.
 
         .. warning::
 
--- a/psycopg/psycopg/_cursor_base.py
+++ b/psycopg/psycopg/_cursor_base.py
@@ -15,6 +15,7 @@ from . import adapt
 from . import errors as e
 from .abc import ConnectionType, Query, Params, PQGen
 from .rows import Row, RowMaker
+from ._capabilities import capabilities
 from ._column import Column
 from .pq.misc import connection_summary
 from ._queries import PostgresQuery, PostgresClientQuery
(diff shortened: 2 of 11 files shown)


docs/api/cursors.rst
```diff
@@ -180,7 +180,14 @@ The `!Cursor` class
         .. __: https://materialize.com/docs/sql/subscribe/
         .. __: https://www.cockroachlabs.com/docs/stable/changefeed-for.html
 
-        The parameters are the same of `execute()`.
+        The parameters are the same of `execute()`, except for `size` which
+        can be used to set results retrieval by chunks instead of row-by-row.
+
+        .. note::
+
+            This `size` parameter is only available from libpq 17, you can use
```
dvarrazzo, line 188:
```suggestion
            This `size` parameter is only available from libpq 17, you can use the
```

dvarrazzo:
Heads up: I gave a forced push to the branch because github didn't detect that #793 was merged and the MR still showed two commits.

psycopg/psycopg/cursor_async.py
```diff
@@ -138,23 +138,32 @@ async def stream(
...
-                    self._stream_send_gen(query, params, binary=binary)
+                    self._stream_send_gen(query, params, binary=binary, size=size)
                 )
                 first = True
                 while await self._conn.wait(self._stream_fetchone_gen(first)):
-                    # We know that, if we got a result, it has a single row.
-                    rec: Row = self._tx.load_row(0, self._make_row)  # type: ignore
-                    yield rec
+                    # We know that, if we got a result, it has a single row or
+                    # a chunk of rows.
```
dvarrazzo, line 161: I think we can drop this comment altogether. The previous one was about assumption about there being exactly one record, but here we are making a careful iteration on >= 0 records.
```suggestion
```

docs/api/cursors.rst
dvarrazzo, line 188:
```suggestion
            This `size` parameter is only available from libpq 17, you can use the
```

dvarrazzo (changes requested):
Just some nits, but otherwise looks good, thank you :slightly_smiling_face:

psycopg/psycopg/_cursor_base.py
```diff
@@ -329,7 +338,7 @@ def _stream_fetchone_gen(self, first: bool) -> PQGen[Optional["PGresult"]]:
             return None
 
         status = res.status
-        if status == SINGLE_TUPLE:
+        if status in (SINGLE_TUPLE, TUPLES_CHUNK):
```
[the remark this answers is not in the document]
dlax: I didn't know that; thank you for sharing.

Fixed, along with other suggestions.

dvarrazzo:
I've rebased this to master and about to merge it.

PR #798, opened 2024-04-19
feat: also handle the timeout in async cancel_safe() with libpq < 17

dlax:
See comments from https://github.com/psycopg/psycopg/pull/780#issuecomment-2062492554

### DIFF
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -304,7 +304,10 @@ class AsyncConnection(BaseConnection[Row]):
             )
         else:
             if True:  # ASYNC
-                await to_thread(self.cancel)
+                try:
+                    await asyncio.wait_for(to_thread(self.cancel), timeout)
+                except asyncio.TimeoutError as exc:
+                    raise e.CancellationTimeout("cancellation timeout expired") from exc
             else:
                 self.cancel()
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -846,8 +846,9 @@ def test_cancel_safe_error(conn_cls, proxy, caplog):
 
 @pytest.mark.slow
 @pytest.mark.timing
-@pytest.mark.libpq(">= 17")
 def test_cancel_safe_timeout(conn_cls, proxy):
+    if pq.version() < 170000:
+        pytest.skip("only for libpq >= 17")
     proxy.start()
     with conn_cls.connect(proxy.client_dsn) as conn:
         proxy.stop()
--- a/tests/test_connection_async.py
+++ b/tests/test_connection_async.py
@@ -850,8 +850,10 @@ async def test_cancel_safe_error(aconn_cls, proxy, caplog):
 
 @pytest.mark.slow
 @pytest.mark.timing
-@pytest.mark.libpq(">= 17")
 async def test_cancel_safe_timeout(aconn_cls, proxy):
+    if False:  # ASYNC
+        if pq.version() < 170000:
+            pytest.skip("only for libpq >= 17")
     proxy.start()
     async with await aconn_cls.connect(proxy.client_dsn) as aconn:
         proxy.stop()


dlax:
Following aforementioned comments in #780, it seems to me that the `cancel_safe(timeout=...)` works even with legacy `PQcancel` but the existing test does not make it possible to check that. For example, I can make it work with the following script:
```python
# file: t.py
import psycopg
import asyncio
import os

print(f"{psycopg.pq.version()=}")


async def main() -> None:
    dsn = os.environ["PSYCOPG_TEST_DSN"]  # host=localhost
    async with await psycopg.AsyncConnection.connect(dsn) as conn:
        print("connected")
        t = asyncio.create_task(conn.execute("select pg_sleep(1)"))
        await conn.cancel_safe(timeout=0.1)
        await t


if __name__ == "__main__":
    asyncio.run(main())
```
and setting `sudo tc qdisc add dev lo root netem delay 150ms` (to be reset with `sudo tc qdisc del dev lo root` after tests):
```python
$ python t.py
psycopg.pq.version()=160002
connected
  File "/usr/lib/python3.11/asyncio/tasks.py", line 490, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

  File "/home/denis/src/psycopg3/psycopg/build/__editable__.psycopg-3.2.0.dev1-py3-none-any/psycopg/connection_async.py", line 308, in cancel_safe
    await asyncio.wait_for(to_thread(self.cancel), timeout)
  File "/usr/lib/python3.11/asyncio/tasks.py", line 492, in wait_for
    raise exceptions.TimeoutError() from exc
TimeoutError

The above exception was the direct cause of the following exception:

  File "/home/denis/src/psycopg3/t.py", line 18, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/denis/src/psycopg3/t.py", line 13, in main
    await conn.cancel_safe(timeout=0.1)
  File "/home/denis/src/psycopg3/psycopg/build/__editable__.psycopg-3.2.0.dev1-py3-none-any/psycopg/connection_async.py", line 310, in cancel_safe
    raise e.CancellationTimeout("cancellation timeout expired") from exc
psycopg.errors.CancellationTimeout: cancellation timeout expired
```

dlax:
So I guess that we could adjust the test to let the proxy delay requests rather than blocking them with `deaf_listen()`; but I'm not sure this is possible with pproxy.

dvarrazzo:
> the existing test does not make it possible to check that

In the changeset where I tried this myself I changed the test this way:

```python
diff --git a/tests/test_connection_async.py b/tests/test_connection_async.py
index f5c355b2..c53c8e5d 100644
--- a/tests/test_connection_async.py
+++ b/tests/test_connection_async.py
@@ -850,8 +850,10 @@ async def test_cancel_safe_error(aconn_cls, proxy, caplog):
 
 @pytest.mark.slow
 @pytest.mark.timing
-@pytest.mark.libpq(">= 17")
 async def test_cancel_safe_timeout(aconn_cls, proxy):
+    if not is_async(__name__) and not psycopg.capabilities.has_cancel_safe():
+        pytest.skip("timeout not supported with legacy cancel in sync mode")
+
     proxy.start()
     async with await aconn_cls.connect(proxy.client_dsn) as aconn:
         proxy.stop()
```

dvarrazzo:
I am trying this again and I'm just verifying that:

```python
await asyncio.wait_for(to_thread(self.cancel), 1)
```

doesn't terminate. It doesn't even respond to ctrl-c.

I think the use of select or poll in `PQcancel()` cannot be interrupted successfully. So, considering that:

- libpq 17 will be soon released (and we will release a binary package including it as soon as available);
- this changeset creates an unpleasant asymmetry between sync and async 
-capabilites (timeout wouldn't still be honoured in sync code, unless going multiprocess or some other [Pythagoras switch](https://youtu.be/wI1sNMmk8oY?list=PLGXaPVaCNjNAdTq7qgiQGVlAALyDmpozY&t=12));

I think we can close this MR.

PR #810, opened 2024-05-06
Avoid possible deadlock (until timeout) on pool closing

dvarrazzo:
With the previous change to avoid finding open connections in the pool (#784), stopping the worker was moved into the critical section. This can create a deadlock in case a worker is in the process of obtaining a new connection, because putting it to the pool requires the lock. The deadlock only last for the default 5s timeout passed to _stop_workers().

Solve the problem by guarding _add_to_pool() to avoid it to try to add the connection if the pool is closed.

However, refactor the pool closing sequence too and close the workers and other resources that now out of the state outside the critical section to keep the operation running under lock to a minimum.

### DIFF
--- a/psycopg_pool/psycopg_pool/pool.py
+++ b/psycopg_pool/psycopg_pool/pool.py
@@ -106,7 +106,8 @@ class ConnectionPool(Generic[CT], BasePool):
         if getattr(self, "_closed", True):
             return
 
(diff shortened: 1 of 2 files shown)


dvarrazzo:
@adriangb I found running the tests stuttering occasionally, hitting the deadlock described here. Do these changes make sense?

PR #811, opened 2024-05-09
Remove type modifier decoding from Column object

### LINKED ISSUE #450: Refactor: move modifier parsing functions to TypeInfo
Currently, the Column object contains convoluted ad-hoc code to be able to parse the type modifier value into something to display. See #449 for some details.

Create equivalent typmodout functions and attach them to builtin data types, giving the possibility to extension types to use theirs as well.

dvarrazzo:
Type modifier decoding is implemented in the `Column` class. The code had to know how to decode the fmod value according to different types, therefore it is a maze of ifs and cannot be extended if not changing the code.

Move the parsing to the `TypeInfo` object, and within it to a `TypeModifier` sub-object. This way the code to decode the fmod is automatically dispatched according to the type. Different `TypeModifier` subclasses can parse the different Postgres types that support modifiers.

Close #450.


dlax (approved):
I'm not very familiar with this part of the code base but those changes look good to me overall.

PR #822, opened 2024-05-14
Add instructions for pgbouncer simple query protocol

### LINKED ISSUE #817: enhancement: Support for pgbouncer SHOW STATS (disabling Extended Query Protocol)
I'm connecting to pgbouncer and trying to perform a `SHOW STATS` (which will be answered by pgbouncer itself, not forwarded to Postgresql). 

It's erroring out with `ProtocolViolation: extended query protocol not supported by admin console`, which is an error generated by pgbouncer. 

As the pgbouncer documentation explains in the [Admin console section]: 

> The admin console currently only supports the simple query protocol. Some drivers use the extended query protocol for all commands; these drivers will not work for this.

I haven't seen any way to "force" simple query protocol in psycopg 3.x. See [my stackoverflow question]

 so I guess here I'm requesting two things

* Update documentation to write somewhere about Extended Query Protocol / Simple Query Protocol , since I don't think it's mentioned , at least googling `site:https://www.psycopg.org/psycopg3 "extended query protocol"` gives no results. Maybe at least mention that psycopg3 is only "Extended Query Protocol" (if that is indeed the case)  
* Add support for simple query protocol for specific commands maybe `cur.execute("", force_simple_query_protocol=True)`, 

[my stackoverflow question]: https://stackoverflow.com/questions/78472610/in-psycopg-3-x-is-it-possible-to-disable-the-extended-query-protocol?noredirect=1#comment138345595_78472610
[Admin console section]: https://www.pgbouncer.org/usage.html#admin-console

### LINKED ISSUE #819: Document how to connect to pgbouncer
Pgbouncer requires the use of the simple query protocol. It can be forced by using a `ClientCursor` but it requires autocommit because commit/rollback are currently sent on advanced protocol.

Add documentation do mention the thing.

see #808, #817

ecerulm:
Closes #819
Closes #817

### DIFF
--- a/docs/advanced/cursors.rst
+++ b/docs/advanced/cursors.rst
@@ -125,6 +125,23 @@ as argument.
     conn = psycopg.connect(DSN)
     cur = psycopg.ClientCursor(conn)
 
+
+You can also use `~ClientCursor` with autocommit mode to disable Extended
+Query Protocol and use Simple Query Protocol instead. This is useful for
+example when issuing `SHOW STATS` to a `pgbouncer` for example, since the
+`Pgbouncer Admin Console  <https://www.pgbouncer.org/usage.html#admin-console>`_
+only support Simple Query Protocol.
+
+.. code:: python
+
+    from psycopg import connect, ClientCursor
+
+    conn = psycopg.connect(DSN, cursor_factory=ClientCursor, autocommit=True)
+    cur = conn.cursor()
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you! Merging this as it is, then figuring out what to do if we implement this change for which we wouldn't require autocommit anymore.

PR #824, opened 2024-05-18
chore: Updates action versions and add dependabot to check monthly for updates

stumpylog:
Updates the actions which could be done with little to minimal changes.. Includes a monthly dependabot check for actions now.

This is motivated by recent updates to the Node 16 availability for runners: https://github.blog/changelog/2024-05-17-updated-dates-for-actions-runner-using-node20-instead-of-node16-by-default/

Changelogs for review:

- https://github.com/actions/cache/releases/tag/v4.0.0
- https://github.com/actions/upload-artifact/releases/tag/v4.0.0
  - Required the most changes due to how uploads work now.  Each upload needs a distinct name now, so each build produces a new artifact
  - See also https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md
  - If desired, I can attempt to merge them all at the end
- https://github.com/docker/setup-qemu-action/releases/tag/v3.0.0
- https://github.com/peter-evans/repository-dispatch/releases
- https://github.com/pypa/cibuildwheel/releases/tag/v2.17.0

Other possible updates:
- [pypa/cibuildwheel@v2.18.0](https://github.com/pypa/cibuildwheel/releases)
  - musllinux was updated to default to 1.2 (1.1 / Alpine 3.12 is long EoL now).  This would require build script updates, etc and might be considered breaking
  - Seemed to be something with MacOS too, it doesn't finish repairing the wheel

Builds seem to be fine:
- Build binary packages: https://github.com/stumpylog/psycopg/actions/runs/9136339177
- Build pool packages: https://github.com/stumpylog/psycopg/actions/runs/9136644822
- Build source packages: https://github.com/stumpylog/psycopg/actions/runs/9136645187

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+# https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem
+
+version: 2
+updates:
+
+  # Enable updates for GitHub Actions
+  - package-ecosystem: "github-actions"
+    target-branch: "master"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"
+    groups:
+      actions:
+        update-types:
+          - "major"
+          - "minor"
+          - "patch"
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -245,3 +245,17 @@ jobs:
 
 
   # }}}
+
+  merge:
+    runs-on: ubuntu-latest
+    needs:
+      - linux
+      - macos-14
+      - macos-12
+      - windows
+    steps:
+      - name: Merge Artifacts
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: psycopg-binary-artifact
+          delete-merged: true
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -27,14 +27,14 @@ jobs:
 
       - name: Set up QEMU for multi-arch build
         # Check https://github.com/docker/setup-qemu-action for newer versions.
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
         with:
           # Note: 6.2.0 is buggy: make sure to avoid it.
           # See https://github.com/pypa/cibuildwheel/issues/1250
           image: tonistiigi/binfmt:qemu-v7.0.0
 
       - name: Cache libpq build
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: /tmp/libpq.build
           key: libpq-${{ matrix.platform }}-${{ matrix.arch }}-${{ env.LIBPQ_VERSION }}-${{ env.OPENSSL_VERSION }}
@@ -43,7 +43,7 @@ jobs:
         run: python3 ./tools/build/copy_to_binary.py
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.5
+        uses: pypa/cibuildwheel@v2.17.0
         with:
           package-dir: psycopg_binary
         env:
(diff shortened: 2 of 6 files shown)


dvarrazzo:
I see that you addressed https://github.com/actions/upload-artifact/issues/478, that's good thank you very much.

Can I see the result of an upload? I would like to know if this will result in dozens of files to download separately or if it still possible to download all of them with one click. In psycopg 2 we had a script to do that, so if it becomes a need in psycopg 3 too we can resume it and adapt it. It was dropped in this commit: https://github.com/psycopg/psycopg2/commit/0b01ded426b7423621af68d9fc33e532a5bda4a7

dvarrazzo:
Hello, thank you, yes, I found [your job](https://github.com/stumpylog/psycopg/actions/runs/9136339177).

You can understand well that it is not feasible that to make a release I should click and download manually all these items. And I am not interested in a workflow pushing artifacts automatically from github to pypi: it's too dangerous.

I would be ok to either have all the artifacts packaged in a single archive or to have a script to download all of them.

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
...
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"
+    labels:
+      - "dependencies"
+    groups:
+      actions:
+        update-types:
+          - "major"
+          - "minor"
+          - "patch"
```
dvarrazzo, line 20: Please make sure that there is a newline at the end of the file.

stumpylog:
Yep, I figured that might be the case.  I'm [testing](https://github.com/stumpylog/psycopg/actions/runs/9142422485) what the merge will look like with the worse case workflow.  That will give each workflow a single artifact/zip.

.github/dependabot.yml
stumpylog: Fixed!

stumpylog:
Alright, all the test runs in `master` are still going as a double check, but I've updated to merge all produced artifacts together at the end.  It will look like [this](https://github.com/stumpylog/psycopg/actions/runs/9142422485), which is pretty much the same result as currently.

I chose the `if: always()` so even if some jobs fail, any produced artifacts still end up merged together.

dvarrazzo:
>
> I chose the if: always() so even if some jobs fail, any produced
> artifacts still end up merged together.
>
This seems a dangerous idea. We might end up making a release with a file
missing. I think that failing hard and rerun the build is a better option,
no?

>

stumpylog:
Certainly up to you, I'm not familiar enough with the release process here.  My thought was someone might want to download the artifact to investigate something, and it would be clear enough that jobs failed it wouldn't be used.  But I'm happen to change that

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
+# https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem
+
+version: 2
+updates:
+
+  # Enable updates for GitHub Actions
+  - package-ecosystem: "github-actions"
+    target-branch: "dev"
```
dvarrazzo, line 8: Sorry, just checked this... is this correct? Shouldn't it be `master`?

.github/workflows/packages-bin.yml
```diff
@@ -235,9 +238,24 @@ jobs:
...
+      - linux
+      - macos-14
+      - macos-12
+      - windows
+    steps:
+      - name: Merge Artifacts
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: psycopg-binary-artifact
+          delete-merged: true
```
dvarrazzo, line 261: Please add an end of line to every file.

dvarrazzo:
Hello! Sorry, I was about to merge this MR because, looking at your artifacts it seems to do the right thing. So I've force-pushed to squash two temporary commits, but then noticed a couple of issues. So please pull before changing them.

Thank you very much!

stumpylog:
No problem, I'll get those fixed up when I'm back at a computer

PR #831, opened 2024-06-02
really drop python<=37 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.


tests/dbapi20.py
```diff
@@ -554,7 +554,7 @@ def _populate(self):
             tests.
         '''
         populate = [
-            "insert into %sbooze values ('%s')" % (self.table_prefix,s)
```
dvarrazzo, line 557: This is not a set of changes I would like to merge. You are simply using a linter that we don't use and we don't have reasons to introduce it in our codebase. So I prefer to close this MR here.

dvarrazzo:
There's a bit of missing info, which I sent by email but maybe github doesn't support it anymore. Most of these changes are already implemented in the [uniform-annotations](https://github.com/psycopg/psycopg/tree/uniform-annotations) branch, which is soon to be merged. Plus, it doesn't break Python 3.8.

PR #832, opened 2024-06-03
Uniform annotations

dvarrazzo:
This is a large but dumb MR, which, fundamentally, drops the usage of several `typing` objects that in more modern Python versions can be done without:

- Avoid using `typing.Dict`, `typing.List` and other aliases, use the builtin `dict`, `list` etc.
- Avoid using `typing.Union` (use `|` instead) and `typing.Optional` (use `| None` instead).

The cleanup is slightly complicated by the need of supporting Python 3.8, for which we must use strings instead of normal expressions in non-annotation context (for instance in `TypeAlias` definitions), but, because it was the version distributed in Ubuntu 20.04, I still don't want to drop it in 3.2.


asqui:
Out of interest, did you use any standard tooling to apply these changes (e.g. something like pyupgrade)?

psycopg/psycopg/pq/_pq_ctypes.pyi
```diff
@@ -175,6 +175,8 @@ def PQexec(arg1: PGconn_struct | None, arg2: bytes) -> PGresult_struct: ...
 def PQexecParams(arg1: PGconn_struct | None, arg2: bytes, arg3: int, arg4: _Pointer[c_uint], arg5: _Pointer[c_char_p], arg6: _Pointer[c_int], arg7: _Pointer[c_int], arg8: int) -> PGresult_struct: ...
 def PQdescribePrepared(arg1: PGconn_struct | None, arg2: bytes) -> PGresult_struct: ...
 def PQdescribePortal(arg1: PGconn_struct | None, arg2: bytes) -> PGresult_struct: ...
+def PQclosePrepared(arg1: PGconn_struct | None, arg2: bytes) -> PGresult_struct: ...
+def PQclosePortal(arg1: PGconn_struct | None, arg2: bytes) -> PGresult_struct: ...
```
dlax, line 179: These additions (and the ones below in that file) seem to be the result of respective lines having been dropped in the previous commit (by mistake, I suspect).

dlax (commented):
LGTM; thanks for upgrading the code base!

dlax:
By the way, `pyupgrade --py38-plus --keep-percent-format $(git ls-files **.py)` suggests a few more changes that might be worth applying. I know you rejected the idea of using pyupgrade in another PR so I'm not suggesting we use it in CI or so; but maybe getting completely upgraded now is worth?

I can suggest fixup commits for that in another PR if that helps, let me know.

psycopg/psycopg/pq/_pq_ctypes.pyi
dlax: Since everything is kept in place at the end, it might not be worth fixing the git history; so feel free to ignore my comment!

dvarrazzo:
> Out of interest, did you use any standard tooling to apply these changes (e.g. something like pyupgrade)?

No, I mostly made changes via `sed -i` and running the test suite to find breakages.

Before I made this MR, but after having pushed a branch with a WIP, someone had a pyupgrade run (#831) but they committed changes [without even looking at what they were doing](https://github.com/psycopg/psycopg/pull/831/files#diff-ec162367f2b494baea5f25f3d81908ca9edbbae2b6f129fd2b3bb3c89d626a20L3), which is not a great way to make me like a MR.

I understand that the change didn't exclude all the usages of the `typing` aliases, because Python 3.8 still needs them outside of annotations (for instance in `TypeAlias`), but I have preferred to fix this problem by specifying TypeAliases as strings and drop the typing aliases in a way that might have been done by pyupgrade probably using Python > 3.8 as parameter.

dvarrazzo:
@dlax 

> but maybe getting completely upgraded now is worth?

Maybe it is, thank you for trying to run pyupgrade without changing every single percent formatting. I'll take a look at you MR :slightly_smiling_face:

PR #834, opened 2024-06-10
refactor: make PGcancelConn.error_message return bytes

### LINKED ISSUE #781: Bad decoding of connection cancel error message
Hi @dlax 

Sorry, something I overlooked in previous reviews:

https://github.com/psycopg/psycopg/blob/3ce90657625636afab287185347d3d5d1c64d65a/psycopg/psycopg/pq/pq_ctypes.py#L961-L963

This can fail if someone connects to a server with non-English localization and encoding non-utf8 (situations like https://github.com/psycopg/psycopg2/issues/1442)

This method should return the unchanged message as bytes, and the decoding should happen keeping into account the server encoding. The pattern is similar to what you can see in the `connect()` generator.

Looking at the connection, one problem we have is that the connection can tell its encoding via `parameter_status()`, which is then used to decode the error message. I assume that the cancel connection doesn't have such luxury. What we could do instead is to add an encoding attribute to PGcancelConn, defaulting to utf8, and assign it in `PGconn.cancel_conn()` e.g. with `PGcancelConn(rv, encoding=self.parameter_status(b"client_encoding"))`.

The reason why I would make the decode happen externally is for symmetry with the PGconn and PGresult message attributes, both returning bytes (the result doesn't have an encoding either, so it's taken from the connection too). But the most important thing is that I prefer that we return a message decoded the wrong way rather than we crash on displaying an error and so we clobber another error.

Yet another way to skin this cat is to add an `encoding` parameter to both PGconn and PGcancelConn, so that we may treat the objects the same `pq.error_message()` branch

https://github.com/psycopg/psycopg/blob/3ce90657625636afab287185347d3d5d1c64d65a/psycopg/psycopg/pq/misc.py#L94-L98

and we can extend the `error_message()` input type to accept a PGcancelConn too.

dvarrazzo:
This is more consistent with PGconn and PGresult.

There doesn't seem to be the need of extending pq.misc.error_message as we never pass it a PGcancelConn object.

Close #781

The MR also includes refactoring of the procedure of getting error messages from various objects as text. Some generic function taking different input types and returning a string have been dropped (or just reduced to a façade to keep backward compatibility) and now all relevant objects have grown a method `get_error_message(encoding: str = ...)` to return an error message decoded either in an encoding known by the object itself (e.g. PGconn, PGcancelConn) or one obtained from the context (e.g. PGresult, broken PGconn).


dvarrazzo:
I am doing a little refactoring to get rid of functions such as `pgconn_encoding()` and `error_message()`. It makes sense to do that in the context of this MR because pretty much it adds support to return an error from PGconn and PGresult as text.

dlax:
> I am doing a little refactoring to get rid of functions such as `pgconn_encoding()` and `error_message()`. It makes sense to do that in the context of this MR because pretty much it adds support to return an error from PGconn and PGresult as text.


Fine by me, thank you!

dvarrazzo:
Ready for review! MR description updated with the added content.

psycopg/psycopg/errors.py
```diff
@@ -92,6 +93,9 @@ def finish(self) -> None:
     def reset(self) -> NoReturn:
         self._raise()
 
+    def get_error_message(self, encoding: str = "") -> str:
+        return "the conenection is closed"
```
dlax, line 97:
```suggestion
        return "the connection is closed"
```

psycopg/psycopg/errors.py
```diff
@@ -544,15 +548,9 @@ def lookup(sqlstate: str) -> type[Error]:
...
-    from psycopg import pq
-
     state = result.error_field(DiagnosticField.SQLSTATE) or b""
-    cls = _class_for_state(state.decode("ascii"))
-    return cls(
-        pq.error_message(result, encoding=encoding),
-        info=result,
-        encoding=encoding,
-    )
+    cls = _class_for_state(state.decode("utf-8", "replace"))
```
dlax, line 552:
```suggestion
    cls = _class_for_state(state.decode(encoding, "replace"))
```
? was "utf-8" here intended?

dlax (commented):
LGTM overall; nice refactoring!

psycopg/psycopg/errors.py
dvarrazzo: This value is only supposed to be a 5 chars sqlstate codes, and any encoding would do to decode it. It was "ascii" before, and would have failed in case for networking reason we end up with random characters here, because there is no "replace".

So I thought, instead of failing with a decoding error here, let's keep going down the rabbit hole and see what get raised, as it might be more informative (`_class_for_state()` would return `DatabaseError` in case you throw random values at it).

dlax (approved):

PR #835, opened 2024-06-13
feat(capabilities): add has_send_close_prepared()

dvarrazzo:
This is a capability we need internally. `has_pgbouncer_prepared()` is pretty much just an alias of it, more outwards-facing. I wonder if we want to keep them both.

### DIFF
--- a/docs/advanced/prepare.rst
+++ b/docs/advanced/prepare.rst
@@ -68,8 +68,8 @@ PgBouncer__ middleware, using the following caveats:
 - PgBouncer version must be version `1.22`__ or newer.
 - PgBouncer `max_prepared_statements`__ must be greater than 0.
 - The libpq version on the client must be from PostgreSQL 17 or newer
-  (you can check the `~Capabilities.has_pgbouncer_prepared` capability to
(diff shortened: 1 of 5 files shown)


dlax (approved):
> I wonder if we want to keep them both.

If we were to drop one (seems reasonable), I'd say we should rather keep `has_send_close_prepared` in order to avoid confusion that this might be pgbouncer-specific.

dvarrazzo:
> If we were to drop one (seems reasonable), I'd say we should rather keep has_send_close_prepared in order to avoid confusion that this might be pgbouncer-specific.

I assume we can tweak the docs and highlight that if pgbouncer users want to check compatibility they should rather check `has_send_close_prepared()`, yes.

dvarrazzo:
As per last commit:

```
chore(capability): drop has_pgbouncer_prepared()

Document to use the `has_send_closed_prepared()` capability instead,
which is a necessary condition for PgBouncer support, and doesn't give
the impression to be sufficient.
```

dlax (approved):

PR #839, opened 2024-06-18
Add RawServerCursor class

dvarrazzo:
I had this changeset in a branch and forgot to make a MR of it


dlax (approved):
Maybe add a note to the changelog?

dvarrazzo:
> Maybe add a note to the changelog?

There is this entry:

https://github.com/psycopg/psycopg/blob/69ad4112cf25d64c3dc2c1272e4a1e0a860141b0/docs/news.rst?plain=1#L21-L22

It just references #560, which has most of the changeset. I will add a reference to this MR too for completeness.

PR #844, opened 2024-06-24
Avoid to dump [multi]range as text when binary is requested

dvarrazzo:
If a range has no type information (e.g. is empty and it is a base class, not a specific range subtype), `upgrade()` would have returned a text dumper. This would have resulted in protocol violation, such as requesting unhealthy amount of memory to the backend... and it was tested too!

Restoring sanity has a drawback. If we have an array of ranges, none of which has type information, we now fail binary dumping. Previously this worked, provided that a cast `::int8range[]` was provided, but I think with an extreme stretch of the protocol: we were producing binary arrays containing text representation of the ranges.

I don't think the latter is quite valid. So I have a preference for dropping the feature in exchange of not violating the protocol. We trade a better dumping of the basic type (an xfail less there) in exchange of losing the ability of dumping an array of ranges, all empty, using a cast, in binary format. Binary dump works using a subclass such as `Int8Range` instead of `Range`.

The lost use case is so marginal that I don't feel the need to document it. We don't even have documentation for the `Range` subclasses.

### DIFF
--- a/tests/types/test_multirange.py
+++ b/tests/types/test_multirange.py
@@ -5,7 +5,6 @@ from decimal import Decimal
 import pytest
 
 from psycopg import pq, sql
-from psycopg import errors as e
 from psycopg.adapt import PyFormat
 from psycopg.types.range import Range
 from psycopg.types import multirange
@@ -196,22 +195,21 @@ def test_dump_builtin_empty_wrapper(conn, wrapper, fmt_in):
     assert rec[2] == dumper.oid
 
 
-@pytest.mark.parametrize("pgtype", mr_names)
-@pytest.mark.parametrize(
-    "fmt_in",
-    [
-        PyFormat.AUTO,
-        PyFormat.TEXT,
-        # There are many ways to work around this (use text, use a cast on the
-        # placeholder, use specific Range subclasses).
-        pytest.param(
-            PyFormat.BINARY,
-            marks=pytest.mark.xfail(
-                reason="can't dump array of untypes binary multirange without cast"
-            ),
+pyformat_array = [
+    PyFormat.AUTO,
+    PyFormat.TEXT,
+    # There are ways to work around this (use text, use specific Range subclasses).
+    pytest.param(
+        PyFormat.BINARY,
+        marks=pytest.mark.xfail(
+            reason="can't dump array of untyped binary multirange without cast"
         ),
-    ],
-)
+    ),
+]
+
+
+@pytest.mark.parametrize("pgtype", mr_names)
+@pytest.mark.parametrize("fmt_in", pyformat_array)
 def test_dump_builtin_array(conn, pgtype, fmt_in):
     mr1 = Multirange()  # type: ignore[var-annotated]
     mr2 = Multirange([Range(bounds="()")])  # type: ignore[var-annotated]
@@ -223,7 +221,7 @@ def test_dump_builtin_array(conn, pgtype, fmt_in):
 
 
 @pytest.mark.parametrize("pgtype", mr_names)
-@pytest.mark.parametrize("fmt_in", PyFormat)
+@pytest.mark.parametrize("fmt_in", pyformat_array)
 def test_dump_builtin_array_with_cast(conn, pgtype, fmt_in):
     mr1 = Multirange()  # type: ignore[var-annotated]
     mr2 = Multirange([Range(bounds="()")])  # type: ignore[var-annotated]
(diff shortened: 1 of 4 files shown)


tests/types/test_multirange.py
```diff
@@ -196,22 +195,21 @@ def test_dump_builtin_empty_wrapper(conn, wrapper, fmt_in):
...
-                reason="can't dump array of untypes binary multirange without cast"
-            ),
+pyformat_array = [
+    PyFormat.AUTO,
+    PyFormat.TEXT,
+    # There are ways to work around this (use text, use specific Range subclasses).
+    pytest.param(
+        PyFormat.BINARY,
+        marks=pytest.mark.xfail(
+            reason="can't dump array of untyped binary multirange without cast"
```
dlax, line 205: When you use `xfail`, it here means that "the code produces an exception, and it's expected"? Or is it something that should work but currently does not (sort of a fixme)?

dvarrazzo: Hard to say. The reason why this particular corner case doesn't work is a "collaboration" between inability to provide enough information to postgres and postgres not understanding the partial information received.

Postgres doesn't accept an array of unknown items in binary (but it does in text). If we pass an array of ranges, and at least one of these ranges has a type, we can infer the type. But if none of them has a type, we need to punt... Currently we pass text as array type, but the result is that postgres understand that the binary representation of the range is a text representation, so it fails with something like:

```
$ pytest --runxfail "tests/types/test_range.py::test_dump_builtin_array_with_cast[b-numrange]" 
///
FAILED tests/types/test_range.py::test_dump_builtin_array_with_cast[b-numrange] - psycopg.errors.InvalidTextRepresentation: malformed range literal: ""
```

```
$ pytest --runxfail "tests/types/test_multirange.py::test_dump_builtin_array_with_cast[b-nummultirange]"
...
FAILED tests/types/test_multirange.py::test_dump_builtin_array_with_cast[b-nummultirange] - psycopg.errors.CharacterNotInRepertoire: invalid byte sequence for encoding "UTF8": 0x00
```

The failing query is (for int4):

```
pytest --runxfail "tests/types/test_range.py::test_dump_builtin_array_with_cast[b-int4range]" --pdb
...
(Pdb) print(query)
select array['empty'::int4range, '(,)'::int4range] = %b::int4range[]
(Pdb) print(params)
([Range(empty=True), Range(None, None, '()')],)
```

the dumping system converts this value to:

```
(Pdb) print(cur._query.types[0])
1009
(Pdb) print(cur._query.params[0])
b'\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x19\x00\x00\x00\x02\x00\x00\x00\x01\x00\x00\x00\x01\x01\x00\x00\x00\x01\x18'
```

where 1009 is the text array oid and the binary representation says:

```
\x00\x00\x00\x01: 1 dimension array
\x00\x00\x00\x00: it has no nulls
\x00\x00\x00\x19: the type oid is 0x19 = 25 = text
\x00\x00\x00\x02: the array has 2 items
\x00\x00\x00\x01: array lower bound is 1
\x00\x00\x00\x01: first item is long 1 byte
\x01            : binary representation of empty range
\x00\x00\x00\x01: second item is long 1 byte
\x18'           : binary representation of `(,)` range
```

The situation is better in text because we can say "unknown oid", therefore the dumpers are configured so that `%s` dumps as text by default. The only way to trigger the bug is with these pathological empty-range-only arrays explicitly dumped as binary using `%b`.

The current state is that, if someone really wants to do that, they should use `Int4Range` (or whatever appropriate), instead of `Range`. This is checked by the `test_dump_builtin_array_wrapper` test, which, unlike `test_dump_builtin_array` and `test_dump_builtin_array_with_cast`, don't have xfail.

Setting a breakpoint after the query in this test:
```
pytest "tests/types/test_range.py::test_dump_builtin_array_wrapper[b-Int4Range]"

# The query
(Pdb) print(cur._query.query.decode())
select '{empty,"(,)"}' = $1

# The value
(Pdb) [r1, r2]
[Int4Range(empty=True), Int4Range(None, None, '()')]

# What gets dumped:
(Pdb) cur._query.types[0]
3905  # int4range array oid

(Pdb) print(cur._query.params[0])
b'\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x0f@\x00\x00\x00\x02\x00\x00\x00\x01\x00\x00\x00\x01\x01\x00\x00\x00\x01\x18'

which contains (annotating only the difference):

```
\x00\x00\x00\x01
\x00\x00\x00\x00
\x00\x00\x0f@: 3904: int4range oid
\x00\x00\x00\x02
\x00\x00\x00\x01
\x00\x00\x00\x01
\x01
\x00\x00\x00\x01
\x18
```

So either we ask the user to provide typing information (by using `Int4Range` instead of `Range`) or we ask pretty please to the Postgres developer to allow arrays of unknown types in binary format.

dlax: I see; thank you for the detailed explanation!

PR #847, opened 2024-06-26
Update openssl to 3.3.1

dvarrazzo:
This is a pretty long jump, previously used version was 1.1.1, which is now unsupported.

In the past, bundling openssl 0.9.8 was causing random segfaults when the python-bundled openssl was used too (e.g. requesting an https url via urllib). See https://github.com/psycopg/psycopg2/issues/543. A lot of time has passed, and hopefully the case shouldn't present again. Unfortunately it's not easy to test (I was never able to reproduce the bug above).

### DIFF
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -12,8 +12,10 @@ jobs:
     if: true
 
     env:
+      # Latest release: https://www.postgresql.org/ftp/source/
       LIBPQ_VERSION: "16.3"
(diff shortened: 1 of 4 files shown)


dvarrazzo:
I would release this changeset in 3.1.20, unless anyone thinks it's a bad idea. I understand OpenSSL 1.1.1 [won't receive further updates](https://www.openssl.org/source/index.html) so going 3.x for some version of x seems advisable.

dlax (approved):

PR #877, opened 2024-07-24
Fix warnings

### LINKED ISSUE #867: Build fails with strict-aliasing violations
I tried to build with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

The -Werror=* flags are important to detect cases where the compiler can try to optimize based on assuming UB cannot happen, and miscompile code that has UB in it. strict-aliasing issues are always bad but LTO can make them even worse.

I got this error:
```
2024-07-15 03:37:17,743 root INFO x86_64-pc-linux-gnu-gcc -fno-strict-overflow -Wsign-compare -DNDEBUG -pipe -march=native -fstack-protector-all -O2 -fdiagnostics-color=always -frecord-gcc-switches -flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing -Wformat -Werror=format-security -Werror=implicit-function-declaration -Werror=implicit-int -Werror=int-conversion -Werror=incompatible-pointer-types -DNDEBUG -fPIC -I/usr/include/postgresql-16 -I/usr/include/python3.12 -c psycopg_c/_psycopg.c -o /var/tmp/portage/dev-python/psycopg-3.1.19/work/psycopg-3.1.19-python3_12/build/temp.linux-x86_64-cpython-312/psycopg_c/_psycopg.o
psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg__array_load_binary’:
psycopg_c/_psycopg.c:34509:16: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
34509 | __pyx_v_dimptr = (__pyx_v_buf + (3 * (sizeof(uint32_t))));
      |                ^
In file included from /usr/include/sys/types.h:176,
                 from /usr/include/stdlib.h:514,
                 from /usr/include/python3.12/Python.h:23,
                 from psycopg_c/_psycopg.c:42:
psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg_20IntervalBinaryLoader_cload’:
psycopg_c/_psycopg.c:49642:52: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
49642 | __pyx_v_val = be64toh((((uint64_t *)__pyx_v_bedata)[0]));
      |                                                    ^
psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg_17FloatBinaryDumper_cdump’:
psycopg_c/_psycopg.c:56365:52: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
56365 | __pyx_v_beval = htobe64((((uint64_t *)(&__pyx_v_d))[0]));
      |                                                    ^
psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg_18Float4BinaryDumper_cdump’:
psycopg_c/_psycopg.c:56557:52: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
56557 | __pyx_v_beval = htobe32((((uint32_t *)(&__pyx_v_f))[0]));
      |                                                    ^
psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg_19NumericBinaryLoader_cload’:
psycopg_c/_psycopg.c:57961:20: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
57961 |   __pyx_v_digitptr = (__pyx_v_data + (sizeof(__pyx_v_behead)));
      |                    ^
```

Reported downstream at https://bugs.gentoo.org/935401
Full build log: [build.log](https://github.com/user-attachments/files/16231403/build.log)

dvarrazzo:
Drop a few warnings compiling the C extension.

Close #867

### DIFF
--- a/psycopg_c/psycopg_c/types/datetime.pyx
+++ b/psycopg_c/psycopg_c/types/datetime.pyx
@@ -971,11 +971,11 @@ cdef class IntervalBinaryLoader(CLoader):
     format = PQ_BINARY
 
     cdef object cload(self, const char *data, size_t length):
-        cdef int32_t bedata[4]
+        cdef int64_t bedata[2]
         memcpy(&bedata, data, sizeof(bedata))
-        cdef int64_t val = endian.be64toh((<uint64_t *>bedata)[0])
-        cdef int32_t days = endian.be32toh(bedata[2])
-        cdef int32_t months = endian.be32toh(bedata[3])
+        cdef int64_t val = endian.be64toh(bedata[0])
+        cdef int32_t days = endian.be32toh((<uint32_t *>bedata)[2])
+        cdef int32_t months = endian.be32toh((<uint32_t *>bedata)[3])
 
         cdef int years
         with cython.cdivision(True):
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Configure memory alignment tests
         if: ${{ matrix.ext == 'align' }}
         run: |
-          echo "CFLAGS=-fsanitize=undefined" >> $GITHUB_ENV
+          echo "CFLAGS=-fsanitize=undefined -Werror=strict-aliasing -Werror=odr -Werror=lto-type-mismatch" >> $GITHUB_ENV
           echo "UBSAN_OPTIONS=halt_on_error=1" >> $GITHUB_ENV
           echo "PYTEST_ADDOPTS=-v" >> $GITHUB_ENV
 
--- a/psycopg_c/psycopg_c/pq/pgconn.pyx
+++ b/psycopg_c/psycopg_c/pq/pgconn.pyx
@@ -710,7 +710,7 @@ cdef int _call_int(PGconn pgconn, conn_int_f func) except -2:
     return func(pgconn._pgconn_ptr)
 
 
-cdef void notice_receiver(void *arg, const libpq.PGresult *res_ptr) with gil:
+cdef void notice_receiver(void *arg, const libpq.PGresult *res_ptr) noexcept with gil:
     cdef PGconn pgconn = <object>arg
     if pgconn.notice_handler is None:
         return
--- a/psycopg_c/psycopg_c/types/array.pyx
+++ b/psycopg_c/psycopg_c/types/array.pyx
@@ -232,7 +232,7 @@ cdef object _array_load_binary(
     cdef Py_ssize_t[MAXDIM] dims
     cdef int i
     cdef uint32_t bedata
-    cdef char *dimptr = buf + 3 * sizeof(uint32_t)
+    cdef const char *dimptr = buf + 3 * sizeof(uint32_t)
     for i in range(ndims):
         # Every dimension is dim, lower bound
         memcpy(&bedata, dimptr, sizeof(bedata))
--- a/psycopg_c/psycopg_c/types/numeric.pyx
+++ b/psycopg_c/psycopg_c/types/numeric.pyx
@@ -393,10 +393,9 @@ cdef class Float4BinaryLoader(CLoader):
         cdef uint32_t bedata
         memcpy(&bedata, data, sizeof(bedata))
         cdef uint32_t asint = endian.be32toh(bedata)
-        # avoid warning:
-        # dereferencing type-punned pointer will break strict-aliasing rules
-        cdef char *swp = <char *>&asint
-        return PyFloat_FromDouble((<float *>swp)[0])
+        cdef float f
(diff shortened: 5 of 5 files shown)


psycopg_c/psycopg_c/types/datetime.pyx
```diff
@@ -971,11 +971,11 @@ cdef class IntervalBinaryLoader(CLoader):
...
 
     cdef object cload(self, const char *data, size_t length):
-        cdef int32_t bedata[4]
+        cdef int64_t bedata[2]
         memcpy(&bedata, data, sizeof(bedata))
-        cdef int64_t val = endian.be64toh((<uint64_t *>bedata)[0])
-        cdef int32_t days = endian.be32toh(bedata[2])
-        cdef int32_t months = endian.be32toh(bedata[3])
+        cdef int64_t val = endian.be64toh(bedata[0])
+        cdef int32_t days = endian.be32toh((<uint32_t *>bedata)[2])
```
SoapGentoo, line 977: @dvarrazzo I think these lines are still UB, you cannot cast a `int64_t*` to a `uint32_t*` and dereference it.

SoapGentoo: here's my suggestion:
declare `bedata` as an object of type
```
struct {
  uint64_t val;
  uint32_t days;
  uint32_t months;
};
```
and then initialize it with your call to `memcpy`

dvarrazzo: @SoapGentoo I understand that the problem comes from unaligned memory. So I understand if aligning 64 bits on 32 bits boundaries can get unaligned, but aligning 32 bits on the 64 bits boundary doesn't. The compiler doesn't report any error. Is there a stricter check you suggest to add?

I'm not sure about the struct; if any, I would use two separate memcpy:

```diff
diff --git a/psycopg_c/psycopg_c/types/datetime.pyx b/psycopg_c/psycopg_c/types/datetime.pyx
index 44b227ed..e2a6ac8a 100644
--- a/psycopg_c/psycopg_c/types/datetime.pyx
+++ b/psycopg_c/psycopg_c/types/datetime.pyx
@@ -971,11 +971,13 @@ cdef class IntervalBinaryLoader(CLoader):
     format = PQ_BINARY
 
     cdef object cload(self, const char *data, size_t length):
-        cdef int64_t bedata[2]
-        memcpy(&bedata, data, sizeof(bedata))
-        cdef int64_t val = endian.be64toh(bedata[0])
-        cdef int32_t days = endian.be32toh((<uint32_t *>bedata)[2])
-        cdef int32_t months = endian.be32toh((<uint32_t *>bedata)[3])
+        cdef int64_t beval
+        cdef int32_t bedm[2]
+        memcpy(&beval, data, sizeof(beval))
+        memcpy(bedm, data + sizeof(beval), sizeof(bedm))
+        cdef int64_t val = endian.be64toh(beval)
+        cdef int32_t days = endian.be32toh(bedm[0])
+        cdef int32_t months = endian.be32toh(bedm[1])
 
         cdef int years
         with cython.cdivision(True):
```

(this actually makes me notice that the original line has an error: it should have been `memcpy(bedata, ...` not `memcpy(&bedata, ...`. How does it even work?)

dvarrazzo: I would be more comfortable making this change than using a struct. I don't know the rules for padding in their most intimate details, but what guarantees that some compiler won't align the data to 64 bits and add 4 bytes of padding between days and months?

SoapGentoo: @dvarrazzo the alignment argument is correct, but `-Wstrict-aliasing` is more than just alignment. You cannot in general cast a `T*` to a `U*` and then dereference it, unless a very specific and constrained set of circumstances is true (for instance, `U = void` or `U = {char, unsigned char}` or some other rules that don't matter here).

Your double-copy is just as fine, if not clearer, so I would go with that route.
`memcpy(bedata, ...` works, because in C arrays often decay into pointers, and since the first argument of `memcpy` is a `void*`, the compiler implicitly converts the array to a pointer to its first element (which is absolutely fine and well defined).

dvarrazzo: Thank you for the feedback, @SoapGentoo. I've pushed a change in #879 for testing.

PR #878, opened 2024-07-25
Update async.rst

Blackeagle52:
Adds an example on how to use `connection.notifes` in combination with `connection.add_notification_handler`, if you want to use the connection based upon the notification.

### DIFF
--- a/docs/advanced/async.rst
+++ b/docs/advanced/async.rst
@@ -335,7 +335,7 @@ Notifications are received as instances of `Notify`. If you are reserving a
 connection only to receive notifications, the simplest way is to consume the
 `Connection.notifies` generator. The generator can be stopped using
 `!close()`. Starting from Psycopg 3.2, the method supports options to receive
-notifications only for a certain time or up to a certain number.
+notifications only for a certain time and/or up to a certain number.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Blackeagle52:
Sorry for creating this PR. I wanted to create a PR within my fork, to discuss the changes with a colleague of mine, and I didn't pay attention to the master.

docs/advanced/async.rst
```diff
@@ -393,6 +397,42 @@ received immediately, but only during a connection operation, such as a query.
     # got this: Notify(channel='mychan', payload='hey', pid=961823)
     # (1,)
 
+If you need to use the connection to handle :sql:`NOTIFY` in a timely manner
+you can combine the two methods.
```
dvarrazzo, line 401: What you describe is your example is not receiving notifications in a timely matter. The notification will be only received when the client is ready to receive it, not when the server sends it.

Blackeagle52: >What you describe is your example is not receiving notifications in a timely matter. The notification will be only received when the client is ready to receive it, not when the server sends it.

That's up for discussion. 

If you need to use the connection for other queries, combining the handler, the generator, and a queue, you will receive notifications as soon as possible. The notifications are stored, and available by First in, first out.
It's the speed (and content) of what the client does based on the notify which will determine whether a client handles the notification in a timely manner.

If someone would receive a Notify every second, and needs more than a second per Notify, he'll never catch up. It doesn't matter what available method the client uses.

But your input does show I need to improve the text. I'll try to clarify. Thanks you for the input.

PR #899, opened 2024-08-27
fix: make make_conninfo()'s first argument positional only

dlax:
From commit c170851ab85f63935c0709ff0bfd7c1e7e9dd0c0, invoking `conninfo.make_conninfo(**kwargs)` (with no `conninfo` argument) produces a typing (Mypy) error:

    error: Argument 1 to "make_conninfo" has incompatible type "**dict[str, str | int]"; expected "str"  [arg-type]

(here with `kwargs: dict[str, str | int]`).

Not specifying `conninfo=""`, the default, is arguably a common pattern and the only workaround on user side is to change the call to `make_conninfo("", **kwargs)`, which is quite cumbersome.

By making `make_conninfo()'`s `conninfo` argument positional only, the typing error is resolved.

Added test serves as a non-regression, but mostly for typing.

### DIFF
--- a/psycopg/psycopg/conninfo.py
+++ b/psycopg/psycopg/conninfo.py
@@ -25,7 +25,7 @@ conninfo_attempts_async = _conninfo_attempts_async.conninfo_attempts_async
 _DEFAULT_CONNECT_TIMEOUT = 130
 
 
-def make_conninfo(conninfo: str = "", **kwargs: ConnParam) -> str:
+def make_conninfo(conninfo: str = "", /, **kwargs: ConnParam) -> str:
     """
     Merge a string and keyword params into a single conninfo string.
 
--- a/tests/test_conninfo.py
+++ b/tests/test_conninfo.py
@@ -39,6 +39,17 @@ def test_make_conninfo(conninfo, kwargs, exp):
     assert conninfo_to_dict(out) == conninfo_to_dict(exp)
 
 
+def test_make_conninfo_kwargs_only():
+    """Check that typing is correct when calling with kwargs only."""
+    kwargs: dict[str, str | int] = {"dbname": "foo", "port": 15432}
+    assert make_conninfo(**kwargs) == "dbname=foo port=15432"
+
+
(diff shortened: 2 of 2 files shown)


dvarrazzo:
Uhm... My fear is that this a non-backward compatible change. These functions [are documented](https://www.psycopg.org/psycopg3/docs/api/conninfo.html) taking a `conninfo` argument, so someone might have used that name.

I wonder if we can solve the problem in some other way (specifying the argument as positional-only seems to solve the problem only by accident...)

dlax:
> Uhm... My fear is that this a non-backward compatible change. These functions [are documented](https://www.psycopg.org/psycopg3/docs/api/conninfo.html) taking a `conninfo` argument, so someone might have used that name.

Yes, that's true, I know. In fact, I hesitated submitting this change some times ago while fixing this issue the first time I encountered it because of this bw compat consideration. But now, after having been forced to fix it in more places, I changed my mind and am suggesting this for discussion :)

> I wonder if we can solve the problem in some other way (specifying the argument as positional-only seems to solve the problem only by accident...)

There is also the [`Unpack`+`TypedDict` pattern](https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs) but I think it cannot be used in this case because the dict for `**kwargs` here is open-ended, and there seems to be no way to tell type checker that "this typed dict has no `conninfo` key" (there are `Required`/`NotRequired` qualifiers, but no `Forbidden`). Also the "positional-only" pattern is documented there.

dvarrazzo:
Let me say something silly: what if we annotate conninfo to be, optionally, an int too (just to raise an explicit TypeError if someone tries to use it that way?)

It's a little white lie that maybe would make mypy happy without breaking compatibility.

dlax:
That's too "silly" for me, sorry :)

dlax:
Yes, I know. Your position on bw compatibility is right and horourable. So I'm not insisting on this; let's just close this PR.

I wish we had better coverage concerning "typing" breakage. In my experience, this can be improved by adding more type hints in the test suite, but that's a bit of a chore...

PR #906, opened 2024-09-14
docs: replace fastapi events with lifespan

VinzenzKlass:
The FastAPI `@app.on_event("startup")` and  `@app.on_event("shutdown")` are deprecated. This PR simply replaces their usage in the AsyncConnectionPool section with the now recommened Lifespan.


dvarrazzo:
Yes, I have noticed that the example was out of date respect as fastapi pattern.

I am not sure anymore about the organization of those examples. However, re-reading that section, your changes leave the meaning of the entire section unchanged and they are a modernization of the pattern shown. So yes, I think it's a good change.

Thank you very much!

PR #915, opened 2024-09-26
test: mark more tests as 'timing'

autra:
# Description 

When the machine is under load, any tests relying on a timeout or a sleep is subject to fail. Let's mark some of them as 'timing'.

I followed the recommandation in [a comment on a similar issue](https://github.com/psycopg/psycopg/issues/692#issuecomment-1838118714).

Context: these failing tests have been discovered when automatically testing a geos update on nixpkgs. The process rebuilds all geos dependencies, and put my laptop under a quite high load. That being said, I reproduce sometimes even when limiting the number of workers `nixpkgs-review` is able to use.

Thanks for the awesome lib!

### DIFF
--- a/tests/test_notify.py
+++ b/tests/test_notify.py
@@ -133,6 +133,7 @@ def test_notify_timeout(conn_cls, conn, dsn):
 
 
 @pytest.mark.slow
+@pytest.mark.timing
 def test_notify_timeout_0(conn_cls, conn, dsn):
     conn.set_autocommit(True)
     conn.execute("listen foo")
@@ -149,6 +150,7 @@ def test_notify_timeout_0(conn_cls, conn, dsn):
 
 
 @pytest.mark.slow
+@pytest.mark.timing
 def test_stop_after(conn_cls, conn, dsn):
     conn.set_autocommit(True)
     conn.execute("listen foo")
@@ -175,6 +177,7 @@ def test_stop_after(conn_cls, conn, dsn):
     assert ns[0].payload == "3"
 
 
+@pytest.mark.timing
 def test_stop_after_batch(conn_cls, conn, dsn):
     conn.set_autocommit(True)
     conn.execute("listen foo")
@@ -196,6 +199,7 @@ def test_stop_after_batch(conn_cls, conn, dsn):
 
 
 @pytest.mark.slow
+@pytest.mark.timing
 def test_notifies_blocking(conn):
 
     def listener():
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Of course: your MR is welcome. Is it ready to merge or shall we let it simmer a bit longer, in case your laptop finds something else to complain about? :grin: 

Also, I have an idea: how about we have a pytest fixture where to import `time()` and `sleep()` from, and we make every test using it tagged `timing` automatically?

autra:
> Of course: your MR is welcome. Is it ready to merge or shall we let it simmer a bit longer, in case your laptop finds something else to complain about? 😁

I think it is ready to merge, builds seems to be reliably ok right now :laughing: But to be very honest I can't be sure, my laptop has its own opinion on things ;-)

More seriously, I know it is not really useful from a dev point of  view. This kind of commit is really to make various CI and other packaging processes happy. Boring, I know...

> Also, I have an idea: how about we have a pytest fixture where to import `time()` and `sleep()` from, and we make every test using it tagged `timing` automatically?

That would be clever. Maybe too magic? You'll have to remember not to use sleep and time directly. That being said, it's not worse than the current situation, where we have to remember to mark tests manually... I'm not the maintainer, not even a regular contributor, so I let you judge and I'll execute :-) 

That being said, I'm actually not sure how this could be achieved. By using [fixture's introspection and the request object](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures)?

dvarrazzo:
> Maybe too magic?

No, we do it for several other fixtures and pytest has all the hooks needed. The pattern is like this:

https://github.com/psycopg/psycopg/blob/af5bfa0ec2690571033e38205a98cfc8c6287134/tests/fix_gc.py#L9-L12

This is pretty much it: I would add a `timing` 

> You'll have to remember not to use sleep and time directly.

That's easy: we can drop all the time-related imports from the top-level module, and hack until the tests work again :upside_down_face: 

After that, I would introduce a `timing` fixture, returning an object exposing methods `time()`, `sleep()`, `asleep()`, which would be the mandatory way to use these functions in the tests. The marker could then be dropped from the tests using this fixture.

Before that, maybe it's worth looking at the tests currently marked `timing` and figure out if they are time-sensitive and yet they don't use those functions (maybe the test relies on a timeout)?

If you would like to try this work it would be welcome. This is unrelated to this MR anyway: I will leave a few other days pass and merge it.

Thank you very much.

PR #920, opened 2024-09-29
Package PostgreSQL 17 libpq with Windows binary packages

### LINKED ISSUE #916: Use the libpq from EDB on Windows
We are currently building windows package using [Postgres 14](https://github.com/psycopg/psycopg/blob/af5bfa0ec2690571033e38205a98cfc8c6287134/.github/workflows/packages-bin.yml#L238-L243) because this is what is [in the runner](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#postgresql).

However [we try](https://github.com/psycopg/psycopg/blob/af5bfa0ec2690571033e38205a98cfc8c6287134/.github/workflows/packages-bin.yml#L231) to install a newer version, using [wheel_win32_before_build.bat](https://github.com/psycopg/psycopg/blob/af5bfa0ec2690571033e38205a98cfc8c6287134/tools/build/wheel_win32_before_build.bat). Either that doesn't work, or the libpq ends up in mysterious places and we can't find it.

Looking better at what choco does, it seems that it installs a package using [these instructions](https://github.com/majkinetor/au-packages/blob/master/postgresql/tools/chocolateyInstall.ps1), which is pretty much taking it from EDB website. So we can instead try to do the same and figure out where things end up.

@kemosh might be able to help?

Related:

- #852
- #912

dvarrazzo:
Attempt to download the EDB package (which seems to be [what choco does](https://github.com/majkinetor/au-packages/blob/master/postgresql/tools/chocolateyInstall.ps1)) and use it for building.

At the time of writing, download works, but unpacking fails silently. Leaving it here for further work (and if anyone with more knowledge of Windows wants to continue with it, you are welcome).

Close #916

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   lint:
     runs-on: ubuntu-latest
-    if: true
+    if: false
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -236,8 +236,8 @@ jobs:
           ls
 
           echo "unpacking"
-          .\package.exe --mode unattended --enable-components commandlinetools,server,pgAdmin,stackbuilder --extract-only 1 --prefix d:\postgres
-          d:\postgres\bin\pg_config.exe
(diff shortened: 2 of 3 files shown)


dvarrazzo:
Closing as overtaken by #966.

It doesn't package libpq 17 yet but hopefully it should happen soon.

Issue #858, opened 2024-07-01
psycopg-binary can't be installed with M1 CPU starting from 3.1.20


maybephilipp:
Hello everyone!

I have issues installing psycopg-binary on macOS with M1 Pro CPU. 3.1.18 and 3.1.19 versions work, but all later ones are broken - pip says "No matching distribution found".

```
ERROR: Could not find a version that satisfies the requirement psycopg-binary==3.1.20 (from versions: 3.1.4, 3.1.5, 3.1.6, 3.1.8, 3.1.9, 3.1.10, 3.1.11, 3.1.12, 3.1.13, 3.1.14, 3.1.15, 3.1.16, 3.1.17, 3.1.18, 3.1.19)
ERROR: No matching distribution found for psycopg-binary==3.1.20
```

Was m1 deprecated in your build system?

Thanks

AshwinSundar:
I think the next version should be `3.2.1`, not `3.20`

https://www.psycopg.org/psycopg3/docs/

maybephilipp:
> I think the next version is called `3.2.1`, not `3.1.20`

https://www.psycopg.org/psycopg3/docs/news.html#psycopg-3-1-20

dvarrazzo:
Which version of Python are you using?

MacOS M1 packages for Psycopg 3.1.20 on are available for Python 3.10, 3.11, 3.12.

If you are you using Python 3.9, unfortunately support was dropped from some of the build infrastructure and we can't release that package anymore.

maybephilipp:
@dvarrazzo Nah, I tried 3.11 and 3.12 - both complaining. Poetry reports:

*I tried without Poetry too, just plain `python3.11 -m venv venv && ./venv/bin/pip install psycopg-binary==3.1.20` doesn't work either*

```
Poetry
Version: 1.8.3
Python:  3.11.7

Virtualenv
Python:         3.11.7
Implementation: CPython
Path:           /Users/ileep/Library/Caches/pypoetry/virtualenvs/fin-snake-pits-nUqOSej8-py3.11
Executable:     /Users/ileep/Library/Caches/pypoetry/virtualenvs/fin-snake-pits-nUqOSej8-py3.11/bin/python
Valid:          True

Base
Platform:   darwin
OS:         posix
Python:     3.11.7
Path:       /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11
[...]

PR #925, opened 2024-10-03
Add the possibility to set an environment variable to set macOS deployment target

thierryba:
This adds the support to set MACOSX_DEPLOYMENT_TARGET before building the psycopg extension.
I am not saying it necessarily is pretty but it seems to be doing the job. I tried locally to run setup.py and had set that variable to different values and I could see that with otool -l, it would show the right minimum version.

BTW, this is related to https://github.com/psycopg/psycopg/issues/858

### DIFF
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -109,6 +109,7 @@ jobs:
 
     env:
       PG_VERSION: "17"
+      MACOSX_DEPLOYMENT_TARGET: "12.0"
 
     steps:
       - uses: actions/checkout@v4
@@ -146,7 +147,7 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: macos-14-${{matrix.pyver}}-macosx_${{matrix.arch}}
+          name: macos-12-${{matrix.pyver}}-macosx_${{matrix.arch}}
           path: ./wheelhouse/*.whl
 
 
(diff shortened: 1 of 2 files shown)


thierryba:
this is supposed to change the environment used to build the extension. You can check the result with for instance:
otool -l ./python3.12/site-packages/psycopg_binary/_psycopg.cpython-312-darwin.so | grep -E -A4 LC_BUILD
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 14.4

So in our case the minos should be 12.0. I have not added a test for this since I was not super familiar with how to integrate it. Also I was very unsure that you would be interested in that change.

dvarrazzo:
I am also not experienced in macOS. So my question is: if this works this way, why is cibuildwheel not doing it?

thierryba:
Good question. I suppose it just installs python and whatever limit was set to build python is applied to the extension that we are building.
I am not saying this patch is ideal of course.

dvarrazzo:
I have no experience with macOS and this changeset doesn't seem to solve the problem as described in https://github.com/pypa/cibuildwheel/discussions/1926. Therefore I don't feel safe to merge this MR unless someone with macOS experience comes along and responsibly states that this change will produce correctly working binary packages.

I think you should suggest this changeset to cibuildwheel and if they accept it we will make use of it.

PR #926, opened 2024-10-05
Add Typing for Pool Stats

max-muoto:
This PR adds improved typing for connection pooling stats - so that it's easier to work with the returned stats without having to look at the specific documentation.

### DIFF
--- a/psycopg_pool/psycopg_pool/base.py
+++ b/psycopg_pool/psycopg_pool/base.py
@@ -192,7 +192,7 @@ class BasePool:
         stats, self._stats = self._stats, Counter()
         rv = dict(stats)
         rv.update(self._get_measures())
-        return rv
+        return cast(_PoolStats, rv)
 
     def _get_measures(self) -> dict[str, int]:
         """
--- a/psycopg_pool/psycopg_pool/base.py
+++ b/psycopg_pool/psycopg_pool/base.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 from time import monotonic
 from random import random
-from typing import Any, TYPE_CHECKING
+from typing import Any, TYPE_CHECKING, TypedDict, cast
 
 from psycopg import errors as e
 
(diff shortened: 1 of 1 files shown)


psycopg_pool/psycopg_pool/base.py
```diff
@@ -19,6 +19,24 @@
     from psycopg._connection_base import BaseConnection
 
 
+class _PoolStats(TypedDict):
```
max-muoto, line 22: Mainly making this private since I'm not sure it makes sense to live here from an interface POV and I'm not seeing an existing convention for `.types` module that contain type aliases (only have seen `psycopg.abc`), but let me know if you think there's a good place this could live and not be private.

dvarrazzo:
Hello,

thank you very much, but I would prefer to not have strict typing on the stats. [As documented](https://www.psycopg.org/psycopg3/docs/advanced/pool.html#pool-stats):

> The following values should be provided, but please don’t consider them as a rigid interface: it is possible that they might change in the future. Keys whose value is 0 may not be returned.

If we add strict typing then we will be forced to maintain this set of value, even if it would stop making sense. The only type information we guarantee (and it might even not have been a good idea, time will tell) is that keys are strings and values are integers, we don't want to make any guarantee about which ones the keys are.

max-muoto:
> Hello,
> 
> thank you very much, but I would prefer to not have strict typing on the stats. [As documented](https://www.psycopg.org/psycopg3/docs/advanced/pool.html#pool-stats):
> 
> > The following values should be provided, but please don’t consider them as a rigid interface: it is possible that they might change in the future. Keys whose value is 0 may not be returned.
> 
> If we add strict typing then we will be forced to maintain this set of value, even if it would stop making sense. The only type information we guarantee (and it might even not have been a good idea, time will tell) is that keys are strings and values are integers, we don't want to make any guarantee about which ones the keys are.

Makes senses! Apologies for missing that.

PR #932, opened 2024-10-15
feat: support loading composite types with keyword arguments

seandstewart:
# Background

## Problem

I like the idea of being able to register an in-app data model with a composite type defined server-side, but the current implementation only supports positional arguments. It's common in my organization to use `dataclasses` with `kw_only=True` - however this is not compatible.

I would consider using a `typing.NamedTuple`, however, we cannot support field inheritance with this solution. For a small codebase that's fine, but with a medium-to-large codebase, it's desirable to use inheritance to factor out common fields.

## Proposed Solution

Introduce an alternative composite loader which can leverage the `field_names` available in the `CompositInfo` during type instantiation.


seandstewart:
@dvarrazzo lmk your thoughts on this proposal. 

Also, looks like a new mypy failure unrelated to my changes:

> psycopg/psycopg/waiting.py:67: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:102: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:163: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:221: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:258: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:314] error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]
psycopg/psycopg/waiting.py:362: error: Incompatible types in assignment (expression has type "Any | None", variable has type "RV")  [assignment]

dvarrazzo:
Hello!

Thank you very much for your proposal. In this moment I have a perfect storm of other tasks to work on, but I will try to get back at yours. I doubt I will be able before two weeks from now: it doesn't mean I'm ignoring this proposal :slightly_smiling_face: 

And yes, those failures are probably related to a newer Mypy version that has some objection on previously written code.

Cheers!

seandstewart:
Hey @dvarrazzo -

Let me know what I can do to help move this along. I'm happy to look into the typing issue as well.

dvarrazzo:
Hello Sean,

the typing problem was solved on master. I have rebased your branch and pushed here to clear the error.

I am sorry for the lack of interaction but I am extremely busy with other projects at the moment. I think that your feature proposal is reasonable and the implementation does the right thing.

I see that the `load()` method is pretty much duplicated between `KeywordCompositeLoader` and `CompositeLoader`. Maybe there is a way to refactor the code to separate the type conversion from the Python object creation. The last part of `CompositeType.load()` can be written as:

```python
        if data == b"()":
            parsed = ()
        else:
            parsed = self._tx.load_sequence(tuple(self._parse_record(data[1:-1])))
        return type(self).factory(*parsed)
```

Where `parsed` is the same you use. So probably we can refactor this into an internal method only returning a parsed tuple and then have a different final step between `*args` and `**kwargs` composite loader? Looking at `CompositeBinaryLoader.load()` ISTM that we want something more similar to that.

At the moment the inheritance tree is:

```
Loader
  + BaseCompositeLoader
      + RecordLoader
          + CompositeLoader
              + KeywordCompositeLoader
  + RecordBinaryLoader
      + CompositeBinaryLoader
      + KeywordCompositeBinaryLoader

```

so there is a marked asymmetry between text and binary. I think the structure is better on the binary side, where `RecordBinaryLoader.load()` does the conversion and the two subclasses just create the Python object. Maybe we can do the same on the text side too?

seandstewart:
Hey @dvarrazzo -

Thanks for the thoughtful review - and no need to apologize, being busy is a good thing 👍 

I made a few changes which make the type tree for both binary and text loaders symmetrical:

```
Loader
  + RecordLoader
      + BaseInstanceLoader[InstanceT]
          + CompositeLoader[InstanceT]
          + KeywordCompositeLoader[InstanceT]
  + RecordBinaryLoader
      + BaseBinaryInstanceLoader[InstanceT]
          + CompositeBinaryLoader[InstanceT]
          + KeywordCompositeBinaryLoader[InstanceT]
```

I made a couple other changes:

1. Some optimizations for the `RecordBinaryLoader`.
    - Use a generator to unpack (oid, value) pairs.
    - Use a membership check instead of `KeyError` to optimize hot loops when looking up transformers. 
2. Generics for the `*InstanceLoader` so mypy can better-track output results.

Looking forward to your thoughts!

psycopg/psycopg/_transformer.py
```diff
@@ -11,6 +11,8 @@
 from . import abc
 from ._cmodule import _psycopg
 
+__all__ = ("Transformer",)
```
dvarrazzo, line 14: Why is this necessary? It seems something unneeded to maintain.

psycopg/psycopg/types/composite.py
```diff
@@ -196,6 +196,10 @@ def _parse_record(self, data: abc.Buffer) -> Iterator[bytes | None]:
 
 
 class RecordLoader(BaseCompositeLoader):
+    def __init__(self, oid: int, context: abc.AdaptContext | None = None):
+        super().__init__(oid, context)
+        self._tx = Transformer(context)
```
dvarrazzo, line 201: The `BaseCompositeLoader` already has a `_tx`.

psycopg/psycopg/types/composite.py
```diff
@@ -219,63 +223,136 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+            return ()
+
+        oids, record = zip(*self._iterunpack(data, nfields))
+        tx = self._get_transformer(oids)
+        return tx.load_sequence(record)
+
+    @staticmethod
+    def _iterunpack(
+        data: abc.Buffer, nfields: int
+    ) -> Iterator[tuple[int, abc.Buffer | None]]:
```
dvarrazzo, line 238: If you did this as an optimization, I doubt it is. You are creating several intermediate tuples and call into the iterator protocol, which requires throwing and catching an exception to terminate.

If you can make a micro-benchmark to prove that this is better than building two lists I can accept this changeset, otherwise it would be better to leave it as it was.

Actually, a micro-optimization would be to preallocate the lists, as we know the lengths, and set the items. Could you maybe try if this doesn't cause a regression and which solution is faster?

```python
oids = [0] * nfields
record = [None] * nfields
for i in range(nfields):
    oids[i], length = _unpack_oidlen(data, offset)
    offset += 8
    if length >= 0:
        record[i] = data[offset : offset + length]
        offset += length
```

psycopg/psycopg/types/composite.py
```diff
@@ -219,63 +223,136 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+    format = pq.Format.BINARY
+
+    def __init__(self, oid: int, context: abc.AdaptContext | None = None):
+        super().__init__(oid, context)
+        self._ctx = context
+        # Cache a transformer for each sequence of oid found.
+        # Usually there will be only one, but if there is more than one
+        # row in the same query (in different columns, or even in different
+        # records), oids might differ and we'd need separate transformers.
+        self._txs: dict[tuple[int, ...], abc.Transformer] = {}
```
dvarrazzo, line 273: All this block seems a repetition of what `super().__init__()` does, including the comment.

.gitignore
```diff
@@ -11,6 +11,7 @@ __pycache__/
 .venv
 .coverage
 htmlcov
+.idea
```
dvarrazzo, line 14: I can commit this to master so we can leave this out of this MR.

psycopg/psycopg/types/composite.py
```diff
@@ -219,63 +223,136 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
 
-        if data == b"()":
-            return type(self).factory()
-
-        return type(self).factory(
-            *self._tx.load_sequence(tuple(self._parse_record(data[1:-1])))
+        args = (
+            ()
+            if data == b"()"
+            else self._tx.load_sequence(tuple(self._parse_record(data[1:-1])))
```
dvarrazzo, line 295: An if-else instead of a ternary is more readable and compact here.

dvarrazzo (changes requested):

psycopg/psycopg/types/composite.py
```diff
@@ -219,63 +223,136 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+        # row in the same query (in different columns, or even in different
+        # records), oids might differ and we'd need separate transformers.
+        self._txs: dict[tuple[int, ...], abc.Transformer] = {}
+
+    def load(self, data: abc.Buffer) -> tuple[Any, ...]:
+        return self._parse_record(data)
+
 
-        return tx.load_sequence(tuple(record))
+InstanceT = TypeVar("InstanceT")
```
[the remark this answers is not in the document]
seandstewart: Within the scope of the `register_composite` and class creation, mypy can track it, which will still provide more robust analysis - in this case, it caught some subtle bugs which were otherwise masked by passing through `Any` rather than the anonymous `InstanceT` typevar.

psycopg/psycopg/types/composite.py
seandstewart: In general I try to avoid `list.append` because it forces the interpreter to check if more memory needs be allocated with every call (not to mention the cost if it does, in fact, need to allocate more memory).

I can compare the two suggestions - I agree creating intermediate tuples incurs its own cost, but tuples are fairly well-optimize and pre-allocating the list is likely to be the winner here.

psycopg/psycopg/_transformer.py
seandstewart: I received an error with mypy without it:

https://github.com/psycopg/psycopg/actions/runs/11784504225/job/32823815099#step:7:12

> Module "psycopg._transformer" does not explicitly export attribute "Transformer"  [attr-defined]

This may be resolved due to other changes I made afterward, so will test without this.

seandstewart: Pushed a change here to remove this addition.

psycopg/psycopg/types/composite.py
seandstewart: I've gone ahead and followed your recommended approach here to pre-allocate the sequences.

psycopg/psycopg/types/composite.py
seandstewart: Fixed!

psycopg/psycopg/types/composite.py
seandstewart: Removed all the extra `super().__init__()` calls, not sure what I was doing here.

psycopg/psycopg/types/composite.py
```diff
@@ -219,63 +223,136 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+            ()
+            if data == b"()"
+            else self._tx.load_sequence(tuple(self._parse_record(data[1:-1])))
         )
+        return self._load_instance(args)
 
     def _config_types(self, data: abc.Buffer) -> None:
         self._tx.set_loader_types(self.fields_types, self.format)
 
+    @classmethod
```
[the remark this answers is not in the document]
seandstewart: Moved back to standard methods 👍

seandstewart:
Hey @dvarrazzo -

I'm back! Sorry I let this sit so long - work took over for some time.

I've gone through and made some adjustments based on your feedback - I look forward to another review.

I believe you'll need to kickstart the CI pipeline, it doesn't seem to have started with this latest set of changes.

dvarrazzo:
Hello, thank you for getting back on this. I was meaning to ping you, but, because we will probably wait for Python 3.14 to release psycopg 3.3, I didn't see too much time pressure :slightly_smiling_face: 

I'll try to review it in the next few days. Cheers!

seandstewart:
Sounds good - no rush on my end. I'll keep an eye out for your review 🫡

dvarrazzo:
Hello

I apologisefor dropping the ball on this MR. I am wrapping up 3.3 and I would like to include this feature.

For a start, I have rebased it on master.

In the meantime, in #1175 I have proposed a refactoring to simplify the inheritance (d955062e32ab327de4d506805810350f0e559760) in which the record loaders are removed from the composite inheritance tree. One reason to do so, apart from offering something easier to work with for the Cython conversion, is the unique need of the binary record loader to have a mapping of transformer in order to accommodate returning different records in the same query, which is a need that composite dumpers subclass don't share (every different composite in the same query would have different loader instantiated). The common implementation of the text and binary parsing has been moved to module-level functions in order to not force inheritance just to reuse implementation.

In pictures, inheritance in 3.2 (where your MR started) was
composite-inheritance-3 2

In the #1175 refactoring the inheritance tree is becoming:
composite-inheritance-1175
(the `Composite[Binary]Loader` classes got an underscore to signify that they are private implementation).

In ths MR instead it's becoming:
composite-inheritance-932

I would like to check if I can plant the new classes you introduced on the new inheritance tree and make use of the other improvements you are proposing here.

Again, I apologise for the slowness in my response. If you have lost interest in this PR I understand. I will try to make use of the most of the changes you propose and keep attribution for your work.

psycopg/psycopg/_transformer.py
```diff
@@ -11,6 +11,8 @@
...
+
```
dvarrazzo: Unneeded, I think? Mypy has no problem with this definition and this is an internal module which people won't use, especially for `import *`.

Does the lack of __all__ causes problem to other static checkers?

psycopg/psycopg/types/composite.py
```diff
@@ -216,64 +201,95 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+        if not record:
+            return ()
 
+        tx = self._get_transformer(tuple(oids))
         return tx.load_sequence(record)
 
+    def _get_transformer(self, key: tuple[int, ...]) -> abc.Transformer:
+        # This could be managed with a custom `dict.__missing__` implementation
+        # micro-optimization - Prefer a membership check to a KeyError
+        #   try/except blocks are expensive in hot loops.
```
dvarrazzo: This is not necessarily true. It used to be, but [since Python 3.11](https://bugs.python.org/issue40222) you pay the cost only if the exception is captured, not when it is set up.

I don't mind the refactoring, let's keep it, but the comment is redundant.

```suggestion
```

psycopg/psycopg/types/composite.py
```diff
@@ -216,64 +201,95 @@ def __init__(self, oid: int, context: abc.AdaptContext | None = None):
...
+    The class is complete but lack information about the fields types and
+    object factory. These will be added by register_composite(), which will
+    create a subclass of this class.
+    """
 
-class CompositeLoader(RecordLoader):
-    factory: Callable[..., Any]
-    fields_types: list[int]
+    fields_types: tuple[int]
+    fields_names: tuple[str]
```
dvarrazzo, line 231: Note: I have dropped `factory` from the loader class. Now the `make_instance()` method knows the factory.

dvarrazzo:
Merging, but it can be argued that something in the other direction should be done.

Dumping a composite assumes that you are dumping a sequence with a 1:1 mapping with composite fields. Any different objects (not a sequence, not same order and number of fields) cannot be naturally dumped.

I am preparing a MR that cover that case too, so we have the possibility of customizing the object-composite mapping in both directions in a similar way.

seandstewart:
Hey @dvarrazzo -

No problem - sorry I missed this, I'm not in a place to do dev work right now as I just had a new addition to the family. Glad to see you got it pushed through!

dvarrazzo:
@seandstewart sounds like you have much more important things in your hands! All the best and thank you again for your contribution! :purple_heart:

PR #940, opened 2024-10-25
docs: ensure operator in text and code example matches

0xTiger:
In the code example demonstrating that `IS %s` isn't possible:  The `IS NOT DISTINCT FROM` operator is used

However in the accompanying text reads:
*"instead is to use the `IS DISTINCT FROM`"*
I've read this a few times and I think the text should read:
*"instead is to use the `IS NOT DISTINCT FROM`"*

Thanks for the work maintaining this btw, psycopg is great :)

### DIFF
--- a/docs/basic/from_pg2.rst
+++ b/docs/basic/from_pg2.rst
@@ -270,8 +270,8 @@ You cannot use :sql:`IS %s` or :sql:`IS NOT %s`::
     LINE 1: SELECT * FROM foo WHERE field IS $1
                                          ^
 
-This is probably caused by the fact that :sql:`IS` is not a binary operator in
-PostgreSQL; rather, :sql:`IS NULL` and :sql:`IS NOT NULL` are unary operators
+This is probably caused by the fact that :sql:`IS` is not a binary predicate in
+PostgreSQL; rather, :sql:`IS NULL` and :sql:`IS NOT NULL` are unary predicates
 and you cannot use :sql:`IS` with anything else on the right hand side.
 Testing in psql:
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
I assume that what I meant there is to use IS [NOT] DISTINCT FROM in lieu
of IS [NOT] NULL, considering one operator and its complementary version
equivalent.

Maybe I would prefer a wording such as "is to use the IS [NOT] DISTICT FROM
predicate".

The wording can also be improved by using "predicate" instead of
"operator", as that's how these constructs are referred to in the postgres
docs page linked.

Does it sound ok?

0xTiger:
Hmmmm sorry I don't understand
I personally think the wording is fine although I agree "predicate" may indeed be better

But my main point is in my opinion there's surely a mixup of `IS DISTINCT FROM` vs `IS NOT DISTINCT FROM`?

Paraphrasing from the docs:
> you cannot use `IS` with anything else on the right hand side.
Testing in psql: SELECT 10 IS 10; gives an error
What you can do instead is to use the `IS DISTINCT FROM` operator, which
will gladly accept a placeholder:
`conn.execute("SELECT * FROM foo WHERE field IS NOT DISTINCT FROM %s", [None])`

You give the example of `10 IS %s` not working
and you seem to suggest that the correct alternative should be `10 IS DISTINCT FROM %s`
but I think the correct alternative should be `10 IS NOT DISTINCT FROM %s`

You even use a code example with `IS NOT DISTINCT FROM %s` just below the text

0xTiger:
Sure, yeah sounds good to me! :slightly_smiling_face: 

I've made the change, I left the explaning of the `IS DISTINCT FROM %s` equivalence as it appears already; Below the code example in the line beginning leaving the "Analogously you can use..."

Feel free to amend further

dvarrazzo:
I made some cleanup to the wording and restored the link reference that was lost. I have chosen a wording that creates a good link to the Postgres documentation and that ties well to the following paragraph:

> Analogously you can use :sql:`IS DISTINCT FROM %s` as a parametric version of
:sql:`IS NOT %s`.

PR #950, opened 2024-11-07
Extend `libpq` install instructions for cross-platform use

gy-mate:
- Add macOS and Windows install instructions for `libpq`
    - This is also required e.g. to run cross-platform zipapps generated with https://github.com/linkedin/shiv that include Psycopg as a dependency, because I think that using any C code (e.g. `psycopg-binary`) is not a viable option in this case

If this is merged, could https://pypi.org/project/psycopg/#description be updated to include this README section? I would've helped me a lot because a zipapp that I generated with `shiv` wasn't working on another platform and I didn't know why.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,11 +17,12 @@ For further information about installation please check `the documentation`__.
 .. __: https://www.psycopg.org/psycopg3/docs/basic/install.html
 
 
-Hacking / Cross-platform use
+.. _Hacking:
+
+Hacking
 -------
 
-In order to work on the Psycopg source code, or to use cross-platform zipapps
-created with ``shiv`` that include Psycopg as a dependency, you must have the
+In order to work on the Psycopg source code, you must have the
 ``libpq`` PostgreSQL client library installed on the system. For instance, on
 Debian systems, you can obtain it by running::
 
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you. I am not familiar with the tools you propose but I assume it might be useful for other platforms than Linux, where I usually develop.

Can you please:

- make sure that there is a link to `shiv` to allow people to install it
- can you create a "cross-compiling" section after the base "hacking" and separate it well from "hacking" that doesn't require cross-compiling support?

Thank you very much

gy-mate:
> Can you please:
> 
> * make sure that there is a link to `shiv` to allow people to install it
> * can you create a "cross-compiling" section after the base "hacking" and separate it well from "hacking" that doesn't require cross-compiling support?

@dvarrazzo Sure, both are done! (MacOS tests seem to fail for unrelated reasons.)

dvarrazzo:
Sure, macOS failure is not related to this :) looks like macOS 12 infrastructure is crumbling :melting_face:

README.rst
```diff
@@ -72,3 +83,13 @@ Now hack away! You can run the tests using::
...
     export PSYCOPG_TEST_DSN="dbname=psycopg_test"
     pytest
+
+
+Cross-compiling
+---------------
+
+To use cross-platform zipapps created with `shiv`__ that include Psycopg
+as a dependency you must also have ``libpq`` installed. See
+:ref:`the section above <Hacking>` for install instructions.
```
dvarrazzo, line 93: This link doesn't render well: see [here](https://github.com/psycopg/psycopg/blob/d2f6590c7a7f3f58c4207d759f83d1a2b08a8173/README.rst). `:ref:` is a Sphinx role; in pure reST you can just use `Hacking_` to link to a title in the same document.

dvarrazzo:
Merged, thank you very much!

gy-mate:
@dvarrazzo Thank _you_! :)

PR #955, opened 2024-11-13
Prepared Connection Wrapper

ohmdelta:
add simple wrapper for connection class to default to prepared statements
specifically for the use case of connection pools


dvarrazzo:
Why? you can use the `kwargs` parameter of the connection pool.

ohmdelta:
thanks got it, missed the comment in the docs

PR #966, opened 2024-12-15
ci: use vcpkg to install libpq on windows

trim21:
ref: https://github.com/psycopg/psycopg/issues/916

It build wheel as expected:

https://github.com/trim21/psycopg/actions/runs/12343652281

### DIFF
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -218,6 +218,10 @@ jobs:
         pyver: [cp38, cp39, cp310, cp311, cp312, cp313]
 
     steps:
+      # there are some other libpq in PATH
+      - run: rm -rf c:/tools/php C:/Strawberry/c/bin
+        shell: bash
+
       - uses: actions/checkout@v4
 
       - name: Start PostgreSQL service for test
@@ -225,6 +229,16 @@ jobs:
           $PgSvc = Get-Service "postgresql*"
           Set-Service $PgSvc.Name -StartupType manual
           $PgSvc.Start()
+
+      - name: Export GitHub Actions cache environment variables
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const path = require('path')
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+            core.addPath(path.join(process.env.VCPKG_INSTALLATION_ROOT, 'installed/x64-windows-release/lib'));
+            core.addPath(path.join(process.env.VCPKG_INSTALLATION_ROOT, 'installed/x64-windows-release/bin'));
 
       - name: Create the binary package source tree
         run: python3 ./tools/build/copy_to_binary.py
@@ -234,6 +248,7 @@ jobs:
         with:
           package-dir: psycopg_binary
         env:
+          VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite" # cache vcpkg
           CIBW_BUILD: ${{matrix.pyver}}-${{matrix.arch}}
           CIBW_ARCHS_WINDOWS: AMD64 x86
           CIBW_BEFORE_BUILD_WINDOWS: '.\tools\build\wheel_win32_before_build.bat'
@@ -245,7 +260,6 @@ jobs:
             pytest {project}/tests -m "not slow and not flakey" --color yes
           CIBW_ENVIRONMENT_WINDOWS: >-
             PSYCOPG_IMPL=binary
-            PATH="C:\\Program Files\\PostgreSQL\\14\\bin;$PATH"
             PSYCOPG_TEST_DSN="host=127.0.0.1 user=postgres"
             PSYCOPG_TEST_WANT_LIBPQ_BUILD=">= 16"
             PSYCOPG_TEST_WANT_LIBPQ_IMPORT=">= 16"
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -247,8 +247,8 @@ jobs:
             PSYCOPG_IMPL=binary
             PATH="C:\\Program Files\\PostgreSQL\\14\\bin;$PATH"
             PSYCOPG_TEST_DSN="host=127.0.0.1 user=postgres"
-            PSYCOPG_TEST_WANT_LIBPQ_BUILD=">= 14"
-            PSYCOPG_TEST_WANT_LIBPQ_IMPORT=">= 14"
+            PSYCOPG_TEST_WANT_LIBPQ_BUILD=">= 16"
+            PSYCOPG_TEST_WANT_LIBPQ_IMPORT=">= 16"
 
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -306,6 +306,9 @@ jobs:
         shell: bash
 
     steps:
+      # there are some extra libpq.dll in PATH
+      - run: rm -rf c:/tools/php C:/Strawberry/c/bin
+
       - uses: actions/checkout@v4
 
       - uses: actions/setup-python@v5
@@ -329,15 +332,28 @@ jobs:
         if: ${{ github.event_name == 'schedule' }}
         run: echo "NOT_MARKERS=$NOT_MARKERS refcount" >> $GITHUB_ENV
 
+      - name: Export GitHub Actions cache environment variables
+      # https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const path = require('path')
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+            core.addPath(path.join(process.env.VCPKG_INSTALLATION_ROOT, 'installed/x64-windows-release/lib'));
+            core.addPath(path.join(process.env.VCPKG_INSTALLATION_ROOT, 'installed/x64-windows-release/bin'));
+
+      - name: Install libpq from vcpkg and install pg_config.exe stub
+        run: .\tools\build\wheel_win32_before_build.bat
+        shell: powershell
+        env:
+          # cache vcpkg
+          VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
+
       - name: Build the C wheel
         if: ${{ matrix.impl == 'c' }}
         run: |
           pip install delvewheel wheel
-
-          # The windows runner is a total mess, with random copies of the libpq
-          # scattered all over the places. Give precedence to the one under our
-          # control (or the illusion of it).
-          export PATH="/c/Program Files/PostgreSQL/14/bin/:$PATH"
 
           # If the wheel is not delocated, import fails with some dll not found
           # (but it won't tell which one).
--- a/tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
+++ b/tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
@@ -1,0 +1,51 @@
+"""
+We use vcpkg in github actions to build psycopg-binary.
+
+This is a stub to work as `pg_config --libdir` or `pg_config --includedir` to make it work with vcpkg.
+
+You will need install vcpkg and set `VCPKG_ROOT `env
+and run `vcpkg install libpq:x64-windows-release` before you use this script
+"""
+
+import os
+import pathlib
+import sys
+import platform
+
+
+def main():
+    # only x64-windows
+    if not (sys.platform == "win32" and platform.machine() == "AMD64"):
+        raise Exception("this script should only be used in x64-windows")
+
+    what = sys.argv[1]
+
+    if what == "--help":
+        print(__doc__)
+        return
+
+    # on github actions it's `VCPKG_INSTALLATION_ROOT`
+    if "VCPKG_INSTALLATION_ROOT" not in os.environ:
+        print("failed to find VCPKG ROOT path", file=sys.stderr)
+        sys.exit(1)
+
+    vcpkg_root = pathlib.Path(os.environ["VCPKG_INSTALLATION_ROOT"])
+    vcpkg_platform_root = vcpkg_root.joinpath("installed/x64-windows-release").resolve()
+    if not vcpkg_root.joinpath("packages/libpq_x64-windows-release").exists():
+        print("libpq not installed with vcpkg", file=sys.stderr)
+        sys.exit(1)
+
+    if what == "--libdir":
+        print(str(vcpkg_platform_root.joinpath("lib")))
+        return
+    if what == "--includedir":
+        print(str(vcpkg_platform_root.joinpath("include")))
+        return
+
+    print(
+        "unexpected command: {!r}\n this maybe out-of-sync between 'psycopg_c/setup.py' and 'tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py'".format(
+            sys.argv[1:]
+        ),
+        file=sys.stderr,
+    )
+    sys.exit(1)
--- a/tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
+++ b/tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
@@ -1,51 +1,79 @@
+#!/usr/bin/env python
 """
 We use vcpkg in github actions to build psycopg-binary.
 
-This is a stub to work as `pg_config --libdir` or `pg_config --includedir` to make it work with vcpkg.
+This is a stub to work as `pg_config --libdir` or `pg_config --includedir` to
+make it work with vcpkg.
 
-You will need install vcpkg and set `VCPKG_ROOT `env
-and run `vcpkg install libpq:x64-windows-release` before you use this script
+You will need install `vcpkg`, set `VCPKG_ROOT` env, and run `vcpkg install
+libpq:x64-windows-release` before using this script.
 """
 
 import os
-import pathlib
 import sys
 import platform
+from pathlib import Path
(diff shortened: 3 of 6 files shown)


dvarrazzo:
Trying to have a chat with vkpkg to figure out if they can add libpq 17...

dvarrazzo:
https://github.com/microsoft/vcpkg/issues/41747 would allow to use libpq 17

dvarrazzo:
[This](https://github.com/trim21/psycopg/actions/runs/12343652281/job/34445062020#step:7:458) shows something strange:

```
libpq used: 140012
libpq compiled: 160004
```

it seems that we do find the vcpkg at build time, but when testing we found an outdated version. I have push a change to upgrade the dependency check.

Maybe we should just delete the offending stale libpq.dll from the runner. I would iterate on every entry in PATH and delete `$dir/libpq.dll` and `$dir/pg_config.exe` if found, before installing the vcpkg version.

dvarrazzo (changes requested):
Thank you very much for this initiative: it seems a welcome change.

.github/workflows/packages-bin.yml
```diff
@@ -226,6 +226,22 @@ jobs:
...
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+            core.exportVariable('VCPKG_ROOT', process.env.VCPKG_INSTALLATION_ROOT || '');
+            core.addPath(path.join(process.env.VCPKG_INSTALLATION_ROOT, 'installed/x64-windows-release/lib'));
+
+      - name: Install libpq from vcpkg
+        run: vcpkg install libpq:x64-windows-release
+        env:
+          # cache vcpkg
+          VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
```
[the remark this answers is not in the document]
trim21: yes, and it's very slow

trim21:
> [This](https://github.com/trim21/psycopg/actions/runs/12343652281/job/34445062020#step:7:458) shows something strange:
> 
> ```
> libpq used: 140012
> libpq compiled: 160004
> ```
> 
> it seems that we do find the vcpkg at build time, but when testing we found an outdated version. I have push a change to upgrade the dependency check.
> 
> Maybe we should just delete the offending stale libpq.dll from the runner. I would iterate on every entry in PATH and delete `$dir/libpq.dll` and `$dir/pg_config.exe` if found, before installing the vcpkg version.

I think it may be possible to use a static libpq here? would that be a option?

tools/build/wheel_win32_before_build.bat
```diff
@@ -19,7 +19,13 @@ REM choco upgrade postgresql
 
 REM On https://community.chocolatey.org/packages/postgresql15/15.0.1#discussion
 REM I found the following command in a comment:
-choco install postgresql15 --version 15.0.1
+
+vcpkg --version >nul 2>&1 && (
+    vcpkg install libpq:x64-windows-release
+) || (
+    choco install postgresql15 --version 15.0.1
```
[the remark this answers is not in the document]
trim21: also I think we may be able to remove `delvewheel` here, it's included in cibuildwheel actions already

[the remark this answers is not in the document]
dvarrazzo: In the past it wasn't. If now it is then, sure, feel free to drop it, thank you :slightly_smiling_face:

.github/workflows/packages-bin.yml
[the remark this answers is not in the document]
dvarrazzo: The build steps in your fork take <30s to build. Out of curiosity, have you got an example of slow build somewhere?

dvarrazzo:
> I think it may be possible to use a static libpq here? would that be a option?

If it works it would not be a problem to distribute a static build.

.github/workflows/packages-bin.yml
```diff
@@ -247,8 +258,8 @@ jobs:
             PSYCOPG_IMPL=binary
             PATH="C:\\Program Files\\PostgreSQL\\14\\bin;$PATH"
```
dvarrazzo, line 259: This entry can be removed. It might be actually the thing that ends up importing the wrong libpq.

trim21, line 259:
```suggestion
```

trim21, line 259:
```suggestion
```

trim21:
surprisingly we are using libpq.dll from php 😅

```text
python .github/scripts/remove_default_libpq.py
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    PSYCOPG_IMPL: python
    DEPS: ./psycopg[test] ./psycopg_pool
    PSYCOPG_TEST_DSN: host=127.0.0.1 dbname=postgres
    NOT_MARKERS: timing proxy mypy
    pythonLocation: C:\hostedtoolcache\windows\Python\[3](https://github.com/psycopg/psycopg/actions/runs/12365600275/job/34510897206?pr=966#step:5:3).11.9\x64
    PKG_CONFIG_PATH: C:\hostedtoolcache\windows\Python\3.11.9\x64/lib/pkgconfig
    Python_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.11.9\x6[4](https://github.com/psycopg/psycopg/actions/runs/12365600275/job/34510897206?pr=966#step:5:4)
    Python2_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.11.9\x[6](https://github.com/psycopg/psycopg/actions/runs/12365600275/job/34510897206?pr=966#step:5:6)4
    Python3_ROOT_DIR: C:\hostedtoolcache\windows\Python\3.11.9\x64
c:\tools\php\libpq.dll
```

trim21:
> [This](https://github.com/trim21/psycopg/actions/runs/12343652281/job/34445062020#step:7:458) shows something strange:
> 
> ```
> libpq used: 140012
> libpq compiled: 160004
> ```
> 
> it seems that we do find the vcpkg at build time, but when testing we found an outdated version. I have push a change to upgrade the dependency check.
> 
> Maybe we should just delete the offending stale libpq.dll from the runner. I would iterate on every entry in PATH and delete `$dir/libpq.dll` and `$dir/pg_config.exe` if found, before installing the vcpkg version.

Just find out that we do not need to dll libpq.dll from path, there is a extra repair wheel step with extra PATH I forget to remove, so it bundle wrong dll into wheel

https://github.com/psycopg/psycopg/blob/c7b8ed5d85e15874eb4dc71d29c1b08800bf9606/.github/workflows/tests.yml#L333

.github/workflows/tests.yml
```diff
@@ -322,16 +324,28 @@ jobs:
         if: ${{ github.event_name == 'schedule' }}
         run: echo "NOT_MARKERS=$NOT_MARKERS refcount" >> $GITHUB_ENV
 
+      - name: Export GitHub Actions cache environment variables
```
dvarrazzo, line 327: I don't see how this is performing any caching. Can you explain? Why not using `actions/cache@v3`?

See [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows).

dvarrazzo (changes requested):

psycopg_c/setup.py
```diff
@@ -21,6 +23,32 @@
...
+        # on github actions it's `VCPKG_INSTALLATION_ROOT`
+        if "VCPKG_ROOT" in os.environ or "VCPKG_INSTALLATION_ROOT" in os.environ:
+            vcpkg_root = (
+                os.environ.get("VCPKG_ROOT") or os.environ["VCPKG_INSTALLATION_ROOT"]
+            )
+            for target in [
+                # "x64-windows-static-md-release",
+                # "x64-windows-static-md",
+                "x64-windows-release",
+                "x64-windows",
```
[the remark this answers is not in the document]
trim21: `x64-windows` is release+debug, and `-release` is release only

.github/workflows/tests.yml
trim21: this export env `ACTIONS_CACHE_URL` `ACTIONS_RUNTIME_TOKEN` which only exists when you call `uses:` but not `run:`, so we need to exprot these to env so vcpkg can use github actions cache.

psycopg_c/setup.py
```diff
@@ -21,6 +23,30 @@
...
+                if vcpkg_platform_root.exists():
+                    if what == "libdir":
+                        if vcpkg_platform_root.joinpath("lib/libpq.lib").exists():
+                            print("using vcpkg", target, "from", vcpkg_root)
+                            return str(vcpkg_platform_root.joinpath("lib"))
+                    if what == "includedir":
+                        if vcpkg_platform_root.joinpath("include/libpq").exists():
+                            print("using vcpkg", target, "from", vcpkg_root)
+                            return str(vcpkg_platform_root.joinpath("include"))
+
```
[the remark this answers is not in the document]
trim21: OK, I'll write a shim for pg_config and put it in `wheel_win32_before_build.bat`

docs/basic/install.rst
```diff
@@ -130,7 +130,10 @@ In order to perform a local installation you need some prerequisites:
 - a C compiler,
 - Python development headers (e.g. the ``python3-dev`` package).
 - PostgreSQL client development headers (e.g. the ``libpq-dev`` package).
-- The :program:`pg_config` program available in the :envvar:`PATH`.
+- The :program:`pg_config` program available in the :envvar:`PATH`,
+- On Windows, you can also install libpq with vcpkg.
+  For example, ``vcpkg install libpq:x64-windows-release``  or ``vcpkg install libpq:x64-windows``,
+  ``psycopg[c]`` will prefer vcpkg installed libpq when it's available.
```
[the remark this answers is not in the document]
trim21: OK, I'll revert this section.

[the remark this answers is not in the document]
trim21: Yes, I forget that this won't work, dll may not exists in PATH.

psycopg_c/setup.py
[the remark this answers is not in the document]
trim21: I add a dummy python package in `tools/build/pg_config_stub` and use pip to install a python script as `pg_config.exe`, this is the simplest way I can think of.

tools/build/pg_config_stub/pg_config_vcpkg_stub.py
```diff
@@ -0,0 +1,37 @@
...
+        vcpkg_platform_root = vcpkg_root.joinpath(
+            "installed/x64-windows-release"
+        ).resolve()
+        if vcpkg_platform_root.exists():
+            if what == "--libdir":
+                if vcpkg_platform_root.joinpath("lib/libpq.lib").exists():
+                    print(str(vcpkg_platform_root.joinpath("lib")))
+            if what == "--includedir":
+                if vcpkg_platform_root.joinpath("include/libpq").exists():
+                    print(str(vcpkg_platform_root.joinpath("include")))
```
dvarrazzo: no `main()` call?

Also, what happens if any of the `if` condition fails (line 24, line 31, unexpected `what`)? The script will succeed but not print anything. Maybe we should throw an exception in these cases

dvarrazzo: Ah I see, you have included a pyproject. ok.

trim21: > Also, what happens if any of the `if` condition fails (line 24, line 31, unexpected `what`)? The script will succeed but not print anything. Maybe we should throw an exception in these cases

OK, I'll add some error message below

.github/workflows/tests.yml
dvarrazzo: I see: vcpkg interacts actively with github action cache as a service, rather than having its artifacts cached.

Weird, but hey.

.github/workflows/tests.yml
```diff
@@ -306,6 +306,8 @@ jobs:
         shell: bash
 
     steps:
+      # there are some extra libpq.dll in PATH
+      - run: rm c:/tools/php C:/Strawberry/c/bin -rf
```
dvarrazzo, line 310:
```suggestion
      - run: rm -rf c:/tools/php C:/Strawberry/c/bin

```

.github/workflows/tests.yml
```diff
@@ -329,16 +331,29 @@ jobs:
         if: ${{ github.event_name == 'schedule' }}
         run: echo "NOT_MARKERS=$NOT_MARKERS refcount" >> $GITHUB_ENV
 
+      - name: Export GitHub Actions cache environment variables
```
dvarrazzo, line 334: This is obscure enough to deserve a comment.
```suggestion
      - name: Export GitHub Actions cache environment variables
      # https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache
```

.github/workflows/packages-bin.yml
```diff
@@ -218,6 +219,10 @@ jobs:
         pyver: [cp38, cp39, cp310, cp311, cp312, cp313]
 
     steps:
+      # there are some other libpq in PATH
+      - run: rm c:/tools/php C:/Strawberry/c/bin -rf
+        shell: bash
```
dvarrazzo, line 224: This order is not the standard way to use `rm`. I also don't think that `shell: bash` is required.

```suggestion
      - run: rm -rf c:/tools/php C:/Strawberry/c/bin
```

tools/build/wheel_win32_before_build.bat
```diff
@@ -1,26 +1,7 @@
...
 
-REM On https://community.chocolatey.org/packages/postgresql15/15.0.1#discussion
-REM I found the following command in a comment:
-choco install postgresql15 --version 15.0.1
-REM which I'm going to randomly try.
+vcpkg install libpq:x64-windows-release
 
-REM See https://community.chocolatey.org/packages/postgresql15#install
-REM for the last package available (bump the version number in the url).
+pipx install .\tools\build\pg_config_vcpkg_stub\
```
dvarrazzo, line 7: Why `pipx`? `pip install` should work as expected.

tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+                print(str(vcpkg_platform_root.joinpath("include")))
+                return
+
+    print(
+        "unexpected command: {!r}\n this maybe out-of-sync between 'psycopg_c/setup.py' and 'tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py'".format(
+            sys.argv[1:]
+        ),
+        file=sys.stderr,
+    )
+    sys.exit(1)
```
dvarrazzo, line 50: This script is written in a very different style from others in the same package and has a few bad behaviours: the error message at line 44 is wrong in cases such as the package not installed.

I will give it a rewrite.

dvarrazzo (changes requested):

tools/build/wheel_win32_before_build.bat
dvarrazzo: Ok, strange. On Linux `pip` works, but of course that's not the intended target.

.github/workflows/packages-bin.yml
```diff
@@ -226,6 +231,17 @@ jobs:
...
           $PgSvc.Start()
 
+      - name: Export GitHub Actions cache environment variables
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const path = require('path')
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+            core.exportVariable('VCPKG_ROOT', process.env.VCPKG_INSTALLATION_ROOT || '');
```
dvarrazzo, line 241: Question: why to export `VCPKG_ROOT` as an alias for `VCPKG_INSTALLATION_ROOT`? AFAICS `VCPKG_ROOT` is only used by the pg_config stub. Are there other uses? If not, this can be dropped and the pg_config stub can use `VCPKG_INSTALLATION_ROOT`.

tools/build/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py
dvarrazzo: I have already rewritten this script. I will push it in this branch.

.github/workflows/packages-bin.yml
dvarrazzo: Thank you for this observation. However this means that, for people who would try to build psycopg_c using vcpkg, this script would fail if we use `VCPKG_INSTALLATION_ROOT`. If this is the case, your previous solution was better.

As I'm rewriting the script, I'll keep both the env vars into account.

dvarrazzo (approved):

dvarrazzo:
Thank you very much for this works: looks good!

I have rebased the branch and squashed your commits and re-enabled all the CI jobs. Running a final test on the workflows our side, and then this is good to go.

.github/workflows/packages-bin.yml
dvarrazzo: Weird: it's not required in `tests.yml`.

dvarrazzo:
You are right, my bad: my local master was ahead one step WRT upstream, because it's sitting on an unrelated MR. Rebasing again...

.github/workflows/packages-bin.yml
dvarrazzo: I see. Probably worth uniforming them... But we can leave it out of this MR.

dvarrazzo:
> qemu is so slow…

Why, what are you trying to do? :smile:

dvarrazzo:
Merged! Thank you very much for this work!

I actually need to do the same for psycopg2: the build system for Windows runs on appveyor and the platform is not very quick to release new runners. As a consequence we haven't provided yet Python 3.13 wheels for Windows (https://github.com/psycopg/psycopg2/issues/1736) :\

I think it will easy to backport this build procedure to psycopg 2 too. Would you be willing to provide a MR for it? If not it's no problem: thanks to your help it shouldn't be difficult for me to do it.

Thank you again!

trim21:
Sorry to reply to a already merged issue, I recently just find that conda ecosystem (conda-force) provide libpq(with a working pg_config.exe) and openssl for windows platform. And doesn't like vcpkg, you can lock version of libpq and openssl.

So maybe it's possible to resolve this comment if we replace vcpkg with conda

https://github.com/psycopg/psycopg/blob/c05c117f5f1e1ef88bcf94a12f861853797143bd/.github/workflows/packages-bin.yml#L26-L30

PR #974, opened 2024-12-22
Convert to toml

dvarrazzo:
This MR gets rid of the `setup.py` and `setup.cfg` and only uses `pyproject.toml` to define metadata.

The interaction with `pg_config` is moved to the build backend pointed at by `psycopg_c`'s pyproject.


pauloxnet (approved):
This seems to me a great advancement 👍

psycopg/pyproject.toml
```diff
@@ -1,3 +1,102 @@
...
+[project.license]
+text = "GNU Lesser General Public License v3 (LGPLv3)"
+
+[project.urls]
+Homepage = "https://psycopg.org/"
+Documentation = "https://psycopg.org/psycopg3/docs/"
+Changes = "https://psycopg.org/psycopg3/docs/news.html"
+Code = "https://github.com/psycopg/psycopg"
+"Issue Tracker" = "https://github.com/psycopg/psycopg/issues"
+Download = "https://pypi.org/project/psycopg/"
```
stankudrow, line 50: What if name it PyPI?  Because it is not only about downloading the package.

dvarrazzo: Is there a standardised set of urls? Sampling some major projects:

- numpy points to https://pypi.org/project/numpy/#files
- Django, SQLAlchemy, setuptools don't have that link.

Considering that the prime place where this link is displayed is the PyPI page, which has a prominent "Download files" link on top of the same RHS column where these links are displayed, I think we can drop it altogether.

Issue #976, opened 2024-12-23
Drop Python 3.8 compatibility


pauloxnet:
Python 3.8 is EOL since 2024-10-07 so I think it would be safe to drop the support for this unmaintained version of Python

dvarrazzo:
Yes, that's true, and it will surely happen.

Usually we drop compatibility on minor releases: as soon as we have a new feature (let's say #932) we would fork `maint-3.2` from `master`. We would then keep `maint-3.2` as bugfix and we would proceed to drop Python 3.8 compatibility from master, which eventually would be released as Psycopg 3.3.

stankudrow:
Can it be split into minor independent PRs like "Drop Py3.8 for psycopg_pool" and so on without links to any features?

dvarrazzo:
@stankudrow It can be done any moment, when it will be done it will be separated from the 3.2 maintenance branch. It doesn't need to be "linked to a feature", maybe I didn't explain myself clearly.

If we do it now, and there is no difference between master and Python 3.2, no new feature, then it's a waste of effort to maintain two otherwise identical branches; as we need to cherry-pick changes and keep the two branches in parallel, at least for bugfux-related changes, so I'd rather delay this moment until needed.

By the way I am considering that the refactoring in #974 is a bit too invasive for the next 3.2.x so maybe we can create masint-3.2 soon and land that on master instead.

As to implement it in several MR, sure, it can be done, but I don't think it's so much work that is necessary. I would drop 3.8 from metadata first, then from the CI, then I would grep around the code looking for `3.?8`, `3.?9` etc and kill dead code branches, dead requirements, etc.

stankudrow:
> @stankudrow It can be done any moment, when it will be done it will be separated from the 3.2 maintenance branch. It doesn't need to be "linked to a feature", maybe I didn't explain myself clearly.
> 
> If we do it now, and there is no difference between master and Python 3.2, no new feature, then it's a waste of effort to maintain two otherwise identical branches; as we need to cherry-pick changes and keep the two branches in parallel, at least for bugfux-related changes, so I'd rather delay this moment until needed.
> 
> By the way I am considering that the refactoring in #974 is a bit too invasive for the next 3.2.x so maybe we can create masint-3.2 soon and land that on master instead.
> 
[...]

PR #977, opened 2024-12-23
Drop Py3.8 dependencies from `pyproject.toml`s and CI/CD workflows

stankudrow:
Partially addresses the issue #976 .
Should go after the PR #974 .

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,32 +26,30 @@ jobs:
       matrix:
         include:
           # Test different combinations of Python, Postgres, libpq.
-          - {impl: python, python: "3.8", postgres: "postgres:17", libpq: oldest}
-          - {impl: python, python: "3.9", postgres: "postgres:16", libpq: master}
-          - {impl: python, python: "3.10", postgres: "postgres:15"}
-          - {impl: python, python: "3.11", postgres: "postgres:14"}
-          - {impl: python, python: "3.12", postgres: "postgres:13", libpq: newest}
+          - {impl: python, python: "3.9", postgres: "postgres:17", libpq: oldest}
+          - {impl: python, python: "3.10", postgres: "postgres:16", libpq: master}
+          - {impl: python, python: "3.11", postgres: "postgres:15"}
+          - {impl: python, python: "3.12", postgres: "postgres:14", libpq: newest}
           - {impl: python, python: "3.13", postgres: "postgres:12"}
 
-          - {impl: c, python: "3.8", postgres: "postgres:12", libpq: master}
-          - {impl: c, python: "3.9", postgres: "postgres:13"}
-          - {impl: c, python: "3.10", postgres: "postgres:14"}
+          - {impl: c, python: "3.9", postgres: "postgres:12", libpq: master}
+          - {impl: c, python: "3.10", postgres: "postgres:13"}
           - {impl: c, python: "3.11", postgres: "postgres:15", libpq: oldest}
           - {impl: c, python: "3.12", postgres: "postgres:16", libpq: newest}
           - {impl: c, python: "3.13", postgres: "postgres:17"}
 
-          - {impl: python, python: "3.8", ext: gevent, postgres: "postgres:17"}
-          - {impl: c, python: "3.12", ext: gevent, postgres: "postgres:14"}
+          - {impl: python, python: "3.9", ext: gevent, postgres: "postgres:17"}
           - {impl: python, python: "3.9", ext: dns, postgres: "postgres:14"}
           - {impl: python, python: "3.9", ext: postgis, postgres: "postgis/postgis"}
           - {impl: python, python: "3.10", ext: numpy, postgres: "postgres:14"}
           - {impl: c, python: "3.11", ext: numpy, postgres: "postgres:15"}
+          - {impl: c, python: "3.12", ext: gevent, postgres: "postgres:14"}
 
           # Test with minimum dependencies versions
           # WARNING: when bumping min version, make sure that the dependencies
           # # in tests/constraints.txt are updated and that binary packages
           # are available for such version.
-          - {impl: c, python: "3.8", ext: min, postgres: "postgres:15"}
+          - {impl: c, python: "3.9", ext: min, postgres: "postgres:15"}
 
           # Test memory alignment
           - {impl: c, python: "3.12", ext: align, postgres: "postgres:16"}
@@ -220,9 +218,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {impl: python, python: "3.8"}
           - {impl: python, python: "3.9"}
-          - {impl: c, python: "3.8"}
           - {impl: c, python: "3.9"}
 
     env:
@@ -281,13 +277,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - {impl: python, python: "3.8"}
           - {impl: python, python: "3.9"}
           - {impl: python, python: "3.10"}
           - {impl: python, python: "3.11"}
           - {impl: python, python: "3.12"}
           - {impl: python, python: "3.13"}
-          - {impl: c, python: "3.8"}
           - {impl: c, python: "3.9"}
           - {impl: c, python: "3.10"}
           - {impl: c, python: "3.11"}
--- a/docs/basic/install.rst
+++ b/docs/basic/install.rst
@@ -22,10 +22,11 @@ Supported systems
 
 The Psycopg version documented here has *official and tested* support for:
 
-- Python: from version 3.8 to 3.13
+- Python: from version 3.9 to 3.13
 
-  - Python 3.6 supported before Psycopg 3.1
+  - Python 3.8 supported before Psycopg 3.3
   - Python 3.7 supported before Psycopg 3.2
+  - Python 3.6 supported before Psycopg 3.1
 
 - PyPy: from version 3.9 to 3.10
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -32,7 +32,7 @@ How to make a psycopg release
   .. __: https://github.com/psycopg/psycopg/actions/workflows/packages-bin.yml
   .. __: https://github.com/psycopg/psycopg/actions/workflows/packages-pool.yml
 
-- Delete the ``wheelhouse`` directory there is one.
+- Delete the ``wheelhouse`` directory if there is one.
 
 - Build m1 packages by running ``./tools/build/run_build_macos_arm64.sh BRANCH``.
   On successful completion it will save built packages in ``wheelhouse``
@@ -110,3 +110,25 @@ When a new Python major version is released
 - Update the list of versions in ``tools/build/build_macos_arm64.sh`` to include
   the new version. Look for both the ``python_versions`` variable and the
   ``CIBW_BUILD`` environment variable.
+
+
+When dropping end-of-life Python versions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Update project metadata, ``requires-python`` and (maybe) package dependencies
+  in ``pyproject.toml`` files of the corresponding ``psycopg`` directories.
+
+- Update GitHub Actions workflow files in the ``.github/workflows/`` directory,
+  e.g., ``tests.yml``, ``.3rd-party-tests.yml``, ``packages-bin.yml``.
+
+- Bump versions in the ``tests/constraints.txt`` file if it is necessary.
+
+- You may grep throughout the project for occurrences of a version to be dropped.
+  However, favouring smaller pull requests is convenient and easy to review.
+  An example for grepping `end-of-life <https://endoflife.date/python>` Python 3.8::
+
+     git grep -E -e '\b3\.8\b' -e '\b(cp)?38\b' -e '\b3, 8\b'
+
+Examples:
+
+- `PR #977 <https://github.com/psycopg/psycopg/pull/977>`_
--- a/psycopg/pyproject.toml
+++ b/psycopg/pyproject.toml
@@ -13,8 +13,7 @@ classifiers = [
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3 :: Only",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
@@ -27,9 +26,8 @@ classifiers = [
     "Topic :: Software Development",
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
-requires-python = ">= 3.8"
+requires-python = ">= 3.9"
 dependencies = [
-    "backports.zoneinfo >= 0.2.0; python_version < '3.9'",
     "typing-extensions >= 4.6; python_version < '3.13'",
     "tzdata; sys_platform == 'win32'",
 ]
--- a/psycopg_pool/psycopg_pool/_acompat.py
+++ b/psycopg_pool/psycopg_pool/_acompat.py
@@ -15,7 +15,7 @@ import queue
 import asyncio
 import logging
 import threading
-from typing import Any, Callable, Coroutine, TYPE_CHECKING
+from typing import Any, Callable, Coroutine
 
 from ._compat import TypeAlias, TypeVar
 
@@ -32,24 +32,6 @@ sleep = time.sleep
 Worker: TypeAlias = threading.Thread
 AWorker: TypeAlias = "asyncio.Task[None]"
 
-# Hack required on Python 3.8 because subclassing Queue[T] fails at runtime.
-# https://stackoverflow.com/questions/45414066/mypy-how-to-define-a-generic-subclass
-if TYPE_CHECKING:
-    _GQueue: TypeAlias = queue.Queue
-    _AGQueue: TypeAlias = asyncio.Queue
-
-else:
-
-    class FakeGenericMeta(type):
-        def __getitem__(self, item):
-            return self
-
-    class _GQueue(queue.Queue, metaclass=FakeGenericMeta):
-        pass
-
-    class _AGQueue(asyncio.Queue, metaclass=FakeGenericMeta):
-        pass
-
 
 def current_thread_name() -> str:
     return threading.current_thread().name
@@ -60,7 +42,7 @@ def current_task_name() -> str:
     return t.get_name() if t else "<no task>"
 
 
-class Queue(_GQueue[T]):
+class Queue(queue.Queue[T]):
     """
     A Queue subclass with an interruptible get() method.
     """
@@ -102,7 +84,7 @@ class ACondition(asyncio.Condition):
             return False
 
 
-class AQueue(_AGQueue[T]):
+class AQueue(asyncio.Queue[T]):
     pass
 
 
--- a/psycopg_pool/psycopg_pool/abc.py
+++ b/psycopg_pool/psycopg_pool/abc.py
@@ -6,8 +6,7 @@ Types used in the psycopg_pool package
 
 from __future__ import annotations
 
-from typing import Awaitable, Callable, TYPE_CHECKING
-from typing import Union  # drop with Python 3.8
+from typing import Awaitable, Callable, TYPE_CHECKING, Union
 
 from ._compat import TypeAlias, TypeVar
 
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -21,7 +21,7 @@ Test options
 
       $ pytest
       ========================= test session starts =========================
-      platform linux -- Python 3.8.5, pytest-6.0.2, py-1.10.0, pluggy-0.13.1
+      platform linux -- Python 3.9, pytest-6.0.2, py-1.10.0, pluggy-0.13.1
       Using --randomly-seed=2416596601
       libpq available: 130002
       libpq wrapper implementation: c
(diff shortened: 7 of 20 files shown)


stankudrow:
@dvarrazzo , problematic jobs:

- https://github.com/psycopg/psycopg/actions/runs/12470290784/job/34805159937?pr=977 - linux, c 3.9, min, pg 15 - OSError: Could not find library geos_c or load any of its variants ['libgeos_c.so.1', 'libgeos_c.so']

A striking solution: installing `libgeos-dev` -> https://stackoverflow.com/questions/19742406/could-not-find-library-geos-c-or-load-any-of-its-variants .

But why it failed, what have I missed?

.github/workflows/tests.yml
```diff
@@ -90,6 +87,7 @@ jobs:
         if: ${{ matrix.impl == 'c' }}
         run: |
           echo "DEPS=$DEPS ./psycopg_c" >> $GITHUB_ENV
+          sudo apt-get install -y libgeos-dev
```
stankudrow, line 90: Worked (but maybe it is a wrong place to have it here), but another error occurred: ModuleNotFoundError: No module named 'distutils.msvccompiler' ([link](https://github.com/psycopg/psycopg/actions/runs/12471615165/job/34808906125?pr=977))

dvarrazzo:
The problem comes from the pinning of [shapely 1.7.0](https://github.com/stankudrow/psycopg/blob/4fd116dbd437e9381788d99b780269ded86b8f5a/tests/constraints.txt#L39C1-L39C8) in the `constraint.txt` file, which is the file we use to test that the code still works with a well defined set of old versions (and we don't accidentally introduce the requirement for a feature that is not available in a declared dependency).

Shapely 1.7.0 is [not available in a binary package](https://github.com/stankudrow/psycopg/blob/4fd116dbd437e9381788d99b780269ded86b8f5a/tests/constraints.txt#L39C1-L39C8) for Python 3.9. So we should bump that constraint to 1.8.0 for example, which [seems to support Python 3.9 and 3.10](https://pypi.org/project/shapely/1.8.0/#files).

dvarrazzo:
For the same reason, in the same file, we should bump numpy to at least 1.20.0.

stankudrow:
> For the same reason, in the same file, we should bump numpy to at least 1.20.0.

One moment...

Could you enlighten me, please, how have you found the problem so quickly and why these versions? I am interested in some rationale for better understanding.

tests/constraints.txt
```diff
@@ -36,7 +36,7 @@ tomli == 2.0.1
 
 # Undeclared extras to "unblock" extra features
 
-shapely == 1.7.0
-# Warning: binary package only available up to python <= 3.8.
+shapely == 1.8.0
+# Warning: binary package only available up to python <= 3.9
```
stankudrow, line 40: Is this message correct?

dvarrazzo: TBH, the placement of that comment seems wrong, as the statement is valid both for numpy and for shapely, therefore I'd move it above the two.

The sentence is not precise because for example shapely 1.8 supports Python 3.10 too. However I think it's important that `3.9` is in the comment, because this allows to grep for it when it's time to remove support from a Python version.

Therefore I propose something like:

```
# Undeclared extras to "unblock" extra features
#
# Warning: the versions specified for these packages are the oldest versions
# offering a binary package for the oldest Python version we support.
#
# When the minimum supported Python version is increased, these dependencies
# might need to be updated.
#
# Grep help: the current minimum supported version is Python 3.9.
shapely == 1.8.0
numpy == 1.20.0
```

dvarrazzo:
@stankudrow we have a documentation file with the list of chores needed when a new version of Python or Postgres is released: see https://github.com/psycopg/psycopg/blob/master/docs/release.rst

Could you maybe add a section "When dropping support to and old Python major version" and write down the operation we are doing now (updating this file, changing metadata, changing ci...)

dvarrazzo:
There is actually a comment pointing at the need of updating the `constraint.txt` file:

https://github.com/psycopg/psycopg/blob/8868a72c8be121be2148f51a0beb545cd8ed332d/.github/workflows/tests.yml#L50-L54

You re-discovered it the hard way :smile:

stankudrow:
> @stankudrow we have a documentation file with the list of chores needed when a new version of Python or Postgres is released: see https://github.com/psycopg/psycopg/blob/master/docs/release.rst
> 
> Could you maybe add a section "When dropping support to and old Python major version" and write down the operation we are doing now (updating this file, changing metadata, changing ci...)

Sure, will do.

stankudrow:
> There is actually a comment pointing at the need of updating the `constraint.txt` file:
> 
> https://github.com/psycopg/psycopg/blob/8868a72c8be121be2148f51a0beb545cd8ed332d/.github/workflows/tests.yml#L50-L54
> 
> You re-discovered it the hard way 😄

Yeah, but I did not understand where to put `libgeos-dev` there 😅

tests/constraints.txt
stankudrow: Ok, I'll just copypaste it.

.github/workflows/tests.yml
stankudrow: @dvarrazzo , so, is it OK to leave here or there is a better place for it?

dvarrazzo: No, we don't need it. The error only appeared because the wheel package for shapely 1.7.0 is not available for Python 3.9, therefore pip tried to install the requested version from source, and failed on the missing library.

Now that we are using a minimum version for which there is a wheel package for Python 3.9 we don't need this library.

stankudrow: Gotcha, understood, removed. Thank you for your patience and explanations.

dvarrazzo (changes requested):
Some of the CI entries dropped drop useful tests, so we need to move them to other versions, as suggested.

A quick grepping for `3\.8` finds work to do in:

- docs/src/install.rst
- compat.py modules and similar
- `build_macos_arm64.sh` (although this file is no more currently in use)
- typing hints here and there (Probably the tweaks added in  9ad4ea6d0bd6ae0e1a3394e03a385dee2198662d).

Grepping for `3\.9` finds:

- `psycopg_c/pyproject.toml`
- `psycopg/pyproject.toml` (`backports.zoneinfo` no more needed)

Grepping for `\b38\b` and `\b39\b` doesn't find much else (a `38` in  `tools/build/build_macos_arm64.sh` again).

Grepping for `3, 9` finds a few hits in compat modules and `tests/utils.py`

I don't have other places where to look at, at the moment.

Are these changes something you want to do in this MR or do you prefer to open a follow-up?

docs/release.rst
```diff
@@ -32,7 +32,7 @@ How to make a psycopg release
   .. __: https://github.com/psycopg/psycopg/actions/workflows/packages-bin.yml
   .. __: https://github.com/psycopg/psycopg/actions/workflows/packages-pool.yml
 
-- Delete the ``wheelhouse`` directory there is one.
+- Delete the ``wheelhouse`` directory if there is one.
```
stankudrow, line 35: or this "if" is put wrong? Sorry for my English.

docs/release.rst
```diff
@@ -110,3 +110,19 @@ When a new Python major version is released
...
+  in ``pyproject.toml`` files of the corresponding ``psycopg`` directories
+
+- Update necessary GitHub Actions workflow files in the ``.github/workflows/`` directory,
+  e.g., ``tests.yml`` and ``.3rd-party-tests.yml`` and so forth.
+
+- Bump needed versions in the ``tests/constraints.txt`` file.
+
+Examples:
+
+- `PR #977 <https://github.com/psycopg/psycopg/pull/977>`_
```
stankudrow, line 128: Looks good? Waiting for your suggestions, will be added ASAP.

docs/release.rst
dvarrazzo: It's right, thank you, it was a typo :slightly_smiling_face:

dvarrazzo:
Ah, just noticed: There are several cp38 in the `packages-bin.yml` workflow.

stankudrow:
Just leave here -> running `python3 -m flake8 .` entails to the following errors:

```shell
╰─➤  python3 -m flake8 .
./tools/async_to_sync.py:244:15: E999 SyntaxError: invalid syntax
./tools/bump_version.py:104:15: E999 SyntaxError: invalid syntax
```

... because of `match-case` syntactic structure was introduced in Py3.10 ([StackOverflow](https://stackoverflow.com/questions/46701063/why-doesnt-python-have-switch-case-update-match-case-syntax-was-added-to-pyt)).

docs/release.rst
```diff
@@ -110,3 +110,19 @@ When a new Python major version is released
 - Update the list of versions in ``tools/build/build_macos_arm64.sh`` to include
   the new version. Look for both the ``python_versions`` variable and the
   ``CIBW_BUILD`` environment variable.
+
+
+When dropping end-of-life Python versions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Update project metadata, ``requires-python`` and (maybe) package dependencies
+  in ``pyproject.toml`` files of the corresponding ``psycopg`` directories
```
dvarrazzo, line 119:
```suggestion
  in ``pyproject.toml`` files of the corresponding ``psycopg`` directories.
```

docs/release.rst
```diff
@@ -110,3 +110,19 @@ When a new Python major version is released
...
+
+- Update necessary GitHub Actions workflow files in the ``.github/workflows/`` directory,
+  e.g., ``tests.yml`` and ``.3rd-party-tests.yml`` and so forth.
```
dvarrazzo, line 122:
```suggestion
  e.g., ``tests.yml``, ``3rd-party-tests.yml``, ``packages-bin.yml``.
```

docs/release.rst
dvarrazzo: I would add (unless you want to do it in a different MR):

```
- Grep the source code for occurrences of the version to be dropped and the
  version following to find code branches depending on the version to remove.
  A few examples (dropping Python 3.8)::

     git grep -E -e '\b3\.8\b' -e '\b(cp)?38\b' -e '\b3, 8\b'
     git grep -E -e '\b3\.9\b' -e '\b(cp)?39\b' -e '\b3, 9\b'
```

dvarrazzo:
> running python3 -m flake8 . entails to the following errors

I guess it depends on which version it's run. The CI runs it in Python 3.11. I personally run it with Python 3.10 (the stock version in ubuntu 22.04, which I'm still using). I don't see the need to keep compatibility of the development tools back to the same minimum version we want to make the library available.

psycopg/pyproject.toml
```diff
@@ -13,8 +13,7 @@ classifiers = [
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3 :: Only",
```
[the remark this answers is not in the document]
stankudrow: https://pypi.org/classifiers/ - just claiming no more python 2 and since Py3 is the only supportable version, it may be fine unless Py4 is issued and both to be maintained.

I can remove this `Only` claim if you like it better.

tests/README.rst
```diff
@@ -21,7 +21,7 @@ Test options
 
       $ pytest
       ========================= test session starts =========================
-      platform linux -- Python 3.8.5, pytest-6.0.2, py-1.10.0, pluggy-0.13.1
+      platform linux -- Python 3.9, pytest-6.0.2, py-1.10.0, pluggy-0.13.1
```
stankudrow, line 24: Should it be a micro part?

tests/acompat.py
```diff
@@ -118,6 +118,7 @@ async def wait_timeout(self, timeout):
         await asyncio.wait_for(self.wait(), timeout)
 
 
+# TODO
```
stankudrow, line 121: Today after some sleep.

tests/README.rst
dvarrazzo: It's not important, it's just an illustrative example.

psycopg/pyproject.toml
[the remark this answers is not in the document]
dvarrazzo: I don't know the semantic difference between `3` and `3 :: Only`. I know that `3` doesn't mean `3.0`, so for me it had the same meaning as "not 2".

I am ok with this change, no problem. I just wanted to know if you had a better "official explanation".

stankudrow:
> * `build_macos_arm64.sh` (although this file is no more currently in use)

So this `build_macos_arm64.sh` can be safely removed.

psycopg/pyproject.toml
```diff
@@ -27,7 +26,7 @@ classifiers = [
     "Topic :: Software Development",
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
-requires-python = ">= 3.8"
+requires-python = ">= 3.9"
 dependencies = [
     "backports.zoneinfo >= 0.2.0; python_version < '3.9'",
```
pauloxnet, line 31:
```suggestion
```
Can we remove `backports.zoneinfo` ?

stankudrow:
Gosh, this should be recorded:
```shell
 diff --git a/tests/crdb/test_cursor.py b/tests/crdb/test_cursor.py
index 198feeb..aed5287 100644
--- a/tests/crdb/test_cursor.py
+++ b/tests/crdb/test_cursor.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import json
-from queue import Queue
+from asyncio import Queue as AQueue
 from typing import Any
 from uuid import uuid4
 
ERROR sync and async files... out of sync!
```

Reference: https://github.com/psycopg/psycopg/actions/runs/12479753071/job/34829429651?pr=977

psycopg_pool/psycopg_pool/_acompat.py
```diff
@@ -60,7 +42,7 @@ def current_task_name() -> str:
     return t.get_name() if t else "<no task>"
 
 
-class Queue(_GQueue[T]):
+class Queue(queue.Queue[T]):
```
stankudrow, line 45: Leave it like this or remove completely?

docs/release.rst
stankudrow: Changed, is it still good?

stankudrow:
@dvarrazzo , finally linting stages passes.

I noticed that you suggested some commits and I didn't accept them because I applied them during other changes. But I don't want you to steal your coauthorship, so, if you don't mind, feel free to suggest some commits, I'll accept them.

stankudrow:
Darn, MacOS fails: https://github.com/psycopg/psycopg/actions/runs/12480024372/job/34830158538?pr=977 .

docs/release.rst
```diff
@@ -110,3 +110,27 @@ When a new Python major version is released
...
+  However, favouring smaller pull requests is convenient and easy to review.
+  An example for grepping `end-of-life <https://endoflife.date/python>` Python 3.8::
+
+     git grep -E -e '\b3\.8\b' -e '\b(cp)?38\b' -e '\b3, 8\b'
+
+- Consider proceeding steps described in the ``.github/workflows/lint.yml`` before ``git push``.
+
+Examples:
+
+- `PR #977 <https://github.com/psycopg/psycopg/pull/977>`_
```
stankudrow, line 136: @dvarrazzo , @pauloxnet , let's have another round.

tests/acompat.py
```diff
@@ -118,17 +118,5 @@ async def wait_timeout(self, timeout):
...
-        # Always specify a timeout to make the wait interruptible.
-        if timeout is None:
-            timeout = 24.0 * 60.0 * 60.0
-        return super().get(block=block, timeout=timeout)
-
-
-class AQueue(asyncio.Queue):  # type: ignore[type-arg]
-    pass
+Queue = queue.Queue
+AQueue = asyncio.Queue
```
stankudrow, line 122: Should it be got `from psycopg.psycopg._acompat import AQueue, Queue` or just leaving it so?

stankudrow:
https://github.com/psycopg/psycopg/actions/runs/12486112485/job/34845795320?pr=977

```shell
Run docker pull postgres:12, libpq: master
"docker pull" requires exactly 1 argument.
See 'docker pull --help'.

Usage:  docker pull [OPTIONS] NAME[:TAG|@DIGEST]

Download an image from a registry
Error: Process completed with exit code 1.
```

stankudrow:
I need help...

tools/build/build_macos_arm64.sh
```diff
@@ -9,7 +9,7 @@
 
 set -euo pipefail
 
-python_versions="3.8.19 3.9.19 3.10.14 3.11.9 3.12.5 3.13.0"
+python_versions="3.9.19 3.10.14 3.11.9 3.12.5 3.13.0"
```
stankudrow, line 12: This file to be deleted in another PR because its deletion has little to do with dropping Python 3.8.

stankudrow:
Totals:

- importing some entities from `collections.abc` in lieu of `typing` module -> to be investigated
- replacing existing equivalents of entities from standard library (like AQueue which is just asyncio.Queue, or asleep which is asyncio.sleep under the hood) -> to be removed/reconsidered in other PRs
- removing `tools/build/build_macos_arm64.sh` - in another PR (this is fine)

@dvarrazzo , @pauloxnet , let's finalise this PR and merge its trade-off version.

pauloxnet (approved):
LGTM

psycopg_pool/psycopg_pool/abc.py
```diff
@@ -28,7 +27,6 @@
 
 # Callbacks to pass the pool to on connection failure
 ConnectFailedCB: TypeAlias = Callable[["ConnectionPool[Any]"], None]
-AsyncConnectFailedCB: TypeAlias = Union[
-    Callable[["AsyncConnectionPool[Any]"], None],
-    Callable[["AsyncConnectionPool[Any]"], Awaitable[None]],
+AsyncConnectFailedCB: TypeAlias = Callable[
+    ["AsyncConnectionPool[Any]"], Optional[Awaitable[None]]
```
dvarrazzo, line 31: I don't think that this definition is better than the previous one. I do understand that my comment that `Union` could be dropped in Python 3.8 was wrong (`Callable | Callable` is only supported from Python 3.10), but I would rather keep an `Union` here.

dvarrazzo (changes requested):
We are pretty close! Just a few adjustments to make, but I'm ok to not add (or remove) anything else to the MR.

docs/release.rst
```diff
@@ -110,3 +110,27 @@ When a new Python major version is released
...
+
+- Bump versions in the ``tests/constraints.txt`` file if it is necessary.
+
+- You may grep throughout the project for occurrences of a version to be dropped.
+  However, favouring smaller pull requests is convenient and easy to review.
+  An example for grepping `end-of-life <https://endoflife.date/python>` Python 3.8::
+
+     git grep -E -e '\b3\.8\b' -e '\b(cp)?38\b' -e '\b3, 8\b'
+
+- Consider proceeding steps described in the ``.github/workflows/lint.yml`` before ``git push``.
```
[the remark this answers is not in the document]
stankudrow: Ok, I remove it from release docs.

psycopg_pool/psycopg_pool/abc.py
stankudrow: Will do...

docs/basic/install.rst
```diff
@@ -22,7 +22,7 @@ Supported systems
 
 The Psycopg version documented here has *official and tested* support for:
 
-- Python: from version 3.8 to 3.13
+- Python: from version 3.9 to 3.13
 
   - Python 3.6 supported before Psycopg 3.1
   - Python 3.7 supported before Psycopg 3.2
```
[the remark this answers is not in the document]
stankudrow: Added and reversed ...

.github/workflows/tests.yml
```diff
@@ -26,32 +26,29 @@ jobs:
       matrix:
         include:
           # Test different combinations of Python, Postgres, libpq.
-          - {impl: python, python: "3.8", postgres: "postgres:17", libpq: oldest}
           - {impl: python, python: "3.9", postgres: "postgres:16", libpq: master}
           - {impl: python, python: "3.10", postgres: "postgres:15"}
           - {impl: python, python: "3.11", postgres: "postgres:14"}
           - {impl: python, python: "3.12", postgres: "postgres:13", libpq: newest}
           - {impl: python, python: "3.13", postgres: "postgres:12"}
```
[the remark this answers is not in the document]
stankudrow: I hope I wrote it right.

.github/workflows/tests.yml
```diff
@@ -26,32 +26,31 @@ jobs:
...
           # Test different combinations of Python, Postgres, libpq.
-          - {impl: python, python: "3.8", postgres: "postgres:17", libpq: oldest}
+          - {impl: python, python: "3.9", postgres: "postgres:17", libpq: oldest}
           - {impl: python, python: "3.9", postgres: "postgres:16", libpq: master}
-          - {impl: python, python: "3.10", postgres: "postgres:15"}
-          - {impl: python, python: "3.11", postgres: "postgres:14"}
-          - {impl: python, python: "3.12", postgres: "postgres:13", libpq: newest}
+          - {impl: python, python: "3.10", postgres: "postgres:16"}
+          - {impl: python, python: "3.11", postgres: "postgres:15"}
+          - {impl: python, python: "3.12", postgres: "postgres:14", libpq: newest}
```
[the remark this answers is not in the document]
stankudrow: Done, leaving [this comment](https://github.com/psycopg/psycopg/pull/977/files#r1896102978) unresolved for cross-validation sake.

stankudrow:
Daniele and Paolo, thank you for your sharp attention, enormous patience and massive help.

dvarrazzo (approved):

dvarrazzo (approved):

dvarrazzo:
Merged! Thank you very much for your patience, @stankudrow!

stankudrow:
> Merged! Thank you very much for your patience, @stankudrow!

Thank you.

Issue #921, opened 2024-09-30
DOC: Dumper.get_key() is documented to return type "]]"


j-carson:
Two closing brackets as a type hint makes no sense

get_key(obj: [Any](https://docs.python.org/3/library/typing.html#typing.Any), format: [PyFormat](https://www.psycopg.org/psycopg3/docs/api/adapt.html#psycopg.adapt.PyFormat)) → ]]

https://www.psycopg.org/psycopg3/docs/api/abc.html#psycopg.abc.Dumper.get_key

dvarrazzo:
It's a Sphinx bug, probably it fails to deal with a recursive type definition, and I don't have the resource to debug it.

If you can forward the issue to the sphinx project I'd be grateful.

j-carson:
I'll have to see if I can make a small reproducible test case.

stankudrow:
Should it look like this?

![Screenshot from 2024-12-25 21-31-32](https://github.com/user-attachments/assets/b9df5b31-aee3-4572-b3b5-a926940053b6)

PR #979, opened 2024-12-26
Fix `Dumper.get_key` method return type

stankudrow:
May fix the issue #921 .

Won't work in Py3.8, follows the #976 .

UPD: Should go after #982 .


stankudrow:
@dvarrazzo , hello.

Still works and the same as it were posted in the issue #921 . Is this solution enough?

Screenshot from 2025-01-04 10-10-30

Screenshot from 2025-01-04 10-11-13

dvarrazzo:
This works. I've changed the commit message.

dvarrazzo:
Thank you very much!

PR #980, opened 2024-12-27
Add pre-commit hooks

### LINKED ISSUE #978: Set up pre-commit hooks
First-time contributors will likely trigger CI lint fail. The set of linting applied is not documented.

Set up [pre-commit hooks](https://pre-commit.com/), to reproduce the same linting.

I'm not sure that adding async-to-sync to the hooks is a good idea anyway, because it requires either a specific Python version or Docker/Podman.

See https://github.com/psycopg/psycopg/pull/977#discussion_r1897423297

dvarrazzo:
This MR adds pre-commit hooks to run black, codespell, flake8, mypy. This allows contributors to find issues before submitting changes that will be rejected by the lint step.

The linters are executed in the dev virtualenv, instead of in a separate virualenv as pre-commit does. In my experience, a separate environment creates problems and makes the behaviour more obscure.

Note that it's not easy to add async-to-sync to the pre-commit hook in a generic way because it depends either on a specific system python version installed or on docker/podman.

Note that running:

- `mypy` (uses the files in `pyproject.toml`)
- `mypy .` (runs on every file)
- `mypy FILE1.py` (typically run by an editor)
- `mypy FILE1.py FILE2.py` (typically run by `pre-commit run -a`)

seem to yield different result. The current state is meta-stable and mypy seems to pass in every case. The amount of typing changes in MR is due to `mypy` working differently if running with a set of files.

Close #978.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,5 +22,5 @@ repos:
       - id: mypy
         name: mypy
         language: system
-        entry: mypy --pretty
+        entry: mypy --pretty --follow-imports=silent
         files: \.py[i]?$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,26 @@
+# See https://pre-commit.com for more information
+repos:
+  - repo: local
+    hooks:
+      - id: black
+        name: black
+        language: system
+        entry: black --check --diff
+        files: \.py[i]?$
+
+      - id: codespell
+        name: codespell
+        language: system
+        entry: codespell
+
+      - id: flake8
+        name: flake8
+        language: system
+        entry: flake8
+        files: \.py$
+
+      - id: mypy
+        name: mypy
+        language: system
+        entry: mypy --pretty
+        files: \.py[i]?$
--- a/psycopg/psycopg/_copy_base.py
+++ b/psycopg/psycopg/_copy_base.py
@@ -216,10 +216,11 @@ class TextFormatter(Formatter):
         self._encoding = encoding
 
     def parse_row(self, data: Buffer) -> tuple[Any, ...] | None:
+        rv: tuple[Any, ...] | None = None
         if data:
-            return parse_row_text(data, self.transformer)
-        else:
-            return None
+            rv = parse_row_text(data, self.transformer)
+
+        return rv
 
     def write(self, buffer: Buffer | str) -> Buffer:
         data = self._ensure_bytes(buffer)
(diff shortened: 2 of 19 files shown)


dvarrazzo:
@stankudrow if you had found this setup (readme, tooling) when you worked on your MR in the previous days, would have you found it useful? Feedback is welcome.

stankudrow:
> @stankudrow if you had found this setup (readme, tooling) when you worked on your MR in the previous days, would have you found it useful? Feedback is welcome.

Aight, I'll have a look at your PR.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
+# See https://pre-commit.com for more information
+repos:
+  - repo: local
+    hooks:
+      - id: black
+        name: black
+        language: system
+        entry: black --check --diff
+        files: \.py[i]?$
```
stankudrow, line 9: What is the purpose for this line? `black` looks for `.py` files itself. Yes, it should also seek for `.pyi` files, but can it be configured in the `pyproject.toml` file?

psycopg/psycopg/_copy_base.py
```diff
@@ -216,10 +216,11 @@ def __init__(self, transformer: Transformer, encoding: str = "utf-8"):
         self._encoding = encoding
 
     def parse_row(self, data: Buffer) -> tuple[Any, ...] | None:
+        rv: tuple[Any, ...] | None = None
```
stankudrow, line 219: can it be just `tuple`?

psycopg/psycopg/_cursor_base.py
```diff
@@ -334,7 +334,7 @@ def _stream_send_gen(
         yield from send(self._pgconn)
 
     def _stream_fetchone_gen(self, first: bool) -> PQGen[PGresult | None]:
-        res = yield from fetch(self._pgconn)
+        res: PGresult | None = yield from fetch(self._pgconn)
         if res is None:
             return None
```
stankudrow, line 339: Maybe just `return res` is enough?

psycopg/psycopg/_copy_base.py
dvarrazzo: No, I don't think it's a valid type annotation.

.pre-commit-config.yaml
dvarrazzo: If you just run `black [DIRECTORY]` it will look for Python files only, yes, but pre-commit will pass it the list of all the files changed, and black will try to work on non-python files too, failing. Like:

```
$ black pyproject.toml 
error: cannot format pyproject.toml: cannot use --safe with this file; failed to parse source file AST: [...]
```

so this argument tells pre-commit to not pass to black any file that is not Python.

psycopg/psycopg/_cursor_base.py
dvarrazzo: No, the error is downstream, if the annotation is misisng:

```
psycopg/psycopg/_cursor_base.py:347: error: Returning Any from function
declared to return "PGresult | None"  [no-any-return]
                return res
                ^~~~~~~~~~
Found 1 error in 1 file (checked 1 source file)
```

PR #981, opened 2024-12-27
Lint the code via `pyupgrade --py38-plus ...`

stankudrow:
Follows the PR #977 (the issue #976).

[Pyupgrade](https://pypi.org/project/pyupgrade/) is a tool (and pre-commit hook) to automatically upgrade syntax for newer versions of the language (desc from project page).

Is good to go after the PR #980 because it a pre-commit hook for pyupgrade can be added as well (perhaps it is more likely to be unnecessary).


dvarrazzo:
Question: Can this tool apply the walrus operator where it makes sense? I.e. converting:
```python
        res = yield from fetch(self._pgconn)
        if res is None:
            return None
```
into
```python
        if (res := yield from fetch(self._pgconn)) is None
            return None
```

dvarrazzo:
I am ok to have a pass with this tool, but adding it as a hook seems unnecessary :slightly_smiling_face:

stankudrow:
> I am ok to have a pass with this tool, but adding it as a hook seems unnecessary 🙂

I agree.

dvarrazzo:
We can suggest a run through pyupgrade as a step following the removal of an old Python version in the notes you wrote a few days ago.

dvarrazzo:
About walrus conversion: I have prepared a script, but still have to test it better: https://gist.github.com/dvarrazzo/a0e9bd4bee4bb64b543332721de294c0

dvarrazzo:
I have noticed that you should have run `--py39-plus` instead, as we dropped support for Python 3.8.

I have run it and tweaked with the result. There are further changes that can be applied because the `version >= (3, 9)` guards that can be included. I'll push them in another MR.

stankudrow:
> I have noticed that you should have run `--py39-plus` instead, as we dropped support for Python 3.8.
> 
> I have run it and tweaked with the result. There are further changes that can be applied because the `version >= (3, 9)` guards that can be included. I'll push them in another MR.

Hello 👋🏽 .

Yes, my mistake, I close this PR in favour of # 982 .

dvarrazzo:
Ah, ok, I thought we would have merged this and then the other changesets in #982. No problem either way.

PR #982, opened 2024-12-28
Cleanup compat

dvarrazzo:
Cleanup after dropping from the `_compat` modules objects that don't need to be imported from different places because we dropped support for Python 3.8.

cc @stankudrow.


tests/dbapi20.py
```diff
@@ -191,31 +191,31 @@ def test_Exceptions(self):
...
         if sys.version[0] == '3': #under Python 3 StardardError no longer exists
-            self.failUnless(issubclass(self.driver.Warning,Exception))
-            self.failUnless(issubclass(self.driver.Error,Exception))
+            self.assertTrue(issubclass(self.driver.Warning,Exception))
+            self.assertTrue(issubclass(self.driver.Error,Exception))
         else:
-            self.failUnless(issubclass(self.driver.Warning,StandardError))  # type: ignore[name-defined]
-            self.failUnless(issubclass(self.driver.Error,StandardError))  # type: ignore[name-defined]
+            self.assertTrue(issubclass(self.driver.Warning,StandardError))  # type: ignore[name-defined]
+            self.assertTrue(issubclass(self.driver.Error,StandardError))  # type: ignore[name-defined]
```
stankudrow, line 198: Daniele, could you explain the purpose of this "else" branch, please?

stankudrow:
Looks good to me.

tests/dbapi20.py
dvarrazzo: It can be dropped, sure.

This test suite is a sort of historic thing, to show basic adherence to the DBAPI standard of a driver. So it's largely been untouched to maintain it as original as possible. But for sure we don't need to support Python 2.

PR #986, opened 2025-01-06
Build libpq on macOS

dvarrazzo:
Build the libpq from source, instead of using Homebrew libraries, on macOS.

This avoids problems with `MACOSX_DEPLOYMENT_TARGET` drifting in builders (see https://github.com/psycopg/psycopg/issues/858).

Largely contributed by @bwoodsend in https://github.com/psycopg/psycopg2/pull/1755. Thank you very much!

### DIFF
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -5,19 +5,19 @@ on:
   schedule:
     - cron: '28 7 * * sun'
 
+env:
+  # Latest release: https://www.postgresql.org/ftp/source/
+  LIBPQ_VERSION: "17.0"
+  # Latest release: https://www.openssl.org/source/
+  # Note: 3.3.2 is broken:
+  # https://github.com/openssl/openssl/issues/25366
+  OPENSSL_VERSION: "3.3.1"
+
 jobs:
 
   linux:  # {{{
     runs-on: ubuntu-latest
     if: true
-
-    env:
-      # Latest release: https://www.postgresql.org/ftp/source/
-      LIBPQ_VERSION: "17.0"
-      # Latest release: https://www.openssl.org/source/
-      # Note: 3.3.2 is broken:
-      # https://github.com/openssl/openssl/issues/25366
-      OPENSSL_VERSION: "3.3.1"
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Hmm... [latest builds](https://github.com/psycopg/psycopg/actions/runs/12624394823/job/35174601987) fail delocating.

@bwoodsend maybe you have an idea about what is going wrong?

Note: main differences between this and psycopg2 is the use of libpq 17 and openssl 3 (which probably should be backported to pg2, especially openssl).

bwoodsend:
Looks like the deployment target is getting lost somewhere. All the libraries being bundled are the ones we built but we've built >=14.0 compatible ones. Can we enable bash's noisy mode like we have in pyschopg2?

bwoodsend:
Nevermind, I'm spinning it up locally...

dvarrazzo:
Echo on is echoing.

bwoodsend:
Hmph, can't reproduce locally. `otool -l /tmp/libpq.build/lib/*.dylib | grep -E '^/tmp/.*|minos'` is giving me the long stream of 11.0s that CI should be getting too.

The [`export MACOSX_DEPLOYMENT_TARGET` line](https://github.com/psycopg/psycopg/actions/runs/12624921082/job/35175843715?pr=986#step:5:256) is running. Not sure what can go wrong after that.

bwoodsend:
Oh, it's pulling `libpq.dylib` [from homebrew](https://github.com/psycopg/psycopg/actions/runs/12624921082/job/35175843715?pr=986#step:5:16266) instead of the one we built. So I'm guessing it's the extension module building step that needs to be told when to pull its `libpq` from.

bwoodsend:
Think we just need to define `LDFLAGS='-L/tmp/libpq.build/lib'` for the `pip wheel` step. That makes it behave for me locally.

dvarrazzo:
Probably the problem was the inclusion in the path of the homebrew postgres bin directory, which was necessary previously to use the right `pg_config`, but now obviously getting in the way. Fixed in 58544a9894bb9cb51bcf43d044ddc75e842f13fc.

Getting there. Now it builds but fails import with [missing `_PQbackendPID` symbol](https://github.com/psycopg/psycopg/actions/runs/12629717479/job/35188156376#step:5:16741). The weird thing is that this symbol is not in postgres codebase nor in psycopg.

I'll run a quick test by dropping the lib64 directory (necessary for linux libpq 17 but maybe not for macOS), then I'll have to stop for the day, because I'm traveling today.

Cheers!

bwoodsend:
That's actually the same failure as before but you're just looking at the other architecture. The two are failing with quite different error messages although I think it's the same root cause: picking up the wrong `libpq` so arm64 gets the wrong deployment target and x86_64 gets the wrong architecture.

`LD_LIBRARY_PATH` is really called `DYLD_LIBRARY_PATH` on macOS. I suspect amending 58544a9894bb9cb51bcf43d044ddc75e842f13fc to use the latter is all that's missing. I'll spin up a fork and try that theory...

bwoodsend:
Ahh, that's what we were missing. Your latest change plus the [LD_LIBRARY_PATH -> DYLIB_LIBRARY_PATH](https://github.com/bwoodsend/psycopg/commit/d26a2d93200e469d2d69f38a42ee5d0a54b76436) change [fixes it](https://github.com/bwoodsend/psycopg/actions/runs/12758547964).

dvarrazzo:
The problem was probably the one fixed here: 9ec29f3fe68a8fb1ea4a3f585eae4b0313895326. The env var is not defined there, so we missed our version of pg_config and ended up using the homebrew one.

bwoodsend:
Good to see it's working.

Feel free to @mention me if it starts playing up in the future.

dvarrazzo:
Thank you very much, @bwoodsend, I hope we will not have to bother you too much.

I'm running the last chores before packaging psycopg 3.2.4, for which we would distribute these packages. Finger crossed...

:crossed_fingers:  :grimacing: :crossed_fingers:

PR #989, opened 2025-01-08
Fix clobbered error on disconnect

### LINKED ISSUE #988: How can I know when Postgre is returning a idle_in_transaction_session_timeout?
Originally posted by @b0uh in https://github.com/psycopg/psycopg/discussions/987

Hello,

I'm migrating some code from psycopg2 to psycopg3 and it is going well so far, thanks for the hard work.
I have some difficulties around the behavior of `idle_in_transaction_session_timeout` setting though.

**With psycopg3, the error/exception is generic:** 
```py
import psycopg
from time import sleep

with psycopg.connect("...") as conn:
    with conn.cursor() as cursor:
        cursor.execute("SET SESSION idle_in_transaction_session_timeout = 1000")
        sleep(2)
        cursor.execute("SELECT * from pg_tables")

# Exception raised:
# OperationalError: consuming input failed: server closed the connection unexpectedly
# 	This probably means the server terminated abnormally
# 	before or while processing the request.
```

**With psycopg2, the error/exception is more explicit:** 
```py
import psycopg2
conn = psycopg2.connect("....")
with conn.cursor() as cursor:
    cursor.execute("SET SESSION idle_in_transaction_session_timeout = 1000")
    sleep(2)
    cursor.execute("SELECT * from pg_tables")
conn.close()

# Exception raised:
# psycopg2.errors.IdleInTransactionSessionTimeout: terminating connection due to idle-in-transaction timeout
# server closed the connection unexpectedly
# 	This probably means the server terminated abnormally
# 	before or while processing the request.
```

While looking at the code of psycopg3 I noticed the [`IdleInTransactionSessionTimeout`](https://github.com/psycopg/psycopg/blob/5cb56350c18c5b74fb6463bbebd43142c1165f68/psycopg/psycopg/errors.py#L1103) exception which makes me wonder if maybe I missed something.

Does someone has been able to have clear error for idle_in_transaction_session_timeout?

Thank you!</div>

dvarrazzo:
Avoid that an exception caused by a disconnection clobbers a server error already received as result. This happens when the server returns an error and immediately closes the connection, such as when hitting an idle-in-transaction timeout.

Close #988.

### DIFF
--- a/psycopg/psycopg/generators.py
+++ b/psycopg/psycopg/generators.py
@@ -180,13 +180,13 @@ def _fetch_many(pgconn: PGconn) -> PQGen[list[PGresult]]:
         try:
             res = yield from _fetch(pgconn)
         except e.DatabaseError:
-            # What might have happened here is that a previuos error
-            # disconnected the connection, for example a idle in transaction
-            # timeout. Check if we had received an error before, and raise it
-            # as exception, because it should contain more details. See #988.
-            for res in results:
-                if res.status == FATAL_ERROR:
-                    raise e.error_from_result(res, encoding=pgconn._encoding) from None
+            # What might have happened here is that a previuos error disconnected
+            # the connection, for example a idle in transaction timeout.
+            # Check if we had received an error before: if that's the case
+            # just exit the loop. Our callers will handle this error and raise
+            # it as an exception.
+            if any(res.status == FATAL_ERROR for res in results):
+                break
             else:
                 raise
 
--- a/psycopg/psycopg/generators.py
+++ b/psycopg/psycopg/generators.py
@@ -46,6 +46,7 @@ COMMAND_OK = pq.ExecStatus.COMMAND_OK
 COPY_OUT = pq.ExecStatus.COPY_OUT
 COPY_IN = pq.ExecStatus.COPY_IN
 COPY_BOTH = pq.ExecStatus.COPY_BOTH
+FATAL_ERROR = pq.ExecStatus.FATAL_ERROR
 PIPELINE_SYNC = pq.ExecStatus.PIPELINE_SYNC
 
 WAIT_R = Wait.R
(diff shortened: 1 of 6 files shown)


psycopg/psycopg/generators.py
```diff
@@ -176,7 +177,20 @@ def _fetch_many(pgconn: PGconn) -> PQGen[list[PGresult]]:
...
-        res = yield from _fetch(pgconn)
+        try:
+            res = yield from _fetch(pgconn)
+        except e.DatabaseError:
+            # What might have happened here is that a previuos error disconnected
+            # the connection, for example a idle in transaction timeout.
+            # Check if we had received an error before: if that's the case
+            # just exit the loop. Our callers will handle this error and raise
+            # it as an exception.
+            for res in results:
```
akrherz, line 188: I am likely wasting your time, sorry in advance, but I don't understand how `results` is not always an empty list from line 178.

dlax: @akrherz, the `results` list is appended below, l. 196.

dvarrazzo:
Interesting fact: after fixing this issue, a disconnection resulting by `pg_cancel_backend()` is now reported as `AdminShutdown` instead of `OperationalError`. Because the former is a subclass of the latter, this change is backward compatible.

This is a welcome change, so I've added a test in df9abaed7ce1de7c2de8a3ed0b03ed766f551a10 to make sure that the behaviour won't regress.

Note that the previous [CI test failure](https://github.com/psycopg/psycopg/actions/runs/12669494230) seem to indicate that Windows and CRDB behave differently. I'm running a [final complete test run](https://github.com/psycopg/psycopg/actions/runs/12670789482), then I will annotate the tests in order to skip them on the platform where this detail of errors-on-disconnection behave differently than the optimal (Postgres on Linux/macOS).

Issue #447, opened 2022-11-30
Write Cython UUID dumper/loader


dvarrazzo:
UUID is more and more used in databases. The fact that the UUID adapter is in Python only is probably a performance penalty when using this data type.

dvarrazzo:
Implemented in #998.

PR #998, opened 2025-01-27
Add Cython UUID Loader

henadzit:
[Related ticket](https://github.com/psycopg/psycopg/issues/447)

Hey! Thanks for such a well maintained package!

This PR adds `UUIDLoader` and `UUIDBinaryLoader` written in Cython that are faster than their Python counterparts. I used CodSpeed to measure the performance in my fork. Below are the results:

Screenshot 2025-02-13 at 17 16 42

[The benchmarks](https://codspeed.io/henadzit/psycopg/branches/cython-uuid-benchmarks-2) and related GitHub Actions are in my repo. I haven't included them with this PR because they are very specific to the changes and I'm not sure if the community wants them, but I suggest to consider using CodSpeed. It makes it very convenient to monitor how PRs affect performance. I have used with great success with [Tortoise ORM](https://github.com/tortoise/tortoise-orm/).


dvarrazzo:
Hello He, thank you for the initiative!

Some time ago I gave a quick look to the UUID optimization but I found it's a standard Python object and haven't seen an immediate way to make it decently faster. Looking at the uuid.py code probably the biggest win is we bypass the init method and set the `int` and `is_safe` slots in a more streamlined way.

About Cython optimization: the first step for me is to check the C code generated and verify if it's creating some sleek C or a big bunch of instruction for a single Cython instruction. you can do it by setting `annotate=True` when building the extension:

https://github.com/psycopg/psycopg/blob/35af223c6454fc34d19145b2af9dd467137254b0/psycopg_c/build_backend/psycopg_build_ext.py#L66

and enjoy the generated html pages. I can give it a look too later.

dvarrazzo:
A quick example of bypassing `__init__`:

```python
>>> import uuid
>>> u = uuid.UUID.__new__(uuid.UUID)
>>> object.__setattr__(u, 'is_safe', uuid.SafeUUID.unknown)
>>> object.__setattr__(u, 'int', 0)
>>> u
UUID('00000000-0000-0000-0000-000000000000')
```

of course we need tests in the test suite to make sure that this implementation details will not change in different versions of Python and we can have a fast version and a slow version of the loaders.

Would you like to check if an implementation of the uuid loaders based on this hack works? We can start from the Python version, and if it works we can port it to Cython.

henadzit:
> Would you like to check if an implementation of the uuid loaders based on this hack works? We can start from the Python version, and if it works we can port it to Cython.

This is awesome! Yes, I'll give it a try and get back to you!

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,66 @@
...
+            j += 1
+        hex_str[32] = 0
+
+        # uuid is slow to import, lazy load it
+        import uuid
+
+        u = uuid.UUID.__new__(uuid.UUID)
+        object.__setattr__(u, 'is_safe', uuid.SafeUUID.unknown)
+        object.__setattr__(u, 'int', PyLong_FromString(hex_str, NULL, 16))
+        return u
```
dvarrazzo, line 52: This code seems vulnerable to malformed input: what happens if the input string doesn't have exactly 36 characters which are not a '-'?

However, as it can be made more robust, it can be also be made massively faster.

1) Create a static lookup table of 256 bytes, all 0 except the uppercase and lowercase hex digits, which will have their value (assign 1 to the char `1`, assign 10 to the chars `A` and `a`, etc). Similar to this:

https://github.com/psycopg/psycopg/blob/be4e811369e446188e3ffd6c000dd32eec594023/psycopg_c/psycopg_c/_psycopg/copy.pyx#L324-L327

2) The algorithm becomes, pseudo:

```
sum = 0
ndigits = 0
for c in data:
    sum *= 16
    sum += lut[c]
    ndigits += 1

if ndigits != 32:
    raise ValueError...

...
object.__setattr__(u, 'int', sum)
```

so you build your int directly as you scan the string and you don't need to call `PyLong_FromString`.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,66 @@
...
+        cdef int j = 0
+        for i in range(36):
+            if data[i] == b'-':
+                continue
+            hex_str[j] = data[i]
+            j += 1
+        hex_str[32] = 0
+
+        # uuid is slow to import, lazy load it
+        import uuid
```
dvarrazzo, line 47: Slower operations might be made in the loader init, not in cload (called once per query), which is the hottest path (called once per datum).

See the Python implementation of the loader, for example.

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: @dvarrazzo thank you for the feedback!

Regarding the lookup table, I'm not sure about how faster it is going to be because we are dealing with 128 bit integers that do not fit in CPU registers and require `PyLong`. On other hand, `PyLong_FromString` relies on heavily optimized [long_from_binary_base](https://github.com/python/cpython/blob/a810cb89f108f026d70c5b20141b709493051d96/Objects/longobject.c#L2636). I'll give it a try though!

dvarrazzo: Good observation. It would be interesting to check if doing the same but in two batches of 64 bits, and then aggregate them at the end (b1 << 64 + b2), would make a measurable difference...

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: Moved to `__cinit__`

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: I added `hex_to_int_map` but used -1 for everything except 0-9a-zA-Z because we actually need to skip "-", "}" and "{". I also added the length check. There doesn't seem to be much of a difference performance wise but this version seems safer against malformed input. Let me know what you think!

There seems to be a failing test, "crdb (c, latest-master-build, 3.13)", not sure if it is related but I'll have a look.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,118 @@
...
+            import uuid
+
+    cdef object cload(self, const char *data, size_t length):
+        cdef uint64_t high = 0
+        cdef uint64_t low = 0
+        cdef int i
+        cdef int ndigits = 0
+        cdef int8_t c
+
+        for i in range(length):
```
dvarrazzo, line 81: This line causes a warning (which can be seen with `pip install -v ./psycopg_c`):
```
  psycopg_c/_psycopg.c: In function ‘__pyx_f_9psycopg_c_8_psycopg_10UUIDLoader_cload’:
  psycopg_c/_psycopg.c:71240:31: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
  71240 |   __pyx_v_i = __pyx_t_3;
        |                               ^
```

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,118 @@
...
+                high = (high << 4) | hex_to_int_map[c]
+            else:
+                low = (low << 4) | hex_to_int_map[c]
+            ndigits += 1
+
+        if ndigits != 32:
+            raise ValueError("Invalid UUID string")
+
+        cdef object py_high = PyLong_FromUnsignedLongLong(high)
+        cdef object py_low = PyLong_FromUnsignedLongLong(low)
```
dvarrazzo, line 96: I wonder if we can infer the `sizeof(unsigned long long)`, and if it's >= 16, execute this operation in C rather than Python.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,118 @@
...
+
+        u = uuid.UUID.__new__(uuid.UUID)
+        object.__setattr__(u, 'is_safe', uuid.SafeUUID.unknown)
+        object.__setattr__(u, 'int', (py_high << 64) | py_low)
```
dvarrazzo, line 100: These operations are relatively inefficient, in cython scale, because they contain repeated lookups.

We can cache the lookups in init and use them with `PyObject_CallFunctionObjArgs` to produce more streamlined C.

The following compiles with no warning and passes the current tests. Could you please check if the speedup is measurable?

```diff
diff --git a/psycopg_c/psycopg_c/types/uuid.pyx b/psycopg_c/psycopg_c/types/uuid.pyx
index 0682b50c..2533b9f9 100644
--- a/psycopg_c/psycopg_c/types/uuid.pyx
+++ b/psycopg_c/psycopg_c/types/uuid.pyx
@@ -65,12 +65,22 @@ static const int8_t hex_to_int_map[] = {
 cdef class UUIDLoader(CLoader):
     format = PQ_TEXT
 
+    cdef PyObject *_uuid_type
+    cdef object _uuid_new
+    cdef object _obj_setattr
+    cdef PyObject *_safeuuid_unknown
+
     def __cinit__(self, oid: int, context: AdaptContext | None = None):
         global uuid
         # uuid is slow to import, lazy load it
         if uuid is None:
             import uuid
 
+        self._uuid_type = <PyObject *>uuid.UUID
+        self._uuid_new = uuid.UUID.__new__
+        self._obj_setattr = object.__setattr__
+        self._safeuuid_unknown = <PyObject *>uuid.SafeUUID.unknown
+
     cdef object cload(self, const char *data, size_t length):
         cdef uint64_t high = 0
         cdef uint64_t low = 0
@@ -94,10 +104,13 @@ cdef class UUIDLoader(CLoader):
 
         cdef object py_high = PyLong_FromUnsignedLongLong(high)
         cdef object py_low = PyLong_FromUnsignedLongLong(low)
+        py_int = (py_high << 64) | py_low
 
-        u = uuid.UUID.__new__(uuid.UUID)
-        object.__setattr__(u, 'is_safe', uuid.SafeUUID.unknown)
-        object.__setattr__(u, 'int', (py_high << 64) | py_low)
+        cdef object u = PyObject_CallFunctionObjArgs(self._uuid_new, self._uuid_type, NULL)
+        PyObject_CallFunctionObjArgs(self._obj_setattr,
+            <PyObject *>u, <PyObject *>"is_safe", self._safeuuid_unknown, NULL)
+        PyObject_CallFunctionObjArgs(self._obj_setattr,
+            <PyObject *>u, <PyObject *>"int", <PyObject *>py_int, NULL)
         return u
```

tests/types/test_uuid.py
```diff
@@ -32,7 +41,13 @@ def test_uuid_load(conn, fmt_out):
...
+    uuid_val = UUID(val)
+    assert res == uuid_val
+    # the C modules bypasses __init__, so checking the state of the UUID object
+    assert res.hex == uuid_val.hex
+    assert res.int == uuid_val.int
+    assert res.bytes == uuid_val.bytes
+    assert res.is_safe == uuid_val.is_safe
 
 
 @pytest.mark.slow
```
dvarrazzo, line 53: I would like to add a test to verify that the UUID structure doesn't change in future versions; if it does we might need to adapt our loaders.

```python
assert uuid.UUID.__slots__ == ("int", "is_safe", "__weakref__"), "UUID structure changed"
```

psycopg_c/psycopg_c/types/uuid.pyx
dvarrazzo: Definitely crdb master is broken independently from these changes, no problem about it :)

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: Will do!

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: Sorry, I'm having a hard time to understand what you mean, could you please elaborate?

dvarrazzo: As you pointed out, calculations in > 8 bytes registry might be slow, so we split the calculation in two 8 bytes var, `low` and `high`

Now we need to put them together in `num = (high << 64) | low`.

If we are certain that `unsigned long long` is at least 128 bits, we could do this operation in C, otherwise we need to do it in Python. Doing it in C saves from creating 2 pythonn ints and to run the calculations.

So I'd like to do the equivalent of:
```
cdef object py_num
#if sizeof(unsigned long long) >= 16
    cdef unsigned long long c_num = (<unsigned long long>high << 64) | low
    py_num =  PyLong_FromUnsignedLongLong(c_num)
#else
    cdef object py_high = PyLong_FromUnsignedLongLong(high)
    cdef object py_low = PyLong_FromUnsignedLongLong(low)
    py_num = py_high << 64 | py_low
#endif
```

but cython doesn't really have a pre-compiler, so we need to insert a snippet of C to do that. I can take a look at it.

dvarrazzo: I cheched and even on linux 64 bits `unsigned long long` is 8 bits, so I doubt there is any platform which would benefit of that.

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: Fixed!

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: Updated, it looks like it improved the benchmark results by 10-15%, here's the CodSpeed comparison between the before and after https://codspeed.io/henadzit/psycopg/runs/compare/679fcea63abf8fbadcd20fad..67a655c4ecb907f9856e0644

tests/types/test_uuid.py
henadzit: Added!

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
+cimport cython
+
+from types import ModuleType
+from cpython.bytes cimport PyBytes_AsString
+from cpython.long cimport PyLong_FromUnsignedLongLong
+
+cdef extern from "Python.h":
+    # PyUnicode_AsUTF8 was added to cpython.unicode in 3.1.x but we still
+    # support 3.0.x
```
dvarrazzo, line 9: Isn't this comment out of date? We only support 3.8+ and we never supported anything before 3.6.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
...
+    const char *PyUnicode_AsUTF8(object unicode) except NULL
+
+
+uuid: ModuleType | None = None
```
dvarrazzo, line 13: I might have to rethink this. From previous experiences with psycopg2, storing Python objects in static C variables is dangerous for environments using multiple interpreters (such as Apache mod_python).

Maybe it would be better to drop this module var and just have `import uuid` in the `cinit`s. Importing an already imported module is a fast operation anyway.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
...
+    cdef PyObject *_uuid_type
+    cdef object _uuid_new
+    cdef object _obj_setattr
+    cdef PyObject *_safeuuid_unknown
+
+    def __cinit__(self, oid: int, context: AdaptContext | None = None):
+        global uuid
+        # uuid is slow to import, lazy load it
+        if uuid is None:
+            import uuid
```
dvarrazzo, line 77: As per above, I propose to do without the module-level name.
```suggestion
        # uuid is slow to import, lazy load it
        import uuid
```

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
...
```
dvarrazzo, line 126: As per above, I propose to do without the module-level name.
```suggestion
        # uuid is slow to import, lazy load it
        import uuid
```

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
...
+
+        self._uuid_type = <PyObject *>uuid.UUID
+        self._uuid_new = uuid.UUID.__new__
+        self._obj_setattr = object.__setattr__
+        self._safeuuid_unknown = <PyObject *>uuid.SafeUUID.unknown
+
+    cdef object cload(self, const char *data, size_t length):
+        cdef object int_value = int.from_bytes(data[:length], 'big')
```
dvarrazzo, line 134: We can load the binary data into an array of two 64 bits uuid, swap them to host order, and then use the same high+low procedure used in the text loader to get a 128 bits integer.

```suggestion
        cdef uint64_t be[2]
        if length != sizeof(be):
            raise ValueError("Invalid UUID data")
        memcpy(&be, data, sizeof(be))

        cdef uint64_t high = endian.be64toh(be[0])
        cdef uint64_t low = endian.be64toh(be[1])
        cdef object int_value = (PyLong_FromUnsignedLongLong(high) << 64) | PyLong_FromUnsignedLongLong(low)
```

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -0,0 +1,138 @@
...
+    def __cinit__(self, oid: int, context: AdaptContext | None = None):
+        global uuid
+        # uuid is slow to import, lazy load it
+        if uuid is None:
+            import uuid
+
+        self._uuid_type = <PyObject *>uuid.UUID
+        self._uuid_new = uuid.UUID.__new__
+        self._obj_setattr = object.__setattr__
+        self._safeuuid_unknown = <PyObject *>uuid.SafeUUID.unknown
```
dvarrazzo, line 131: The `__cinit__` is identical for the two loaders. Probably we should introduce a base class from which to derive both.

Probably we can also have an internal method `_uuid_from_int`, so that in the `cload()` we can just deal with parsing the data into the `int_value`, and return self._uuid_from_int(int_value)`.

tests/types/test_uuid.py
```diff
@@ -32,7 +41,19 @@ def test_uuid_load(conn, fmt_out):
...
 
-    assert res == UUID(val)
+    uuid_val = UUID(val)
+    assert res == uuid_val
+    # the C modules bypasses __init__, so checking the state of the UUID object
+    assert res.hex == uuid_val.hex
+    assert res.int == uuid_val.int
+    assert res.bytes == uuid_val.bytes
+    assert res.is_safe == uuid_val.is_safe
+    assert getattr(UUID, "__slots__") == (
```
dvarrazzo, line 51: What is wrong with `UUID.__slots__`? :slightly_smiling_face:

henadzit: mypy complaints about it

Screenshot 2025-02-08 at 08 41 22

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: That's about the Cython version, not the Python version. `{impl: c, python: "3.9", ext: min, postgres: "postgres:15"}` uses it, see https://github.com/psycopg/psycopg/blob/master/tests/constraints.txt#L34.

I probably need to mention it in the comment if it isn't clear.

psycopg_c/psycopg_c/types/uuid.pyx
henadzit: I tested this but it wasn't any faster than just using `int.from_bytes`, so I opted for a the option that is less code. The two 64 bit numbers approach wasn't faster in the case of `UUIDLoader` either but it seemed less error prone that what I initially proposed, I mentioned it [here](https://github.com/psycopg/psycopg/pull/998#discussion_r1938580621). Let me know if you think we need to implement it anyway.

tests/types/test_uuid.py
```diff
@@ -19,9 +19,18 @@ def test_uuid_dump(conn, fmt_in):
...
 @pytest.mark.parametrize("fmt_out", pq.Format)
-def test_uuid_load(conn, fmt_out):
+@pytest.mark.parametrize(
+    "val",
+    [
+        "12345678123456781234567812345679",
+        "12345678-1234-5678-1234-567812345679",
+        "0123456789abcdef0123456789abcdef",
+        "01234567-89ab-cdef-0123-456789abcdef",
+        "{a0eebc99-9c0b4ef8-bb6d6bb9-bd380a11}",
```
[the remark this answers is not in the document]
henadzit: > These are array markers and should be parsed by the array loader before passing the single elements to the uuid loader,

Hm, interesting! where does it happen? I just tried to simulate what the UUID class is doing [here](https://github.com/python/cpython/blob/80b9e79d84e835ecdb5a15c9ba73e44803ca9d32/Lib/uuid.py#L200). Also [Postgres](https://www.postgresql.org/docs/current/datatype-uuid.html) supports the `{a0eebc99-9c0b4ef8-bb6d6bb9-bd380a11}` format for UUID.

henadzit:
There is a intermittent test on MacOS, it looks like an actual issue that happens when `Faker` hits the right configuration. I'll look into it!

tests/types/test_uuid.py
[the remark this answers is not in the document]
dvarrazzo: I see, it must be some standard but uncommon uuid representation. As Postgres docs say, output is always in the standard form, so we don't expect to receive them in the loader input.

However, the algorithm is pretty fine this way: I'm ok to treat any non-hex-digit as noise and discard it, I don't see the need to look for malformed input except if the size is not right.

psycopg_c/psycopg_c/types/uuid.pyx
dvarrazzo: Cool, thank you for measuring it!

psycopg_c/psycopg_c/types/uuid.pyx
dvarrazzo: Wasn't the conversation you linked about the text loader? It seems strange that 16 bytes of memcopy are faster than all the Python bureaucracy necessary to call and to execute [that method](https://github.com/python/cpython/blob/80b9e79d84e835ecdb5a15c9ba73e44803ca9d32/Objects/clinic/longobject.c.h#L397).

So I think this is more an error in the measurement and I would rather use the memcpy implementation of this method :slightly_smiling_face:

dvarrazzo:
> There is a intermittent test on MacOS, it looks like an actual issue that happens when `Faker` hits the right configuration. I'll look into it!

FYI you can repeat the randum numbers generated using `--randomly-seed=NNN` (the value is printed at the beginning of a test run), or `--randomly-seed=last` to repeat the last one used.

tests/types/test_uuid.py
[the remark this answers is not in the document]
dvarrazzo: > where does it happen?

In the `ArrayLoader`, which parses the `{...}` array format and loads the elements according to the loader configured.

https://github.com/psycopg/psycopg/blob/be4e811369e446188e3ffd6c000dd32eec594023/psycopg/psycopg/types/array.py#L384-L389

Receuvubg an uuid in square brackets wouldn't be ambiguous, because we get the oid of the array of uuids:

```
piro=# select array[gen_random_uuid()];
                 array                  
----------------------------------------
 {705fcd0b-2e65-49c0-958a-0927a4afeb58}

piro=# select pg_typeof( array[gen_random_uuid()]);
 pg_typeof 
-----------
 uuid[]
```

```python
>>> conn.execute("select array[gen_random_uuid()]").fetchone()[0]
[UUID('c6a65bbb-1763-41c0-85c3-dc566954b303')]
```

Postgres never emits brackets in the uuid output, as per docs, but even if it did, it wouldn't be problematic, as arrays use double quotes to escape probematic items:
```
piro=# select array['{fake-uuid}'::text];
      array      
-----------------
 {"{fake-uuid}"}
```

Trying hard to pass brackets to the uuid loader...

```python
cur = conn.cursor()
cur.adapters.register_loader("text", psycopg.types.uuid.UUIDLoader)
cur.execute("select array['{705fcd0b-2e65-49c0-958a-0927a4afeb58}'::text]").fetchone()
>>> [UUID('705fcd0b-2e65-49c0-958a-0927a4afeb58')]
```

dvarrazzo:
I have made another test in [this commit](https://github.com/psycopg/psycopg/commit/edd892dbeffc20b373c9524e60961b1e22a8d46b): I have introduced a writable version of UUID, so that we can set the attributes with a function faster than calling object.__setattr__ explicitly, and then changed the class of the object, which sounds illegal, but it's actually possible under certain conditions (memory compatibility).

This has shaved some performance time. I run some micro-benchmark using:

```
python -m timeit -s '
import psycopg
fmt = psycopg.pq.Format.TEXT  # or BINARY
uuid_oid = psycopg.adapters.types["uuid"].oid
loader = psycopg.adapters.get_loader(uuid_oid, fmt)(uuid_oid)
data = [b"12345678-1234-5678-1234-567812345678", b"\x12\x34\x56\x78\x12\x34\x56\x78\x12\x34\x56\x78\x12\x34\x56\x78"][fmt]
' 'loader.load(data)'
```

results are, TEXT/BINARY:

- 770/620 ns/loop: Python code (can be obtained by setting `PSYCOPG_IMPL=python`)
- 210/196 ns/loop: your branch at 99ad7c0d5392113a8989e64986e7fe52e932f31c
- 170/150 ns/loop: writable uuid at edd892dbeffc20b373c9524e60961b1e22a8d46b

Can you confirm this speedup?

This commit also includes some of the changes mentioned in the review: dropping the `uuid` module-level var, refactoring the code in a single base class.

henadzit:
> Can you confirm this speedup?

The `UUIDBinaryLoader` benchmark showing 37% improvement and the `UUIDLoader` benchmark shows 28% improvement comparing to the previous commit which is amazing! [The results](https://codspeed.io/henadzit/psycopg/runs/compare/67a9ea960c95e1369ec25ad5..67a9fd01ed35717c51fe8e27) with your commit applied. If you don't mind, I'll apply your commit.

I updated the benchmarks to run 100 iterations of `load` instead of a single iteration. It greatly improved the consistency. Previously I've seen +-10% fluctuations, now it's consistent results with each rerun.

[The overall improvement at the moment](https://codspeed.io/henadzit/psycopg/branches/cython-uuid-benchmarks-2):

Screenshot 2025-02-10 at 14 41 01

I'll look into MacOS test failures and other comments in coming days, I was traveling and was quite busy in general, hence a delay.

dvarrazzo:
>  If you don't mind, I'll apply your commit.

Sure, sounds great.

> I'll look into MacOS test failures and other comments in coming days, I was traveling and was quite busy in general, hence a delay.

No worries, and thank you very much for your contribution so far!

Let me know when you can about the macOS instability, otherwise I think we are in a good shape here.

tests/types/test_uuid.py
henadzit: Looks like `__slots__` aren't really present much in typeshed, here's [an open ticket](https://github.com/python/typeshed/issues/8832). Will update with a type ignore.

tests/types/test_uuid.py
```diff
@@ -32,7 +41,19 @@ def test_uuid_load(conn, fmt_out):
...
+    assert res.hex == uuid_val.hex
+    assert res.int == uuid_val.int
+    assert res.bytes == uuid_val.bytes
+    assert res.is_safe == uuid_val.is_safe
+    assert getattr(UUID, "__slots__") == (
+        "int",
+        "is_safe",
+        "__weakref__",
+    ), "UUID structure changed"
+    assert getattr(uuid_val, "__slots__") == getattr(UUID, "__slots__")
```
dvarrazzo, line 56: Thank you for looking into the typeshed issue. I propose the following change.

```suggestion
    #  https://github.com/python/typeshed/issues/8832
    slots = ("int", "is_safe", "__weakref__")
    assert (
        UUID.__slots__ == slots  # type: ignore[attr-defined]
    ), "UUID structure changed"
```

I don't think we need to check the equality on the instances.

henadzit:
@dvarrazzo [this commit ](https://github.com/psycopg/psycopg/pull/998/commits/9b71a188f0aac1bf0dd1f14f66c31f1150995a3c) introduced some test failures because high and low parts didn't take into account the endianness. I made a fix, let me know what you think!

I was able to pinpoint intermittent test failures to the Cython version of `UUIDDumper`. Basically it missed `.encode()`. With `.encode()` added, it was slower than the Python version. I decided to remove both `UUIDDumper` and `UUIDBinaryDumper` from this PR because they weren't really improving anything. I'm thinking about having another try with dumpers later in another PR.

psycopg_c/psycopg_c/types/uuid.pyx
dvarrazzo: Sorry, it was me to misread that. We can bump to a later version of Cython, being on 3.0 it's not a major requirement. But we can leave it out of this MR.

psycopg_c/psycopg_c/types/uuid.pyx
```diff
@@ -61,25 +58,37 @@ static const int8_t hex_to_int_map[] = {
...
+
+        self._object_new = object.__new__
+        self._uuid_type = _uuid.UUID
+        self._wuuid_type = <PyObject *>_uuid._WritableUUID
+        self._safeuuid_unknown = _uuid.SafeUUID_unknown
+
+    cdef object _return_uuid(self, uint64_t low, uint64_t high):
+        cdef object py_low = PyLong_FromUnsignedLongLong(low)
+        cdef object py_high = PyLong_FromUnsignedLongLong(high)
+        cdef object py_value = (py_low << 64) | py_high
```
dvarrazzo, line 57: Uhm, this seems an error...

dvarrazzo:
Uhm, something seems strange: your previous code didn't take endianness into account either, and it seems it's right:
```
        cdef object int_value = (PyLong_FromUnsignedLongLong(high) << 64) | PyLong_FromUnsignedLongLong(low)
```
mine doesn't either... but it looks like I wrote it backwards :flushed: 
```
        cdef object py_value = (py_low << 64) | py_high
```
I suspect that I wrote it that way to test if the tests would have detected the error and I committed it that way by mistake... 

Fixed in dfd182cb, running tests [here](https://github.com/psycopg/psycopg/actions/runs/13264496822).

dvarrazzo:
(you know, when you introduce a bunch of changes and tests pass at the first try, you suspect that the tests are testing the wrong thing...)

henadzit:
> By the look of it, that was the right change. We don't need to sniff for endianness to tell high from low half, because, whatever is the endian, x << 64 means "multiply x by 2 ^ 64", whether this really shifts bits left or right. So it's always the high half that needs to be shifted. The endianness is taken care by the be64toh() function (big-endian to host-endian, which will be a mirror or a no-op according to the platform).

Oh, you're right, thanks for explaining! I removed the endianness check. Looks like all the tests pass now except the crdb test.

dvarrazzo:
I'll disable the crdb test in master, it's probably using a too unstable image, and it's unrelated to this :slightly_smiling_face:

dvarrazzo:
Hello @henadzit

I have rebased this branch on master, cleaned up the history a bit by collapsing small fixes, added a news entry.

I am satisfied with the speedup of the loaders, and really grateful for this collaboration. If the dumpers speedup is marginal I am ok to leave it out.

If you are ok we can merge this branch at this state. Do you have a final figure we can put about the loaders speedup, to add it to [the release notes](https://github.com/psycopg/psycopg/pull/998/files#diff-87f7a7bb7fe8dbed76e16b67b87f27bda7b779b56e7d93ed3350d77ee4a3f217R22)?

henadzit:
> Hello @henadzit
> 
> I have rebased this branch on master, cleaned up the history a bit by collapsing small fixes, added a news entry.
> 
> I am satisfied with the speedup of the loaders, and really grateful for this collaboration. If the dumpers speedup is marginal I am ok to leave it out.
> 
> If you are ok we can merge this branch at this state. Do you have a final figure we can put about the loaders speedup, to add it to [the release notes](https://github.com/psycopg/psycopg/pull/998/files#diff-87f7a7bb7fe8dbed76e16b67b87f27bda7b779b56e7d93ed3350d77ee4a3f217R22)?

@dvarrazzo thank you so much for being such a great reviewer and collaborator, I learnt a lot about psycopg and Cython through this experience! I'll try to contribute again, hopefully, with less hiccups! 

I updated the PR description with the latest benchmark results!

dvarrazzo:
> I updated the PR description with the latest benchmark results!

Thank you! So looks like we can declare a 3x speed improvement in UUID loading :money_mouth_face: News entry updated accordingly.

dvarrazzo:
@henadzit It can be merged. I am sorry, I am busy with other projects, but I will try to merge it tomorrow or, at most, within this week.

PR #1006, opened 2025-02-12
Use isort to standardise imports

dvarrazzo:
Use isort.

The configuration used is similar to the one informally used so far, with the imports in order of length. Just because it pleases my brain.

Ideally I would have liked the import by length, but the items in a `from module import ...` in alphabetical order, but I haven't seen an obvious way to configure it so, and I don't think it's possible to use a plugin to make this distinction, because ISTM that a plugin ordering function is used both for the module names and for the object without the possibility to tell them apart.

isort is also integrated the in pre-commit hook and in async-to-sync, and it seems to play well with all the other tools (black, flake8...).

Note that if you have used `async_to_sync {--docker|--podman}` you will likely need to rebuild the image.

The first commit in this MR is the configuration, the second is the isort run on the whole codebase.


dlax (approved):
Thank you for taking care of this! That'd be surely be useful.

dvarrazzo:
Ready to merge.

dlax (approved):

tools/async_to_sync.py
```diff
@@ -12,14 +12,14 @@
...
-from concurrent.futures import ProcessPoolExecutor
 import sys
 import logging
 import subprocess as sp
 from copy import deepcopy
 from typing import Any, Literal
 from pathlib import Path
 from argparse import ArgumentParser, Namespace, RawDescriptionHelpFormatter
+from concurrent.futures import ProcessPoolExecutor
 from importlib.metadata import version
```
stankudrow, line 23: Hm, why out of order?

PR #1011, opened 2025-02-22
Rename tools/build library "ci"

dvarrazzo:
The name `build` is problematic because we typically want to avoid tools to go exploring them (mypy, isort, black...) but this leaves this directory unexplored.


dvarrazzo:
Will merge after #1012, because a contributor is making changes to the `build` directory.

PR #1013, opened 2025-02-22
fix: win32 libpq missing error message should be more clear

trim21:
If we can't find `libpq.dll` on windows, `libname` will be `None`, and psycopg will raise a very unreadable error about `None` missing `__fspath__` instead of expected `ImportError("libpq library not found")`

see here, for example: https://github.com/psycopg/psycopg/actions/runs/13475336675/job/37653819602?pr=1012


dvarrazzo:
Thank you, cherry-picked.

PR #1023, opened 2025-03-13
Fix how `status` is inferenced in `connection_summary`

sobolevn:
Hi! I am working on mypy in https://github.com/python/mypy/pull/18797

And I've noticed a small thing in your code. Ideally, `SomeEnum.Item.name` should be inferenced as `Literal[NAME_ONE, NAME_TWO, ...]`

But, since you later modify `status` (which is the name of a enum item), it should be explicitly annotated as `str`, not `Literal`.


dvarrazzo:
Thank you very much! I have reworded the comment a bit and will merge your proposa.

PR #1024, opened 2025-03-13
Add support for server_cursor_factory to Connection.connect.

apollo13:
I cannot run tests locally so I am using the CI, please excuse -- the changes are most likely to work :D

The error I get is:
```
  File "/home/florian/.local/share/pipx/venvs/pytest/lib64/python3.13/site-packages/_pytest/config/__init__.py", line 858, in import_plugin
    __import__(importspec)
    ~~~~~~~~~~^^^^^^^^^^^^
  File "/home/florian/.local/share/pipx/venvs/pytest/lib64/python3.13/site-packages/_pytest/assertion/rewrite.py", line 184, in exec_module
    exec(co, module.__dict__)
    ~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/home/florian/sources/psycopg/tests/fix_db.py", line 13, in <module>
    from psycopg import pq, sql
ImportError: cannot import name 'pq' from 'psycopg' (unknown location)
```

No idea what is happening there :D Works fine in a shell but not with pytest, sigh.

### DIFF
--- a/docs/api/connections.rst
+++ b/docs/api/connections.rst
@@ -42,6 +42,8 @@ The `!Connection` class
             :ref:`row-factories` for details.
         :param cursor_factory: Initial value for the `cursor_factory` attribute
             of the connection (new in Psycopg 3.1).
+        :param server_cursor_factory: Initial value for the `server_cursor_factory`
+            attribute of the connection (new in Psycopg 3.3).
         :param prepare_threshold: Initial value for the `prepare_threshold`
             attribute of the connection (new in Psycopg 3.1).
 
@@ -67,6 +69,9 @@ The `!Connection` class
 
         .. versionchanged:: 3.1
             added `!prepare_threshold` and `!cursor_factory` parameters.
+
+        .. versionchanged:: 3.3
+            added `!server_cursor_factory` parameter.
 
     .. automethod:: close
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -14,7 +14,7 @@ Python 3.3.0 (unreleased)
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Drop support for Python 3.8.
-
+- Add ``server_cursor_factory`` parameter to `Connection.connect`.
 
 Current release
 ---------------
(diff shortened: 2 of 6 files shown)


dvarrazzo:
To be super, super honest, I actually think that specifying named cursors using `conn.cursor(name)` was a mistake, it was a design choices coming from psycopg 2 and I didn't think hard enough about rethinking it.

It is much easier to create cursors by instantiating and passing the connection: `WhateverCursor(conn)`, and `WhateverServerCursor(conn, name)` work, and I think we should encourage this syntax.

So I'm kinda wary about adding more configuration support to the named cursor factores...

Would just using the cursor constructor work for your use case?

apollo13:
> Would just using the cursor constructor work for your use case?

Guess so, will have to shuffle a few things around in https://github.com/django/django/pull/19269/files but generally it should be doable. Just have to find the best location for it so it doesn't end up ugly (since we currently allow the users to override `cursor_factory`.

dvarrazzo:
TBH using `.cursor()` internally has created problems here and there exactly because it might have been customized, and we ended up replacing it with `Cursor(conn)`. See d68ae0d78eeb5e8b01ba1fb1d9ca9037520eea42 for example. Or even the historical https://github.com/psycopg/psycopg2/commit/21d323d2c8b0f0e606ce8dd46d9a6278770014b3

apollo13:
Fair enough, I just wanted to provide it as datapoint since you asked whether the cursor constructor would work for me. 

Personally I am leaning slightly towards the factory methods but I can understand your reservations about them. I'll further play around and see what I can come up with. My Django PR is only a cleanup after all, further complicated by the fact that we support psycopg2 as well (though we might wanna think about dropping it)

dvarrazzo:
I think that we should have been more aggressive in departing from psycopg2 behaviour in this area and we shouldn't have introduced `Conn.cursor(name=)` to create server-side cursors, but rather let people use the `ServerCursor` constructor.

We will keep the support for this `cursor()` method overload, but I don't wish to have more people rely on them, therefore I prefer not merging this feature which would allow to customize the class returned.

Thank you very much for the proposal though :slightly_smiling_face:

PR #1032, opened 2025-03-21
Fix ci trigger 

trim21:


dvarrazzo:
Thank you very much for the correction!

PR #1033, opened 2025-03-23
Shapely SRID support

### LINKED ISSUE #1028: register_shapely ignores SRID causing SRID missmatch errors
register_shapely ignores SRID when dumping geometries.

I was able to resolve the issue by updating the dumper logic as follows:

```py
class BaseGeometryBinaryDumper(Dumper):
    format = Format.BINARY

    def dump(self, obj: BaseGeometry) -> Buffer:
        return to_wkb(obj, include_srid=True)


class BaseGeometryDumper(Dumper):
    def dump(self, obj: BaseGeometry) -> Buffer:
        return to_wkb(obj, True, include_srid=True).encode()
```

For geometries without SRID, the output remains unchanged. I have switched to using to_wkb (which is [a wrapper around](https://github.com/shapely/shapely/blob/3dbaa2a856249b09434750f8cade9a9aaf40b301/shapely/wkb.py#L65) dumps) for better performance. The `Buffer | None` type hint was also incorrect because this overridden method never returns None.

While working on this, I made some small optimizations to the loader:

```py
class GeometryBinaryLoader(Loader):
    format = Format.BINARY

    def load(self, data: Buffer) -> BaseGeometry:
        return from_wkb(bytes(data))


class GeometryLoader(Loader):
    def load(self, data: Buffer) -> BaseGeometry:
        # it's a hex string in binary
        return from_wkb(bytes(data))
```

I replaced loads usage with from_wkb (which is [a wrapper around](https://github.com/shapely/shapely/blob/3dbaa2a856249b09434750f8cade9a9aaf40b301/shapely/wkb.py#L20) loads). In the binary case, it avoids an isinstance call + conditional check, as bytes(bytes) is [very efficient and avoids copying](https://github.com/python/cpython/blob/8ad4646c675211dc1df0254a3160f50a18e4c6c3/Objects/object.c#L870) in the CPython implementation. In the text case, it avoids bytes copy + conditional check, and simply reuses the binary implementation: from my testing, from_wkb seems to be very flexible, accepting bytes, hex strings, and hex bytes:

Image

> [!NOTE]
> If you're planning to use any part or all of the suggested code, please add me to the commit's Co-authored-by list. I enjoy collecting contributions to open source projects 🙂.

Zaczero:
Closes #1028


Zaczero:
Idk how to go best about the mypy issues here. Seems like the project is missing type hints for shapely. Perhaps install typeshed stubs https://github.com/python/typeshed/tree/main/stubs/shapely/shapely? Also, forcing all imports to use noqa: E402 sounds like a misconfiguration issue.

dvarrazzo:
> Also, forcing all imports to use noqa: E402 sounds like a misconfiguration issue.

If there was a way to tag `noqa` the `importofskip()` I would do it, but I haven't found any :\

dvarrazzo:
> Seems like the project is missing type hints for shapely.

I pushed a commit to this branch adding `types-shapely` to the dev requirements.

Zaczero:
> > Also, forcing all imports to use noqa: E402 sounds like a misconfiguration issue.
> 
> If there was a way to tag `noqa` the `importofskip()` I would do it, but I haven't found any :\

Maybe it would be possible to just disable this inspection in all tests?

dvarrazzo:
> Maybe it would be possible to just disable this inspection in all tests?

No, but the test can be skipped for a selection of files. I have skipped it for the shapely test file. Pushed a change to this branch.

Other two test files use importorskip, but because they don't have further import configuration is not necessary.

Zaczero:
1. Tested on Python 3.9.21, shapely 2.0.7 (same as the CI):

```py
from shapely import to_wkb, from_wkb, Point, get_srid, set_srid
get_srid(from_wkb(to_wkb(set_srid(Point(0,0), 4326), include_srid=True))
# np.int32(4326)
```

image

2. Added compatibility imports. **dumps** uses **WKBWriter** which [support include_srid](https://github.com/shapely/shapely/blob/maint-1.8/shapely/geos.py#L517-L524) kwarg. Checking the [1.8.5 docs](https://shapely.readthedocs.io/en/1.8.5/project.html), I couldn't find a way to get/set SRID via shapely, so the tests simply ignore missing imports (not ideal). Generally, shapely before 2.0 had very minimal support for SRID, so it's highly unlikely there will be any issues with that.

Zaczero:
> Maybe handling the srid depends on some library that doesn't get installed in the test environment?

It was a mistake on my part. set_srid returns a new modified geometry and doesn't modify it inplace. So the test was just broken.

dvarrazzo:
I have made a few changes to this branch:

- added shapely explicit version check rather than relying on import failure - see 7b9bce89. In this commit I have also renamed the objects by dropping the `_compat` suffix - was there a specific reason for it?
- rebased on master
- squashed oops commits
- adjusted commit messages to the project style
- added news entry

The current state looks ok for me. Ok to merge?

dvarrazzo:
(I'm a bit confused about how in shapely < 2 `dumps()` supports `include_srid=True` although as no-op...)

Zaczero:
> renamed the objects by dropping the _compat suffix - was there a specific reason for it?

Personal preference. I like to clearly indicate methods that may not exactly be what they look like - just to be careful.

> I'm a bit confused about how in shapely < 2 dumps() supports include_srid=True

My guess is that since shapely is generally a wrapper around GEOS, some other compatible libraries may use the SRID, so they want to handle such cases gracefully. Even if they don't expose operations on SRID explicitly (with the single exception being the dumps method).

dvarrazzo:
I have massaged the tests to make better use of the fixtures and running tests with Python 3.12 in addition to Python 3.9 (which runs in the `min` job in CI). After the above tests have finished we can merge this branch.

PR #1035, opened 2025-03-27
refactor: use the walrus operator with generosity

dvarrazzo:
This refactoring was generated mostly automatically using [this walrusify script](https://gist.github.com/dvarrazzo/a0e9bd4bee4bb64b543332721de294c0) (not added to the repository because it's meant to be used only once).

The result of the script was rinsed through our linters/checkers and committed chunk by chunk to avoid unneeded changes (caused by the fact that the script operates on the AST) and cases in which the walrus operator created less readable code.


psycopg/psycopg/_cursor_base.py
```diff
@@ -467,15 +466,14 @@ def _check_results(self, results: list[PGresult]) -> None:
 
         for res in results:
             status = res.status
-            if status != TUPLES_OK and status != COMMAND_OK and status != EMPTY_QUERY:
+            if status != TUPLES_OK and status != COMMAND_OK and (status != EMPTY_QUERY):
```
dlax, line 469: Unnecessary parentheses added?

dlax (commented):
Only (quickly) read the application code; LGTM.

psycopg/psycopg/_cursor_base.py
dvarrazzo: Indded, thank you for spotting it. They were added by the AST transformation and I committed them by mistake.

dlax (approved):
I have to admit I didn't read all changes in tests :P

pauloxnet:
I'm sincerely curious: other than code shortness, you observed some speed improvements or other benefits from this bulk changes?

PR #1036, opened 2025-03-27
Rename some news.rst section headers from python to psycopg

LuiMoiPer:
I seemed like these section headers were supposed to be psycopg not python in the webpage about releases.


dvarrazzo:
Thank you very much! Your markup was broken but the titles underline too short. Fixed it, will merge.

PR #1053, opened 2025-04-30
Add support for Python 3.14

dvarrazzo:
Work in progress, until the release gets to a more stable state.

### DIFF
--- a/.github/workflows/packages-bin.yml
+++ b/.github/workflows/packages-bin.yml
@@ -50,7 +50,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x86_64, ppc64le, aarch64]
-        pyver: [cp310, cp311, cp312, cp313]
+        pyver: [cp310, cp311, cp312, cp313, cp314-dev]
         platform: [manylinux, musllinux]
 
     steps:
@@ -130,7 +130,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x86_64, arm64]
-        pyver: [cp310, cp311, cp312, cp313]
+        pyver: [cp310, cp311, cp312, cp313, cp314-dev]
 
     steps:
       - name: Checkout repos
@@ -183,7 +183,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [win_amd64]
-        pyver: [cp310, cp311, cp312, cp313]
+        pyver: [cp310, cp311, cp312, cp313, cp314-dev]
 
     defaults:
       run:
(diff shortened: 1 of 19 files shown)


ulgens:
@dvarrazzo Any idea how to fix failing CI? I'd like to help if there is anything I can help with.

edgarrmondragon:
> @dvarrazzo Any idea how to fix failing CI? I'd like to help if there is anything I can help with.

Looking at the error I think it's a cibuildwheel misconfig. Fix might be to use 3.1, which builds 3.14 wheels by default.

.github/workflows/packages-bin.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x86_64, ppc64le, aarch64]
-        pyver: [cp310, cp311, cp312, cp313]
+        pyver: [cp310, cp311, cp312, cp313, cp31]
```
edgarrmondragon:
```suggestion
        pyver: [cp310, cp311, cp312, cp313, cp314]
```

.github/workflows/packages-bin.yml
```diff
@@ -130,7 +130,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x86_64, arm64]
-        pyver: [cp310, cp311, cp312, cp313]
+        pyver: [cp310, cp311, cp312, cp313, cp31]
```
edgarrmondragon:
```suggestion
        pyver: [cp310, cp311, cp312, cp313, cp314]
```

.github/workflows/packages-bin.yml
dvarrazzo: Yup, lready fixed, thank you. Now there seems to be a regression on delocate on musllinux...

edgarrmondragon: https://github.com/psycopg/psycopg/pull/1144 might fix the musllinux builds:

https://github.com/edgarrmondragon/psycopg/actions/runs/17304961288/job/49125377448

edgarrmondragon:
The only build failure now is on windows, caused by a deprecation warning for `WindowsSelectorEventLoopPolicy`. It's gonna be removed in 3.16 but I don't think there's a good alternative at the moment.

dvarrazzo:
Working on it. The alternative is to use asyncio.run(..., loop_factory=...), which is available from Python 3.12.

dvarrazzo:
Last hurdle: python-proxy is not compatible with Python 3.14

https://github.com/qwj/python-proxy/issues/200

If there's another library we can use (this one seems unmaintained) I'm happy to switch.

dvarrazzo:
Finally made to work for master. Backporting the branch to 3.2 for testing.

PR #1061, opened 2025-05-01
Fix return value of `DateFromTicks` and `TimeFromTicks`

### LINKED ISSUE #1058: test_date_from_ticks fails
Running pytest I get:

```python
FAILED tests/test_psycopg_dbapi20.py::test_date_from_ticks[0-1970-01-01] - assert datetime.date(1969, 12, 31) in [datetime.date(1970, 1, 1)]
```

Tested with:

- psycopg 3.2.7
- postgresql 17.4
- python 3.13.3
- macOS 15.4
- Apple M4

This issue was also reported here for riscv64: https://github.com/psycopg/psycopg/issues/883

I think the problem is in `TimestampFromTicks(ticks)` in dbapi20.py. It uses `time.localtime(0)` to find the epoch start, _with offset to local timezone_. This is turned into a timezone-aware datetime object.

I'm in timezone UTC-7 so `TimestampFromTicks(0)` returns:

```python
datetime.datetime(1969, 12, 31, 17, 0, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=61200)))
```

`DateFromTicks(0)` calls `.date()` on that object, which ignores the timezone and returns `datetime.date(1969, 12, 31)`, which is incorrect.

`TimeFromTicks(0)` returns `datetime.time(17, 0)` which is also incorrect. However, in `test_time_from_ticks`, the hour is forced to 0 so the test passes.

According to the [Python documentation](https://docs.python.org/3/library/time.html) (referred to from [PEP-249](https://peps.python.org/pep-0249/#timestampfromticks)):

> The epoch is the point where the time starts, the return value of time.gmtime(0). It is January 1, 1970, 00:00:00 (UTC) on all platforms.

I think the solution is to use `time.gmtime(0)` in `TimestampFromTicks` and set the timezone to UTC.

For reference, SQLAlchemy sets epoch like this:

```python
epoch = dt.datetime.fromtimestamp(0, dt.timezone.utc).replace(tzinfo=None)
```

dvarrazzo:
These function are based on the return value of `TimestampFromTicks`. Because this value returned by this function is timezone-aware and localized to the current timezone, truncating the value (and implicitly losing the timezone) might return the wrong result.

For example, at tick 0, in NY, it was 1969-12-31T19:00-05:00. By truncating the timestamp on the date, the date would be the 1969-12-31 instead of the expected 1970-01-01.

This changeset actually changes `TimestampFromTicks` to return an UTC timestamp instead of a local timestamp. We believe that this change is not a problem because we don't change neither the type (tz-aware datetime) nor the value of the returned object (old and new values compare equal). An alternative would have been to decouple the Timestamp function from the Date/Time functions, but I think it would be confusing if the Timestamp function returned a datetime with the date in 1969 but the Date function returned a date in 1970.

Adjust the tests to be *stricter* than before, making use of the fact that now the output is not affected by the local timezone.

Close #1058.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -15,6 +15,14 @@ Psycopg 3.3.0 (unreleased)
 
 - Drop support for Python 3.8.
 
+Psycopg 3.2.8 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix `DateFromTicks` and `TimeFromTicks` return values to return date and
+  time referred as UTC rather than the local timezone. Change
+  `TimestampFromTicks` to return a datetime in UTC rather than in the local
+  timezone (:ticket:`#1058`).
+
 
 Psycopg 3.2.7
 ^^^^^^^^^^^^^
--- a/psycopg/psycopg/dbapi20.py
+++ b/psycopg/psycopg/dbapi20.py
@@ -6,15 +6,15 @@ Compatibility objects with DBAPI 2.0
 
 from __future__ import annotations
 
-import time
 import datetime as dt
-from math import floor
 from typing import Any
 from collections.abc import Sequence
 
 from . import _oids
 from .abc import AdaptContext, Buffer
 from .types.string import BytesBinaryDumper, BytesDumper
+
+EPOCH = dt.datetime(1970, 1, 1, tzinfo=dt.timezone.utc)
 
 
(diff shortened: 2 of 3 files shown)


dvarrazzo:
This MR was merged OK; the ticket state is wrong (because I merged from my laptop and pushed to master before pushing here).

PR #1062, opened 2025-05-02
Clarify docstring for BaseCursor.statusmessage

fedorkobak:
The original definition can be confusing (at least it was for me) because if you run several commands separated by `;` in a single `cursor.execute`, the `cursor.statusmessage` attribute retains the tag from the first command. For example:

```py
query = """
CREATE TABLE table1 (val INT, val2 INT);
INSERT INTO table1 (val, val2) VALUES (10, 20);
"""
cursor.execute(query=query)
print("Execute 1:", cursor.statusmessage)

cursor.execute(query="DROP TABLE table1;")
print("Execute 2:", cursor.statusmessage)
```

This will result in:

```
Execute 1: CREATE TABLE
Execute 2: DROP TABLE
```

But according to the original description, it should be:

```
Execute 1: INSERT
Execute 2: DROP TABLE
```

### DIFF
--- a/docs/api/cursors.rst
+++ b/docs/api/cursors.rst
@@ -263,10 +263,9 @@ The `!Cursor` class
     .. attribute:: pgresult
         :type: Optional[psycopg.pq.PGresult]
 
-        The result returned by the last query and currently exposed by the
-        cursor, if available, else `!None`.
+        Representation of the current result set, if available, else `!None`.
 
-        It can be used to obtain low level info about the last query result
+        It can be used to obtain low level info about the current result set
         and to access to features not currently wrapped by Psycopg.
 
 
--- a/docs/api/cursors.rst
+++ b/docs/api/cursors.rst
@@ -244,7 +244,7 @@ The `!Cursor` class
             for record in cursor:
                 ...
 
-        syntax will iterate on the records in the current recordset.
+        syntax will iterate on the records in the current result set.
 
     .. autoattribute:: row_factory
 
--- a/docs/api/cursors.rst
+++ b/docs/api/cursors.rst
@@ -232,7 +232,7 @@ The `!Cursor` class
 
     .. rubric:: Methods to retrieve results
 
-    Fetch methods are only available if the last operation produced results,
+    Fetch methods are only available if the current result set contains results,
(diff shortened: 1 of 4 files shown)


dvarrazzo:
You are right, the docstring was likely a copypaste from psycopg2, which only shows the last result.

However, psycopg 3 receives all the results and allows to iterate via `nextset()` the statusmessage reflects the current result in the list of results returned:

```python
>>> cursor.execute(query=query)
>>> cursor.statusmessage
'CREATE TABLE'

>>> cursor.nextset()
>>> cursor.statusmessage
'INSERT 0 1'
```

so this is the behaviour we should actually document.

Similarly affected is `rowcount`; I don't remember about other attribute affected. Maybe we want to make sure that the picture of nextset() - rowcount - statusmessage is clear from the docs. Would you like to take a look if there are inconsistencies?

Thank you very much!

fedorkobak:
Ok, I'll take a look.

fedorkobak:
@dvarrazzo
Two terms are used in the documentation. _Result set_ ([example](https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/cursor.py#L182)) and _recordset_ ([example](https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/_cursor_base.py#L152)). From my point, it looks that they are refer to the same - information about pg response that corresponds to the particular executed SQL command. Are they?

dvarrazzo:
As I was grepping the code and looking at messages, it occurred to me that the error resulting from calling `fetch*()` after a command not returning tuples could have been improved:

> the last operation didn't produce a result

The last operation did produce a result, just it didn't produce any row/record.

While at it, it also occurred me that sometimes users are confused by that error and can't quite tell what the last operation was. So I thought about adding the statusmessage to the error message.

The result is #1063. What do you think of it?

fedorkobak:
Looks good to me

fedorkobak:
I created new descriptions for the `rowcount`, `statusmessage` and `pgresult` attributes and replaced some terms as "result set" that I think are related.

dvarrazzo:
Thank you for the care, looks good!

PR #1063, opened 2025-05-02
Improve error message trying to fetch from a no recordset result

dvarrazzo:
Print the last command status with the error. For example:

    the last operation didn't produce records (command status: NOTIFY)

or, if the result doesn't even have a command status:

    the last operation didn't produce records (result status: EMPTY_QUERY)

Related to #1062.


dlax (approved):
Useful addition.

PR #1064, opened 2025-05-03
Add support for `next(cursor)`

dvarrazzo:
This MR transforms the cursor from an iterable (supporting `iter()`) to an iterator (supporting `iter()` and `next()`).

`next(cursor)` behaves like `cursor.fetchone()`, but its result is `Row`, not `Row | None`. That's because, if there is no row to fetch, it will raise `StopIteration`.

This makes simpler to write strictly typed code using the result when there is a 100% guarantee that a query will return a record (or will raise an exception), for example when fetching `count(*)`, which always returns a result even if the counted records are 0. The type checker is not aware of that, therefore you need to guard for None with an `if`, or an `assert`, or you need to do something horrible like `(rec,) = cur.fetchall()` in order to convince Mypy that `rec` will never be none.

### DIFF
--- a/docs/advanced/typing.rst
+++ b/docs/advanced/typing.rst
@@ -38,6 +38,8 @@ annotations such as `!Connection[Any]` and `!Cursor[Any]`.
 
    rec = cur.fetchone()     # type is tuple[Any, ...] | None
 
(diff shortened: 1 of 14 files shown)


dlax (approved):
Very useful addition!!

PR #1067, opened 2025-05-04
Allow integration of psycopg pool with SQLAlchemy

### LINKED ISSUE #1046: Enhance `ConnectionPool` to support use in SQLAlchemy
This is a follow-up to [SQLAlchemy discussion #12522](https://github.com/sqlalchemy/sqlalchemy/discussions/12522).

The `max_idle` connection-closing feature of `psycopg_pool.ConnectionPool` is really cool! Our project uses SQLAlchemy, so I would like to use Psycopg 3's pooling in SQLAlchemy. The testcase is described in more detail in the linked discussion above, but I will include the script itself here for completeness:

```python
#! /usr/bin/env python
import logging
import time

logging.basicConfig(level=logging.DEBUG, format="%(asctime)s %(levelname)s %(name)s: %(message)s")
logging.getLogger("psycopg.pool").setLevel(logging.DEBUG)
logging.getLogger("sqlalchemy.pool").setLevel(logging.DEBUG)

import psycopg_pool
import sqlalchemy

db_url = "postgresql+psycopg://postgres:postgres@hostname:port/database"  # CHANGEME

# Create a psycopg_pool connection pool
mypool = psycopg_pool.ConnectionPool(
    conninfo=db_url.replace("+psycopg", ""),
    min_size=0,
    max_size=1,
    max_idle=5,
)

# Create a SQLAlchemy engine that uses the psycopg_pool connection pool
engine = sqlalchemy.create_engine(
    url=db_url,
    poolclass=sqlalchemy.pool.NullPool,  # disable SQLAlchemy's default connection pool
    creator=mypool.getconn,              # use psycopg_pool to create connections
)

# Register a 'checkin' event listener to return connections to psycopg_pool
# (https://docs.sqlalchemy.org/en/20/core/events.html#sqlalchemy.events.PoolEvents.checkin)
def return_to_pool(dbapi_connection, connection_record):
    mypool.putconn(dbapi_connection)

sqlalchemy.event.listen(engine, "checkin", return_to_pool, named=True)

# Define a function that asks the server for the current time
def do_stuff():
    with engine.connect() as connection:
        result = connection.execute(sqlalchemy.text("SELECT now()"))
        print(f"Current time: {result.fetchone()[0]}")

# Perform two transactions, using the same pool-sourced server connection
print("Doing do_stuff() #1...")
do_stuff()
print("Doing do_stuff() #2...")
do_stuff()
print("Done doing stuff! Waiting 10 seconds...")

# Allow idle pool connection to disconnect from server
time.sleep(10)

# Perform another transaction, requiring a new pool connection to the server
print("Doing do_stuff() #3...")
do_stuff()
```

The issue with this combination is that SQLAlchemy's `NullPool` class issues a final `connection.close()` when it is done, which results in the connection actually being closed and thus no longer . Note the last line here:

```
2025-04-12 06:26:00,564 DEBUG sqlalchemy.pool.impl.NullPool: Closing connection <psycopg.Connection [IDLE] (...) at 0x7fffdd180250>
                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

According to one of the SQLAlchemy maintainers in the linked discussion, this combination requires `ConnectionPool` to handle `connection.close()` for pool-sourced connections by simply ensuring it is back in the pool instead of actually closing it.

Is this something that can be done, either by default or with an option?

Thanks in advance!

dvarrazzo:
Add `close_returns` pool parameter. If set, `conn.close()` will behave like `pool.putconn(conn)`. This is the behaviour expected by SQLAlchemy's NullPool to integrate an external pool: see https://github.com/sqlalchemy/sqlalchemy/discussions/12522

Close #1046

Note that the feature will be released with psycopg_pool 3.3, and will work with any psycopg version, but, with psycopg connections < 3.3, they must be the vanilla psycopg connections, not subclasses. For psycopg < 3.3, the user will need to subclass `close()` to integrate with SA, instead of setting `close_returns`.

Please look at the changes in the documentation for the details.


dvarrazzo:
@chrispy-snps @CaselIT this branch should be ready to merge. The documentation now has a section for SQLAlchemy integration, the url should be <https://www.psycopg.org/psycopg3/docs/advanced/pool.html#pool-sqlalchemy for a workaround>, which will be available as soon as this branch is merged to master and before the package is released.

A review is welcome, and if you wish you can adapt https://github.com/sqlalchemy/sqlalchemy/pull/12540 relying on the documentation here.

chrispy-snps:
@dvarrazzo - thank you, this is great!

Can I make a merge request to suggest some small grammatical fixes? (I was a technical writer in a former life.) If yes, I can base it on this branch, or on the main branch once this is merged.

docs/advanced/pool.rst
```diff
@@ -429,3 +429,70 @@ Metric                  Meaning
...
+                await super().close()
+
+
+Using a connection implementing this method you will not need to specify
+`!close_returns` in the pool constructor, but just `!connection_class`; the
+example above would be modified as:
+
+.. code:: python
+
+    pgpool = psycopg_pool.ConnectionPool(conninfo=db_url, connection_class=MyConnection)
```
CaselIT, line 498: this likely should use `AsyncConnectionPool`

what would likely require a reword also of `the example above would be modified as:` since the async pool would not work with a sync sqlalchemy engine.

Maybe it makes more sense to show the async example also for sqlalchemy since it's more involved (it requires awaiting for `open()`)?

CaselIT (approved):

docs/advanced/pool.rst
dvarrazzo: You are right, I have been inconsistent between sync and async in my example.

I have rewritten the sample code to use consistently async examples. I haven't actually tested it though. I'd like to be sure that the idioms to create the engine is correct and it is consistent with your documentation. I would add some pointer to SA docs for example to say "this example is async; the differences with a sync engine are documented at https://sqlalchemy.org/...".

docs/advanced/pool.rst
```diff
@@ -451,22 +451,35 @@ Since `!psycopg_pool` version 3.3, the `!close_returns` parameter of the pool
...
 
-    engine = sqlalchemy.create_engine(
-        url=db_url.replace("postgresql://", "postgresql+psycopg://"),
-        poolclass=sqlalchemy.pool.NullPool,
-        creator=pgpool.getconn,
-    )
+        # Create a SQLAlchemy engine using the pool just created
+        engine = sqlalchemy.ext.asyncio.create_async_engine(
+            url=db_url.replace("postgresql://", "postgresql+psycopg://"),
+            poolclass=sqlalchemy.pool.NullPool,
```
dvarrazzo, line 474: I copied and adapted this example from https://github.com/sqlalchemy/sqlalchemy/discussions/12522. Does `sqlalchemy.pool.NullPool` support both sync and async engine/pool?

CaselIT: Yes, null pool works with both versions

docs/advanced/pool.rst
CaselIT: I'll take a second look later

CaselIT (approved):

docs/advanced/pool.rst
```diff
@@ -451,22 +451,35 @@ Since `!psycopg_pool` version 3.3, the `!close_returns` parameter of the pool
...
+        await mypool.open()
 
-    engine = sqlalchemy.create_engine(
-        url=db_url.replace("postgresql://", "postgresql+psycopg://"),
-        poolclass=sqlalchemy.pool.NullPool,
-        creator=pgpool.getconn,
-    )
+        # Create a SQLAlchemy engine using the pool just created
+        engine = sqlalchemy.ext.asyncio.create_async_engine(
+            url=db_url.replace("postgresql://", "postgresql+psycopg://"),
```
CaselIT, line 473: since you pass a creator you can also just hardcode `url="postgresql+psycopg://"`

chrispy-snps:
@dvarrazzo - that was a kind way of saying "RTFM," thank you. :)  The instructions were easy to follow.

I tried the new functionality and it works great! I updated the examples [here](https://github.com/sqlalchemy/sqlalchemy/discussions/12522#discussioncomment-13024666) to reflect the new support.

zhcn000000:
On lines 512 to 516 of pool.py
```python
            if conn._expire_at <= monotonic():
                logger.info("discarding expired connection %s", conn)
                conn.close()
                self.run_task(AddConnection(self))
                continue
```
should be changed to 
```python
            if conn._expire_at <= monotonic():
                logger.info("discarding expired connection %s", conn)
                conn._pool = None
                conn.close()
                self.run_task(AddConnection(self))
                continue
```
Otherwise, when close_returns is opened, it will cause an infinite loop. The conn._pool property is not set to empty. self.putconn is triggered again, and the timeout is triggered and conn.close() is executed. At the same time, a new connection is added, causing the database to overload
The same applies to other places where conn.close() is executed without setting self._pool to None

dvarrazzo:
@zhcn000000 thank you for your observation, but this ticket is the wrong place to address it. I have opened #1124 for it. If you have anything else to contribute, or if you want to prepare a MR we can continue there.

dvarrazzo:
Hello peeps! Is there anything else to change to this branch? With the code or the docs? Thank you!

chrispy-snps:
@dvarrazzo - I must defer the technical assessment to others, but if you are interested, I could contribute a light grammatical edit of the .rst content.

AdityaSoni19031997:
Thanks for adding this feature! Thanks for filling in the gap and making lives simpler!
QQ: Are we planning to merge it to master anytime soon? 

Thanks!

dvarrazzo:
Time to merge this!

@chrispy-snps documentation review is welcome. I you want to contribute to it please make a new MR for the master branch.

Thank you very much everyone for your contributions!

PR #1074, opened 2025-05-07
fixed loading of rows when row maker returns None

### LINKED ISSUE #1073: Cursors do not load rows properly when row maker returns None
A `RowMaker` returning `None` can be usefull, for example:
```python
from dataclasses import dataclass
from psycopg import Connection
from psycopg.rows import args_row
from typing import Any


@dataclass(kw_only=True)
class User:
    id: int | None = None
    name: str


def create_or_ignore_user(connection: Connection[Any], user: User) -> bool:
    def on_user_created(id: int) -> None:
        user.id = id

    query = 'INSERT INTO "user" ("name") VALUES (%s) ON CONFLICT ("name") DO NOTHING RETURNING "id"'
    params = user.name,
    row_factory = args_row(on_user_created)

    @args_row
    with connection.cursor(row_factory=row_factory) as cursor:
        cursor.execute(query, params)
        cursor.fetchone()

        return cursor.rowcount == 1
```
Hence `Cursor` resp. `AsyncCursor` should load rows properly even when the `RowMaker` returns  `None`.
For example, I would expect, that all results in the following example are `[None, 0]`:
```python
with connection.cursor(name, row_factory=scalar_row) as cursor:
    query = 'VALUES (%s), (%s)'
    params = [None, 0]

    # cursor.fetchone
    cursor.execute(query, params)
    rows = [cursor.fetchone() for _ in range(len(params))]
    print(rows)  # [None, 0] if name else [None, None]

    # cursor.fetchmany
    cursor.execute(query, params)
    rows = cursor.fetchmany(len(params))
    print(rows)  # [None, 0]

    # cursor.fetchall
    cursor.execute(query, params)
    rows = cursor.fetchall()
    print(rows)  # [None, 0]

    # cursor.__iter__
    cursor.execute(query, params)
    rows = [row for row in cursor]
    print(rows)  # [None, 0] if name else []

    # cursor.stream
    stream = cursor.stream(query, params)
    rows = [row for row in stream]
    print(rows)  # [0]
```
This is not the case of the current implementation, please see outputs in comments.

embecka:
Closes #1073.


psycopg/psycopg/_py_transformer.py
```diff
@@ -313,12 +313,11 @@ def load_rows(self, row0: int, row1: int, make_row: RowMaker[Row]) -> list[Row]:
...
 
-    def load_row(self, row: int, make_row: RowMaker[Row]) -> Row | None:
-        if not (res := self._pgresult):
-            return None
+    def load_row(self, row: int, make_row: RowMaker[Row]) -> Row:
+        assert (res := self._pgresult)
 
         if not 0 <= row < self._ntuples:
-            return None
+            raise IndexError
```
dvarrazzo, line 320: A message could do here
```suggestion
            raise IndexError("record index out of range")
```

dvarrazzo (changes requested):
I appreciate the fix. However I would also like to see tests in the unit test that shows that `values (None), (0)` is indeed fetched well using the scalar row with the various methods available:

- repeated fetchone()
- fetchmany()
- fetchall()
- __iter__
- stream()

dvarrazzo:
> I would also like to see tests in the unit test that shows...

Tests should be performed on both client-side cursors and server-side cursors, because the latter have a different implementation. I understand by your description in #1073 that you have taken them in consideration already. Thank you very much!

psycopg/psycopg/cursor.py
```diff
@@ -187,9 +186,12 @@ def fetchone(self) -> Row | None:
         """
         self._fetch_pipeline()
         self._check_result_for_fetch()
-        if (record := self._tx.load_row(self._pos, self._make_row)) is not None:
+        assert self.pgresult
```
dvarrazzo, line 189: If we make `self._check_result_for_fetch` return `self.pgresult` we can get rid of these assets.

psycopg/psycopg/_py_transformer.py
embecka: I also forgot to change `transform.pyx`. I should unify error handling, I will do it later.

psycopg/psycopg/cursor.py
embecka: The only reason for that assertion is the linting tool. I need to use `self.pgresult.ntuples`, but `self.pgresult` can be `None`. Do you have a suggestion how to change that?

dvarrazzo: What I suggested: if `self.pgresult` is `None`, then `_check_result_for_fetch()` will raise `ProgrammingError` (the user has used `fetch*()` before `execute()`. So we can change `_check_result_for_fetch()` to reutrn `self.pgresult` on success:
```diff
--- a/psycopg/psycopg/_cursor_base.py
+++ b/psycopg/psycopg/_cursor_base.py
@@ -569,7 +569,7 @@ class BaseCursor(Generic[ConnectionType, Row]):
                 name, pgq.params, param_formats=pgq.formats, result_format=fmt
             )
 
-    def _check_result_for_fetch(self) -> None:
+    def _check_result_for_fetch(self) -> PGresult:
         if self.closed:
             raise e.InterfaceError("the cursor is closed")
 
@@ -577,7 +577,7 @@ class BaseCursor(Generic[ConnectionType, Row]):
             raise e.ProgrammingError("no result available")
 
         if (status := res.status) == TUPLES_OK:
-            return
+            return res
         elif status == FATAL_ERROR:
             raise e.error_from_result(res, encoding=self._encoding)
         elif status == PIPELINE_ABORTED:
```

and then perform all the operations we need on the result of this function rather than on `self.pgresult`, for example (in my branch):

```diff
--- a/psycopg/psycopg/cursor.py
+++ b/psycopg/psycopg/cursor.py
@@ -200,13 +200,12 @@ class Cursor(BaseCursor["Connection[Any]", Row]):
         :rtype: Sequence[Row], with Row defined by `row_factory`
         """
         self._fetch_pipeline()
-        self._check_result_for_fetch()
-        assert self.pgresult
+        res = self._check_result_for_fetch()
 
         if not size:
             size = self.arraysize
         records = self._tx.load_rows(
-            self._pos, min(self._pos + size, self.pgresult.ntuples), self._make_row
+            self._pos, min(self._pos + size, res.ntuples), self._make_row
         )
         self._pos += len(records)
         return records
```

psycopg/psycopg/_py_transformer.py
embecka: Now errors are raised in the same way as in `Transformer.load_rows`

psycopg/psycopg/cursor.py
embecka: Nice, I changed it in that way, there are no asserts now.

tests/test_cursor_common_async.py
```diff
@@ -900,3 +900,20 @@ async def test_error_no_result(aconn):
...
+    await cur.execute(query)
+    assert [await cur.fetchone() for _ in range(len(recs))] == recs
+    await cur.execute(query)
+    assert await cur.fetchmany(len(recs)) == recs
+    await cur.execute(query)
+    assert await cur.fetchall() == recs
+    await cur.execute(query)
+    assert [row async for row in cur] == recs
+    stream = cur.stream(query)
+    assert [row async for row in stream] == recs
```
dvarrazzo, line 919: We have an `alist` helper you can use here:
```suggestion
    assert await alist(cur) == recs
    assert await alist(cur.stream(query)) == recs
```
which will be converted to `list` by async_to_sync.

embecka: Done.

dvarrazzo (changes requested):
Pretty good, thank you very much :slightly_smiling_face:

Issue #1055, opened 2025-04-30
Psycopg stops execution on connection


thedomeffm:
psycopg stops the execution somehow. I am just helping out my co-worker with some sql so I have not the abilities to debug or deep dive into the problem.

Ref: https://github.com/psycopg/psycopg/issues/1041

**Example Code**
```python
def example():
    dsn = (
        f"host={os.getenv('POSTGRES_HOST')} "
        f"dbname={os.getenv('POSTGRES_DATABASE')} "
        f"user={os.getenv('POSTGRES_USER')} "
        f"password={os.getenv('POSTGRES_PASSWORD')} "
        f"port={os.getenv('POSTGRES_PORT')}"
    )
    print("1")

    with psycopg.connect(dsn) as sql_driver:
        print("2")
        with sql_driver.cursor(row_factory=dict_row) as cursor:
            print("3")
            cursor.execute("SELECT * FROM core_users").fetchone()
    print("4")
```

## With `psycopg 3.1.20`
```bash
1
2
3
4
```

## With `psycopg 3.2.6`
```bash
1
// **Execution STOPS HERE!**
// Than I cancel the process with Ctrl+C
2
3
4
```

- Appears on W11
- Tested on Python 3.9 and 3.13

dvarrazzo:
Hello! I can only debug this issue with some help from your part.

If I understand you report

- the connection attempts block
- you press ctrl-c
- the connection attempt restarts

it makes me think that some libpq call forks a process. Could you take a look in the task manager if you see anything being forked off the python process?

I have just released psycopg 3.2.7 where I have added some logging statements to help debugging this endless problem. Could you please run your test after enabling debug logging on the `psycopg` logger? AKA:

```
>>> import logging
>>> logging.basicConfig(level=logging.DEBUG, format='%(asctime)s %(levelname)s %(message)s')
>>> logging.getLogger("psycopg").setLevel(logging.DEBUG)
>>> import psycopg
>>> psycopg.connect("host=localhost")
2025-04-30 15:13:30,928 DEBUG connection started: <psycopg_c.pq.PGconn [STARTED] at 0x7b85f20f98a0>
2025-04-30 15:13:30,928 DEBUG connection polled: <psycopg_c.pq.PGconn [MADE] at 0x7b85f20f98a0>
2025-04-30 15:13:30,929 DEBUG connection polled: <psycopg_c.pq.PGconn [SSL_STARTUP] at 0x7b85f20f98a0>
[...]

PR #1076, opened 2025-05-07
fix: show all messages associated with connection failures

### LINKED ISSUE #1069: Connection errors suppressed by subsequent connection attempts
When connecting to a domain name which has multiple associated IP addresses, the current connection algorithm can produce pretty confusing situations where the relevant error message is suppressed by subsequent connection attempts.

This caused us a problem recently in https://github.com/aws-samples/aurora-dsql-samples/pull/113#discussion_r2072316690. An SSL certificate error was causing the initial IPv4 connection attempt to fail, but only the subsequent IPv6 connection failure was shown. In this case IPv6 routing was not expected to work on the host.

The exception we could see was this one:

```
Failed: Unexpected exception: connection is bad: connection to server at "2600:1f18:692c:303:31c4:3b43:ed6d:d04d", port 5432 failed: Network is unreachable
Is the server running on that host and accepting TCP/IP connections?
```

But the host running the code did not have a functioning IPv6 address, so such a failure would be expected. There was no indication that an IPv4 address was being retrieved for this connection attempt, even though the DNS server is supposed to return both an IPv4 and an IPv6 address. We spent a fair amount of time trying to figure out why the wrong IP address was being used, even though in this case it turned out all of the DNS and IP operations were working correctly.

The code [here](https://github.com/psycopg/psycopg/blob/3.2.7/psycopg/psycopg/connection.py#L96) in the Psycopg `connection.py` file iterates through all IP addresses returned from the DNS request. In this case the IPv4 address was being attempted first, but it was failing due to an SSL certification error. The IPv6 address was attempted next, but that also failed due to the missing IPv6 support on the host. Each subsequent exception overwrites the previous in the current logic, and only the most recent exception is shown to the user. This results in scenarios where the relevant exception (in this case the first IPv4 attempt) is suppressed, and the user is confused about a misleading exception.

This was particularly confusing in this case since it implied an IP layer issue, or some issue with the DNS lookup. In practice this was an SSL certificate issue but there was no way to determine this without deeper investigation. After enabling debug logging I could see the failed IPv4 attempt, but I didn't realize that was possible until after digging through the Psycopg source code to find the connection logic:

```
DEBUG - psycopg - connection attempt failed: host: '***' port: '5432', hostaddr '54.158.101.233': connection failed: connection to server at "54.158.101.233", port 5432 failed: root certificate file "./root.pem" does not exist
Either provide the file, use the system's trusted roots with sslrootcert=system, or change sslmode to disable server certificate verification.

DEBUG - psycopg - connection attempt failed: host: '***' port: '5432', hostaddr '2600:1f18:692c:303:31c4:3b43:ed6d:d04d': connection is bad: connection to server at "2600:1f18:692c:303:31c4:3b43:ed6d:d04d", port 5432 failed: Network is unreachable
	Is the server running on that host and accepting TCP/IP connections?
```

The current logic may suppress critical diagnostic information when multiple connection attempts fail. While displaying all exceptions from
every attempted IP address could be verbose (especially for domains resolving to numerous IPs), selectively showing only the last error risks hiding the root cause.

Would it be possible to show all exceptions in the case of failure here? I'm not sure if there is a better option to determine which single exception to show, like maybe a hierarchical approach where some exceptions have more weight?

Thanks in advance

danielfrankcom:
This PR fixes #1069.

It updates the connection failure logic to preserve all error messages for the various attempts. Previously, only the message from the final attempt was shown the user, which created confusing situations as described in #1069.

Below is an example of the new message format:

```
psycopg.OperationalError: connection failed: connection to server at "1.2.3.4", port 5432 failed: FATAL:  unable to accept connection, access denied
DETAIL:  Session Id: ghijklmnp4yquucuzkab7hhka
Multiple connection attempts failed. All failures were:
- host: 'redacted.for.github', port: None, hostaddr: '2200:1f18:1f72:a404:61e8:e38a:6be2:fake': connection failed: connection to server at "2200:1f18:1f72:a404:61e8:e38a:6be2:fake", port 5432 failed: FATAL:  unable to accept connection, access denied
DETAIL:  Session Id: abcdefloavxpiw6oweks7x7fcu
- host: 'redacted.for.github', port: None, hostaddr: '1.2.3.4': connection failed: connection to server at "1.2.3.4", port 5432 failed: FATAL:  unable to accept connection, access denied
DETAIL:  Session Id: ghijklmnp4yquucuzkab7hhka
```

I made a few decisions which could be discussed:
- I decided to keep the `-` pushed to the left rather than indent it, as some messages contain multiple lines. We shouldn't need to worry about artificially indenting multi-line messages, and this is made more complicated by the fact that on my terminal the first line of these messages is long enough to wrap itself. I think it's easier to see the delineation if the `-` is pushed up against the left rather than indented in these cases of multi-line or wrapping messages.
- I replaced the debug message and included the connection information in the message itself. As a user that was trying to use these messages to diagnose a problem in #1069, I think having that information is super valuable for diagnostic purposes. Though the example I posted above includes duplicate information, not all messages do and it depends on the type of failure. A timeout exception for example doesn't show any information about which host/port it was trying to connect to. If you end up with different exceptions for different hosts, it's extremely valuable to know which host produced which error.

### DIFF
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -95,29 +95,43 @@ class Connection(BaseConnection[Row]):
         timeout = timeout_from_conninfo(params)
         rv = None
         attempts = conninfo_attempts(params)
+        connection_errors: list[tuple[e.Error, str]] = []
         for attempt in attempts:
             try:
                 conninfo = make_conninfo("", **attempt)
                 gen = cls._connect_gen(conninfo, timeout=timeout)
                 rv = waiting.wait_conn(gen, interval=_WAIT_INTERVAL)
             except e.Error as ex:
-                if len(attempts) > 1:
-                    logger.debug(
-                        "connection attempt failed: host: %r port: %r, hostaddr %r: %s",
-                        attempt.get("host"),
-                        attempt.get("port"),
-                        attempt.get("hostaddr"),
-                        str(ex),
-                    )
-                last_ex = ex
+                attempt_details = "host: {}, port: {}, hostaddr: {}".format(
+                    repr(attempt.get("host")),
+                    repr(attempt.get("port")),
+                    repr(attempt.get("hostaddr")),
+                )
+                connection_errors.append((ex, attempt_details))
             except e._NO_TRACEBACK as ex:
                 raise ex.with_traceback(None)
             else:
                 break
 
         if not rv:
-            assert last_ex
-            raise last_ex.with_traceback(None)
+            last_exception, _ = connection_errors[-1]
+            if len(connection_errors) == 1:
+                raise last_exception.with_traceback(None)
+            else:
+                formatted_attempts = []
+                for error, attempt_details in connection_errors:
+                    formatted_attempts.append(f"- {attempt_details}: {error}")
+                # Create a new exception with the same type as the last one, containing
+                # all attempt errors while preserving backward compatibility.
+                last_exception_type = type(last_exception)
+                message_lines = [
+                    f"{last_exception}",
+                    "Multiple connection attempts failed. All failures were:",
+                ]
+                message_lines.extend(formatted_attempts)
+                enhanced_message = "\n".join(message_lines)
+                enhanced_exception = last_exception_type(enhanced_message)
+                raise enhanced_exception.with_traceback(None)
 
         rv._autocommit = bool(autocommit)
         if row_factory:
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -111,29 +111,45 @@ class AsyncConnection(BaseConnection[Row]):
         timeout = timeout_from_conninfo(params)
         rv = None
         attempts = await conninfo_attempts_async(params)
+        connection_errors: list[tuple[e.Error, str]] = []
         for attempt in attempts:
             try:
                 conninfo = make_conninfo("", **attempt)
                 gen = cls._connect_gen(conninfo, timeout=timeout)
                 rv = await waiting.wait_conn_async(gen, interval=_WAIT_INTERVAL)
             except e.Error as ex:
-                if len(attempts) > 1:
-                    logger.debug(
-                        "connection attempt failed: host: %r port: %r, hostaddr %r: %s",
-                        attempt.get("host"),
-                        attempt.get("port"),
-                        attempt.get("hostaddr"),
-                        str(ex),
-                    )
-                last_ex = ex
+                attempt_details = "host: {}, port: {}, hostaddr: {}".format(
+                    repr(attempt.get("host")),
+                    repr(attempt.get("port")),
+                    repr(attempt.get("hostaddr")),
+                )
+                connection_errors.append((ex, attempt_details))
             except e._NO_TRACEBACK as ex:
                 raise ex.with_traceback(None)
             else:
                 break
 
         if not rv:
-            assert last_ex
-            raise last_ex.with_traceback(None)
+            last_exception, _ = connection_errors[-1]
+            if len(connection_errors) == 1:
+                raise last_exception.with_traceback(None)
+            else:
+                formatted_attempts = []
+                for error, attempt_details in connection_errors:
+                    formatted_attempts.append(f"- {attempt_details}: {error}")
+
+                # Create a new exception with the same type as the last one, containing
+                # all attempt errors while preserving backward compatibility.
+                last_exception_type = type(last_exception)
+
+                message_lines = [
+                    f"{last_exception}",
+                    "Multiple connection attempts failed. All failures were:",
+                ]
+                message_lines.extend(formatted_attempts)
+                enhanced_message = "\n".join(message_lines)
+                enhanced_exception = last_exception_type(enhanced_message)
+                raise enhanced_exception.with_traceback(None)
 
         rv._autocommit = bool(autocommit)
         if row_factory:
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -24,6 +24,8 @@ from ._test_connection import testctx  # noqa: F401  # fixture
 from ._test_connection import conninfo_params_timeout, tx_params, tx_params_isolation
 from ._test_connection import tx_values_map
 
+MULTI_FAILURE_MESSAGE = "Multiple connection attempts failed. All failures were:"
+
 
 def test_connect(conn_cls, dsn):
     conn = conn_cls.connect(dsn)
@@ -35,6 +37,45 @@ def test_connect(conn_cls, dsn):
 def test_connect_bad(conn_cls):
     with pytest.raises(psycopg.OperationalError):
         conn_cls.connect("dbname=nosuchdb")
+
+
+@pytest.mark.slow
+def test_connect_error_single_host_original_message_preserved(conn_cls, proxy):
+    with proxy.deaf_listen():
+        with pytest.raises(psycopg.OperationalError) as e:
+            conn_cls.connect(proxy.client_dsn, connect_timeout=2)
+
+    msg = str(e)
+    assert "connection timeout expired" in msg
+    assert MULTI_FAILURE_MESSAGE not in msg
+
+
+@pytest.mark.slow
+def test_connect_error_multi_hosts_each_message_preserved(conn_cls):
+    # IPv4 address blocks reserved for documentation.
+    # https://datatracker.ietf.org/doc/rfc5737/
+    args = {"host": "192.0.2.1,198.51.100.1", "port": "1234,5678"}
+    with pytest.raises(psycopg.OperationalError) as e:
+        conn_cls.connect(**args, connect_timeout=2)
+
+    msg = str(e.value)
+    assert MULTI_FAILURE_MESSAGE in msg
+
+    host1, host2 = args["host"].split(",")
+    port1, port2 = args["port"].split(",")
+
+    msg_lines = msg.splitlines()
+
+    expected_host1 = f"host: '{host1}', port: '{port1}', hostaddr: '{host1}'"
+    expected_host2 = f"host: '{host2}', port: '{port2}', hostaddr: '{host2}'"
+    expected_error = "connection timeout expired"
+
+    assert any(
+        (expected_host1 in line and expected_error in line for line in msg_lines)
+    )
+    assert any(
+        (expected_host2 in line and expected_error in line for line in msg_lines)
+    )
 
 
 def test_connect_str_subclass(conn_cls, dsn):
--- a/tests/test_connection_async.py
+++ b/tests/test_connection_async.py
@@ -21,6 +21,8 @@ from ._test_connection import testctx  # noqa: F401  # fixture
 from ._test_connection import conninfo_params_timeout, tx_params, tx_params_isolation
 from ._test_connection import tx_values_map
 
+MULTI_FAILURE_MESSAGE = "Multiple connection attempts failed. All failures were:"
+
 
 async def test_connect(aconn_cls, dsn):
     conn = await aconn_cls.connect(dsn)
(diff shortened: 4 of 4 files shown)


psycopg/psycopg/errors.py
```diff
@@ -1757,3 +1757,4 @@ class IndexCorrupted(InternalError,
 # instance in a certain `Cursor.execute()`.
 
 _NO_TRACEBACK = (Error, KeyboardInterrupt, CancelledError)
+_NoTracebackException = Union[Error, KeyboardInterrupt, CancelledError]
```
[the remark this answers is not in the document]
dvarrazzo: Unrelated, but I just noticed (https://github.com/psycopg/psycopg/issues/1055#issuecomment-2859787149) that it's wrong here to trap KeyboardInterruplt: instead of stopping we just move to the following attempt.

So the except should probably just trap `psycopg.Error`, not the other we trap in queries.

psycopg/psycopg/connection.py
```diff
@@ -95,27 +95,44 @@ def connect(
         timeout = timeout_from_conninfo(params)
         rv = None
         attempts = conninfo_attempts(params)
+        connection_errors: list[tuple[e._NoTracebackException, str]] = []
```
dvarrazzo, line 98: You could use `list[tuple[BaseException, str]]` or even `list[tuple[Exception, str]]` if we stop catching `KeyboardInterrupt/CancelledError` here.

psycopg/psycopg/connection.py
```diff
@@ -95,27 +95,44 @@ def connect(
...
-                        attempt.get("host"),
-                        attempt.get("port"),
-                        attempt.get("hostaddr"),
-                        str(ex),
-                    )
-                last_ex = ex
+                attempt_details = "host: {}, port: {}, hostaddr: {}".format(
+                    repr(attempt.get("host")),
+                    repr(attempt.get("port")),
+                    repr(attempt.get("hostaddr")),
```
dvarrazzo, line 108: Is this the most useful order in which to show these 3 details or you would rather see them in a different error to spot problems more easily (hostaddr first maybe?)

psycopg/psycopg/connection.py
```diff
@@ -95,27 +95,44 @@ def connect(
...
+                )
+                connection_errors.append((ex, attempt_details))
             else:
                 break
 
         if not rv:
-            assert last_ex
-            raise last_ex.with_traceback(None)
+            num_errors = len(connection_errors)
+            assert num_errors > 0
```
dvarrazzo, line 116: I don't think this assert is useful. `assert last_ex` was necessary to please MyPy, but this list will never be empty, and even if it was, IndexError or AssertError don't make any difference.

psycopg/psycopg/connection.py
```diff
@@ -95,27 +95,44 @@ def connect(
...
+                # all attempt errors while preserving backward compatibility.
+                last_exception_type = type(last_exception)
+                message_lines = [
+                    f"{last_exception}",
+                    "Multiple connection attempts failed. All failures were:",
+                ]
+                message_lines.extend(formatted_attempts)
+                enhanced_message = "\n".join(message_lines)
+                enhanced_exception = last_exception_type(enhanced_message)
+                raise enhanced_exception.with_traceback(None) from last_exception
```
dvarrazzo, line 135: Does it make sense to raise `from last_exception`? I don't think it adds much value. Maybe a simple exception is better.

tests/test_connection.py
```diff
@@ -37,6 +37,46 @@ def test_connect_bad(conn_cls):
         conn_cls.connect("dbname=nosuchdb")
 
 
+def test_connect_error_single_host_original_message_preserved(conn_cls):
+    with pytest.raises(psycopg.OperationalError) as e:
+        conn_cls.connect("dbname=nosuchdb")
+
+    msg = str(e)
+    assert 'database "nosuchdb" does not exist' in msg
+    assert "Multiple connection attempts failed" not in msg
```
dvarrazzo, line 46: Please put this string in a const and use the same const for both the tests, otherwise, if we change the string this test would become invalid (it would pass even if we break something).

psycopg/psycopg/connection_async.py
```diff
@@ -111,27 +111,46 @@ async def connect(
...
+            num_errors = len(connection_errors)
+            assert num_errors > 0
+
+            last_exception, _ = connection_errors[-1]
+            if num_errors == 1:
+                raise last_exception.with_traceback(None)
+            else:
+                formatted_attempts = []
+                for error, attempt_details in connection_errors:
+                    formatted_attempts.append(f"- {attempt_details}: {str(error)}")
```
dvarrazzo, line 140: `str()` is implicit
```suggestion
                    formatted_attempts.append(f"- {attempt_details}: {error}")
```

dvarrazzo (changes requested):

danielfrankcom:
Recent pushes are attempting to fix the tests for `crdb`. Haven't taken a look through comments yet.

psycopg/psycopg/connection.py
danielfrankcom: Agreed, removing those other exceptions would make the types a lot simpler here

psycopg/psycopg/connection.py
danielfrankcom: I'm not sure - my thought was there may be multiple hosts, and then multiple addresses associated with each host. I tried to do it in the order of what I thought was likely to have the least variations first, moving to the most variations on the right.

psycopg/psycopg/connection.py
danielfrankcom: Agreed, I can remove this.

psycopg/psycopg/connection.py
danielfrankcom: I think this is a hold-over from before you mentioned they were already flattened, so I was trying to preserve the origin of the exception. Since they're flat, I don't think this makes sense. Will remove.

psycopg/psycopg/connection.py
danielfrankcom: Should be addressed in [982dda4](https://github.com/psycopg/psycopg/commit/982dda4f5c99832d95e0e447ba6c16eb9122a24a). I've left the `_NO_TRACEBACK` variable since it's used in quite a few places and may still be needed in the future.

psycopg/psycopg/connection.py
danielfrankcom: Addressed in [982dda4](https://github.com/psycopg/psycopg/commit/982dda4f5c99832d95e0e447ba6c16eb9122a24a)

psycopg/psycopg/connection.py
danielfrankcom: Addressed in [982dda4](https://github.com/psycopg/psycopg/commit/982dda4f5c99832d95e0e447ba6c16eb9122a24a)

tests/test_connection.py
danielfrankcom: Addressed in [982dda4](https://github.com/psycopg/psycopg/commit/982dda4f5c99832d95e0e447ba6c16eb9122a24a)

psycopg/psycopg/connection_async.py
danielfrankcom: Addressed in [982dda4](https://github.com/psycopg/psycopg/commit/982dda4f5c99832d95e0e447ba6c16eb9122a24a)

danielfrankcom:
In the latest push I modified the first unit test to depend on a timeout error instead of the invalid DB error. It seems CRDB doesn't produce an error when the DB doesn't exist, so I need to reply on a different type of failure for that case.

This was a bit confusing as the original test directly above the 2 I added gets an `OperationalError` from CRDB, but it seems like it's because the DSN isn't configured. If I set the `dsn` and override the `dbname`, the CRDB connection still succeeds which isn't what I want for that unit test.

dvarrazzo:
I am making the change in #1077 where we log and continue attempts only when a database error is returned, and we exit immediately if a ctrl-c is received. I would say let's first merge that MR, then we can rebase this on top, so the type annotation can be simplified.

dvarrazzo:
> Recent pushes are attempting to fix the tests for `crdb`. Haven't taken a look through comments yet.

We can just skip the test on CRDB. This is a purely client behaviour, different vendors will not make a relevant difference.

psycopg/psycopg/connection.py
dvarrazzo: Ok, let's leave it this way.

danielfrankcom:
> > Recent pushes are attempting to fix the tests for `crdb`. Haven't taken a look through comments yet.
> 
> We can just skip the test on CRDB. This is a purely client behaviour, different vendors will not make a relevant difference.

We can do that if you would prefer, but there should be no need with the latest PR version. I switched the forced exception type to a timeout, which worked and passed all tests including on CRDB.

Regarding #1077 and rebasing this one afterwards - that sounds good to me.

tests/test_connection.py
```diff
@@ -37,6 +39,51 @@ def test_connect_bad(conn_cls):
...
+    assert MULTI_FAILURE_MESSAGE not in msg
+
+
+def test_connect_error_multi_hosts_each_message_preserved(conn_cls, monkeypatch):
+    # This is unreasonably short as it is expected to fail as fast as possible.
+    monkeypatch.setattr(psycopg.conninfo, "_DEFAULT_CONNECT_TIMEOUT", 0.1)
+
+    # IPv4 address blocks reserved for documentation.
+    # https://datatracker.ietf.org/doc/rfc5737/
+    args = {"host": "192.0.2.1,198.51.100.1", "port": "1234,5678"}
```
dvarrazzo, line 63: This is cool, I didn't know about these unroutable addresses! Thank you for the reference.

So far, in order to test with a host that doesn't reply, we opened a listening socket with [proxy.deaf_port](https://github.com/psycopg/psycopg/blob/24ad0eb967188702b6fccabb2d1b46a7f6a8b7c7/tests/fix_proxy.py#L105-L118) but if these IPs work reliably we can probably do without.

You can set `connect_timeout` in the connection string instead of monkeypatching. However, 0.1s is a slow test :slightly_smiling_face: so we need to mark it so.

```suggestion
@pytest.mark.slow
def test_connect_error_multi_hosts_each_message_preserved(conn_cls):
    # IPv4 address blocks reserved for documentation.
    # https://datatracker.ietf.org/doc/rfc5737/
    args = {"host": "192.0.2.1,198.51.100.1", "port": "1234,5678", "connect_timeout": 0.1}
```

same for both tests, both sync and async.

dvarrazzo (changes requested):

tests/test_connection.py
danielfrankcom: I didn't use the proxy in this case since I needed 2 hosts and 2 ports that would reliably fail and it seemed like the proxy only exposed 1 of each. If you have a suggestion for how to do that with the proxy then I'm happy to swap, but the documentation IPs should work reliably and have done so in my testing.

I'll remove the monkeypatch and mark the tests accordingly, thanks for the tips on this.

psycopg/psycopg/errors.py
```diff
@@ -1756,4 +1755,4 @@ class IndexCorrupted(InternalError,
 # to get the high level information about where the exception was raised, for
 # instance in a certain `Cursor.execute()`.
 
-_NO_TRACEBACK = (Error, KeyboardInterrupt, CancelledError)
+_NO_TRACEBACK = (Error,)
```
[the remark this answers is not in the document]
danielfrankcom: Ah I see, I misunderstood our previous discussion. I'll rebase onto the other change now since that should remove any need for me to worry about this.

tests/test_connection.py
dvarrazzo: These tests hosts are ok. In other tests we have in the test suite probably we monkeypatch the dns to inject expected failures, but I don't think it's needed here.

danielfrankcom:
Latest push rebases onto #1077. I will push again shortly after addressing the test comments.

tests/test_connection.py
danielfrankcom: Should be addressed by [24c2995](https://github.com/psycopg/psycopg/commit/24c299561e3f13793029fa98706aaac637fce92d)

danielfrankcom:
Latest push addresses unit test comments. I removed the monkeypatching, and used the deaf proxy for the first test since it only required 1 host/port. Both tests were updated to use the `connect_timeout` parameter, though I had to set them to a 2 second timeout due to the logic [here](https://github.com/psycopg/psycopg/blob/3.2.7/psycopg/psycopg/conninfo.py#L142) which the monkeypatch was bypassing.

psycopg/psycopg/errors.py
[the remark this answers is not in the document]
dvarrazzo: This was addressed by #1077.

dvarrazzo:
Thank you very much, merged!

danielfrankcom:
Excellent, thank you

Issue #685, opened 2023-11-27
Deadlock when using pipeline mode


wemoloh:
When inserting a large amount of data, psycopg hangs when using pipeline mode.  It does not hang if pipeline mode is not used (but it is much slower when the amount of inserted data is small).

Platform: Windows Server 2019
psycopg version: 3.1.13
PostgreSQL version: 15.2
Python version: 3.11.5

For example, given the following definitions:

```sql
create table test (bigdata text[]);
```

```python
from psycopg import connect

def insert_a_lot(cur, nrows, nchars, nvals):
    for _ in range(nrows):
        data = ['a' * nchars] * nvals
        cur.execute('insert into test (bigdata) values (%s)', (data,))
```

...the following code executes as expected:

```python
with connect(TESTDB) as conn, conn.cursor() as cur:
    insert_a_lot(cur, 10, 1000, 1000)
```

...but the following code just hangs forever.  The database reports the connection is left in 'ClientRead' state:

```python
with connect(TESTDB) as conn, conn.cursor() as cur:
    with conn.pipeline():
        insert_a_lot(cur, 10, 1000, 1000)
```

dvarrazzo:
Hello,

I have played with the issue. On Linux, with psycopg 3.1.17, using the C extension the script never blocks, using the Python implementation, the script blocks about the 50% of the times.

When it blocks, a ctrl-c reveals:

```
Traceback (most recent call last):
  File "/home/piro/dev/psycopg3/test-685.py", line 12, in <module>
    with conn.pipeline():
  File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/home/piro/dev/psycopg3/psycopg/psycopg/connection.py", line 958, in pipeline
    with pipeline:
  File "/home/piro/dev/psycopg3/psycopg/psycopg/_pipeline.py", line 248, in __exit__
    self._conn.wait(self._exit_gen())
  File "/home/piro/dev/psycopg3/psycopg/psycopg/connection.py", line 974, in wait
    return waiting.wait(gen, self.pgconn.socket, timeout=timeout)
  File "/home/piro/dev/psycopg3/psycopg/psycopg/waiting.py", line 342, in wait_poll
    fileevs = poll.poll(timeout)
KeyboardInterrupt
```

[...]

PR #1081, opened 2025-05-09
Add a result iterator to all cursor classes

### LINKED ISSUE #1080: Add a result iterator to all cursor classes
After executing the `executemany` function, it is needed to iterate over the results for each set of parameters. It would be useful if cursor classes explicitly provided such an iterator.
```python
def results(self) -> Iterator[Self]:
    # if params_seq is not empty
    if cursor.pgresult:
        while True:
            yield cursor

            if not cursor.nextset():
                break
```
For example, instead of writing every time
```python
if cursor.pgresult:
    while True:
        process_result(cursor)

        if not cursor.nextset():
            break
```
one could write
```python
map(process_result, cursor.results())
```
If you agree that this is a good idea, I can prepare PR including tests.

embecka:
Closes #1080.


tests/test_cursor_server_async.py
```diff
@@ -572,7 +572,7 @@ async def test_stolen_cursor_close(aconn):
 
 
 async def test_row_maker_returns_none(aconn):
-    cur = aconn.cursor(row_factory=rows.scalar_row)
+    cur = aconn.cursor("test", row_factory=rows.scalar_row)
```
dvarrazzo, line 575: Ah I missed this. It is something to review outside this MRm

docs/news.rst
```diff
@@ -32,6 +32,8 @@ Psycopg 3.2.8 (unreleased)
 - Don't process further connection attempts after Ctrl-C (:ticket:`#1077`).
 - Fix cursors to correctly iterate over rows even if their row factory
   returns `None` (:ticket:`#1073`).
+- Add `~Cursor.sets()` to iterate over the result sets of the queries
+  executed though `~Cursor.executemany()` or `~Cursor.execute()`.
```
dvarrazzo, line 36: This change would be released in 3.3 because it is a new feature. You mat mention ticket #1080, where you proposed the idea in first place.

tests/test_cursor_server_async.py
dvarrazzo: On it.

dvarrazzo: Addressing it in 00e67ff4. Only the test was wrong, the library behaviour is correct.

docs/api/cursors.rst
```diff
@@ -266,6 +262,18 @@ The `!Cursor` class
...
+
+        .. note::
+
+            You can use::
+
+                for _ in cursor.results():
+                    for row in cursor:
+                        ...
+
+            to iterate over result sets and over rows in the current result set.
```
dvarrazzo, line 275: To expand on the docstring I propose the following. Also adding `versionadded`.

```suggestion
        The iterator yields the cursor itself upon iteration, but the cursor
        state changes, in a way equivalent to calling `nextset()` in a loop.
        Therefore you can ignore the result of the iteration if you are consuming
|        `!results()` in a loop::

                for _ in cursor.results():
                    for row in cursor:
                        ...

        or make use of it for example using `map()` to consume the iterator::

                def cursor_consumer(cur: Cursor) -> Any:
                    ...

                map(cursor_consumer, cursor.results())
                
        .. versionadded:: 3.3
```

docs/news.rst
```diff
@@ -16,6 +16,9 @@ Psycopg 3.3.0 (unreleased)
 - Cursors are now iterators, not only iterables. This means you can call
   ``next(cur)`` to fetch the next row (:ticket:`#1064`).
 - Drop support for Python 3.8 (:ticket:`#976`) and 3.9 (:ticket:`#1056`).
+- Add `~Cursor.results()` to iterate over the result sets of the queries
```
dvarrazzo, line 19: For better context, let's make the first `Cursor` visible:

```suggestion
- Add `Cursor.results()` to iterate over the result sets of the queries
```

psycopg/psycopg/_cursor_base.py
```diff
@@ -164,6 +165,20 @@ def nextset(self) -> bool | None:
         else:
             return None
 
+    def results(self) -> Iterator[Self]:
```
dvarrazzo, line 168: As discussed in #1080, my preference, and I understand yours too, would be to make this cursor async.

I do understand that consistency with `nextset()` is relevant. However that would be consistency with a design error, which would just contribute to make the mistake more serious and mitigating it more costly. For example if we amy decide to make `nextset()` async in an eventual Psycopg 4; now people would have both `nextset()` and `results()` to chase and fix.

Happy to receive other people's opinion too, but my view is that this method should be async.

dvarrazzo (changes requested):

dvarrazzo:
@dlax I would like to hear your opinion about a topic related to this improvement.

As you surely remember, initially, in psycopg 3.1 the pipeline mode was accumulating all the results in the cursor and we changed this behaviour in #604 (we warned against using the feature in 3.1.10 and changed the behaviour in 3.2). However the lack of reliability might have been related to other errors that have been since fixed (like #685, fixed in 3.1.18).

I see some utility in receiving more than one result, especially considering that - I seem to remember - in pipeline mode you can't use more than one query in the same statement. People could use the pipeline mode to request a batch of objects from different tables and then receive all the result to aggregate on the client:

```python
with cur.pipeline():
    cur.execute("select * from parent where id = %s", [rec_id])
    cur.execute("select * from child where parent_id = %s", [rec_id])

    parent = cur.fetchone()
    cur.nextset()
    children = cur.fetchall()
```

So I was thinking to enable this behaviour on request: what do you think about a parameter on the `pipeline()` method, such as `all_results=True`, which would make the results aggregate in the cursor, like in psycopg 3.1 instead of clobbering the current one (which should remain the default behaviour to avoid changing the cursor behaviour between pipeline and non-pipeline mode)?

I haven't reviewed completely the conversations in the tickets linked. Do you have in mind any reasons why this couldn't obviously work?

docs/api/cursors.rst
```diff
@@ -122,17 +122,13 @@ The `!Cursor` class
...
-                    if not cur.nextset():
-                        break
+                ids = [cur.fetchone()[0] for _ in cur.results()]
 
         .. warning::
 
             More explicitly, `!fetchall()` alone will not return all the
             values returned! You must iterate on the results using
-            `!nextset()`.
+            `!results()`.
```
dvarrazzo, line 131: Looking at the docs, `nextset()` is also used in the [from_pg2 page](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#multiple-results-returned-from-multiple-statements).

Do we want to change that example? Like:

```
using the usual `!fetch*()` methods. In order to access all the results,
you can use the `~Cursor.results()` method::

    >>> cur_pg3.execute("SELECT 1; SELECT 2")
    >>> for _ in cur_pg3:
    ...     print(cur.fetchone())
    (1,)
    (2,)
```

I don't feel the need to keep the `nextset()` example, in all its clunkiness. If anything we could add a paragraph like:

```
Before Psycopg 3.3 you can use `~Cursor.nextset()` instead.
```

dvarrazzo (changes requested):

tests/test_cursor_server.py
```diff
@@ -579,3 +579,10 @@ def test_row_maker_returns_none(conn):
         assert list(cur) == recs
         stream = cur.stream(query)
         assert list(stream) == recs
+
+
+@pytest.mark.parametrize("count", [1])
+def test_results_after_execute(conn, count):
```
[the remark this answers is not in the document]
dvarrazzo: No need to raise unsupported error: the method works but it is basically no-op (it always return a one-element generator). This is consistent with `nextset()` which works, but will just return `False`.

tests/test_cursor_server_async.py
```diff
@@ -585,3 +585,10 @@ async def test_row_maker_returns_none(aconn):
         assert await alist(cur) == recs
         stream = cur.stream(query)
         assert await alist(stream) == recs
+
+
+@pytest.mark.parametrize("count", [1])
+async def test_results_after_execute(aconn, count):
+    async with aconn.cursor() as cur:
```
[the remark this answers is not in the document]
dvarrazzo: Fixed

docs/api/cursors.rst
dvarrazzo: Fixed.

Issue #888, opened 2024-08-02
psycopg 3.2 causing SEVERE performance degredation


davidgrohmann:
We recently pulled in psycopg3.2 through a regular pipenv upgrade cycle and it has killed performance across a few apps. Unfortunately I don't have any exact measurements to point to, but something has changed, and reverting to 3.1.x solved the problem.

I don't know if this is the queries themselves returning slower or  if it is establishing a connection or what.

I'm posting this here as a warning and hopefully if others see same problem and can post more details.

I apologize for not bringing any evidence

dvarrazzo:
Compared to which version?

jameschristopher:
This may or may not be related but I've noticed a new error `connection timeout expired` in pure python psycopg >3.1.19. This occurs when running under gevent. Pinning to ==3.1.19 does not cause this error. It's connecting to pgbouncer so I doubt it's a postgres side problem. Thank you @dvarrazzo for your continue work on this project.

mtvx:
I'm witnessing similar(ish) behavior bumping to 3.2, seeing a bunch of randomish `connection timeout expired`.

The problem should be along the lines https://github.com/psycopg/psycopg/blob/00c6625d4735b7fe774e517e87f7daadc41a82f6/psycopg/psycopg/connection.py#L101-L103 which got refactored in https://github.com/psycopg/psycopg/pull/766.

I added timing information to the lines before and after the connection attempts, and the code got into https://github.com/psycopg/psycopg/blob/00c6625d4735b7fe774e517e87f7daadc41a82f6/psycopg/psycopg/connection.py#L118-L119 without actually waiting for the given `connect_timeout`. For example, with 10s given as connect timeout, the code got into raising state in just some milliseconds.

Didn't go deeper as I'm not really familiar with that code.

Anyway, something flawed in the new waiting logic?

jameschristopher:
That's interesting, I have also found it to be random. I thought maybe their was truly some kind of constraint on my pgBouncer instance, but it is managed by Heroku, so that seemed unlikely. And it wasn't during any kind of high load either.

jameschristopher:
As a follow up, after more analysis I determined that the `connection timeout expired` errors were occurring when my app was over my Heroku dyno's memory soft limit. I was able to solve the problem by increasing my Heroku dyno size to get more memory. Not sure what the underlying mechanics are of this, but incase anyone stumbles across this issue I want them to have a solution, such as it is.

jhrr:
[...]

PR #1085, opened 2025-05-10
Better logging

dvarrazzo:
After the struggle in figuring out slowness on connection (#888, #1055), improving and documenting connection logging.

See the docs change for an example of the logging produced. It should help to figure out the connection attemtps and where any stops.

### DIFF
--- a/docs/basic/usage.rst
+++ b/docs/basic/usage.rst
@@ -236,3 +236,55 @@ integration between your Python program and your PostgreSQL database:
 - If you want to customise how Python values and PostgreSQL types are mapped
   into each other, beside the :ref:`basic type mapping <types-adaptation>`,
   you can :ref:`configure your types <adaptation>`.
+
+
+.. _logging:
+
+Connection logging
+------------------
+
+Psycopg uses the `logging` module to report the operations happening at
+connection time. If you experience slowness or random failures on connection
+you can set the ``psycopg`` logger at ``DEBUG`` level to read the operations
+performed.
+
+A very simple example of logging configuration may be the following:
+
+.. code:: python
+
+    import logging
+    import psycopg
+
+    logger = logging.getLogger()
+    logging.basicConfig(level=logging.DEBUG, format="%(asctime)s %(levelname)s %(message)s")
+
+    logging.getLogger("psycopg").setLevel(logging.DEBUG)
+
+    psycopg.connect("host=192.0.2.1,localhost connect_timeout=10")
+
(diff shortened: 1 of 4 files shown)


docs/basic/usage.rst
```diff
@@ -236,3 +236,55 @@ integration between your Python program and your PostgreSQL database:
...
   into each other, beside the :ref:`basic type mapping <types-adaptation>`,
   you can :ref:`configure your types <adaptation>`.
+
+
+.. _logging:
+
+Connection logging
+------------------
+
+Psycopg uses the `logging` module to report the operations happening at
```
asqui, line 246: Is it worth clarifying that this is the _standard library_ logging module (as opposed to a psycopg internal logging module)?

Is it easy to link to https://docs.python.org/3/library/logging.html

docs/basic/usage.rst
```diff
@@ -236,3 +236,55 @@ integration between your Python program and your PostgreSQL database:
...
+performed.
+
+A very simple example of logging configuration may be the following:
+
+.. code:: python
+
+    import logging
+    import psycopg
+
+    logger = logging.getLogger()
```
asqui, line 258: This line is not needed? `logger` is not used in the example.

asqui (commented):
Nice docs, as always.

Made a couple of unsolicited drive-by-review comments 😜

docs/basic/usage.rst
dvarrazzo: The above links to the stdlib docs indeed (see [rendered version](https://www.psycopg.org/psycopg3/docs/basic/usage.html#connection-logging)). No hurt in making it
```
uses the stdlib `logging` module"
```
though. Thank you!

docs/basic/usage.rst
dvarrazzo: Ah you are right. WIll remove it. Thank you for the "dive-by" review! :smile:

docs/basic/usage.rst
asqui: Ah nice, is that Sphinx magic? (I'm big on documentation, in general, but am a complete Sphinx newb!)

dvarrazzo: Yes, "[intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)"

https://github.com/psycopg/psycopg/blob/48a151b805224f090577341e7809cd7fc9fcc3a3/docs/conf.py#L95-L99

Issue #1091, opened 2025-05-14
Memory leak when listening to Notifications - with long time connections


carleson:
This memory leak was introduced in version 3.2.4 and remains in current version (3.2.8).

In the ambition to resolve issue #962, a new memory leak was introduced in the handling of NOTIFY events.

[dvarrazzo: Dec 10, 2024](https://github.com/psycopg/psycopg/issues/962)

> This can become a leak, but only if the user calls notifies() once on a connection and 
> then uses it for a long time, never closing it, never calling notifies() again, and 
> without unlistening the channel. 

In our usecase we use long running connections to listen for NOTIFY events.
The number of NOTIFY events can exceed 10.000 events/minute.
In our case the internal memory use increase very quickly (around 200Mb/hour) .

We suspect that the **connection._notifies_backlog** buffer is not consumed correctly.
It seems like every notify event is added to this variable, is this really correct behaviour?

Test example code.

```python
import psycopg
import sys
import time


def connect() -> psycopg.connection.Connection:
    connect_string = "postgresql://my_user@dbserver/dbname"

    conn = psycopg.connect(
        connect_string,
        cursor_factory=psycopg.ClientCursor,
        autocommit=True,
        keepalives=1,
        keepalives_idle=5,
        keepalives_interval=2,
        keepalives_count=2,
    )
    return conn

def handle_message(message: psycopg.Notify):
    pass

def main():
    sql = "SELECT 1;"
    with connect() as conn:
        conn.add_notify_handler(handle_message)
        conn.execute("LISTEN notify_name;")
        while True:
            conn.execute(sql)
            time.sleep(1)
            backlog_size = sys.getsizeof(conn._notifies_backlog)
            backlog_objects = len(conn._notifies_backlog)
            print(f'Size: {backlog_size}, objects: {backlog_objects}')


if __name__ == "__main__":
    main()
```

dvarrazzo:
Hello,

yes, you are doing exactly what I said that an user should do to have a leak: you are calling LISTEN on a connection and then you are not reading the notifications using `conn.notifies()`. For example, adding `print(list(conn.notifies(timeout=0)))` resets the backlog to 0.

Is your expectation is that because you have registered a message handler then the backlog doesn't get filled? This might be reasonable...

dvarrazzo:
I have opened #1092 for an attempt to change the behaviour of the system and avoid to collect notifies in the backlog if there is any notify handler registered.

[...]

Issue #873, opened 2024-07-18
Failing to update `notifies` while changing from 3.1 to 3.2


Blackeagle52:
Hi Daniele,

This issue gonna have some examples, so I want to preface with an excuse for the length of this issue.

Currently I'm using psycopg3.1 within a Django app stack, and I fail to update to version 3.2. My app (quite generally described) has services notifying of tasks in a channel, and another service listening to those channels. But I've reduced my problem to a generic psycopg3.2 only example.

First I want to show you how the app currently uses LISTEN/NOTIFY (with version 3.1)

```python
    def listen(self):
        connection.ensure_connection()
        real_con = connection.connection
        real_con.autocommit = True
        with connection.cursor() as cursor:
            for name in self.tasks:
                cursor.execute(sql.SQL("LISTEN {};").format(sql.Identifier(name)))
        return real_con

    def poller(self):
        real_con = self.listen()
        while True:
            if select.select([real_con], [], [], self.timeout) == ([], [], []):
                logger.debug("Polling timeout, time to check for periodics to execute")
                yield None
            else:
                for notify in real_con.notifies():
                    logger.debug(
                        "Got NOTIFY: %s, %s, %s", notify.pid, notify.channel, notify.payload
                    )
                    yield notify.channel
```

Depending the yield-value of the poller certain CRUD stuff happened in a transaction. Having read #760, it appears it is bad practice to use the connection while the generator is being consumed, and a lock has been introduced while the `notifies` generator runs. 

This lead to a rewrite of the above code, but my attempts have been failures. My current solution does not receive all notifies. To reduce the problem I've created a small setup to reproduce the problem. I have the following two files, `main.py` and `notifier.py`. 
First step is to start `main.py` to activate the listeners. Then run the notifier a few times. 
After the files are the logs the files print, and the `postgres -E` log.

I hope the problem is properly described, and you can help me find out what's going wrong. As proper examples how to use `notifies` are hard to find. 

Thanks in advance,
Peter

 

`main.py`
```python
import datetime
import time

import psycopg
from psycopg import sql

channels = ['channel_1', 'channel_2']


def log(msg):
    print(datetime.datetime.utcnow().isoformat(), msg)


def __listen(connection):
    with connection.cursor() as cursor:
        for channel in channels:
            log(f"Listen on channel {channel}")
            cursor.execute(sql.SQL("LISTEN {};").format(sql.Identifier(channel)))


def __poller(connection, timeout):
    while True:
        n = None
[...]

PR #1092, opened 2025-05-14
Collect notifies only if no handler is registered

dvarrazzo:
Opening this MR to look into this way to fix the issue is reasonable and if it doesn't cause problems somewhere else.

If someone is listening to notifications by using an handler, the notifies backlog would fill without ever being emptied: see #1091.

This change has the risk of breaking something if someone is relying on notifies being received both via callback and via generator, but I don't know how to satisfy it without creating a leak to users who don't use the generator.

The MR changes the tests because `test_notify_query_notify` doesn't pass anymore. The test is now split in two testing notifies and generator separately.

### DIFF
--- a/docs/advanced/async.rst
+++ b/docs/advanced/async.rst
@@ -325,16 +325,33 @@ Because of the way transactions interact with notifications (see |NOTIFY|_
 documentation), you should keep the connection in `~Connection.autocommit`
 mode if you wish to receive or send notifications in a timely manner.
 
-Notifications are received as instances of `Notify`. If you are reserving a
-connection only to receive notifications, the simplest way is to consume the
-`Connection.notifies` generator. The generator can be stopped using
-`!close()`. Starting from Psycopg 3.2, the method supports options to receive
-notifications only for a certain time or up to a certain number.
-
 .. note::
 
     You don't need an `AsyncConnection` to handle notifications: a normal
     blocking `Connection` is perfectly valid.
+
+Notifications are received as instances of the `Notify` object. You can
+receive notifications using either the :ref:`notifies generator
+<notifies-generator>` or a :ref:`notifies handler <notifies-handlers>`.
+
+.. warning::
+
+    You should use only one of the two methods to receive notifications.
+
+    Since Psycopg 3.2.10 using the generator and handlers at the same time
+    raises a runtime warning.
+
+
+.. _notifies-generator:
+
+Notifies generator
+^^^^^^^^^^^^^^^^^^
+
+If you are reserving a connection only to receive notifications, the simplest
+way to receive them is to consume the `Connection.notifies()` generator. The
+generator can be stopped using its `!close()` method, or using the parameters
+`!timeout` or `!stop_after` to receive notifications only for a certain time
+or up to a certain number.
 
 The following example will print notifications and stop when one containing
 the ``"stop"`` message is received.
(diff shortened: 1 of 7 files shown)


dvarrazzo:
Hello @codeech, @Blackeagle52, @carleson

We have been reported an issue with notifies in #1091: if someone calls LISTEN and uses a handler to listen to notifications, the backlog implemented to play back notifications in the `notifies()` fills indefinitely.

It comes to think that if someone has registered a handler then they wouldn't be interested in using the generator to manage the notifications too. However maybe this assumption doesn't hold?

As you have worked with notifications in your systems (#873, #962) I would like to ask you for your feedback: if we changed the behaviour of the adapter and we didn't collect notifies anymore in case a handler is registered, would it break the code you have written?

In practice, I think that the change in this MR would cause the following behaviour:

- receiving notifications via handler will work as before, no change;
- receiving notifications via generator will work as after 3.2.4, with the notification received out of a `notifies()` run delivered at the beginning of the `notifies()` call, **but only if no notifies handler has been registered**;
- if a notifies handler is registered then the generator will only return notifies received while it is running, as it was before 3.2.4.

In practice, I would say that we should discourage to use both the notifier handler and the generator at the same time on a connection.

What do you think about these changes? Is there anything else do you think we could do better?

Thank you in advance for the feedback.

carleson:
@dvarrazzo I think this is a reasonable compromise which should apply to most cases.

If you want to use both notify handler and generator you can create separate connection objects/Listeners, right?.
I've looked at the patch and I think it looks like a good solution. Nice, good and quickly fixed! :+1:

dvarrazzo:
I am a bit concerned to make this change so halfway in a bugfix release knowing it is a breaking change. So my proposal is:

- we will release this change in 3.3
- in next 3.2 bugfix release we will start raising a warning when `notifies()` is used with a handler installed.
- we will change the documentation to discourage to use the handler and the generator on the same connection.

@carleson, would you be able to change your systems and add some periodic `list(conn.notifies(timeout=0))` to clean up the buffer and "survive" until 3.3 is released (which will likely happen when Python 3.14 is released)? Of course this would be exactly the situation that would raise the warning above; therefore we can allow setting up an env var, such as `PSYCOPG_NO_NOTIFIES_WARNING` to avoid the warning.

Thoughts?

dvarrazzo:
Another option could also be to enable the 3.3 behaviour already, guarded by an env var, such as `PSYCOPG_NO_NOTIFIES_BACKLOG`. Let me think about that a bit more...

carleson:
@ dvarrazzo As for me, you can do what's best for you.

Our solution is to use version 3.2.3 until the issue is finally resolved.
Since this works for us, we will not use feature flags to enable upcoming functionality.

/thx

Blackeagle52:
I have no issue with the change. We had issues with the 3.2.4 release because we used `notifies` and the handler together. Because of the addition of the collector it was possible for us to remove the handler from our code, and solely depend on `notifies`.

I kinda complained about a breaking change in 3.2.4, I'm sorry about that. But, right now, I think this fix should be released in 3.2.5. Although @carleson is the first who created the issue, there can be more users who have a memory leak and don't know about it. Ignorance is bliss.

>- in next 3.2 bugfix release we will start raising a warning when notifies() is used with a handler installed.
>- we will change the documentation to discourage to use the handler and the generator on the same connection.

I think these warnings would be a nice addition to the MR. As I stated in #962, when I used `notifies` and the handler together in 3.2.4, it was possible to receive duplicate `Notify` events. By adding a warning other (ignorant) people could be helped.

dvarrazzo:
Hey,

sorry for the long silence on this. I'm clearing my psycobacklog and trying to wrap a bugfix release soon, including this.

I have implemented the behaviour described in this MR. To repeat it (as I'm aware I've gone a bit back and forth on it):

- receiving notifications via handlers will work as before, no change;
- receiving notifications via generator will work as after 3.2.4, with the notification received out of a `notifies()` run delivered at the beginning of the `notifies()` call, **but only if no notifies handler has been registered**;
- if a notifies handler is registered then the generator will only return notifies received while it is running, as it was before 3.2.4.

Because of the awkwardness of the interaction between generator and handlers, starting from 3.2.10 we will raise a `RuntimeWarning` if the generator and the handlers are used on the same connection.

This MR also describes the changes in notifications reception (both the ones happened in 3.2.4 and the ones happening in 3.2.10).

Let me know if you have any feedback. Cheers!

Blackeagle52:
Thanks for the update! Looks like a solid solution and a good way forward.

PS. There is a small typo in the `RuntimeWarning`; "thees".

dvarrazzo:
> PS. There is a small typo in the RuntimeWarning; "thees".

Thank you, will fix :)

PR #1116, opened 2025-07-05
Add PostgreSQL 18 OAuth support

dvarrazzo:
This is a work in progress, there is still a lot to understand with this functionality and how it can interact with Python.


dvarrazzo:
PR closed in favour of #1190.

Issue #1132, opened 2025-08-17
Support running psycopg test suite with pytest-xdist


bluetech:
On my computer (a recent AMD Ryzen 9 HX 370), `pytest` takes about 8 minutes. When run with `-n auto` (this computer has 12 cores/24 threads), it takes 50s. So it seems worthwhile to make pytest-xdist work, but feel free to close the issue if you are not interested in this.

There are two issues:

1. The [`segfault` detection](https://github.com/psycopg/psycopg/blob/7671556413ca4b94911ef2032d5f1be6140c28e1/tests/conftest.py#L59-L69) in conftest.py is not compatible with xdist; the cache is shared among the the controller and all workers, so the workers immediately assume a segfault occurred and refuse to run. The solution is to only do it on the controller.

2. Some tests fail, presumably because of interference from other concurrently-running tests (see example list below). The solution is either to make the tests not interfere with each other, or just make each worker run in its own DB instead of all running in against a single shared DB (this is the approach pytest-django uses for example). I haven't looked at this yet.

<details><summary>Example test failures with 24 workers</summary>
<p>

```
ERROR tests/test_transaction.py::test_force_rollback_successful_exit[asyncio-pipeline=on] - psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
ERROR tests/test_transaction.py::test_preserves_autocommit[asyncio-pipeline=off-True] - psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
ERROR tests/test_transaction.py::test_rollback_on_exception_exit[asyncio-pipeline=off] - psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
ERROR tests/test_transaction.py::test_named_savepoints_exception_exit[asyncio-pipeline=on] - psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
ERROR tests/test_transaction.py::test_interaction_dbapi_transaction[asyncio-pipeline=on] - psycopg.errors.UndefinedTable: table "test_table" does not exist
ERROR tests/test_transaction.py::test_out_of_order_exit_same_name[asyncio-pipeline=off-Rollback] - psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "pg_type_typname_nsp_index"
[...]

PR #1133, opened 2025-08-17
test: make `segfault` detection compatible with pytest-xdist

bluetech:
Fixes point 1 of #1132.


bluetech:
Also pushed a commit for point 2, but closing per https://github.com/psycopg/psycopg/issues/1132#issuecomment-3194427417.

PR #1139, opened 2025-08-26
Set gssencmode default to disable in binary package builds

### LINKED ISSUE #1136: Psycopg under multiprocessing on MacOS crash when kerberos authentication enabled (which is by default).
On an M2 Mac, I was having an issue where using postgres (via django), would crash Celery (usually, though not every time).

It turns out there can be an issue on Macs where fork unsafe code can cause issues under python multiprocessing.

I  tried disabling kerberos (which i'm not using in my local dev environment) and things started working.
`gssencmode=disable`.

This affects people using psycopg on Macs and some form of multiprocessing (my own is Celery, but others are available).

Until this is fixed in Postgres this should probably be documented here, and possibly defaulted to off.
I'm also not entirely sure where to report the bug upstream, so this seems like a good place to start.

I've attached a stacktrace of the crash.
[stacktrace-psycopg.txt](https://github.com/user-attachments/files/21942216/stacktrace-psycopg.txt)

dvarrazzo:
The MR builds on top of #1137, but it targets the master branch.
Tentatively scheduled for release in 3.3.
 
Fix #1136

### DIFF
--- a/tests/pq/test_pq.py
+++ b/tests/pq/test_pq.py
@@ -1,4 +1,5 @@
 import os
+import sys
 
 import pytest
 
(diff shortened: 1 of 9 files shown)


tests/pq/test_pq.py
```diff
@@ -1,4 +1,5 @@
 import os
+import sys  # noqa: F401  # used in skipif
```
dlax: Can't we a boolean instead of a "condition string" in this `skipif` so as to avoid this `# noqa` here?

dlax (approved):

tests/pq/test_pq.py
dvarrazzo: Yes, but then we need to specify a reason, whereas, with a string, pytest will do it for you.

But in this case I added a sort of reason as comment, so yes,you are right: better as a bool here :)

dvarrazzo:
Ready to merge. In 3.2.x we will raise a DeprecationWarning, in 3.3 it will be a RuntimeWarning, and the spelling of the message is reworded accordingly.

Issue #645, opened 2023-09-19
Excessive CPU usage on Windows


dvarrazzo:
Reported [in the mailing list](https://www.postgresql.org/message-id/DB7PR07MB3916A230837909499A87EF3287E8A%40DB7PR07MB3916.eurprd07.prod.outlook.com) by David Raymond:

---

I've noticed that CPU usage for queries seems to have drastically increased going from psycopg2 to psycopg3.

In psycopg2 a connection would sit quietly while waiting for the results, using next to no CPU.
In psycopg it looks like the threads are using their core at full power, continuously checking for the result as fast as possible, and never sleeping or relaxing.

See below for a simple example script. During the psycopg2 part while the 8 threads are waiting on results my processor usage sits at the normal idling 2-3%. But in the psycopg3 part, while the 8 threads are waiting on results my processor usage jumps up to 62-63% or so.

Is there a setting or configuration I'm missing?
This is on Windows, where I install with "pip install psycopg2-binary" and "pip install psycopg[binary]"
Current versions shown by pip list:
psycopg            3.1.10    (This gets installed even when telling it to install psycopg[binary])
psycopg-binary     3.1.10
psycopg2-binary    2.9.7

And their reported .__version__
psycopg.__version__: 3.1.10
psycopg2.__version__: 2.9.7 (dt dec pq3 ext lo64)

I've got several scripts where I open up multiple connections to multiple servers and run a number of queries on them, where 98% of my machine's time should be just relaxing and waiting for the server to finish the next 1 to 15 minute long query. I don't need to continue the very microsecond the query completes. Is there anything I can do to bring CPU usage back down with psycopg (3) ?

Thank you,

```python
import sys
import threading


import psycopg
import psycopg2


def connectAndRunSleep3():
    conn = psycopg.connect(host = "localhost", dbname = "...", user = "...", password = "...", autocommit = True)
    with conn.cursor() as cur:
        cur.execute("select pg_sleep(15);")
    conn.close()


def connectAndRunSleep2():
    conn = psycopg2.connect(host = "localhost", dbname = "...", user = "...", password = "...")
    conn.autocommit = True
    with conn.cursor() as cur:
        cur.execute("select pg_sleep(15);")
    conn.close()


if __name__ == "__main__":
    print("Python version:", sys.version)
    print("psycopg.__version__:", psycopg.__version__)
    print("psycopg2.__version__:", psycopg2.__version__)
    
    numThreads = 8
    
[...]

PR #1141, opened 2025-08-27
Fix polling for read+write

Arshia001:
Hi!

This change fixes an issue with using `poll` when checking for both read and write at the same time.

According to my understanding of `poll`, `events` contains the inputs to the `poll` function, whereas `revents` contains the actual results. This means that, with the previous code, when a poll would happen for both read and write (`events = 3`), but only one would be ready afterwards (e.g. `revents = 2`), the code would mistakenly assume both reads and writes were ready.

This issue breaks our [WASIX](https://wasix.org) fork of [psycopg-binary](https://pythonindex.wasix.org/simple/psycopg-binary/index.html) when a socket is ready to be written but no data is available for reading; the code gets a read-ready flag by mistake and makes a socket receive call, which times out.

TBH, I'm surprised as to how this issue hasn't been discovered. Perhaps most platforms use the `select`-based variant of `wait_c_impl`, or otherwise something in the socket flags causes the mistaken receive call to not time out? I find it strange enough that I did multiple searches to make sure I'm not wrong about `revents` vs. `events`.

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/waiting.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/waiting.pyx
@@ -88,8 +88,8 @@ retry_eintr:
 
     rv = 0;  /* success, maybe with timeout */
     if (select_rv >= 0) {
-        if (input_fd.events & POLLIN) { rv |= SELECT_EV_READ; }
-        if (input_fd.events & POLLOUT) { rv |= SELECT_EV_WRITE; }
+        if (input_fd.revents & POLLIN) { rv |= SELECT_EV_READ; }
+        if (input_fd.revents & POLLOUT) { rv |= SELECT_EV_WRITE; }
     }
 
 #else


dvarrazzo:
Hello,

Thank you for your report. Eyeballing, it looks that you are right about the missing use of `revents`.  I am trying to find the bits of code that affected my implementation, but ISTM that I got them from Postgres libpq, which never waits for RW. I'll try to dig more later.

Running a quick `pytest -m not\ slow` on my machine, I found a test failings, which can be reproduced by the following script:

```python
import socket
from psycopg import waiting

wait = waiting.Wait.R
ready = waiting.Ready.R
waitfn = waiting.wait_c  # doesn't work
# waitfn = waiting.wait_poll  # works

def gen():
    r = yield wait
    return r

with socket.socket() as s:
    r = waitfn(gen(), s.fileno())
assert r & ready
```

so maybe there's more to change? (Or is it the test methodology wrong?)

Windows users reported excessive CPU usage after introducing this change (#645), but I think that windows is not a `HAVE_POLL` platform, so it is not affected by this change. Does it sound correct?

Arshia001:
I'm not super-familiar with psycopg or even with Python as a whole, but I'll try my best:

> which never waits for RW

This is the specific piece of code that was getting stuck and timing out:

https://github.com/psycopg/psycopg/blob/c69395bb1054c59f6695af45f8bba8514b1c3e05/psycopg/psycopg/generators.py#L240-L254

> which can be reproduced by the following script

I don't think I understand what the test is supposed to do exactly... To my (untrained) eye, it seems as though it's opening a socket, and just waiting for a read event to arrive for it? I'm not quite sure what's the expected behavior when polling an unconnected, unbound socket. If I bind the socket:

```python
with socket.socket() as s:
    s.bind(("127.0.0.1", 50505))
    s.listen(1000)
    print("waiting...")
    r = waitfn(gen(), s.fileno())
    print("done:", r)
```

Then what happens is:

* with `wait_c`, the socket properly waits for a connection to come in, then reports it's ready with a `1`.
* with `wait_poll`, it just returns `0` immediately, and the assertion fails. If I also specify a timeout, _then_ `wait_poll` works as intended and returns `1`.

Now, changing `wait` to `RW`, like so:

```python
import socket
from psycopg import waiting

wait = waiting.Wait.RW
ready = waiting.Ready.R
waitfn = waiting.wait_c
# waitfn = waiting.wait_poll

def gen():
    r = yield wait
    return r

with socket.socket() as s:
    s.bind(("127.0.0.1", 50506))
    s.listen(1000)
    print("waiting...")
    r = waitfn(gen(), s.fileno(), 1000)
    print("done:", r)
assert r & ready

```

and waiting for a connection to come in on the socket, we get:

* with `wait_c`, the reported result is `3`, which is R+W. This is wrong, and the subject of this PR.
* with `wait_poll`, the reported result is `1`, which is correct.

Note, this is with an unpatched psycopg I installed off of pip, and running python natively to avoid any potential errors in the implementation of WASIX.

> Windows users reported excessive CPU usage

TBH, no idea... AI tells me Windows doesn't have poll at all.

--------------

To summarise:

* I'm not sure what the output from the code snippet you posted should be, but I think polling an unbound, unconnected socket is not a well-defined operation (though I could be wrong)
* With a properly bound socket, `wait_c` reports the wrong value of RW when just an R or W comes in
* `wait_poll` behaves differently to `wait_c` in that, if a timeout is not specified, it returns a zero immediately.
  * I would expect a zero timeout value to mean infinity though?

dvarrazzo:
> - I'm not sure what the output from the code snippet you posted should be, but I think polling an unbound, unconnected socket is not a well-defined operation (though I could be wrong)

Or you could be right. I'm not sure how robust that test is. The script I posted is just a stand-alone version of a test in the test suite that exercises the various wait functions.

https://github.com/psycopg/psycopg/blob/c69395bb1054c59f6695af45f8bba8514b1c3e05/tests/test_waiting.py#L51-L65

All the other wait implementation return 1 on read, only the C implementation returns 0. As suggested earlier, it might also be the test methodology to be wrong. However, `wait_c()` internally uses `poll()` or `select()` so I don't really expect, or desire, a radically different behaviour.

> - With a properly bound socket, wait_c reports the wrong value of RW when just an R or W comes in

Yup, that is wrong, The goal is to fix it.

> - wait_poll behaves differently to wait_c in that, if a timeout is not specified, it returns a zero immediately.
>
>   - I would expect a zero timeout value to mean infinity though?

Just an immediate polling in case of 0 seems more useful, otherwise you'll need an epsilon-short time in case you need to poll. Waiting forever is not very useful as Python wouldn't be able to interrupt it and the whole interpreter would freeze.

I am not sure whether in our codebase we ever make use of a 0 timeout, and I'm reasonably sure that we never use an infinite timeout, so maybe neither of these behaviours are exercised anyway by the codebase using a wait function. But I'll take a look at the consistency of the various implementations.

Arshia001:
> Just an immediate polling in case of 0 seems more useful

This makes sense. When I commented on the difference in behaviour between `wait_c` and `wait_poll`, I failed to realize that leaving a value out would set it to `None`, not 0. Again, not very good with Python, sorry!

However, the difference is still there; leaving the timeout value out causes `wait_c` to wait (probably indefinitely?), whereas `wait_poll` returns immediately. I guess this is one more thing that needs to be fixed in `wait_c`.

jerch:
@dvarrazzo I have looked up several of the defined generators and most of them do something like this for the hot data-in-transport path:

```python
while not (yield WAIT_RW):
    continue
```

Now looking at the loop here: https://github.com/psycopg/psycopg/blob/93ae9094232d01532da34ae3c57bd3d0477019e2/psycopg_c/psycopg_c/_psycopg/waiting.pyx#L200-L213

we see, that the send function gets always called, even if PY_READY_NONE was returned, which puts the generator in an "almost busy" loop iteration. It is only an almost busy loop, because in cython the poll timeout would still put a handbrake on it.

Together with the bug found above, this creates some extra CPU rounds for no good reasons. Note that the bug found above would always return something in rv due to reading the events instead of revents, so this would add additional generator iterations.
What I dont get yet - why did that bug above not cause issues before? I have only a guess here - maybe you shaped the generators versatile enough to just deal with this edge case...

Proposed solution:

Beside the revent fix above, I'd change the following:
- move the goto retry_eintr jump into GIL released section (no need to handle that with re-aquiring GIL in between causing additional stops) or even better - use glibc's TEMP_FAILURE_RETRY macro (see https://sourceware.org/glibc/manual/2.42/html_node/Interrupted-Primitives.html)
- if possible, shortwire the READY_NONE state in the poll loop, so the generator does not iterate in its busy loop anymore (this only works, if there are no generators explicitly waiting for that condition)

These changes should at least lift some CPU usage, but I dont expect it to be a huge difference, that would make ppl complain about it. In a second step I'd suggest to have a closer look at the details of poll & select and whether there are unhandled rarely occuring conditions, that might lead huge CPU usage (real busy loop). As I wrote in another thread, POLLHUP can lead to this under certain circumstances, but I am not yet sure, if it is a possible issue here.

jerch:
On a sidenote - I dont quite understand your architecture design regarding the postgres connection and the waiting yet. In a multithreading application using multiple sync connections - would each thread do its polling on its own (within that thread), or do you collect the polling needs somehow and delegate that to a dedicated polling thread? Sorry, if this question sounds dumb, but I didnt find anything in that regard yet.

dvarrazzo:
Hi @jerch,

> we see, that the send function gets always called, even if PY_READY_NONE was returned, which puts the generator in an "almost busy" loop iteration. It is only an almost busy loop, because in cython the poll timeout would still put a handbrake on it.

Yes: I seem to remember we have a 0.1s timeout, so we iterate 10 times/sec. This is required to be able to break the loop with a signal: if we use no timeout and we only yield when there is actual activity the client will appear stuck (like in psycopg2 in non-green mode, i.e. a path where we use the `PQexec` function and we let the lipbq to select/poll in C code unrelated to Python.

> In a multithreading application using multiple sync connections - would each thread do its polling on its own

Yes, correct.

dvarrazzo:
@Arshia001 thank you for pointing out that the tests for the wait functions were misleading. I have improved the test suite and better checked for the behaviour of the different implementations.

I have removed support for "waiting forever" with an interval of None, which was unused in our codebase (I think... tests will prove me wrong if that's not the case), made the behaviour of tomeout=0 consistent and made sure that all returned values are what should be expected.

Running the tests in [this branch](https://github.com/dvarrazzo/psycopg/tree/fix/wait-rw) to release the change as a 3.2 bugfix and then adding the changes made there to this branch for merging to master.

Thank you very much!

Arshia001:
Cheers! Happy to help.

dvarrazzo:
I cannot push changes here, so I have opened #1183. Please set this branch on that head, so I can merge this PR. Otherwise I'll just merge #1183 instead.

Arshia001:
@dvarrazzo done. But feel free to just merge #1183 in if it's easier.

dvarrazzo:
Thank you very much. This issue triggered a long quest about why we hadn't spot the issue before and led to massive tests improvement and fixing several other wait-related issues.

It was tough but worth it.

I'll check about other pending bugfixes and make a release soon to include this change. Thank you very much!

Arshia001:
Always happy to help! :)

PR #1144, opened 2025-08-28
Add `krb5-libs` for alpine/musl

edgarrmondragon:


dvarrazzo:
Looks like it somewhat worked, but not on x86_64

dvarrazzo:
Observation: when I tested with 3.1.0 the build passed, but I also tweaked the cache key: see https://github.com/psycopg/psycopg/commit/f174fe7e869f768cc86d4350063e0cedc2a727ac

This makes me think that the issue is with caching and that we are failing to include the depending libraries in the cache.

Trying a build with the latest cibuildwheel and cache deactivated [here](https://github.com/psycopg/psycopg/actions/runs/17326771043)... and it seems that [build passed](https://github.com/psycopg/psycopg/actions/runs/17326771043/job/49192741659).

So I don't think this is the solution but we should rather review what we are caching.

dvarrazzo:
Ok, I think the problem is that, in the build_libpq script, if we found a cache restored we didn't install the dependencies.

The changeset https://github.com/psycopg/psycopg/commit/08e7975284c0e451f69a29b8b7eca43b6bf84944 seems to solve the problem: install the dependencies before checking if the libpq is already build and hence bail out.

Closing this MR as likely unneeded. Thank you very much all the same!

PR #1145, opened 2025-08-29
Fixing a small typo

frankliuao:
`.format()` was incorrectly placed out of `.SQL()`.

### DIFF
--- a/docs/basic/from_pg2.rst
+++ b/docs/basic/from_pg2.rst
@@ -94,8 +94,8 @@ use the `psycopg.sql` module::
     # This will quote the user and the password using the right quotes
     # e.g.: ALTER USER "foo" SET PASSWORD 'bar'
     >>> conn.execute(
-    ...     sql.SQL("ALTER USER {} SET PASSWORD {}"
-    ...     .format(sql.Identifier(username), password)))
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Format is precisely [a `psycopg.SQL` method](https://www.psycopg.org/psycopg3/docs/api/sql.html#psycopg.sql.SQL.format).

frankliuao:
Thanks, @dvarrazzo !
I was confused because I realized it's a bit inconsistent with how `.format()` was used throughout the doc.
I then realized after seeing your comment that the **other one** is probably the one we need to reformat? 🤦 
```>>> from psycopg import sql

# This will quote the user and the password using the right quotes
# e.g.: ALTER USER "foo" SET PASSWORD 'bar'
>>> conn.execute(
...     sql.SQL("ALTER USER {} SET PASSWORD {}"
...     .format(sql.Identifier(username), password)))```

dvarrazzo:
Yep, that looks wrong. Would you like to amend your MR and fix It instead?

frankliuao:
Done with a new commit. @dvarrazzo thanks a lot for the instructions.

dvarrazzo:
Thank you very much!

PR #1149, opened 2025-09-01
Fix binary copy

dvarrazzo:
Avoid generating spurious exceptions in binary copy if there is an error in a dumper.

Close #1147

### DIFF
--- a/psycopg/psycopg/_copy.py
+++ b/psycopg/psycopg/_copy.py
@@ -136,8 +136,9 @@ class Copy(BaseCopy["Connection[Any]"]):
         using the `Copy` object outside a block.
         """
         if self._direction == COPY_IN:
-            if data := self.formatter.end():
-                self._write(data)
+            if not exc:
+                if data := self.formatter.end():
+                    self._write(data)
             self.writer.finish(exc)
             self._finished = True
         else:
--- a/psycopg_c/psycopg_c/_psycopg/copy.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/copy.pyx
@@ -34,10 +34,7 @@ def format_row_binary(
     else:
         pos = PyByteArray_GET_SIZE(out)
 
-    # let's start from a nice chunk
-    # (larger than most fixed size; for variable ones, oh well, we'll resize it)
-    cdef char *target = CDumper.ensure_size(
-        out, pos, sizeof(berowlen) + 20 * rowlen)
+    cdef char *target = CDumper.ensure_size(out, pos, sizeof(berowlen))
 
     # Write the number of fields as network-order 16 bits
     memcpy(target, <void *>&berowlen, sizeof(berowlen))
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -39,6 +39,7 @@ Psycopg 3.2.10 (unreleased)
 - Fix memory leak when lambda/local functions are used as argument for
   `~.psycopg.types.json.set_json_dumps()`, `~.psycopg.types.json.set_json_loads()`
   (:ticket:`#1108`).
+- Fix bad data on error in binary copy (:ticket:`#1147`).
 - Fix `psycopg_binary.__version__`.
 
 
--- a/psycopg/psycopg/_copy_async.py
+++ b/psycopg/psycopg/_copy_async.py
@@ -133,8 +133,9 @@ class AsyncCopy(BaseCopy["AsyncConnection[Any]"]):
         using the `Copy` object outside a block.
         """
         if self._direction == COPY_IN:
-            if data := self.formatter.end():
-                await self._write(data)
+            if not exc:
+                if data := self.formatter.end():
+                    await self._write(data)
             await self.writer.finish(exc)
             self._finished = True
         else:
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -643,6 +643,18 @@ def test_description(conn):
         assert cur.description[2].name == "column_3"
 
 
+def test_binary_partial_row(conn):
+    cur = conn.cursor()
+    ensure_table(cur, "id serial primary key, num int4, arr int4[][]")
+    with pytest.raises(
+        psycopg.DataError, match="nested lists have inconsistent depths"
+    ):
+        with cur.copy("copy copy_in (num, arr) from stdin (format binary)") as copy:
+            copy.set_types(["int4", "int4[]"])
+            copy.write_row([15, None])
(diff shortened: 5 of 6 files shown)


psycopg/psycopg/_copy.py
```diff
@@ -136,8 +136,9 @@ def finish(self, exc: BaseException | None) -> None:
         using the `Copy` object outside a block.
         """
         if self._direction == COPY_IN:
-            if data := self.formatter.end():
-                self._write(data)
+            if not exc:
+                if data := self.formatter.end():
+                    self._write(data)
```
jerch, line 141: This change already fixed it for me, so not sure if the cython change is needed at all?

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -34,10 +34,7 @@ def format_row_binary(
     else:
         pos = PyByteArray_GET_SIZE(out)
 
-    # let's start from a nice chunk
-    # (larger than most fixed size; for variable ones, oh well, we'll resize it)
-    cdef char *target = CDumper.ensure_size(
-        out, pos, sizeof(berowlen) + 20 * rowlen)
+    cdef char *target = CDumper.ensure_size(out, pos, sizeof(berowlen))
```
jerch, line 37: As wrote above, not quite sure if this change is really needed.

On a sidenote - I still wonder, why the buffer would go out of hand here. Since `pos` indicates the last buffer write end, shouldn't oversizing things at the beginning have no effect at all? I dont know much about the dumper code, just wondering if `pos` gets in all cases adjusted before or after dumper execution. If always done afterwards, an exception within the dumper should not report a too big buffer, as pos would still be the value from the last successful field adaption?

jerch: Oh wait - since the exception will exit the code block early, the final pos-based size adjustment gets skipped and eventually the bigger current size of the buffer reported back.

Got it, sorry for the noise.

psycopg/psycopg/_copy.py
dvarrazzo: Either is needed. The cython error causes returning a buffer with some random data, so it can be fixed by either not generating junk or not using it. 

Generating junk is a bug in itself and I don't feel comfortable to leave it unfixed, even if not sending to the server already solves the random exceptions.

Looks like we can easily div both the errors though.

jerch:
@dvarrazzo Here the issue I suspected:

```python
import psycopg

conn = psycopg.connect('dbname=postgres user=postgres password=mysecretpassword host=localhost port=5432')
conn.cursor().execute('CREATE TEMPORARY TABLE test (id serial PRIMARY KEY, int2d integer[][])')

def repro():
    with conn.cursor() as cur:
        with conn.transaction():
            with conn.transaction():
                with cur.copy(b'COPY "test" ("int2d") FROM STDIN (FORMAT BINARY)') as copy:
                    copy.set_types([1007])
                    copy.write_row([None])
                    # user tries himself to handle likely to fail rows...
                    try:
                        copy.write_row([[[None], None]])
                    except:
                        pass
                    copy.write_row([None])
                    copy.write_row([[123, 456]])

repro()

```
The above weird looking handling again surfaces to wrong bytearray size/aligment in the c version and I get the following postgres error:

```text
psycopg.errors.BadCopyFileFormat: unexpected EOF in COPY data
CONTEXT:  COPY test, line 2, column int2d
```

It is possible to fix this by post-patching the write buffer similar to this:
```patch
diff --git a/psycopg/psycopg/_copy_base.py b/psycopg/psycopg/_copy_base.py
index 05180950..57dd4a8a 100644
--- a/psycopg/psycopg/_copy_base.py
+++ b/psycopg/psycopg/_copy_base.py
@@ -285,7 +285,13 @@ class BinaryFormatter(Formatter):
             self._write_buffer += _binary_signature
             self._signature_sent = True
 
-        format_row_binary(row, self.transformer, self._write_buffer)
+        pos = len(self._write_buffer)
+        try:
+            format_row_binary(row, self.transformer, self._write_buffer)
+        except Exception as e:
+            self._write_buffer = self._write_buffer[:pos]
+            raise e
+
         if len(self._write_buffer) > BUFFER_SIZE:
             buffer, self._write_buffer = self._write_buffer, bytearray()
             return buffer
```
But this is a quite awkward workaround with delayed exception propagation, so more like a hack.

A possibly better fix would be to make `pos` an outer variable passed by reference. This way a dumper can freely fail and would not update `pos`. Also the exception can freely bubble this way. (Well `pos` would have to be the end position in the buffer of the last successfully written row, which is in line with the granularity of `write_row`.)

Edit: Side effect of such an outer length handling - it would make the "avoid the reallocs" idea possible in c version.

jerch:
Follow-up to my last post - did a rewrite for COPY/binary c-version with a single shared buffer and external `bytelength` tracking and it indeed is able to fix the original and that follow-up issue.
The change basically creates a single oversized bytearray as a chunk of memory to work with. `bytelength` gets tracked  on the formatter instance and applied to `format_row_binary`, which returns the new value on a successful run.
For write-out it returns `memoryview(self._write_buffer)[:byte_length]`.

But that change is very invasive (and not applied to the python version yet and also wont work with the queued writer that way), so I would want to hear your opinion first on that follow-up issue and whether the interim exception handling is good enough.

jerch:
@dvarrazzo Made a PR with my exception suggestion #1150, so you can compare them in their behavior.

Note that PR does not contain the oversized single buffer idea, as this turned out to be a lot of code rewrite for no substantial perf benefit. Also see https://github.com/psycopg/psycopg/discussions/1147#discussioncomment-14292818.

dvarrazzo:
Hello! Thank you very much for the issue you found and the proposed fix. I haven't looked into it well yet and I won't be able to for a few days. 

If you think your fix is awkward I'll try to consider if we can do something else instead, but, in the meantime, thank you for identifying the problem.

jerch:
@dvarrazzo No pressure from my side. Feel free to look at it whenever you want. In the meantime I gonna use a workaround in my lib until a proper fix has landed here. (The OOM from postgres is too scary for me to just ignore it...)

About my other PR - it uses the exception trick from above for the c-version, as there is no global byte_length handling place currently. It also fixes things on row level for the python version, but there I only had to flip 2 lines of code to make sure, that the dumpers would raise before the buffer is touched.
Imho that row level fix is the correct idea here (minus the ugly exception treatment), as `write_row` exposes the row level to the user.

PR #1150, opened 2025-09-03
fix buffer in binary copy

jerch:
@dvarrazzo Here my attempt to fix the potential buffer corruption. It ensures buffer integrity for both the c and the python version.

### DIFF
--- a/psycopg/psycopg/_copy_base.py
+++ b/psycopg/psycopg/_copy_base.py
@@ -285,7 +285,22 @@ class BinaryFormatter(Formatter):
             self._write_buffer += _binary_signature
             self._signature_sent = True
 
-        format_row_binary(row, self.transformer, self._write_buffer)
+        if _psycopg:
+            # the c version does inplace writes thus has to increase the buffer prehand
+            # before a dumper might throw, which may corrupt the output buffer
+            # to ensure buffer integrity on row level we take the last good row end
+            # and reset the buffer stripping the faulty row data
+            current_length = len(self._write_buffer)
+            try:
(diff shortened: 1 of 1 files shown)


jerch:
@dvarrazzo If you agree with the row level idea, we prolly should also ensure the row level integrity for the text format (also needs a line flip and the exception clause).

dvarrazzo:
I'm not sure yet that this is the best thing to do and I can't focus on this right now, I'm working of 4 projects at the same time.

I think that the right fix should be in the C layer: if we get an exception there we re-size the buffer to the initial size and propagate the exception. More or less what you have done here but in the code that generates the bad data, not in code using it.

jerch:
Yupp, makes sense. Will see, how far I come on my own (never done cython before) and ping you back.

jerch:
@dvarrazzo Moved the exception delay to c. Also implemented it for the text format there, but it surfaces there not to the same degree - seems the text dumpers rely more on errors sent back from postgres, thus row level integrity cannot be provided there with the same certainty.
This MR will also be affected by #1158, which might select slightly different dumpers from `set_types` for the text format, so I suggest to hold this MR back until #1158 was added.

dvarrazzo:
Sorry again for the delay in reviewing this.

I have rebased on master and solved the conflicts. I have done some lightweight refactoring (such as `cdef`'ing the private copy functions for faster call).

I have added a test that verifies the corruption and the fix discussed in this MR.

I also noticed that there was an unused code path where the copy format function were called with None as buffer and returned a new buffer. It was unused, so I dropped it, to simplify the code flow.

Running the tests, then this can be merged.

PR #1154, opened 2025-09-04
test: add test to reproduce set_types being ignored

dvarrazzo:
It has been reported ignored in the C implementation in binary format. See #1153

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,9 +6,7 @@ on:
     # on.<push|pull_request>.<branches|tags> GitHub Actions docs.
     branches:
       - "*"
-  # Note: trying to exclude pull requests because it results in duplicate job
-  # on the commits belonging to pull requests.
-  # pull_request:
+  pull_request:
   schedule:
     - cron: '48 6 * * *'
 


jerch:
Was able to fix it here: https://github.com/jerch/psycopg/commit/3497867dd0e556e94c66f84ca3f078b98376ed8b

(Sorry idk how to push things back into your already existing PR ...)

Basic idea:
If there are dumpers defined in `tx._row_dumpers` use those. Otherwise determine from values. And that second part is where idk how to solve that properly, which I think needs a conceptual decision from your side:
- either let the first value determine the dumper for the whole column (commented out, thats how it is done in BINARY)
- stick to the more flexible variable dumper approach based on current value (currently active code, that is how it was before for TEXT)

From my own TEXT protocol impl I know that it is still somewhat flexible, since postgres still does limited value conversion, so my gut feeling tells me to stick with the second approach. From an API viewpoint one could argue to also harmonize that between BINARY and TEXT, but I think postgres would just error out on mixed-up types for a column in BINARY?

Edit: If my last question is true (postgres being very picky in binary) - wouldn't it make sense to make `set_types` mandatory on the API for binary? (I know API changes need proper justification and such, so a clear indication in the docs might be sufficient...)

dvarrazzo:
> a conceptual decision from your side:

I think that if someone decides to use `set_types()` their choice should have the precedence on the psycopg types mapping.

If I understand correctly, if we implement the TEXT branch the same way of the BINARY there will be a change in behaviour:

- currently, every value is adapted as in normal querying
- in the future, the first non-null value of a column will decide the dumper for all the following values of the same column.

TBH, when I saw the current state yesterday I thought that implementing the second behaviour would also give a bit of performance because we look up for types less often. But yes, it might result in a subtle change.

Personally it is a change I would attempt. I would try to change the code to make the TEXT side similar to binary and I would check if any test in the test suite fails.

> wouldn't it make sense to make set_types mandatory on the API for binary? (I know API changes need proper justification and such, so a clear indication in the docs might be sufficient...)

No, because with the right objects (Int4, JsonB wrapper...) using `set_types()` is not necessary. Binary copy is difficult to use and who use it has to be responsible of what they are doing.

jerch:
> If I understand correctly, if we implement the TEXT branch the same way of the BINARY there will be a change in behaviour:
>     * currently, every value is adapted as in normal querying
>     * in the future, the first non-null value of a column will decide the dumper for all the following values of the same column.

Yupp, correct (except for None's ofc). So far this was not the case for TEXT, it would always deduct the dumper from the python type at hand.

> Personally it is a change I would attempt. I would try to change the code to make the TEXT side similar to binary and I would check if any test in the test suite fails.

Ok, then I'd change it to work the same way as in binary. Just a mild warning, as this might need some docs updates as well to clearly document this "frozen col dumper" behavior also for TEXT. And hopefully you dont get flooded with issue reports afterwards :smile_cat:

jerch:
@dvarrazzo Ive created a new MR for this here: https://github.com/psycopg/psycopg/pull/1158

dvarrazzo:
Closing in favour of #1158.

PR #1158, opened 2025-09-04
respect set_types in c-version in text format

jerch:
`set_types` is currently ignored in the c version in text format. This PR harmonizes the behavior with the binary version:
- respect `set_types`
- if no `set_types`, set fxied col dumper from first non-None value

Should fix https://github.com/psycopg/psycopg/discussions/1153

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/copy.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/copy.pyx
@@ -130,6 +130,11 @@ def format_row_text(
     cdef PyObject *fmt = <PyObject *>PG_TEXT
     cdef PyObject *row_dumper
 
+    # try to get preloaded dumpers from set_types
+    #if not tx._row_dumpers:
+    #    tx._row_dumpers = PyList_New(rowlen)
+    dumpers = tx._row_dumpers
+
     for i in range(rowlen):
         # Include the tab before the data, so it gets included in the resizes
         with_tab = i > 0
@@ -139,7 +144,16 @@ def format_row_text(
             _append_text_none(out, &pos, with_tab)
             continue
 
-        row_dumper = tx.get_row_dumper(<PyObject *>item, fmt)
+        #row_dumper = PyList_GET_ITEM(dumpers, i)
+        #if not row_dumper:
+        #    row_dumper = tx.get_row_dumper(<PyObject *>item, fmt)
+        #    Py_INCREF(<object>row_dumper)
+        #    PyList_SET_ITEM(dumpers, i, <object>row_dumper)
+        if dumpers:
+            row_dumper = PyList_GET_ITEM(dumpers, i)
+        else:
+            row_dumper = tx.get_row_dumper(<PyObject *>item, fmt)
+
         if (<RowDumper>row_dumper).cdumper is not None:
             # A cdumper can resize if necessary and copy in place
             size = (<RowDumper>row_dumper).cdumper.cdump(
--- a/psycopg_c/psycopg_c/_psycopg/copy.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/copy.pyx
@@ -131,8 +131,8 @@ def format_row_text(
     cdef PyObject *row_dumper
 
     # try to get preloaded dumpers from set_types
-    #if not tx._row_dumpers:
-    #    tx._row_dumpers = PyList_New(rowlen)
+    if not tx._row_dumpers:
+        tx._row_dumpers = PyList_New(rowlen)
     dumpers = tx._row_dumpers
 
     for i in range(rowlen):
@@ -144,15 +144,11 @@ def format_row_text(
             _append_text_none(out, &pos, with_tab)
             continue
 
-        #row_dumper = PyList_GET_ITEM(dumpers, i)
-        #if not row_dumper:
-        #    row_dumper = tx.get_row_dumper(<PyObject *>item, fmt)
-        #    Py_INCREF(<object>row_dumper)
-        #    PyList_SET_ITEM(dumpers, i, <object>row_dumper)
-        if dumpers:
-            row_dumper = PyList_GET_ITEM(dumpers, i)
-        else:
+        row_dumper = PyList_GET_ITEM(dumpers, i)
+        if not row_dumper:
             row_dumper = tx.get_row_dumper(<PyObject *>item, fmt)
+            Py_INCREF(<object>row_dumper)
+            PyList_SET_ITEM(dumpers, i, <object>row_dumper)
 
         if (<RowDumper>row_dumper).cdumper is not None:
             # A cdumper can resize if necessary and copy in place
--- a/tests/test_copy_async.py
+++ b/tests/test_copy_async.py
@@ -120,6 +120,17 @@ async def test_rows(aconn, format):
     assert aconn.info.transaction_status == pq.TransactionStatus.INTRANS
 
 
+@pytest.mark.parametrize("format", pq.Format)
+async def test_set_types(aconn, format):
+    cur = aconn.cursor()
+    await ensure_table_async(cur, "id serial primary key, data jsonb")
+    async with cur.copy(
+        f"copy copy_in (data) from stdin (format {format.name})"
+    ) as copy:
+        copy.set_types(["jsonb"])
+        await copy.write_row([{"foo": "bar"}])
+
+
 async def test_set_custom_type(aconn, hstore):
     command = """copy (select '"a"=>"1", "b"=>"2"'::hstore) to stdout"""
     cur = aconn.cursor()
--- a/tests/test_copy_async.py
+++ b/tests/test_copy_async.py
@@ -123,12 +123,12 @@ async def test_rows(aconn, format):
 @pytest.mark.parametrize("format", pq.Format)
 async def test_set_types(aconn, format):
     cur = aconn.cursor()
-    await ensure_table_async(cur, "id serial primary key, data jsonb")
+    await ensure_table_async(cur, "id serial primary key, data jsonb, data2 bigint")
     async with cur.copy(
-        f"copy copy_in (data) from stdin (format {format.name})"
+        f"copy copy_in (data, data2) from stdin (format {format.name})"
     ) as copy:
-        copy.set_types(["jsonb"])
-        await copy.write_row([{"foo": "bar"}])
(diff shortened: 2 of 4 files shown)


jerch:
@dvarrazzo The last commit fixes an issue with numpy integer types. For some reason I dont grasp the frozen row dumpers now refer to `_IntOrSubclassDumper`, which does an isinstance check in the dump function. After adding the numpy types there tests pass again.

(I dont want to open too many issues, but I think there is a module loading race condition hidden in `numerc.pyx` with that `_int_classes` global.)

tests/test_copy_async.py
```diff
@@ -120,6 +120,17 @@ async def test_rows(aconn, format):
...
+@pytest.mark.parametrize("format", pq.Format)
+async def test_set_types(aconn, format):
+    cur = aconn.cursor()
+    await ensure_table_async(cur, "id serial primary key, data jsonb, data2 bigint")
+    async with cur.copy(
+        f"copy copy_in (data, data2) from stdin (format {format.name})"
+    ) as copy:
+        copy.set_types(["jsonb", "bigint"])
+        await copy.write_row([{"foo": "bar"}, 123])
+
```
dvarrazzo, line 152: It would be wise to fetch back the data to verify it was written correctly (sorry, I missed doing it in my initial MR). Can you add it please?

jerch: Yupp, done.

jerch:
The changes here indeed cover all my findings described in https://github.com/psycopg/psycopg/discussions/1153 (so also True/False work now in integer columns in text format). So I think this is ready for a code review.

Whether docs changes are needed is ofc up to you. I have 2 aspects that might be worth some doc notes for COPY_IN:
- `set_types` always freezes dumpers on cols and thus the expected python types (diff to before: in the c version set_types was only respected for binary format and ignored for text format)
- without `set_types` the first not `None` value determines the dumper for a col and all follow-up col values (diff to before: for text format in the c version the dumper was solely determined by current python value)

jerch:
@dvarrazzo Found an issue with this access here https://github.com/psycopg/psycopg/blob/93ae9094232d01532da34ae3c57bd3d0477019e2/psycopg_c/psycopg_c/_psycopg/copy.pyx#L64

It will segfault when the user supplied dampers via `set_types`, but a row contains more values than entries in the dampers list. Reason is, that the item was retrieve without error checking (PyList_GetItem vs. PyList_GET_ITEM), and some memory behind the list gets treated as a pointer to a damper:
```gdb
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5a4da7b in __pyx_pf_9psycopg_c_8_psycopg__format_row_binary (__pyx_self=<optimized out>, __pyx_v_out=<optimized out>, __pyx_v_tx=<optimized out>, __pyx_v_row=<optimized out>) at psycopg_c/_psycopg.c:15266
15266         __pyx_v_size = __pyx_t_10;
```

While this certainly is a misuse of `set_types` and the actual data provided, imho it should be handled gracefully. Maybe just error out, if the dampers length doesnt match the rowlen?

dvarrazzo:
> While this certainly is a misuse of set_types and the actual data provided, imho it should be handled gracefully. Maybe just error out, if the dampers length doesnt match the rowlen?

Yeah, checking for rowlen consistency should be needed.

jerch:
> Yeah, checking for rowlen consistency should be needed.

It is now in place like this:
- if there are dumpers from `set_types`, it is "authoritative" for the expected rowlen
- otherwise rowlen of first data row is used and "authoritative" for all later rows

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -52,8 +52,9 @@ def format_row_binary(
 
     if not tx._row_dumpers:
         tx._row_dumpers = PyList_New(rowlen)
-
     dumpers = tx._row_dumpers
+    if len(dumpers) != rowlen:
```
jerch: The `len` call here costs ~10% additional conversion runtime. If we can ensure, that `tx._row_dumpers` is always of a list type, we get the length check with `PyList_GET_SIZE` almost for free.

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -130,6 +131,13 @@ def format_row_text(
     cdef PyObject *fmt = <PyObject *>PG_TEXT
     cdef PyObject *row_dumper
 
+    # try to get preloaded dumpers from set_types
+    if not tx._row_dumpers:
+        tx._row_dumpers = PyList_New(rowlen)
+    dumpers = tx._row_dumpers
+    if len(dumpers) != rowlen:
```
jerch: Same here, should be `PyList_GET_SIZE`.

jerch (commented):
A few notes to myself...

psycopg_c/psycopg_c/_psycopg/copy.pyx
dvarrazzo: Do we have this certainty?

jerch:
@dvarrazzo I am currently trying to get some profiling running with a very much reduced snippet first:

```python
from psycopg._copy_base import format_row_binary, _binary_signature, _binary_trailer
def bench3():
    with conn.cursor() as cur:
        with conn.transaction():
            with cur.copy(b'COPY "numbers" (i1,i2,i3,i4,i5) FROM STDIN (FORMAT BINARY)') as copy:
                copy.set_types(['smallint'] * 5)
                ba = bytearray(_binary_signature)
                tx = copy.formatter.transformer
                for row in npdata:
                    format_row_binary(row, tx, ba)
                ba += _binary_trailer
```
This disables most of the copy class machinery to get a better picture of the raw conversion runtime (still have to figure out, how to properly profile into cython code). At least I already found, that `len` is pretty expensive compared to the rest of `format_row_binary`. The issue above basically also applies to `len(row)` - if I switch here to `PyList_GET_SIZE` as well, the runtime is like ~10% faster. Which boils down to the question, whether we can rewrite parts of `format_row_binary` more C-ish without getting too strict in the data type assumptions.

dvarrazzo:
It depends if a piece of data comes from psycopg or from the user. For example the dumpers row is an internal object and we can make sure it is specifically a list, or we can cast to one, because it would be an operation that we would do only once at the beginning of a copy. But if the row comes from `write_row()`, and hence from the user, we can't make too many assumptions about what it is, because we declared that the function just takes a sequence, and converting every sequence to a list just to be able to use the `PyList_*` fast functions is clearly going to be slower.

A micro-optimization could be the equivalent in Cython of:
```
if type(row) is list:
    length = PyList_GET_SIZE(row)
elif type(row) is tuple:
    length = PyTuple_GET_SIZE(row)
else:
    length = len(row)
```

A list or a tuple are the most likely objects to receive, and in Cython, `type(obj) is t` is only a struct lookup and pointer comparison, so it's pretty fast.

Another question about your `if len(dumpers) != rowlen`: does the comparison happen in C or in Python? This is easy to answer by setting `annotate=True` in the cython build:

https://github.com/psycopg/psycopg/blob/faedeeb66d1f0925a835206c7c2ae0c6a1110d57/psycopg_c/build_backend/psycopg_build_ext.py#L59-L66

This will make Cython [emit html pages](http://docs.cython.org/en/stable/src/userguide/source_files_and_compilation.html#Cython.Compiler.Options.annotate) showing the generate C and whether it's generating an excess of Python boilerplate which can be reduced by specifying some type or by using a cpython api function. In the code I have on master, where the line is `cdef Py_ssize_t rowlen = len(row)`, there is enough type information to generate the following, which is nimble for Cython standards:

image

> still have to figure out, how to properly profile into cython code

When I started with the project and was looking into psycopg 2 vs. 3 and where in pg3 time was spent I used [py-spy](https://github.com/benfred/py-spy). At the time I wrote [an article](https://www.varrazzo.com/blog/2020/05/19/a-trip-into-optimisation/) about it.

jerch:
> Another question about your `if len(dumpers) != rowlen`: does the comparison happen in C or in Python?

Ah good question. I read in the cython docs, that if you want fast C-ish handling it should get bound to a cdef variable declaration. But I dont do that here, so maybe cython went the more expensive python way of things. Gonna check the html output then (so far I always went to the c file cython generated :sweat_smile:). So thx for the pointers, gonna help a lot to get a better picture...

jerch:
Update: I churned the code of `format_row_binary` doing unspeakable things (precalc of needed bytearray space, inlining dumper code, pulling row iteration to c etc.) - the raw conversion of 5 smallint fields gets **4-5 times faster** with that (from 0.09 +- 0.01 down to 0.02 +- 0.01 for 100k rows).

To me it seems that this is kinda the baseline with cython and value by value dumping. Imho from here on only heavy lifting with SIMD tricks and approaching uniform data in bigger blocks can gain more perf (might be interesting for numpy).

Ofc nothing of my code bending is directly applicable to the current internal APIs. Well lets see if something can be carried over...

jerch:
@dvarrazzo Seems the current c build is always `-O2`? Is `-O3` worth considering (would do really aggressive optimization including auto vectorization), or is that doomed to fail?

dvarrazzo:
My understanding is that the optimization level, as well as other compile parameters, are the one Python was compiled with, and it seems unsafe to do otherwise.

jerch:
Oh how convenient: there is a [PySequence_Fast_GET_SIZE](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast_GET_SIZE) for tuples|lists :smile_cat:

jerch:
Kinda derailed this MR with my perf babbling - sorry for that. Still need to cleanup stuff here...

jerch:
Cleanup is done - ready for review.

psycopg_c/psycopg_c/_psycopg/copy.pyx
jerch: Placed a note, where _row_dumpers is set.

dvarrazzo:
Merging!

I have improved the test and the check for row length mismatch in 8244668751676eddb4ea1d28ca044e57a7e16e9f. I assume you wrote that test only to check for a segfault, but the excessively wide `except Exception` was masking other errors. Now mismatch in text mode raises DataError too.

Sorry for taking so long, but other bugs have been an immense can of worms...

jerch:
@dvarrazzo Nice, thank you! Yes the change on the test case make totally sense to not mask other exceptions.

PR #1161, opened 2025-09-06
fix buffer sizing for Int2BinaryDumper in c version

jerch:
@dvarrazzo There is no testcase for this, as I dont know how to test this (it is a cdef function, the value gets immediately corrected by the only reachable outer dump function).


dvarrazzo:
I think it's ok as it is. This bug only results in the buffer enlarged slightly more than what strictly needed (because the int16 will be smaller than the pointer on any supported platform - we are still working on the Commodore 64 porting...) the return value of the function is correct, so functions like the one to format a row for copy will not lose alignment.

jerch:
@dvarrazzo A C64 port is highly appreciated, ofc! But dont let the amiga fans in the rain :rofl:

PR #1163, opened 2025-09-09
fix speed regression in fetchall

### LINKED ISSUE #1155: Performance degradation compared to psycopg2
I noticed some performance issues when trying to migrate a project from psycopg2 to psycopg 3.

### Scenario 1

Inserting one huge row (100 MB). Newer psycopg is ~6 times slower:

```
=== testing psycopg2
inserting data...
1.273s
fetching data...
0.001s
=== testing psycopg
inserting data...
7.920s
fetching data...
0.001s
```

Code:

```py
import time

import psycopg  # installed with psycopg[binary]
import psycopg2

ROW_COUNT = 1
TEXT_LENGTH = 100_000_000

DATA = [(i, 'a' * TEXT_LENGTH) for i in range(ROW_COUNT)]


def measure(func):
    t = time.time()
    func()
    print(f'{(time.time() - t):.3f}s')


def insert():
    print("inserting data...")
    cursor.execute("CREATE TEMPORARY TABLE tmp_table (id int, t text)")
    cursor.execute(' '.join([
        "INSERT INTO tmp_table VALUES",
        ', '.join(str(row) for row in DATA),
        "RETURNING id",
    ]))


def fetch():
    print("fetching data...")
    rows = cursor.fetchall()
    assert [row[0] for row in rows] == [row[0] for row in DATA]


for module in [psycopg2, psycopg]:
    print('=== testing', module.__name__)

    with module.connect(dbname='postgres', user='postgres', password='', host='localhost', port=5432) as conn:
        with conn.cursor() as cursor:
            measure(insert)
            measure(fetch)
```

The difference is not that big with a smaller amount of data, but still noticeable.

I also tried 10000 rows * 10 kB (which is closer to how my data looks in practice), but the results were similar.

### Scenario 2

Inserting 1 milion tiny rows (strings of length 1). Here the insertion times don't differ much, but fetching the identifiers is ~3 times slower:

```
=== testing psycopg2
inserting data...
2.079s
fetching data...
0.194s
=== testing psycopg
inserting data...
2.264s
fetching data...
0.571s
```

Same code as before, just with different parameters:

```py
ROW_COUNT = 1_000_000
TEXT_LENGTH = 1
```

### System details

* Python 3.11.6 with:
```
psycopg                       3.2.9
psycopg-binary                3.2.9
psycopg2                      2.9.10
```
* PostgreSQL 15.5
* Windows 10

### Afterthought

Having written this, I also tried the new `cursor.copy`. This in practice turned out to be faster than bare `INSERT` in psycopg2, despite having to create an additional temporary table for `RETURNING` etc. (https://stackoverflow.com/questions/37704119/how-to-return-primary-keys-generated-from-a-copy-from-statement-in-postgresql). However, I'm still creating this issue since:

* The first scenario of creating one huge row is still much slower (with `cursor.copy`) than the `INSERT` in psycopg2.
* The slower fetching of identifiers in scenario 2 is still relevant.
* In general, I think the traditional way of building direct SQLs should not be slower in the new version of the library, regardless of the existence of a faster solution.

Possibly related: https://github.com/psycopg/psycopg/discussions/411

jerch:
Should fix #1155.

Changes in c version:
- remove precreation of tuples
- create tuples in-the-fly instead
- change iteration to col --> row access

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/transform.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/transform.pyx
@@ -428,6 +428,11 @@ cdef class Transformer:
         return out
 
     def load_rows(self, int row0, int row1, object make_row) -> list[Row]:
+        if self._load_rows_page_size == 1:
+            return self._load_rows_by_row(row0, row1, make_row)
+        if self._load_rows_page_size == 0:
+            return self._load_rows_by_row_loop(row0, row1, make_row)
+
         if self._pgresult is None:
             raise e.InterfaceError("result not set")
 
@@ -504,6 +509,78 @@ cdef class Transformer:
                 Py_INCREF(record)
                 PyList_SET_ITEM(records, i, record)
                 Py_DECREF(<object>brecord)
+        return records
+
+    def _load_rows_by_row_loop(self, int row0, int row1, object make_row) -> list[Row]:
+
+        cdef int row
+        cdef object record
+        cdef object records = PyList_New(row1 - row0)
+
+        for row in range(row0, row1):
+            record = self.load_row(row, make_row)
+            Py_INCREF(record)
+            PyList_SET_ITEM(records, row - row0, record)
+
+        return records
+
+    def _load_rows_by_row(self, int row0, int row1, object make_row) -> list[Row]:
+        if self._pgresult is None:
+            raise e.InterfaceError("result not set")
+
+        if not (0 <= row0 <= self._ntuples and 0 <= row1 <= self._ntuples):
+            raise e.InterfaceError(
+                f"rows must be included between 0 and {self._ntuples}"
+            )
+
+        if self._load_rows_page_size <= 0:
+            raise ValueError(f"_load_rows_page_size  must be positive")
+
+        cdef libpq.PGresult *res = self._pgresult._pgresult_ptr
+        # cheeky access to the internal PGresult structure
+        cdef pg_result_int *ires = <pg_result_int*>res
+
+        cdef int row, col
+        cdef PGresAttValue *attval
+        cdef object record
+
+        cdef object records = PyList_New(row1 - row0)
+
+        cdef PyObject *loader  # borrowed RowLoader
+
+        row_loaders = self._row_loaders  # avoid an incref/decref per item
+
+        for row in range(row0, row1):
+            # Ownership of 'brecord' transferred to 'records'
+            # so brecord is borrowed.
+            record = PyTuple_New(self._nfields)
+            Py_INCREF(record)
+            PyList_SET_ITEM(records, row - row0, record)
+
+            for col in range(self._nfields):
+                loader = PyList_GET_ITEM(row_loaders, col)
+                attval = &(ires.tuples[row][col])
+                if attval.len == -1:  # NULL_LEN
+                    pyval = None
+                    continue
+                if (<RowLoader>loader).cloader is not None:
+                    pyval = (<RowLoader>loader).cloader.cload(
+                        attval.value, attval.len)
+                else:
+                    b = PyMemoryView_FromObject(
+                        ViewBuffer._from_buffer(
+                            self._pgresult,
+                            <unsigned char *>attval.value, attval.len))
+                    pyval = PyObject_CallFunctionObjArgs(
+                        (<RowLoader>loader).loadfunc, <PyObject *>b, NULL)
+
+                Py_INCREF(pyval)
+                PyTuple_SET_ITEM(record, col, pyval)
+
+            if make_row is not tuple:
+                record = PyObject_CallFunctionObjArgs(make_row, <PyObject *>record, NULL)
+                PyList_SET_ITEM(records, row - row0, record)
+
         return records
 
     def load_row(self, int row, object make_row) -> Row:


jerch:
@dvarrazzo There is a huge portion of the code intersecting with the method below, idk - maybe reshape that to an inline function to avoid code duplication?

dvarrazzo:
Wow, I'm surprising that transposing the row-col order makes such a visible difference. What is the size of the dataset you are testing with?

Note for other readers: in #1155 the OP reported a slowdown on insert (to be verified if really depending on us) and secondarily a slowdown on select, and this MR only addresses the latter.

It seems that the idea of preallocating tuples was implemented in b8304f029d0a5ef3d5ca9f1faa1daa4e61a262a6, where and old version of meself stated that looking up adapters only once was a net win.

```
    More efficient rows loading
    
    Iterate by columns first, so that choices regarding the loader can be
    done only once instead of once per row.
```

Trying to get a visual of the code difference, using 

```
meld \
    <(git show b8304f0:psycopg3_c/psycopg3_c/transform.pyx) \
    <(git show 3.2.10:psycopg_c/psycopg_c/_psycopg/transform.pyx)
```
and cleaning up the diff for trivial refactorings the differences between then and now are down to:

image

Where I mention "visible churn" I probably meant something noticeable in py-spy, which is the tool I was using to peep into the behaviour of the Cython code. That predates even the first 3.0 release, so it could easily be that such lookup got cheaper later and the advantage is lost...

I don't see a lot of difference between then and now which may justify why transposing back the order now is a win. There are just some incref/decref less AFAICS, but maybe that's enough to tip the balance?

Looking at the generated C, the difference around the `pyloader = loader.pyloader` (and similar cloader) is not relevant, in both versions it's just C structures access; the main difference seems coming from `loader = self._row_loaders[col]` which triggered the Python lookup machinery and type checking, and now is just a macro: that's probably what I wanted to push out of the innermost loop.

However it gets me to think that all this is very dependent on the number of cols/rows. Maybe it's worth to get a comparison grid with different dataset size, and test before and after this changeset? What do you think?

dvarrazzo:
> There is a huge portion of the code intersecting with the method below, idk - maybe reshape that to an inline function to avoid code duplication?

Yeah, it occurred to me too that now that the looping order is transposed the `load_rows()` is little more than a repeated `load_row()`.

A refactoring of some common code into a single C function might be useful if we are convinced about the numbers produced by this changeset.

jerch:
> Wow, I'm surprising that transposing the row-col order makes such a visible difference. What is the size of the dataset you are testing with?
It seems that the idea of preallocating tuples was implemented in https://github.com/psycopg/psycopg/commit/b8304f029d0a5ef3d5ca9f1faa1daa4e61a262a6, where and old version of meself stated that looking up adapters only once was a net win.

I started with the simple sequence from the issue report, but when I saw the difference of v2 vs. v3 I suspected, that you might have done the col-row access for a reason, so went with an actual ~temp~ table containing 10 int cols * 100000 rows with different queries:

`SELECT * FROM int_test` (getting 11 cols)
- before (master)
```text
rows: 1000000 repeats: 10
psycopg: 21.538s
psycopg2: 14.432s

rows: 10 repeats: 10000
psycopg: 2.739s
psycopg2: 3.438s
```
- this MR
```text
rows: 1000000 repeats: 10
psycopg: 14.064s
psycopg2: 14.723s

rows: 10 repeats: 10000
psycopg: 3.144s
psycopg2: 3.298s
```
The numbers indicate, that this might be cache related - note that for the lower rowcount the runtime slightly increased (the preselection of the loader shows a benefit, data from pgresult still in cache line). Still for the big table of 11M values this MR seems to perform much better. Here the data is def. not in the cache due to the huge row distances.

testing with selecting just one column - `SELECT i1 FROM int_test`:
- before (master)
```text
rows: 1000000 repeats: 10
psycopg: 6.707s
psycopg2: 3.738s

rows: 10 repeats: 10000
psycopg: 2.224s
psycopg2: 3.945s
```
- this MR
```text
rows: 1000000 repeats: 10
psycopg: 3.590s
psycopg2: 3.706s

rows: 10 repeats: 10000
psycopg: 2.641s
psycopg2: 3.859s
```
Since thise queried just 1 column it basically resembles the issue findings for the big rowcount. For the low rowcount (grabbing only 10 int values) the difference of master vs. this MR is a bit bigger than above.

jerch:
> However it gets me to think that all this is very dependent on the number of cols/rows. Maybe it's worth to get a comparison grid with different dataset size, and test before and after this changeset? What do you think?

Yes I think so (partly did that above). Question is - should this be perf tested more rigorously, like creating special benchmark tests with different table "matrices"? Btw I also played around with `cursor._tx.set_loader_types([23], format)` and switched formats, but the differences stayed the same (binary being slightly faster overall).

Furthermore - to really squeeze max perf out of the method, it could even change its strategy for the data container construction. But for that reliable perf numbers would be needed.

dvarrazzo:
Thank you, these are interesting findings.

So your refactoring is a clear winner for a lot of rows, about 30% faster, but a bit slower for few rows, about 20%. Because psycopg2 changes too, I assume some of this could be noise.I wouldn't be surprise, at this point, if the machine architecture and cache size also affected these results.

At least the number of column doesn't change this finding substantially.

I assume there is a break-even point somewhere between 100 and 1000 rows (with your architecture). What do you think about this plan:

- add your implementation alongside the current one, let's call the old one `_load_rows_by_col()` and the one you propose `_load_rows_by_row()`
- add a module-level, protected variable, `_load_rows_by_row_above` with a default value of 500
- the `load_rows()` function will chose the `_by_row` variant if `col1 - col0 > _load_rows_by_row_above`, otherwise the `_by_col` variant.
- optional: just to be open to the possibility that `_by_row` might be faster on less rows than `_by_col` we could allow `_load_rows_by_row_above` to take a negative value and reverse the check. Like:
```python
def load_rows():
    if _load_rows_by_row_above >= 0:
        if col1 - col0 > _load_rows_by_row_above:
            _load_rows_by_row()
        else:
            _load_rows_by_col()
    else:
        if col1 - col0 > -_load_rows_by_row_above:
            _load_rows_by_col()
        else:
            _load_rows_by_row()
```
- add the script you are using for benchmark to the `test/scripts` dir so that we, or other might experiment with different threshold on different architectures without recompiling.

dvarrazzo:
A different idea - I think I like this more.

My interpretation of these findings is that the by col algorithm is intrinsically faster, but becomes a losing choice when the input/output size kills cache locality.

How about we decide on a page size to work on (let's say 500, and that might be exposed by a protected var too) and we paginate the transformation of the col1 - col0 interval? We might preallocate a list of col1 - col0 elements but limit to build 500 tuples at time, so the per-column work will be amortized.

jerch:
Hmm both interesting ideas, my first shot also would have been your first suggestion.

Your second suggestion sounds even better, but I think I dont fully understand it yet. I'm not sure yet, whether the higher cache pressure comes from the scattered tuple access (like random access) - if so your second idea would help yes.

But there is also the cache on the pgresult with the `ires.tuples[row][col]` lookup. Since libpq has `col` here as 2. index, we again would lose locality by moving `row` too far. But I dont know yet, if the result set plays a minor or a major role in the cache awkwardities here.

dvarrazzo:
My gut feeling about the pagination proposal is that by moving in col we both move quickly over the long list of tuples in the result and in the PGresult memory buffer (which is organised by row, so we move in strides when scanning a col).

So if the PGresult and the Python result are large we give no way for the data to reside in the processor L3 cache, we cache-miss after a few rows, have to load another bunch of tuples or another segment of PGresul buffer, and then _whoosh_ we stride over it. And then we need to do this operation again for the next field.

If we work on a short enough sub-section of the result (both PGresult rows and Python tuples) we might be able to do all the work on them in L3, then loading in cache new data to work on, without the need of returning on data already visited.

If the page is too short we end up having to repeat some per-row overhead, the limit being with a 1-tuple page, meaning working row-by-row. We can reduce that overhead to the 1% if we work in 100-tuples page and probably that's little enough data to fit all in L3.

jerch:
Yes sounds reasonable. Well lets just do it, if the var is easy changeable for tests we might even find a proper sweet spot (although machine dependent).

About the pgresult data organization - do you know, whether pointered values (like string data) reside row-col ordered in the data blob (so basically written as read in the row-col progression / near jump from one string value to the next)? Guess with that we have actually 3 cache portions, that might thrash (python tuples, pgresult array, pgresult pointered data values).

jerch:
New commit with actually 2 versions:
- `load_rows_` with pagination (currently hardcoded to 100) - this is indeed able to find the middle ground
- currently active variant with alloca - this is the fastest in all tests (idk if this is too hacky for your taste, sadly cython does not support VLAs directly)

(Just flip the _ to switch the impls for testing...)

Enough for today :smiley_cat:

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -440,50 +445,56 @@ cdef class Transformer:
...
-                                <unsigned char *>attval.value, attval.len))
-                        pyval = PyObject_CallFunctionObjArgs(
-                            (<RowLoader>loader).loadfunc, <PyObject *>b, NULL)
+        while row0 < row1:
+            # TODO: make advance configurable
+            advance = 100 if row1 - row0 > 100 else row1 - row0
+            for row in range(advance):
+                record = PyTuple_New(self._nfields)
+                Py_INCREF(record)
+                PyList_SET_ITEM(records, row, record)
```
dvarrazzo: `row` always go from 0 to 99 here, it is expected to go from 0 to rowcount - 1.

How about introducing 2 vars: pr0 and pr1 as page row 0 and 1, and so iterate first on the page, then on the page record?

```
for pr0 in range(row0, row1, 100):
    pr1 = min(pr0 + 100, row1)
    for row in range(pr0, pr1):
        ...
```

jerch: Woops, yeah thats a mistake :sweat_smile:

dvarrazzo:
`alloca()` is not quite standard. [On windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170) it is defined in `malloc.h` and it is called `_alloca`. In order to use it we'd need a little snippet of C in a `cdef extern from *`.

I am not entirely opposed to that, but this would be better defined in a module of its own and used in more places than just this, so I'd leave it out for a follow-up MR.

jerch:
Both fixed. Really have to call it a day for today.

Well the `alloca` thingy is the fastest in all of my tests, but I have to check that more in detail - tomorrow, lol.

dvarrazzo:
Another issue with `alloca()` is that people sending (maybe maliciously) a query with enough columns may overflow the stack.

Because the memory allocated is directly controlled by user input it seems a bit too dangerous. Let's leave it out (at last here).

jerch:
> Another issue with `alloca()` is that people sending (maybe maliciously) a query with enough columns may overflow the stack.

Yes that would be the biggest concern with alloca. So if nfields is not contrained by any means, it is a clear no-go imho.

jerch:
I gonna try yet another approach by localizing the paged tuple pointers - not with alloca on the stack, but with a dedicated preallocated piece of memory on the transformer object. Maybe this shows a similar benefit and has the advantage that we can control the needed size for it. (It is a bit like writing your own block allocator...)

dvarrazzo:
> with a dedicated preallocated piece of memory on the transformer object

That would be just the `PyObject**` array, but the tuple created will be in a different part of the heap every time the function is called. I assume that creating all the tuples together, as we are doing now (or just the page's tuples, if that makes a changes) gives some soft guarantee that they will be allocated in the same region and may benefit of the L3 cache for repeated access.

I propose that we try unrelated changes in separate MRs where we can measure their own performances. I would like to isolate only this cache locality issue in this MR so we can check what benefit it gives us independently from other gains we might achieve with unrelated improvements.

jerch:
> That would be just the `PyObject**` array, but the tuple created will be in a different part of the heap every time the function is called.

Yes thats clear. My assumption is, since that trick worked on the loaders (which prolly live in really spreaded memory locations), we might lose cycles on the PyList_GET_ITEM indirection and can benefit for the tuples as well.

> I assume that creating all the tuples together, as we are doing now (or just the page's tuples, if that makes a changes) gives some soft guarantee that they will be allocated in the same region and may benefit of the L3 cache for repeated access.

To some degree maybe. Python has its own pymalloc allocator with different backends, where the default one uses malloc under the hood. Which is known to fragment after some time. Sadly python doesnt offer a bulk creation call (already looked it up), so we cannot really do anything about that (not keen to get into allocator business for python objects nope).

> I propose that we try unrelated changes in separate MRs where we can measure their own performances.

Yes - all in all I think the paginated version is the best bed here currently, only to find a useful general purpose page size is left to be done with the help of rigorous tests. I try not to pollute the MRs too much, still since I am at that weird micro optimization cache level here it is a good opportunity to try different approaches (wont push them though, maybe it is material for later MRs).

_Update:_ Nope, it is actually worse by a tiny amount. Seems the tuple objects are fresh enough to still reside in cache, so the indirection just adds on top.

_On a sidenote:_ Postgres itself seems to limit the max allowed columns in a result to 1664 - https://www.postgresql.org/docs/current/limits.html. Means we would use ~13312 bytes stack space max on a 64bit machine with `alloca`.

_for the record:_ Found an even faster version by extracting the loaders' function pointers prehand. To further benefit from that, this logic prolly would have to go into `_c_get_loader`.

jerch:
@dvarrazzo Ok I gonna try to get a semi-automatic test script up to find the proper page size. With this we can then see, if it varies a lot from different cache sizes, col values etc. and try to find a lower bound for most peeps.

jerch:
@dvarrazzo here a few numbers with different page_size:
```text
v2_select_all_1M: 1.77516 ±0.09748 (cv: 0.055)
v2_select_i1_1000: 0.0024 ±0.00047 (cv: 0.196)
v3_select_all_1M_best 10: 1.73998 ±0.05405 (cv: 0.031)
v3_select_all_1M_best 20: 1.66791 ±0.03443 (cv: 0.021)
v3_select_all_1M_best 40: 1.67214 ±0.03055 (cv: 0.018)
v3_select_all_1M_best 80: 1.69877 ±0.0453 (cv: 0.027)
v3_select_all_1M_best 160: 1.6896 ±0.04118 (cv: 0.024)
v3_select_all_1M_best 320: 1.68812 ±0.02276 (cv: 0.013)
v3_select_all_1M_best 640: 1.75336 ±0.03851 (cv: 0.022)
v3_select_all_1M_best 1280: 1.84953 ±0.04391 (cv: 0.024)
v3_select_i1_1000_best 10: 0.00273 ±0.00052 (cv: 0.189)
v3_select_i1_1000_best 20: 0.00267 ±0.00045 (cv: 0.169)
v3_select_i1_1000_best 40: 0.00241 ±0.00049 (cv: 0.202)
v3_select_i1_1000_best 80: 0.00277 ±0.00037 (cv: 0.135)
v3_select_i1_1000_best 160: 0.00278 ±0.00018 (cv: 0.064)
v3_select_i1_1000_best 320: 0.00284 ±0.00011 (cv: 0.038)
v3_select_i1_1000_best 640: 0.00255 ±0.00081 (cv: 0.318)
v3_select_i1_1000_best 1280: 0.00294 ±9e-05 (cv: 0.031)
v3_select_i10_1000_best 10: 0.00343 ±0.00024 (cv: 0.069)
v3_select_i10_1000_best 20: 0.00339 ±8e-05 (cv: 0.024)
v3_select_i10_1000_best 40: 0.00323 ±0.00016 (cv: 0.05)
v3_select_i10_1000_best 80: 0.0032 ±0.00012 (cv: 0.039)
v3_select_i10_1000_best 160: 0.00317 ±0.00013 (cv: 0.042)
v3_select_i10_1000_best 320: 0.00322 ±0.0005 (cv: 0.155)
v3_select_i10_1000_best 640: 0.00333 ±0.00017 (cv: 0.05)
v3_select_i10_1000_best 1280: 0.00328 ±0.00018 (cv: 0.053)
```
I tested it with [this script](https://gist.github.com/jerch/3236986daade5097bdaf4e3087457f00), which is still work in progress...

It tells a few things already:
- a mass select (many cols, many rows) performs typically faster in v3
- a tiny select (one col, few rows only) is way faster in v2
- from pagesize 320 to 640 the mass select gets way more expensive on my machine
- I dont see a significant difference for small selects within the tested page sizes beside a tiny trend of a speedup for bigger pagesizes (prolly all still within L3)

Imho the fact, that small selects are more expensive in v3 than v2 points to a more expensive machinery setup, not the row/value chewing logic itself.

jerch:
Update - if I change columns to text type and put some longish strings into them, then the cache pressure thingy changes massively (prolly thrashing cache much earlier). I am not sure anymore, if the pagination is a good idea. It loses ground to the direct col-row conversion (the difference is not significant in 1-sigma though, cv is too high for that):

row-col based (pagesize 100)
```
v2_select_all_1M: 1.90508 ±0.08177 (cv: 0.043)
v2_select_i1_1000: 0.01254 ±0.00373 (cv: 0.297)
v2_select_i10_1000: 0.01888 ±0.00553 (cv: 0.293)
v3_select_all_1M: 1.87559 ±0.08822 (cv: 0.047)
v3_select_i1_1000: 0.01268 ±0.00367 (cv: 0.29)
v3_select_i10_1000: 0.01436 ±0.00579 (cv: 0.403)
```

col-row based (still with `alloca`)
```text
v2_select_all_1M: 1.92794 ±0.05644 (cv: 0.029)
v2_select_i1_1000: 0.01333 ±0.00257 (cv: 0.193)
v2_select_i10_1000: 0.01793 ±0.00473 (cv: 0.264)
v3_select_all_1M: 1.82669 ±0.0676 (cv: 0.037)
v3_select_i1_1000: 0.01246 ±0.00223 (cv: 0.179)
v3_select_i10_1000: 0.01351 ±0.00365 (cv: 0.271)
```

Long story short - idk if we can find a proper pagesize at all (beside a very lowish one), as it seems to depend on the actual data in pgresult thrashing the cache much earlier for bigger data portions.

dvarrazzo:
Hello!

I seem to understand that a small page size (~100-200 rows) offers some benefit for small rows/data, no benefit (but also not worse) on wider rows, right?

What is your machine's cache size? `lscpu` should show it.

dvarrazzo:
> Imho the fact, that small selects are more expensive in v3 than v2 points to a more expensive machinery setup, not the row/value chewing logic itself.

Yes, this is something I am aware of. In this MR we are mostly trying to optimise the loading,
Paycopg 3 does comparatively more operation than 2, and most of them on python:

- query manipulation
- prepared statements
- the whole execute/fetch cycle
- not using generators

Rewriting some of them in C or compiling using mypyc might deliver some %, at the price of a lot of work.

For that we should run some complete profiling of small queries and see where time is spent.

jerch:
> I seem to understand that a small page size (~100-200 rows) offers some benefit for small rows/data, no benefit (but also not worse) on wider rows, right?

Basically yep, I still have to check against overly wide row data (like very long string data) and if that changes the picture drastically. (Since idk the memory layout of a pgresult, this might or might not be an issue).

> What is your machine's cache size? `lscpu` should show it.

Ah good to know - L3 is 6 MB here (quite oldish Haswell CPU). Cache timings are here: https://www.7-cpu.com/cpu/Haswell.html

Edit: I start to wonder, if measured the L2-L3 transition instead, as my numbers dont quite add up. Guess I'll have to check the value pointer distances in the pgresult...

jerch:
Had to zoom out to get a better picture and a smaller standard deviation (also lowering setup effects):
```text
v2_select_all_1M: 1.86617 ±0.06401 (cv: 0.034)
v2_select_i1_1M: 0.379 ±0.01518 (cv: 0.04)
v2_select_i10_1M: 0.45752 ±0.02055 (cv: 0.045)
v3_select_all_1M_best 10: 1.77426 ±0.05395 (cv: 0.03)
v3_select_all_1M_best 100: 1.7586 ±0.06167 (cv: 0.035)
v3_select_all_1M_best 1000: 1.89871 ±0.06068 (cv: 0.032)
v3_select_all_1M_best 10000: 2.03368 ±0.02542 (cv: 0.012)
v3_select_all_1M_best 100000: 2.23753 ±0.05076 (cv: 0.023)
v3_select_all_1M_best 1000000: 2.36053 ±0.04331 (cv: 0.018)
v3_select_i1_1M_best 10: 0.36319 ±0.01523 (cv: 0.042)
v3_select_i1_1M_best 100: 0.36742 ±0.01567 (cv: 0.043)
v3_select_i1_1M_best 1000: 0.45727 ±0.01677 (cv: 0.037)
v3_select_i1_1M_best 10000: 0.52696 ±0.01766 (cv: 0.034)
v3_select_i1_1M_best 100000: 0.56654 ±0.0178 (cv: 0.031)
v3_select_i1_1M_best 1000000: 0.59471 ±0.02475 (cv: 0.042)
v3_select_i10_1M_best 10: 0.43293 ±0.01935 (cv: 0.045)
v3_select_i10_1M_best 100: 0.42528 ±0.01719 (cv: 0.04)
v3_select_i10_1M_best 1000: 0.5243 ±0.02176 (cv: 0.042)
v3_select_i10_1M_best 10000: 0.59439 ±0.01394 (cv: 0.023)
v3_select_i10_1M_best 100000: 0.6367 ±0.01691 (cv: 0.027)
v3_select_i10_1M_best 1000000: 0.65462 ±0.02828 (cv: 0.043)
```
(done with integers in TEXT format, where iX denotes the integer chars, thus the more expensive handling of i10 vs i1)

_Findings:_
- optimal pagesize is  <<1000 (likely around 100)
- 10 is pretty close to 100 (no significant differences)
- for >1k things seriously degrade

I am not sure anymore about the L3 story here, to me it looks more like L2 or even L1 (note the very lowish pagesize). Next step is to zoom into 10-1000 again. Then repeat everything with huge string data...

jerch:
There is a clear tendency for pagesize to settle around 50 - 150 (though difference in that range are not significant anymore):
```text
v2_select_all_1M: 1.79261 ±0.05998 (cv: 0.033)
v2_select_i1_1M: 0.36546 ±0.01574 (cv: 0.043)
v2_select_i10_1M: 0.439 ±0.01827 (cv: 0.042)
v3_select_all_1M_best 10: 1.74557 ±0.05397 (cv: 0.031)
v3_select_all_1M_best 20: 1.71996 ±0.04424 (cv: 0.026)
v3_select_all_1M_best 50: 1.70851 ±0.03382 (cv: 0.02)
v3_select_all_1M_best 100: 1.71159 ±0.05105 (cv: 0.03)
v3_select_all_1M_best 150: 1.73996 ±0.05476 (cv: 0.031)
v3_select_all_1M_best 200: 1.72852 ±0.05573 (cv: 0.032)
v3_select_all_1M_best 300: 1.72497 ±0.04507 (cv: 0.026)
v3_select_all_1M_best 400: 1.76487 ±0.04864 (cv: 0.028)
v3_select_all_1M_best 500: 1.73817 ±0.04234 (cv: 0.024)
v3_select_i1_1M_best 10: 0.35726 ±0.01645 (cv: 0.046)
v3_select_i1_1M_best 20: 0.36909 ±0.02054 (cv: 0.056)
v3_select_i1_1M_best 50: 0.3737 ±0.01626 (cv: 0.044)
v3_select_i1_1M_best 100: 0.36398 ±0.01908 (cv: 0.052)
v3_select_i1_1M_best 150: 0.36979 ±0.01729 (cv: 0.047)
v3_select_i1_1M_best 200: 0.37551 ±0.01399 (cv: 0.037)
v3_select_i1_1M_best 300: 0.3876 ±0.01399 (cv: 0.036)
v3_select_i1_1M_best 400: 0.39422 ±0.01647 (cv: 0.042)
v3_select_i1_1M_best 500: 0.40858 ±0.01933 (cv: 0.047)
v3_select_i10_1M_best 10: 0.4331 ±0.01826 (cv: 0.042)
v3_select_i10_1M_best 20: 0.42802 ±0.02203 (cv: 0.051)
v3_select_i10_1M_best 50: 0.42348 ±0.01956 (cv: 0.046)
v3_select_i10_1M_best 100: 0.43535 ±0.01912 (cv: 0.044)
v3_select_i10_1M_best 150: 0.44011 ±0.02212 (cv: 0.05)
v3_select_i10_1M_best 200: 0.43545 ±0.01854 (cv: 0.043)
v3_select_i10_1M_best 300: 0.45063 ±0.01862 (cv: 0.041)
v3_select_i10_1M_best 400: 0.45905 ±0.01864 (cv: 0.041)
v3_select_i10_1M_best 500: 0.4694 ±0.01887 (cv: 0.04)
```
Gonna do a last check against 2^n in that range, maybe there is a weird optimization for those in place :smile_cat:

jerch:
No clear tend here (besides 64 being the fastest in single col select):
```text
v2_select_all_1M: 1.84604 ±0.04524 (cv: 0.025)
v2_select_i1_1M: 0.37298 ±0.01418 (cv: 0.038)
v2_select_i10_1M: 0.45653 ±0.01755 (cv: 0.038)
v3_select_all_1M_best 16: 1.76325 ±0.05997 (cv: 0.034)
v3_select_all_1M_best 32: 1.75823 ±0.04649 (cv: 0.026)
v3_select_all_1M_best 64: 1.79471 ±0.05685 (cv: 0.032)  # micro-benchmark artifact?
v3_select_all_1M_best 128: 1.7745 ±0.05 (cv: 0.028)
v3_select_i1_1M_best 16: 0.36206 ±0.01539 (cv: 0.042)
v3_select_i1_1M_best 32: 0.36478 ±0.01561 (cv: 0.043)
v3_select_i1_1M_best 64: 0.36151 ±0.01733 (cv: 0.048)
v3_select_i1_1M_best 128: 0.36728 ±0.01555 (cv: 0.042)
v3_select_i10_1M_best 16: 0.43498 ±0.01734 (cv: 0.04)
v3_select_i10_1M_best 32: 0.43576 ±0.01417 (cv: 0.033)
v3_select_i10_1M_best 64: 0.43003 ±0.01876 (cv: 0.044)
v3_select_i10_1M_best 128: 0.45032 ±0.0205 (cv: 0.046)
```
On to big value testing...

jerch:
Changed all columns to text and applied a 100-char string to each. Numbers:
```text
v2_select_all_1M: 4.28963 ±0.32921 (cv: 0.077)
v2_select_i1_1M: 0.91463 ±0.04001 (cv: 0.044)
v3_select_all_1M_best 10: 4.38628 ±0.32566 (cv: 0.074)
v3_select_all_1M_best 100: 4.2746 ±0.07208 (cv: 0.017)
v3_select_all_1M_best 1000: 5.01355 ±0.75432 (cv: 0.15)
v3_select_all_1M_best 10000: 5.83632 ±0.39503 (cv: 0.068)
v3_select_all_1M_best 100000: 6.37899 ±1.0821 (cv: 0.17)
v3_select_all_1M_best 1000000: 6.15586 ±0.79565 (cv: 0.129)
v3_select_i1_1M_best 10: 1.26806 ±0.3671 (cv: 0.289)
v3_select_i1_1M_best 100: 1.00461 ±0.19694 (cv: 0.196)
v3_select_i1_1M_best 1000: 1.36772 ±0.36705 (cv: 0.268)
v3_select_i1_1M_best 10000: 1.16946 ±0.18425 (cv: 0.158)
v3_select_i1_1M_best 100000: 1.42481 ±0.3904 (cv: 0.274)
v3_select_i1_1M_best 1000000: 1.35209 ±0.33835 (cv: 0.25)
```
Thats a somewhat funny result, my guess is that the system did something in the background during `v3_select_i1_1M_best 1000`. Overall it again suggests, that pagesize should be <<1000. (The high CVs make the results harder to interpret, prolly memeory pessure from malloc for the strings pay its toll here...)

What is abit bothersome - none of the v2 value comes close to the v2 baseline for 1-col select. (string loader being more expensive in v3?)

Next zoom into 10-1000 again...

jerch:
Zoomed values:
```text
v2_select_all_1M: 4.39172 ±0.2742 (cv: 0.062)
v2_select_i1_1M: 0.93971 ±0.03465 (cv: 0.037)
v3_select_all_1M_best 10: 4.50083 ±0.31198 (cv: 0.069)
v3_select_all_1M_best 50: 4.39303 ±0.08919 (cv: 0.02)
v3_select_all_1M_best 100: 4.84736 ±1.0376 (cv: 0.214)
v3_select_all_1M_best 150: 5.32258 ±1.34698 (cv: 0.253)
v3_select_all_1M_best 200: 4.72765 ±0.66844 (cv: 0.141)
v3_select_all_1M_best 400: 5.5341 ±1.42666 (cv: 0.258)
v3_select_all_1M_best 600: 6.01928 ±1.81246 (cv: 0.301)
v3_select_i1_1M_best 10: 1.07253 ±0.24392 (cv: 0.227)
v3_select_i1_1M_best 50: 1.20459 ±0.34877 (cv: 0.29)
v3_select_i1_1M_best 100: 1.10213 ±0.23386 (cv: 0.212)
v3_select_i1_1M_best 150: 1.12543 ±0.18851 (cv: 0.167)
v3_select_i1_1M_best 200: 1.21449 ±0.33451 (cv: 0.275)
v3_select_i1_1M_best 400: 1.17071 ±0.28822 (cv: 0.246)
v3_select_i1_1M_best 600: 1.13849 ±0.19689 (cv: 0.173)
```
The variance of the values is too high to make an educated guess (cv should stay <<0.1). A mild trend may suggest between 50-150 again.

**In summary:**
For my machine the optimal pagesize is most likely between 50-150. Greater than 500 things degrade pretty fast, while 10-50 is not significantly slower than 50-150, so even 50 might be a good default. The numbers might differ for your machine though, this is all my_machine(TM).

dvarrazzo:
I can try and run your benchmarks on my machine, which I think is similar to yours:
```
Vendor ID:                GenuineIntel
  Model name:             12th Gen Intel(R) Core(TM) i7-1255U
Caches (sum of all):      
  L1d:                    352 KiB (10 instances)
  L1i:                    576 KiB (10 instances)
  L2:                     6.5 MiB (4 instances)
  L3:                     12 MiB (1 instance)
```
as well as on a server that is pretty different
```
Vendor ID:                       AuthenticAMD
Model name:                      AMD EPYC 7542 32-Core Processor
L1d cache:                       64 KiB
L1i cache:                       64 KiB
L2 cache:                        512 KiB
L3 cache:                        16 MiB
```
maybe some other server here and there too.

jerch:
And last but not least, compared to ~col-row~ row-col iteration in v3:
```text
v2_select_all_1M: 4.38393 ±0.29761 (cv: 0.068)
v2_select_i1_1M: 0.94838 ±0.03692 (cv: 0.039)
v3_select_all_1M: 4.26938 ±0.41847 (cv: 0.098)
v3_select_i1_1M: 0.96365 ±0.14228 (cv: 0.148)
```
I know you dont like that version, but it is faster in every aspect and shows lower variance (cv much lower).

Edit: Fixed mixed up iteration.

dvarrazzo:
Sorry, not getting it... what is the col-row iteration? Isn't it the current implementation?

jerch:
Well I can offer these on top:

```text
  Model name:             AMD Ryzen 9 3900 12-Core Processor
...
Caches (sum of all):      
  L1d:                    384 KiB (12 instances)
  L1i:                    384 KiB (12 instances)
  L2:                     6 MiB (12 instances)
  L3:                     64 MiB (4 instances)
```

```text
Architecture:                aarch64
Vendor ID:                   ARM
  Model name:                Neoverse-N1
```
The latter does not list cache sizes, but has L2: 1 MB per Core and L3: 32 MB (its an Ampere Altra 80 Core CPU).

jerch:
> Sorry, not getting it... what is the col-row iteration? Isn't it the current implementation?

Woops typo - I meant row-col descent...

dvarrazzo:
> Woops typo - I meant row-col descent...

I thought we had established that it's slower when we have few rows? [here](https://github.com/psycopg/psycopg/pull/1163#issuecomment-3272193823) you reported 2.7 on master vs. 3.1 per-row, right?

jerch:
Yes that was before `alloca` made it to the stage. With `alloca` it the fastest on every test. I know that we should not go with alloca, but I still hope to work around the stack issue by providing a dedicated piece of heap memory. (Not tested that yet...)

dvarrazzo:
`alloca` stands no chance on an input from the network. The fact that the server has a limit on the columns it returns doesn't mean someone can't craft a malicious payload and shoot a bad value at us. `alloca` use [is discouraged](https://www.man7.org/linux/man-pages/man3/alloca.3.html) in its man page. So I would like to close this topic.

If there is an improvement that turns out positive on any result shape we can make a separate MR for it, put that cat in the bag, and move on back here. Let's limit the scope of this MR: here I would like to establish only the iteration strategy. Our options are:

1. we leave it as it is, col-row
2. we swap it to row-col
3. we make it paginated col-row, with a page size defaulting to 100 and settable from the outside
4. we make it either 1 or 2 according to a threshold settable from the outside.

I understand that 3 is the most promising improvement we can make. 4 feels clumsy. If 2 makes us slower on small dataset it's not an option.

So, if there is any other general improvement to make, let's merge a MR before this, or after this, using whatever is the future current state as baseline, but please let's not make this choice more complex and the implementation more complex than choosing from one of the 4 above.

jerch:
Guess I phrased it wrong - I dont want to go with alloca either, the real danger being that you never know or can check how deeply stacked someone might call a library function (although python does not use the stack the same way as  C/C++, it still does stack allocations). The idea is to provide a bytearray memory segment instead to hold the pointers. Benefit - it would be automatically GCed once done. But I have not tested that yet, so I am unsure about its perf implications.
Ok I wont pursue this idea further, instead cleanup the MR with a default pagesize of 100.

jerch:
MR is ready for review. (Edit: I was not sure, where to put a basic testcase, thus created a new test file. Just move the test elsewhere, if you have a more appropriate file for this particular test...)

tests/test_row_pagination_c.py
```diff
@@ -0,0 +1,16 @@
+import pytest
```
dvarrazzo, line 1: Weird that we don't have a test module for the transformer yet - most tests are in `test_adapt.py`.

However tests are oriented per object, or per functionalities. So we may call this file `test_transformer.py`

tests/test_row_pagination_c.py
```diff
@@ -0,0 +1,16 @@
...
+
+@pytest.mark.parametrize('pagesize', [1, 2, 3, 5, 7])
+def test_pagesize_c(conn, pagesize):
+    from psycopg._cmodule import _psycopg
+    if not _psycopg:
+        return
```
dvarrazzo: The right pattern for this is a decorator like:
```
@pytest.mark.skipif("not psycopg._cmodule._psycopg")
```

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -108,6 +110,7 @@ cdef class Transformer:
 
         self.types = self.formats = None
         self._none_oid = -1
+        self._page_size = 100
```
dvarrazzo: I expect this default to be at module level, so tweaking it may be done once per script, not once per query.

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -441,46 +444,49 @@ cdef class Transformer:
...
-                    attval = &(ires.tuples[row][col])
-                    if attval.len == -1:  # NULL_LEN
-                        pyval = None
-                    else:
-                        pyval = (<RowLoader>loader).cloader.cload(
-                            attval.value, attval.len)
-
-                    Py_INCREF(pyval)
-                    PyTuple_SET_ITEM(<object>brecord, col, pyval)
+        cdef int page_size = self._page_size
```
dvarrazzo: `self._page_size` should be just a struct member so it's probably not worth making it a local var. I'll take a look at the C transformation to confirm that.

tests/test_row_pagination_c.py
```diff
@@ -0,0 +1,16 @@
...
+    cur = conn.cursor()
+    cur.execute("SELECT *, 'abc' FROM generate_series(1, 10)")
+    cur._tx._page_size = pagesize
+    result = cur.fetchall()
+    expected = [
+        (1, 'abc'), (2, 'abc'), (3, 'abc'), (4, 'abc'), (5, 'abc'),
+        (6, 'abc'), (7, 'abc'), (8, 'abc'), (9, 'abc'), (10, 'abc')
+    ]
```
dvarrazzo:
```suggestion
    expected = [(i, "abc") for i in range(1, 11)]
```

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -441,46 +444,49 @@ cdef class Transformer:
...
+        for pr0 in range(row0, row1, page_size):
```
dvarrazzo: We need a comment about why we have chosen this paginated transformation. 
```suggestion
        # In order to convert the PGresult content into a list of tuples,
        # we convert batches of `page_size` records at time, and we convert
        # the records in a page col-by-col.
        # - Iterating rows first is the most natural way of doing it. However
        #   there is a certain overhead in the per-column work that we need to
        #   do for every record.
        # - Iterating cols first allows to minimize the per-col overhead.
        #   however this means scanning both the pgresult buffer and the
        #   allocated tuples several times; for large datasets this has a
        #   measurable detrimental effect in the use of the processor's cache.
        # - Splitting the dataset in small pages and iterating by column allows
        #   better cache locality and amortizes the per-column cost. See #1163.
        for pr0 in range(row0, row1, page_size):
```

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -97,6 +97,8 @@ cdef class Transformer:
 
     cdef dict _oid_types
 
+    cdef public int _page_size
```
dvarrazzo: Also, `page_size` is too much of a generic name: the concept of page comes up everywhere. It maybe be good as a local var name in the `load_rows()` function, but it's already too generic as a `Transformer` member name and even less so as a module-level name.
```suggestion
    cdef public int _load_rows_page_size
```
This name is probably also suitable at module level.

dvarrazzo:
A question about [your test script](https://gist.github.com/jerch/3236986daade5097bdaf4e3087457f00):
```
select_i1 = """SELECT i1 FROM int_testtable LIMIT %(limit)s"""
select_i10 = """SELECT i10 FROM int_testtable LIMIT %(limit)s"""
```
why do you expect that fetching the 1st column or fetching the 10th would make a relevant difference? Or am I reading this wrong?

dvarrazzo:
I have noticed that the MR in this state produces a lot of Pythonish code because the 3 args range is surprisingly not converted in C and because the pr0 and pr1 vars are not annotated. Therefore I'm working with these changes in the main loop:

```diff
diff --git a/psycopg_c/psycopg_c/_psycopg/transform.pyx b/psycopg_c/psycopg_c/_psycopg/transform.pyx
index 0be56b7f..ffdcf4d8 100644
--- a/psycopg_c/psycopg_c/_psycopg/transform.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/transform.pyx
@@ -447,11 +447,13 @@ cdef class Transformer:
 
         cdef PyObject *loader  # borrowed RowLoader
         cdef PyObject *brecord  # borrowed
+
+        cdef int pr0, pr1
         row_loaders = self._row_loaders  # avoid an incref/decref per item
 
-        cdef int page_size = self._page_size
-        for pr0 in range(row0, row1, page_size):
-            pr1 = min(pr0 + page_size, row1)
+        pr0 = row0
+        while pr0 < row1:
+            pr1 = min(pr0 + self._load_rows_page_size, row1)
             for row in range(pr0, pr1):
                 record = PyTuple_New(self._nfields)
                 Py_INCREF(record)
@@ -488,6 +490,8 @@ cdef class Transformer:
                         Py_INCREF(pyval)
                         PyTuple_SET_ITEM(<object>brecord, col, pyval)
 
+            pr0 += self._load_rows_page_size
+
         if make_row is not tuple:
             for i in range(row1 - row0):
                 brecord = PyList_GET_ITEM(records, i)
```

psycopg_c/psycopg_c/_psycopg/transform.pyx
dvarrazzo: Still a segfault:
```
$ pytest tests/test_cursor_common.py::test_rownumber[asyncio-Cursor]
...
Segmentation fault (core dumped)
```

```suggestion
                PyList_SET_ITEM(records, row - row0, record)
```

dvarrazzo:
So...

- I cleaned up the benchmark script to better time only the part we care here (loading rows): [`bench-fetchall.py`](https://gist.github.com/dvarrazzo/c544aa848f30ce7ad9ba3302e0b86cce). For example it was including the query time in the timing (which is the same in all versions, but still adds a bias to the comparison).
- The test uses text fields because numbers are managed as text by pg2 and as binary by pg3 so the latter has an advantage (but small numbers are shorter - but text must be parsed - maybe no clear winner, but an unneeded and unrelated difference).
- I have fixed the issues with the Cython types mentioned above, to produce more streamlined C, and have rebased the MR on 3.2 in the [`fix/rows-load-order-3.2`](https://github.com/psycopg/psycopg/tree/fix/rows-load-order-3.2) branch, in order to compare it with the latest release.
- I have added a simple row-col load function, which kicks in when the page size is set to 1. Just for a test I also added a row-col implementation that just calls `load_row()` in a loop.

tl; dr: the row-col loading algorithm is the winner.

More details: I have tested:
- **psycopg2 2.9.10** compiled on my laptop
- **psycopg[binary] 3.2.10** compiled on my laptop from the git release tag
- **psycopg[c] 3.2.11.dev1** compiled on my laptop from the `fix/rows-load-order-3.2` branch.

In the results here there are tests with 1M rows x 10 cols, 10M rows x 1 col, 1000 rows x 500 cols. The results are independent from the shape.

Commands and results below. The first command tests the latest psycopg 2 and 3 as they were released (pg2 is row-col, pg3 is col-row), the second tests the variations discussed in this PR (p=1: row-col, p=0: row-col as trivial for-loop, p>1: paginated col-row).

- 1M rows x 10 cols:
```
$ PSYCOPG_IMPL=binary python ./scrap/bench-fetchall.py -r 1000000 -c 10  --repeat 5 --relax 0.1
creating table, cols: 10, rows: 1000000
testing psycopg2 2.9.10 (dt dec pq3 ext lo64)
rows: 1000000 cols: 10: 0.53687 ±0.01612 (cv: 0.03, repeats: 5)
testing psycopg[binary] 3.2.10
rows: 1000000 cols: 10: 0.88966 ±0.02251 (cv: 0.025, repeats: 5)

$ for p in 0 1 10 50 100 200 1000; do PSYCOPG_IMPL=c python ./scrap/bench-fetchall.py -r 1000000 -c 10 --repeat 5 -m psycopg -p $p --no-schema --relax 0.1; done
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 0: 0.50602 ±0.01159 (cv: 0.023, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 1: 0.46616 ±0.00396 (cv: 0.008, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 10: 0.51055 ±0.00946 (cv: 0.019, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 50: 0.48619 ±0.00693 (cv: 0.014, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 100: 0.56099 ±0.06536 (cv: 0.117, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 200: 0.62421 ±0.01763 (cv: 0.028, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10 page: 1000: 0.72172 ±0.02362 (cv: 0.033, repeats: 5)
```

- 10M rows x 1 col
```
$ PSYCOPG_IMPL=binary python ./scrap/bench-fetchall.py -r 10000000 -c 1  --repeat 5 --relax 0.1
creating table, cols: 1, rows: 10000000
testing psycopg2 2.9.10 (dt dec pq3 ext lo64)
rows: 10000000 cols: 1: 1.05277 ±0.05457 (cv: 0.052, repeats: 5)
testing psycopg[binary] 3.2.10
rows: 10000000 cols: 1: 2.52348 ±0.20573 (cv: 0.082, repeats: 5)

$ for p in 0 1 10 50 100 200 1000; do PSYCOPG_IMPL=c python ./scrap/bench-fetchall.py -r 10000000 -c 1 --repeat 5 -m psycopg -p $p --no-schema --relax 0.1; done
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 0: 0.96195 ±0.01916 (cv: 0.02, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 1: 0.92507 ±0.18681 (cv: 0.202, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 10: 1.13982 ±0.26683 (cv: 0.234, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 50: 1.11442 ±0.00585 (cv: 0.005, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 100: 1.28656 ±0.0464 (cv: 0.036, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 200: 1.66517 ±0.07202 (cv: 0.043, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 10000000 cols: 1 page: 1000: 4.02077 ±0.14138 (cv: 0.035, repeats: 5)
```

- 1000 rows x 500 cols
```
$ PSYCOPG_IMPL=binary python ./scrap/bench-fetchall.py -r 1000 -c 500  --repeat 5 --relax 0.1
creating table, cols: 500, rows: 1000
testing psycopg2 2.9.10 (dt dec pq3 ext lo64)
rows: 1000 cols: 500: 0.02528 ±0.00207 (cv: 0.082, repeats: 5)
testing psycopg[binary] 3.2.10
rows: 1000 cols: 500: 0.03248 ±0.00048 (cv: 0.015, repeats: 5)

$ for p in 0 1 10 50 100 200 1000; do PSYCOPG_IMPL=c python ./scrap/bench-fetchall.py -r 1000 -c 500 --repeat 5 -m psycopg -p $p --no-schema --relax 0.1; done
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 0: 0.02295 ±0.00168 (cv: 0.073, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 1: 0.02191 ±0.00147 (cv: 0.067, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 10: 0.03225 ±0.00424 (cv: 0.131, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 50: 0.03422 ±0.00215 (cv: 0.063, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 100: 0.03617 ±0.0017 (cv: 0.047, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 200: 0.03544 ±0.00261 (cv: 0.074, repeats: 5)
testing psycopg[c] 3.2.11.dev1
rows: 1000 cols: 500 page: 1000: 0.03277 ±0.00108 (cv: 0.033, repeats: 5)
```

Testing smaller dataset just loses the signal in the noise, results are consistent for every shape tested, so these should do.

Conclusions:

- Psycopg 3 load iteration is indeed slower than psycopg 2 - the regression is confirmed
- The simple row-col algorithm (the `_load_row_by_row()` function in 7d62391b) is consistently faster than psycopg2
- Just calling `load_row()` in a loop has a 10% of overhead compared to the above
- Pagination has some visible effect but - maybe overhead, maybe cache locality - almost never beats the simple row-col iteration, not consistently.

So yes, @jerch: you were right a few messages upthread saying that row-col is faster; the situation was confusing to read for me because of the other optimizations that were ongoing but the picture now is much cleaner.

Action points:

- Discard the paginated version and the load-row-in-a-loop version of the functions, just leave the row-col iteration, renamed as `load_rows()`
- Clean up remaining points of this PR as per review above
- Merge into master and maint 3.2 (there are no conflicts).

@jerch do you agree on these results and conclusion? Do you want me to clean up the MR before merging or do you want to do it yourself (I don't mind either way)?

Thank you very much again!

jerch:
@dvarrazzo Oh wow - sorry was in the weekend holidays, still have to catch up...

I already read your final conclusion, but let me also answer a few things from above first:

> why do you expect that fetching the 1st column or fetching the 10th would make a relevant difference? Or am I reading this wrong?

The i1 vs i10 difference was to test for different values sizes in text format, as i1 was just one char wide and i10 10 chars wide. I did this to see, if the pgresult was an important part of the cache pressure, but got no confirmation from the measured results. Thus I went with 1000-chars wide string values in the last tests, which were totally overshadowed by malloc/GC (my guess), as the runtimes had a huge standard deviation not leading to conclusive results regarding the cache limits I was fighting with (actually tried to find out, whether I was at L2 or L3 limit). So in the end all I had were increasing runtimes with bigger pagesizes with a small benefit in the 50-150 area.

Having only a tiny benefit at lower end of the pagesizes was actually a red flag to me (in the sense - maybe the pagination is not that good of an idea after all?) and the reason why I measured the row-col iterations again, but this was still polluted with my weird alloca trickery, so my numbers for that were still flawed. Sorry for not communicating this more clearly.

About your test cleanup - yeah I thought the same halfway, it is basically not a good idea to derive anything meaningful outside of the `load_rows` if we only try to fix that one for perf reasons. In fact I thought about creating an isolated microbenchmark just for that method alone (pretty much what I would do in any of my own project benchmarks) and throw precrafted data at it for comparison. But then I realised that I have no clue, how to do the same for the psycopg2 counterpart, so I did not go that way.

> tl; dr: the row-col loading algorithm is the winner.
...
So yes, @jerch: you were right a few messages upthread saying that row-col is faster; the situation was confusing to read for me because of the other optimizations that were ongoing but the picture now is much cleaner.

Well that was my suspicion after half of the tests, sorry for not making this clearer above. I got confused here myself by 2 things:
- The very first tests I did were on very small rowcounts (beside the initial testcase from @adamsol) - these were all much worse than in v2, by that time I was not aware of the more expensive machinery setup in v3. Also for some weird reason, the paginated version had here a small benefit over row-col iteration. For bigger rowcounts row-col was already better for any select.
- So I went on to find the CPU cycle eater in the row-col based version to also speed up the small rowcount - and ended up with the alloca trick. This indeed helped alot, but I agree - it is not useable that way. You kinda settled very early on the pagination idea and it felt to me, that you dont want to look at the other idea at all anymore. So I tried to clean that one up, also under the impression of some urge - here we are lol.

> @jerch do you agree on these results and conclusion?

Yes I agree with your conclusions, also with the point higher up, that if there is more room for optimizations, it prolly should be done in a separate MR (the pointered loaders do make things faster on top, but I kinda sidetracked things here with it, if addressed, it also should be done earlier in `_c_get_loader`).

> Do you want me to clean up the MR before merging or do you want to do it yourself (I don't mind either way)?

If you dont mind - plz clean up the MR yourself. This way I can see & learn on a subject I understand well enough now, how you do things here and hopefully can do a better shaped MR next time.

dvarrazzo:
I would like to close this PR in favour of #1187.

This PR has some analysis that might come handy in the future, therefore I would like to keep it as it is. #1187 instead has just the transposition, without all the page fine-tuning faffing.

@jerch I have left your authorship on the commit in #1187. Thank you very much for your help with this improvement! Planning to merge it as soon as tests pass and I've done another last-minute perf paranoia check.

jerch:
@dvarrazzo No worries, feel free to close it and use the cleaned up version. :+1:

The history will still be around to extract some of the sidelined ideas. (Like I think that prelayouting the dump/load functions via function pointers can gain further perf, at least thats what the alloca trick indicates...)

dvarrazzo:
Released as #1187.

PR #1167, opened 2025-09-10
Introduce a way to update connection pool credentials

dsuhinin:
Fix for #851

### DIFF
--- a/docs/api/pool.rst
+++ b/docs/api/pool.rst
@@ -48,9 +48,10 @@ The `!ConnectionPool` class
    arguments have meaningful defaults and can probably be tweaked later, if
    required.
 
-   :param conninfo: The connection string. See
-                    `~psycopg.Connection.connect()` for details.
-   :type conninfo: `!str`
+   :param conninfo: The connection string. See `~psycopg.Connection.connect()`
+                    for details. If it is a callable it will be called at
+                    every connection attempt.
+   :type conninfo: `!str` or `!Callable[[], str]`
 
    :param connection_class: The class of the connections to serve. It should
                             be a `!Connection` subclass.
@@ -58,9 +59,9 @@ The `!ConnectionPool` class
 
    :param kwargs: Extra arguments to pass to `!connect()`. Note that this is
                   *one dict argument* of the pool constructor, which is
-                  expanded as `connect()` keyword parameters.
-
-   :type kwargs: `!dict`
+                  expanded as `connect()` keyword parameters. If it is a
+                  callable it will be called at every connection attempt.
+   :type kwargs: `!dict[str, Any]` or `!Callable[[], dict[str, Any]]`
 
    :param min_size: The minimum number of connection the pool will hold. The
                    pool will actively try to create new connections if some
@@ -181,6 +182,9 @@ The `!ConnectionPool` class
    .. versionchanged:: 3.3
         added `!close_returns` parameter to the constructor.
 
+   .. versionchanged:: 3.3
+        `conninfo` and `kwargs` can be callable.
+
    .. warning::
 
         At the moment, the default value for the `!open` parameter is `!True`;
@@ -284,9 +288,17 @@ Only the functions and parameters with different signature from
 
 .. autoclass:: AsyncConnectionPool
 
+   :param conninfo: The connection string. It can be an async function too.
+   :type conninfo: `!str` or `!Callable[[], str]` or `async Callable[[], str]`
+
    :param connection_class: The class of the connections to serve. It should
                             be an `!AsyncConnection` subclass.
    :type connection_class: `!type`, default: `~psycopg.AsyncConnection`
+
+   :param kwargs: Extra arguments to pass to `!connect()`. It can be an async
+                  function too.
+   :type kwargs: `!dict[str, Any]` or `!Callable[[], dict[str, Any]]`
+                 or `async !Callable[[], dict[str, Any]]`
 
    :param check: A callback to check that a connection is working correctly
                  when obtained by the pool.
@@ -312,6 +324,9 @@ Only the functions and parameters with different signature from
 
    .. versionchanged:: 3.3
         added `!close_returns` parameter to the constructor.
+
+   .. versionchanged:: 3.3
+        `conninfo` and `kwargs` can be callable (sync or async).
 
    .. warning::
 


psycopg_pool/psycopg_pool/pool.py
```diff
@@ -42,6 +43,11 @@ def __init__(
         self,
         conninfo: str = "",
         *,
```
nkraemer2, line 44: Should conninfo be typed as `str | None`?

psycopg_pool/psycopg_pool/pool.py
```diff
@@ -59,6 +65,10 @@ def __init__(
         reconnect_failed: ConnectFailedCB | None = None,
         num_workers: int = 3,
     ):
+        if conninfo is None and get_config is None:
```
nkraemer2: What happens if they are both non-None? I read that get_config takes precedence, but this should probably be documented somewhere and/or raised as an exception

dvarrazzo:
There's also the `kwargs` parameter that will likely contain connection parameters, and it might even be easier to manipulate (because I assume only `password` will change, or something like that).

I would rather modify the signature from:
```
conninfo: str = "",
kwargs: dict[str, Any] | None = None,
```
to
```
conninfo: str | Callable[[], str] = "",
kwargs: dict[str, Any] | Callable[[], dict[str, Any]] | None = None,
```
i.e. conninfo can be either a string or a function returning a string, kwargs  can be either a dict or a function returning a dict. Async pools may handle the async version of these functions too.

dsuhinin:
> str | Callable[[], str] = "",



> There's also the `kwargs` parameter that will likely contain connection parameters, and it might even be easier to manipulate (because I assume only `password` will change, or something like that).
> 
> I would rather modify the signature from:
> 
> ```
> conninfo: str = "",
> kwargs: dict[str, Any] | None = None,
> ```
> 
> to
> 
> ```
> conninfo: str | Callable[[], str] = "",
> kwargs: dict[str, Any] | Callable[[], dict[str, Any]] | None = None,
> ```
> 
> i.e. conninfo can be either a string or a function returning a string, kwargs can be either a dict or a function returning a dict. Async pools may handle the async version of these functions too.

so you mean instead of introducing `get_config` as a separate parameter we can use `conninfo` where we can configure string or callback, right? so we don't need to handle dedicated parameters for that in both `pool` and `async_pool`?

dvarrazzo:
Yes, allowing conninfo and kwargs to be callable would be less ambiguous than having both conninfo+kwargs and get_config. The `tuple[str, dict]` return value of get_config makes it kinda awkward to set up.

Whenever the pool requires a new connection it would call them if they are callables and await on the results if they are awaitable.

Re-reading the original use cases and SO post, most set-ups only need to refresh the password. Applications may configure the pool with a fixed conninfo and a kwargs function returning only a dict containing the `password` key.

For people wanting to replace at the same time conninfo and kwargs they may implement it as an object which would periodically refresh its state and offer two functions, `get_conninfo()` and `get_kwargs()` so that both the params may be returned consistently.

dsuhinin:
> Yes, allowing conninfo and kwargs to be callable would be less ambiguous than having both conninfo+kwargs and get_config. The `tuple[str, dict]` return value of get_config makes it kinda awkward to set up.
> 
> Whenever the pool requires a new connection it would call them if they are callables and await on the results if they are awaitable.
> 
> Re-reading the original use cases and SO post, most set-ups only need to refresh the password. Applications may configure the pool with a fixed conninfo and a kwargs function returning only a dict containing the `password` key.
> 
> For people wanting to replace at the same time conninfo and kwargs they may implement it as an object which would periodically refresh its state and offer two functions, `get_conninfo()` and `get_kwargs()` so that both the params may be returned consistently.

yep and we don't to handle situation when both `conninfo` and `get_config` provided.

dsuhinin:
ok, address PR remarks. Please take a look.

dvarrazzo:
Please note that this MR didn't trigger any test job. I have noticed a few days ago that this is because of the CI configuration

https://github.com/psycopg/psycopg/blob/242ffb9189f250a8f947cc8fe51ca490a181352b/.github/workflows/tests.yml#L4-L8

I understand that `*` doesn't match branches with a `/` in the name. I understand from [here](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#push) that this should be `**` in order to match branch names with a `/`.

Could you please add it in your MR so that the tests run? I'll try to make the change on master too, at which we will rebase this MR, but I can't work on it immediately.

psycopg_pool/psycopg_pool/pool.py
```diff
@@ -204,6 +205,30 @@ def getconn(self, timeout: float | None = None) -> CT:
...
+
+    def _resolve_kwargs(self) -> dict[str, Any]:
+        if callable(self.kwargs):
+            value = self.kwargs()
+            if not isinstance(value, dict):
+                raise TypeError("kwargs callable must return a dict")
+            return value
+        return self.kwargs
+
+    def _getconn_params(self) -> tuple[str, dict[str, Any]]:
```
dvarrazzo: I think that this function doesn't need to exist. We can call the two functions in `_connect()`.

psycopg_pool/psycopg_pool/pool.py
```diff
@@ -204,6 +205,30 @@ def getconn(self, timeout: float | None = None) -> CT:
                 f"couldn't get a connection after {timeout:.2f} sec"
             ) from None
 
+    def _resolve_conninfo(self) -> str:
```
dvarrazzo: This code seems written manually. You should generate it automatically by editing `pool_async.py` file and then run `./tools/async_to_sync.py` (using the `--docker|--podman` option or the Python version expected, which will be printed in a warning if it doesn't match the one you are running the script with).

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -232,6 +233,29 @@ async def getconn(self, timeout: float | None = None) -> ACT:
                 f"couldn't get a connection after {timeout:.2f} sec"
             ) from None
 
+    async def _resolve_conninfo(self) -> Callable | str | Any:
+        """Resolve conninfo (static string, sync callable, or async callable)."""
+        if callable(self.conninfo):
+            if inspect.iscoroutinefunction(self.conninfo):
+                return await self.conninfo()
```
dvarrazzo: Instead of using this pattern you can use `ensure_async()` See `reconnect_failed()` for example.
```suggestion
            return await ensure_async(self.conninfo)
```
This will be transformed to `return self.conninfo()` in the sync version.

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -232,6 +233,29 @@ async def getconn(self, timeout: float | None = None) -> ACT:
...
+        return self.conninfo
+
+    async def _resolve_kwargs(self) -> dict[str, Any]:
+        """Resolve kwargs (static dict, sync callable, or async callable)."""
+        if not self.kwargs:
+            return {}
+
+        if callable(self.kwargs):
+            if inspect.iscoroutinefunction(self.kwargs):
+                return await self.kwargs()
```
dvarrazzo: Same as above:
```suggestion
            return await ensure_async(self.kwargs)
```

tests/pool/test_pool_async.py
```diff
@@ -1077,3 +1078,26 @@ async def close(self) -> None:
...
+        config_rotation_counter += 1
+        return dsn
+
+    async with pool.AsyncConnectionPool(
+        conninfo=rotating_config,
+        min_size=2,
+        max_lifetime=0.2,
+    ) as p:
+        # At pool start, no calls yet
+        assert config_rotation_counter == 0
```
dvarrazzo: This is a race condition, because the pool is opening. For this test you can create the pool closed, with:

```python
p = pool.AsyncConnectionPool(..., open=False)
try:
    assert config_rotation_counter == 0
    await p.open(wait=True)
    assert config_rotation_counter == 2
finally:
    await p.close()
...
```

But I don't think the check for 0 is really meaningful: this is basically testing that the pool doesn't connect if `open=False`, which is something tested elsewere. So I'd just check that after wait the counter is 2:

```suggestion
    ) as p:
        await p.wait()
        assert config_rotation_counter == 2
```

tests/pool/test_pool_async.py
```diff
@@ -1077,3 +1078,26 @@ async def close(self) -> None:
...
+
+        # Let connections expire
+        await asleep(0.3)
+        await p.check()  # trigger pool maintenance
+
+        # Since min_size=2, get_config must have been called twice
+        assert config_rotation_counter == 2, "Connections should have rotated with new config"
```
dvarrazzo: There is a race condition here too: check is creating connections async. We want to use `wait()` to make sure we have 2 conns again:

```suggestion
        await p.wait()
        assert config_rotation_counter == 4
```

psycopg_pool/psycopg_pool/pool.py
dvarrazzo: Nit: in this codebase functions are written, when possible, top-down, from the outer to the innermost.

Please move the definition of these functions after `_connect()`

dvarrazzo:
> Could you please add it in your MR so that the tests run? I'll try to make the change on master too, at which we will rebase this MR, but I can't work on it immediately.

I have pushed a fix for this issue on master, as well as solving other unrelated CI issues. Please rebase this PR and have a lint/test run

psycopg_pool/psycopg_pool/pool.py
dsuhinin: yep, ok. I didn't know. thank you for pointing me.

dsuhinin: of course! thank you.

psycopg_pool/psycopg_pool/pool.py
dsuhinin: done.

psycopg_pool/psycopg_pool/pool.py
dsuhinin: fixed.

psycopg_pool/psycopg_pool/pool_async.py
dsuhinin: done.

tests/pool/test_pool_async.py
dsuhinin: done.

psycopg_pool/psycopg_pool/pool.py
dsuhinin: done.

tests/pool/test_pool_async.py
dsuhinin: done.

dsuhinin:
> > Could you please add it in your MR so that the tests run? I'll try to make the change on master too, at which we will rebase this MR, but I can't work on it immediately.
> 
> I have pushed a fix for this issue on master, as well as solving other unrelated CI issues. Please rebase this PR and have a lint/test run

yep, this is also has been fixed.

dvarrazzo:
Strange that I see all the changes from master in this MR. Maybe you need to rebase the branch?

dsuhinin:
> Strange that I see all the changes from master in this MR. Maybe you need to rebase the branch?

I actually did it. Maybe re create the new PR from latest main will be better?

dsuhinin:
> Strange that I see all the changes from master in this MR. Maybe you need to rebase the branch?

@dvarrazzo what do you think?

dvarrazzo:
It was merged in some strange ways, not rebased. I have rebased and force-pushed here.

Are you able to run the lint/tests in CI?

dsuhinin:
> It was merged in some strange ways, not rebased. I have rebased and force-pushed here.
> 
> Are you able to run the lint/tests in CI?

ook. thank you! How they should be run? automatically? If so, then I still can't see any progress with that.

dvarrazzo:
Ok, trying to run them. I have pushed a commit to enable workflows on pull requests.

Workflows on PR were disabled because they result in running job twice. Maybe I can avoid that by creating PR from a personal repository instead of the project repository moving on.

dsuhinin:
> Ok, trying to run them. I have pushed a commit to enable workflows on pull requests.
> 
> Workflows on PR were disabled because they result in running job twice. Maybe I can avoid that by creating PR from a personal repository instead of the project repository moving on.

thank you! Ill check what is failing right now.

dvarrazzo:
Everything ok here?

I think you should just run:
```
./tools/async_to_sync.py --docker
pre-commit run -a
```
If you installed `./psycopg[dev]` you should have all the dependencies. See README.

If you have problems I cn run the linters my did an push here.

dsuhinin:
> Copied!



> Everything ok here?
> 
> I think you should just run:
> 
> ```
> ./tools/async_to_sync.py --docker
> pre-commit run -a
> ```
> 
> If you installed `./psycopg[dev]` you should have all the dependencies. See README.
> 
> If you have problems I cn run the linters my did an push here.

yes yes. I need to a bid adjust `async_to_sync.py` script so it will correctly remove `async` notations. I feel that it will be ready today. Thank you for asking!

dsuhinin:
> Everything ok here?
> 
> I think you should just run:
> 
> ```
> ./tools/async_to_sync.py --docker
> pre-commit run -a
> ```
> 
> If you installed `./psycopg[dev]` you should have all the dependencies. See README.
> 
> If you have problems I cn run the linters my did an push here.

Hey! @dvarrazzo  It seems like pipeline OK now. Can you please take a look into `async_to_sync.py`. I did minor changes to properly genrated from `async` -> `sync` so please check them.

dsuhinin:
@dvarrazzo let me know if something is needed on my end.

dvarrazzo:
I asked if you could add some tests [a few replies above](https://github.com/psycopg/psycopg/pull/1167#pullrequestreview-3216132144). It seems that the previous tests even shrunk after refactoring: now you are only checking that the connection is called twice, no more count check after reconnection.

dsuhinin:
> I asked if you could add some tests [a few replies above](https://github.com/psycopg/psycopg/pull/1167#pullrequestreview-3216132144). It seems that the previous tests even shrunk after refactoring: now you are only checking that the connection is called twice, no more count check after reconnection.

ahh, sorry. I didn't see it before really. only now, when you pointed a link. Ill try to add related tests and maybe extend current ones.

dsuhinin:
> I asked if you could add some tests [a few replies above](https://github.com/psycopg/psycopg/pull/1167#pullrequestreview-3216132144). It seems that the previous tests even shrunk after refactoring: now you are only checking that the connection is called twice, no more count check after reconnection.

Hey @dvarrazzo I added missing test and a bit improved existing one like you requested. Please take a look when you will have some time. I'm not a Python expert, so my code might not be in the style you prefer.

dsuhinin:
@dvarrazzo did you have a chance to take a look into my last changes?

dvarrazzo:
Hi @dsuhinin,

I have other priorities in these days, from work contracts. I will be able to get back on Psycopg maintenance in the next weeks.

dsuhinin:
> Hi @dsuhinin,
> 
> I have other priorities in these days, from work contracts. I will be able to get back on Psycopg maintenance in the next weeks.

got it! then please let me know.

dsuhinin:
@dvarrazzo did you have chance to take a look?

dvarrazzo:
I need to release first psycopg 3.2.11, then psycopg 3.3.0, then I can look at pool improvements.

- 3.2.11 is important because we fix bug first, then we add features, and there are a few serious issues open.
- 3.3.0 is important because introduces support for Python 3.14 Templates Strings, and possibly some PostgreSQL libpq functions to build OAuth authentication flows.

This changeset is lower priority and my bandwidth is limited. If you are in hurry to have it released you can sponsor the task and I will release it in a day, otherwise it will have to wait.

dsuhinin:
no problem at all. I can wait. thank you!

c-rindi:
Ciao @dvarrazzo.
Happy to say that we appreciate your work and have already been sponsoring you (as G-Research-OSS) since March 2022. @dsuhinin is part of the GR-OSS team. 😄

psycopg_pool/psycopg_pool/base.py
```diff
@@ -44,9 +44,14 @@ class BasePool:
...
-        conninfo: str = "",
+        conninfo: str | Callable[[], str] | Callable[[], Awaitable[str]] | None = None,
         *,
-        kwargs: dict[str, Any] | None,
+        kwargs: (
+            dict[str, Any]
+            | Callable[[], dict[str, Any]]
+            | Callable[[], Awaitable[dict[str, Any]]]
+            | None
+        ) = None,
```
dvarrazzo: These types are a bit verbose. I propose to add the following types to the `psycopg_pool.abc` module:

```python
# Types of the connection parameters
ConninfoParam: TypeAlias = Union[str, Callable[[], str]]
AsyncConninfoParam: TypeAlias = Union[ConninfoParam, Callable[[], Awaitable[str]]]

KwargsParam: TypeAlias = Union[dict[str, Any], Callable[[], dict[str, Any]]]
AsyncKwargsParam: TypeAlias = Union[
    KwargsParam, Callable[[], Awaitable[dict[str, Any]]]
]
```
so these may become:
```python
        conninfo: ConninfoParam | AsyncConninfoParam | None = None,
        *,
        kwargs: KwargsParam | AsyncKwargsParam | None = None,
```

psycopg_pool/psycopg_pool/base.py
```diff
@@ -68,7 +73,12 @@ def __init__(
             raise ValueError("num_workers must be at least 1")
 
         self.conninfo = conninfo
-        self.kwargs: dict[str, Any] = kwargs or {}
+        self.kwargs: (
+            dict[str, Any]
+            | Callable[[], dict[str, Any]]
+            | Callable[[], Awaitable[dict[str, Any]]]
+            | None
+        ) = (kwargs or {})
```
dvarrazzo: It's probably better to just store the `kwargs` from the arguments here, and return the fallback `{}` from `_resolve_kwargs()`

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -648,13 +647,13 @@ async def worker(cls, q: AQueue[MaintenanceTask]) -> None:
     async def _connect(self, timeout: float | None = None) -> ACT:
         """Return a new connection configured for the pool."""
         self._stats[self._CONNECTIONS_NUM] += 1
-        kwargs = self.kwargs
+        conninfo, kwargs = await self._resolve_conninfo(), await self._resolve_kwargs()
```
dvarrazzo: No need to pack and unpack a tuple hee.
```suggestion
        conninfo = await self._resolve_conninfo()
        Kwargs = await self._resolve_kwargs()
```

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -677,6 +676,23 @@ async def _connect(self, timeout: float | None = None) -> ACT:
         self._set_connection_expiry_date(conn)
         return conn
 
+    async def _resolve_conninfo(self) -> str | Any:
```
dvarrazzo: Why returning Any?
```suggestion
    async def _resolve_conninfo(self) -> str:
```
We may declare to return `str | None`, or just to return str and return `self.conninfo or ""` at the end.

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -677,6 +676,23 @@ async def _connect(self, timeout: float | None = None) -> ACT:
...
+        """Resolve conninfo (static string, sync callable, or async callable)."""
+        if callable(self.conninfo):
+            return await ensure_async(self.conninfo)
+
+        return self.conninfo
+
+    async def _resolve_kwargs(self) -> dict[str, Any] | Any:
```
dvarrazzo:
```suggestion
    async def _resolve_kwargs(self) -> dict[str, Any]:
```

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -40,10 +40,10 @@ class AsyncConnectionPool(Generic[ACT], BasePool):
 
     def __init__(
         self,
-        conninfo: str = "",
+        conninfo: str | Callable[[], Awaitable[str]] | None = None,
```
dvarrazzo: In these signatures we are not considering the case of an async connection using a sync conninfo/kwargs function. Using the proposed typedefs this would be supported, and `ensure_async()` will do the right thing.

tests/pool/test_pool_async.py
```diff
@@ -1168,3 +1168,96 @@ async def test_override_close_no_loop_subclass(dsn):
...
+            assert name2[0] in app_names
+
+            # Make sure that names are different.
+            assert name1 != name2
+
+    finally:
+        await p.close()
+
+
+@pytest.mark.slow
```
dvarrazzo, line 1232: Note: this test is slow (2s), so I've marked it as such.

tools/async_to_sync.py
```diff
@@ -347,6 +347,7 @@ class RenameAsyncToSync(ast.NodeTransformer):  # type: ignore
     _skip_imports = {
         "acompat": {"alist", "anext"},
         "_acompat": {"ensure_async"},
+        "typing": {"Awaitable"},
```
dvarrazzo: If we introduce the typedefs we don't need this anymore, but we will need to handle the new renames.

dvarrazzo (changes requested):
I have rebased on master and squashed your epic fight against the linter. Wouldn't have been just easier, less frustrating, more environmental friendly, to follow the README and install the dev extra and pre-commit hooks on you laptop? 😄

The MR requires documentation, including a note in the `news_pool.rst` file.

psycopg_pool/psycopg_pool/pool.py
```diff
@@ -627,6 +627,23 @@ def _connect(self, timeout: float | None = None) -> CT:
...
         return conn
 
+    def _resolve_conninfo(self) -> str | Any:
+        """Resolve conninfo (static string, sync callable, or async callable)."""
+        if callable(self.conninfo):
+            return self.conninfo()
+
+        return self.conninfo
+
+    def _resolve_kwargs(self) -> dict[str, Any] | Any:
```
dvarrazzo: I don't expect Any to be needed. But maybe there is a problem in the sync pool because we sort of declare that these callbacks might be async, whereas the reality is that they can only be sync in the sync pool?

If that's the case maybe we should move the type annotations from the base class to the subclasses.

```suggestion
    def _resolve_conninfo(self) -> str:
        """Resolve conninfo (static string, sync callable, or async callable)."""
        if callable(self.conninfo):
            return self.conninfo()

        return self.conninfo

    def _resolve_kwargs(self) -> dict[str, Any]:
```

dsuhinin: thank you! Ill address everything.

psycopg_pool/psycopg_pool/base.py
dsuhinin: done!

psycopg_pool/psycopg_pool/base.py
dsuhinin: moved out `conninfo` and `kwargs` from base.

psycopg_pool/psycopg_pool/pool_async.py
dsuhinin: done.

psycopg_pool/psycopg_pool/pool_async.py
dsuhinin: you are right. fixed.

tests/pool/test_pool_async.py
dsuhinin: thank you!

psycopg_pool/psycopg_pool/pool_async.py
dsuhinin: should be done.

tools/async_to_sync.py
dsuhinin: yep, remove this stuff because of new typedefs.

psycopg_pool/psycopg_pool/pool.py
dsuhinin: yes, you are right. so I just moved out `coninfo` and `kwargs` from the `base` to avoid such a problem.

dsuhinin:
> I have rebased on master and squashed your epic fight against the linter. Wouldn't have been just easier, less frustrating, more environmental friendly, to follow the README and install the dev extra and pre-commit hooks on you laptop? 😄
> 
> The MR requires documentation, including a note in the `news_pool.rst` file.

yep, configured it already. thank you for pointing me! 

about doc, what I have to do? In `docs` `README.rst` I only can se:
```
Psycopg documentation build
===========================

Quick start::

    make env
    source .venv/bin/activate
    make serve

and connect to <http://localhost:8000>.
```
should I somehow re generate/generate it? or did you mean just add few words into -> `Psycopg 3.3.0 (unreleased)` ?

dsuhinin:
> > I have rebased on master and squashed your epic fight against the linter. Wouldn't have been just easier, less frustrating, more environmental friendly, to follow the README and install the dev extra and pre-commit hooks on you laptop? 😄
> > The MR requires documentation, including a note in the `news_pool.rst` file.
> 
> yep, configured it already. thank you for pointing me!
> 
> about doc, what I have to do? In `docs` `README.rst` I only can se:
> 
> ```
> Psycopg documentation build
> ===========================
> 
> Quick start::
> 
>     make env
>     source .venv/bin/activate
>     make serve
> 
> and connect to <http://localhost:8000>.
> ```
> 
> should I somehow re generate/generate it? or did you mean just add few words into -> `Psycopg 3.3.0 (unreleased)` ?

ah, it seems like yes, just to add manually few words about current changes, right?
Screenshot 2025-10-24 at 18 55 41

dvarrazzo:
> ah, it seems like yes, just to add manually few words about current changes, right?

Yes, just add a line manually to the `news_pool.rst` file. Something like:
```
- Allow `!conninfo` and `!kwargs` to be callable to allow connection
  parameters# update (:ticket:`#851`).
```

psycopg_pool/psycopg_pool/abc.py
```diff
@@ -34,3 +32,12 @@
...
 ]
+
+# Types of the connection parameters
+ConninfoParam: TypeAlias = Union[str, Callable[[], str]]
+AsyncConninfoParam: TypeAlias = Union[str, Callable[[], Awaitable[str]]]
+
+KwargsParam: TypeAlias = Union[dict[str, Any], Callable[[], dict[str, Any]]]
+AsyncKwargsParam: TypeAlias = Union[
+    dict[str, Any], Callable[[], Awaitable[dict[str, Any]]]
+]
```
dvarrazzo, line 48: These types are not covering the case that an async pool might still use a non-async function to get their credentials. In a previous suggestion I suggested to make the async version to be the union of the sync version and an async function. The types might also be defined this way:

```suggestion
ConninfoParam: TypeAlias = Union[str, Callable[[], str]]
AsyncConninfoParam: TypeAlias = Union[
    str,
    Callable[[], str],
    Callable[[], Awaitable[str]],
]

KwargsParam: TypeAlias = Union[dict[str, Any], Callable[[], dict[str, Any]]]
AsyncKwargsParam: TypeAlias = Union[
    dict[str, Any],
    Callable[[], dict[str, Any]],
    Callable[[], Awaitable[dict[str, Any]]],
]
```

Note: you end up with importing both the sync and the async definitions in the async pool module but you need only the sync ones in the sync pool module. If renaming doesn't work as expected and you end up with duplicate imports that isort doesn't fix you may guard the import with something like:

```python
from .abc import ConninfoParam, KwargsParam
if True:  # ASYNC
    from .abc import AsyncConninfoParam, AsyncKwargsParam
```
the `if` statement will be dropped in the sync module.

psycopg_pool/psycopg_pool/null_pool_async.py
```diff
@@ -7,12 +7,18 @@
...
 from psycopg.pq import TransactionStatus
 
-from .abc import ACT, AsyncConnectFailedCB, AsyncConnectionCB
+from .abc import (
+    ACT,
+    AsyncConnectFailedCB,
+    AsyncConnectionCB,
+    AsyncConninfoParam,
+    AsyncKwargsParam,
+)
```
dvarrazzo: nit: Uhm... hasn't isort fixed/complained about this import style? It should be more like:

```suggestion
from .abc import ACT, AsyncConnectFailedCB, AsyncConnectionCB, AsyncConninfoParam
from .abc import AsyncKwargsParam
```

psycopg_pool/psycopg_pool/null_pool_async.py
```diff
@@ -23,12 +29,13 @@
 
 
 class AsyncNullConnectionPool(_BaseNullConnectionPool, AsyncConnectionPool[ACT]):
+
     def __init__(
         self,
-        conninfo: str = "",
+        conninfo: AsyncConninfoParam | None = None,
```
dvarrazzo: Any specific reason to change the default from `""` to `None`? I don't think it causes any side effect (both are supported by the connection) but maybe it's a change we can avoid doing.

The default for `kwargs` need to be `None` because `{}` is mutable, but for the string it is not needed.

```suggestion
        conninfo: AsyncConninfoParam = "",
```

psycopg_pool/psycopg_pool/pool_async.py
```diff
@@ -70,7 +76,8 @@ def __init__(
                     " https://www.psycopg.org/psycopg3/docs/advanced/pool.html"
                     "#pool-sqlalchemy for a workaround."
                 )
-
+        self.conninfo = conninfo
+        self.kwargs: AsyncKwargsParam | None = kwargs
```
dvarrazzo: I don't think this type definition is necessary as it's the same in the function argument.

```suggestion
        self.kwargs = kwargs
```

psycopg_pool/psycopg_pool/_acompat.py
```diff
@@ -174,6 +174,12 @@ async def ensure_async(
 async def ensure_async(f: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T: ...
 
 
+@overload
+def ensure_async(
+    f: Callable[P, Awaitable[T]], *args: P.args, **kwargs: P.kwargs
+) -> Coroutine[Any, Any, T]: ...
+
+
```
dvarrazzo: Is this new definition needed?

dvarrazzo (changes requested):

psycopg_pool/psycopg_pool/_acompat.py
dsuhinin: Yes, otherwise the linter complains.

psycopg_pool/psycopg_pool/null_pool_async.py
dsuhinin: nope. this is what `isort` did probably and pipeline is green now.

psycopg_pool/psycopg_pool/abc.py
dsuhinin: Ill take a look. thank you!

dsuhinin: emmm, it seems like not. in async pool I only import:
```
AsyncConninfoParam,
    AsyncKwargsParam,
```
and in sync pool:
```
ConninfoParam, KwargsParam
```
maybe I misunderstood you?

psycopg_pool/psycopg_pool/null_pool_async.py
dsuhinin: I double checked and this is what really isort did for me.

psycopg_pool/psycopg_pool/null_pool_async.py
dsuhinin: yep. you are right. will be fixed.

psycopg_pool/psycopg_pool/pool_async.py
dsuhinin: yep. fixed.

dsuhinin:
> > ah, it seems like yes, just to add manually few words about current changes, right?
> 
> Yes, just add a line manually to the `news_pool.rst` file. Something like:
> 
> ```
> - Allow `!conninfo` and `!kwargs` to be callable to allow connection
>   parameters# update (:ticket:`#851`).
> ```

yep, added few lines. hope it is correct and would work. thank you.

psycopg_pool/psycopg_pool/_acompat.py
dvarrazzo: I found the issue in the definition of
```python
@overload
async def ensure_async(
    f: Callable[P, Coroutine[Any, Any, T]], *args: P.args, **kwargs: P.kwargs
) -> T: ...
```
which doesn't match the return value of `Awaitable[T]`, which I understand it is a superset.

The definition you proposed, `def ensure_async(...) -> Coroutine[T]`, is equivalent of `async def ensure_async(...) -> T` - but I see it is what mypy output suggests.

Therefore I have fixed the first definition instead of introducing a third (8e5ce9ae2e7f9267e77ccd9cd2cf1bbcd7dbb6ad).

psycopg_pool/psycopg_pool/null_pool_async.py
dvarrazzo: I have seen that isort is stable on the multi-line import, yes. However I've tried to reformat all one one line and re-running isort, and now it is stable on the proper style. So I have commited it that way.

psycopg_pool/psycopg_pool/abc.py
dvarrazzo: No, you are right, the sync part is embedded in the async definition. Sorry for the noise.

dvarrazzo:
I have squashed the history and rebased on current master.

I have also mentioned the params as callable in the docs api: see 02d28597f99d5f08756d4c7451ddb39e9684b6a9

There are two pages to document the pool: [the API page](psycopg.org/psycopg3/docs/api/pool.html) and [a discursive page](https://www.psycopg.org/psycopg3/docs/advanced/pool.html). I have only updated the former. Is there any documentation you would like to suggest for the discursive page? A meaningful example of integration with some cloud provider to dynamically update the password.

dsuhinin:
> I have squashed the history and rebased on current master.
> 
> I have also mentioned the params as callable in the docs api: see [02d2859](https://github.com/psycopg/psycopg/commit/02d28597f99d5f08756d4c7451ddb39e9684b6a9)
> 
> There are two pages to document the pool: [the API page](psycopg.org/psycopg3/docs/api/pool.html) and [a discursive page](https://www.psycopg.org/psycopg3/docs/advanced/pool.html). I have only updated the former. Is there any documentation you would like to suggest for the discursive page? A meaningful example of integration with some cloud provider to dynamically update the password.

I don't have a particular example really. @nkraemer2 maybe you want to add something?

dsuhinin:
@dvarrazzo thank you for your review. Please let me know if something else needed or we are good to go?

dvarrazzo:
Will merge soon, thank you very much. This will be released with psycopg-pool 3.3, happening in the next few weeks.

If there's nothing you want to add to the documentation that's ok: if there's any idea for the docs it can be added later.

Thank you very much!

JAC2703:
This implementation solves my issue beautifully. @dvarrazzo do you have an approx. target date for releasing 3.3?

Issue #545, opened 2023-04-07
 Binary protocol transform failed in load_sequence for query returning both RecordArray and Record 


exrok:
Getting `psycopg.ProgrammingError: cannot load sequence of 1 items: 2 loaders registered`
when using the binary protocol with query returning both RecordArray and Record.

See minimal reproduction and error traceback logs included below. 

## Observations
- Changing the query to either of the following resolves the problem:
  - `SELECT ARRAY[ROW('e656b0f5-fa8e-4cbe-98db-525f9ab1cacc'::uuid, 32)]`
  - `SELECT ROW('abc')`
- The bug seems to exist in both `psycopg_binary` and the native python implementation.
- The rust crate `tokio-postgres` doesn't have this issue using the binary protocol, with the same setup.
- The problem only occurs with the binary protocol.

My hypothesis is that it is using the same type info from the first `Record` to parse the second `Record`.

## Env
- Python: v3.10
- OS: Arch Linux
- DB: Postgres 15.2
- psycopg 3.1.8 both with and without `binary` extra
- Don't know if it is used but I have `libpq.so.5.15` on my system.

## Minimal Reproduction

Running Postgres v15.2
```sh
docker run -d -p 5432:5432 --name testdb -e POSTGRES_PASSWORD=postgres postgres:15.2
```

Python script to trigger failure.
```python
import psycopg

with psycopg.connect("host=localhost user=postgres password=postgres") as conn:
    conn.execute(
        "SELECT ARRAY[ROW('e656b0f5-fa8e-4cbe-98db-525f9ab1cacc'::uuid, 32)], ROW('abc')",
        binary=True,
    ).fetchall()
```

## Error Traceback Logs
<details>
<summary>
Traceback when using only <code>psycopg[binary,pool]</code> version 3.1.8
</summary>

```python
Traceback (most recent call last):
  File "/home/user/code/bug/src/repro.py", line 22, in <module>
    rows = cursor.fetchall()
  File "/home/user/code/bug/.venv/lib/python3.10/site-packages/psycopg/cursor.py", line 851, in fetchall
    records = self._tx.load_rows(self._pos, self.pgresult.ntuples, self._make_row)
  File "psycopg_binary/_psycopg/transform.pyx", line 481, in psycopg_binary._psycopg.Transformer.load_rows
  File "/home/user/code/bug/.venv/lib/python3.10/site-packages/psycopg/types/composite.py", line 159, in load
    return self._tx.load_sequence(
[...]

PR #1175, opened 2025-10-04
Implement RecordBinaryLoader in Cython

martinlehoux:
Follow-up of https://github.com/psycopg/psycopg/discussions/1174 

I can also provide the HTML output so that people more used to it can help ^^

[composite.html](https://github.com/user-attachments/files/22696030/composite.html)

A few questions:

1. Is there a difference between `Py_ssize_t` and `size_t`
2. And between `object` and `PyObject` or `PyObject *`
3. I took the `raise e.DataError` from `copy.pyx`, but the Python version doesn't have this check, is it the kind of checks specific to bound checking in C?

## Benchmark

I don't know if there are standard benchmarks for this project, but I use a very simple one (see below).

The baseline was `0.3467s ± 0.0034s` on my computer, and the updated version is `0.2410s ± 0.0091s`.

```py
def benchmark_binary_load():
    with psycopg.connect(CONNINFO) as conn, conn.cursor(binary=True) as cur:
        cur.execute(
            f"""
                SELECT
                    i,
                    (
                        i::integer,
                        (i * 1000)::bigint,
                        (i * 3.14159)::double precision,
                        'row_' || i::text,
                        '2025-06-06T14:25:00'::timestamp,
                        (i % 2 = 0)::boolean,
                        i * '3.00'::numeric
                    ) as record
                FROM generate_series(1, {N_RECORDS}) i
            """.encode()
        )

        records = cur.fetchall()
        assert records[0][1][0] == 1
        assert records[0][1][1] == 1000
        assert records[0][1][2] == 3.14159
        assert records[0][1][3] == "row_1"
        assert records[0][1][4] == datetime(2025, 6, 6, 14, 25, 0)
        assert records[0][1][5] is False
        assert records[0][1][6] == Decimal("3.00")
        return len(records)
```


dvarrazzo:
Thank you @martinlehoux, great result!

I only gave a cursor look at the code: it seems fundamentally ok. One thing that we can save is:
```
cdef tuple key = PyList_AsTuple(oids)
```

Unlike in Python, it is possible to set items on a tuple. Therefore you can just create `oids = PyTuple_New(nfields)` and use `PyTuple_SET_ITEM()` to set the elements and use it both in `set_loader_types()` and as key for `self._txs`.

Another thing that doesn't convince me is that the class is declared `@cython.final`. This class is meant to be subclassed:

https://github.com/psycopg/psycopg/blob/cd6d50bc8cbbe424000e5a4af728c336c5d01c87/psycopg/psycopg/types/composite.py#L264

What I think is happening here is that this module is not actually making use of the C version, because using the optimized adapters only happens in the adapters map, not at module level:

https://github.com/psycopg/psycopg/blob/cd6d50bc8cbbe424000e5a4af728c336c5d01c87/psycopg/psycopg/_adapters_map.py#L277-L287

When we generate a type-specific loader we should find a way to use the Cython class, here:

https://github.com/psycopg/psycopg/blob/cd6d50bc8cbbe424000e5a4af728c336c5d01c87/psycopg/psycopg/types/composite.py#L362-L368

---

_mumble mumble_

---

I gave it a thought: I think that there is something we can simplify here and make your loader even more efficient.

Currently `register_loader()` subclasses the thin subclass `CompositeBinaryLoader`, which just adds a `factory` member to `RecordBinaryLoader`:
```
  RecordBinaryLoader
         ^
         |
CompositeBinaryLoader    # adds the `.factory` member
         ^
         |
CustomTypeBinaryLoader   # created by `_make_binary_loader()` in `register_composite()`
```

The `_txs` map is required to solve #545 to allow to load different types of `record` fields in the same query, but this issue only exists when `RecordBinaryLoader` is used directly. When we use `CustomTypeBinaryLoader` we are not in the case of the same loader asked to load records of different types - all the records will have the same types - so `_txs` will always contain exactly one item.

So I think we should refactor and separate the two:
```
                      _CRecursiveLoader
                             ^
                             |
RecordBinaryLoader  CompositeBinaryLoader
                             ^
                             |
                   CustomTypeBinaryLoader
```
The differences are:

- `RecordBinaryLoader` is not meant to be subclassed, it only loads the `record` type.
- `CompositeBinaryLoader` is not meant to be used as it is: it must be subclassed and a factory specified.
- `RecordBinaryLoader` must manage a map of transformers.
- `CompositeBinaryLoader` must manage a single transformer. So it can inherit from `_CRecursiveLoader`, which is precisely a loader subclass managing an internal transformer (currently used by the arrays loader in `array.pyx`).

I assume we can provide the C version of both `RecordBinaryLoader` and `CompositeBinaryLoader`; in `register_composite()` we should look up dynamically if the C module is available, like:

```python
from ._cmodule import _psycopg

def _make_binary_loader(name: str, factory: Callable[..., Any]) -> type[abc.Loader]:
    base = _psycopg.CompositeBinaryLoader if _psycopg else CompositeBinaryLoader
    return type(f"{name.title()}BinaryLoader", (base,), {"factory": factory})
```

---

I am not sure if everything is clear here: this is quite a deep dive. I can try and give some help here (for example by making the refactoring above in Python and then move to the C optimization).

In terms of timeline, this is probably too invasive for the next 3.2.x bugfix and I'd be happy to consider it for 3.3.

I am currently finishing some other work, after which (next week) I would like to release 3.2.11 soon because there are a few serious bugfixes to address, and then move quickly to the last features to wrap 3.3.0 (ideally in October).

---

About your questions:

> Is there a difference between Py_ssize_t and size_t

I understnad that `size_t` used to not be defined in a standard way and `Py_ssize_t` solves the problem. See [PEP 0353](https://peps.python.org/pep-0353/).

> And between object and PyObject or PyObject *

`object` has refcount semantics for Cython, `PyObject *` can be used to create borrowed pointers.

> I took the raise `e.DataError` from copy.pyx, but the Python version doesn't have this check, is it the kind of checks specific to bound checking in C?

There could be some inconsistencies between C and Python in those functions, there are a few bugs open scheduled to be addressed in 3.2.11 (#1150 and likely others). I'll check if there's anything to clean up.

Thank you very much!

martinlehoux:
OK, let me rephrase this:
1. A single `RecordLoader` is created for parsing a query. It can encounter several Record oids sequence, and caches the transformer required for each sequence in `self._txs`
2. This abstraction is too much for the `CompositeLoader` that only manage a single type of record, described ahead of time (but it seems there is still an oid sequence, because the `load` is used as is): we pay the cache lookup for each record when it should be a simple member

martinlehoux:
So yes, I guess I can work on splitting the two implementations in Python before continuing the C versions. Are the two TEXT version subject to the same idea? It looks different: `CompositeLoader -> RecordLoader -> BaseCompositeLoader`

dvarrazzo:
> OK, let me rephrase this: [...]

Yes, your rephrasing is spot-on correct.

The text side is currently different because the record text loader has no oid information. We receive something like `b"(10,hello)"` and no further information of what to do with it. So, for the `record` type, we only convert the record fields as strings and return the Python object `("10", "hello")`; the binary loader knows per-fields oids instead. When we create a specific composite class we also have OID information from the types registry, so we can make a more precise cast.

So in text format the classes roles are:

```
BaseCompositeLoader  # Has a _tx but doesn't use it directly.
         ^           # Implements `_parse_record()`, used by subclasses.
         |
   RecordLoader      # Parses the record and convert resulting fields as text
         ^
         |
  CompositeLoader    # Abstract class, calling `_parse_record()` and using
         ^           # `field_types` and `factory` (not specified here)
         |
 CustomTypeLoader    # Dynamically created by `_make_loader()`,
                     # adds `field_types` and `factory`
```

So here `RecordLoader` doesn't need the `_txs` because it only need one Transformer (from which it only uses the `TextLoader`).

I have prepared [a refactoring branch](https://github.com/dvarrazzo/psycopg/tree/composite-refactoring) with the following changes:

- Avoid a common base class to reuse the code, use module-level functions.
- Make the inheritance lines of text and binary loaders similar.
- Add documentation and notes.

Maybe this could be a better-looking base to convert the code to Cython. I think it would be a good idea to rebase your branch onto it and to use the Python code as guideline for Cython code. Now it should be clearer and better documented.

martinlehoux:
Just a quick status update: I have not forgotten about this, and am working at my pace. I'm currently working on the Record text parsing, that feels like more than translation because the Python version uses regex. I am taking inspiration from `_array_load_text` but it doesn't feel like exactly the same rules apply

dvarrazzo:
Thank you for the heads-up, @martinlehoux! Take your time: this improvement is welcome but not urgent.

> I am taking inspiration from _array_load_text but it doesn't feel like exactly the same rules apply

Rules are slightly different yes. For [arrays](https://www.postgresql.org/docs/current/arrays.html#ARRAYS-IO):

> The array output routine will put double quotes around element values if they are empty strings, contain curly braces, delimiter characters, double quotes, backslashes, or white space, or match the word NULL. Double quotes and backslashes embedded in element values will be backslash-escaped. For numeric data types it is safe to assume that double quotes will never appear, but for textual data types one should be prepared to cope with either the presence or absence of quotes.

In arrays the delimiter can be different for different types (It's usually `,`, it is `;` for a few geometric types, and in general it is a property of the type).

For [composite](https://www.postgresql.org/docs/current/rowtypes.html#ROWTYPES-IO-SYNTAX):

> The composite output routine will put double quotes around field values if they are empty strings or contain parentheses, commas, double quotes, backslashes, or white space. (Doing so for white space is not essential, but aids legibility.) Double quotes and backslashes embedded in field values will be doubled.

So they are somewhat similar but annoyingly different. Arrays use escaping to represent chars in quotes, composite double the chars:

```
=# select ARRAY['"', '\'] as array, ROW('"', '\') as composite;
┌─────────────┬─────────────┐
│    array    │  composite  │
├─────────────┼─────────────┤
│ {"\"","\\"} │ ("""","\\") │
└─────────────┴─────────────┘
```

In arrays, NULLs are written as the literal `NULL`, so the string `NULL` must be quoted as `"NULL"`. In composites NULLs are represented by an empty string, so the empty string must be quoted as `""`.

```
=# select ARRAY['null', null, ''] as array, ROW('null', null, '') as composite;
┌──────────────────┬────────────┐
│      array       │ composite  │
├──────────────────┼────────────┤
│ {"null",NULL,""} │ (null,,"") │
└──────────────────┴────────────┘
```

I think that `_parse_token()` in `array.pyx` is a good starting point anyway: you should mostly drop the special case for NULL and the delimiter, account for terminators to be `()` instead of `{}`, and undouble only `\` and `"`, instead of unescaping everything that is escaped.

The unit test should contain tests for every possible character, so don't worry: you will be scolded if something is not right!

martinlehoux:
Thanks for the advice, it was a great help to getting this to a working status!
1. On a practical topic, how do you intend to manage your branch and mine? For now your refactor is included in this PR
2. I have a few code/performance points that may need improvement:
	1. `RecordLoader.cload` has a few Python hotspots around the cast using TEXT_OID loader. I feel there is room for better there
	2. A specific question: I see `record.append(xxx)` is compiled to `__Pyx_PyList_Append`, so does it need specific Cython calls or is that ok?
3. Otherwise I think the next step for me is to work on Composite

dvarrazzo:
Hello @martinlehoux 

I have no intention to maintain my branch further, it was only meant to give you a (hopefully) cleaner starting point. If the changeset it contain is now integrated in your branch my branch can be closed.

> RecordLoader.cload has a few Python hotspots around the cast using TEXT_OID loader. I feel there is room for better there

My suggestion is to use [`self._tx._c_get_loader()`](https://github.com/psycopg/psycopg/blob/01779fd9a22427f20839d35b4597ddb481c3a8f1/psycopg_c/psycopg_c/_psycopg/transform.pyx#L568) to get an object that is of type [`RowLoader`](https://github.com/psycopg/psycopg/blob/01779fd9a22427f20839d35b4597ddb481c3a8f1/psycopg_c/psycopg_c/_psycopg/transform.pyx#L47). This is a C path to get a structure that contains either a `CLoader` or a Python `Loader`, according to whether the loader for given oid and type is implemented in C or in Python.

Note that you can do this in the constructor of the `RecordLoader`, because there is a guarantee that these object are instantiated once per query and the loaders map won't change in the meantime. 

Once you have the `RowLoader` you can convert the buffer into a Python object checking whether `cloader` is None or not, see [this pattern](https://github.com/psycopg/psycopg/blob/01779fd9a22427f20839d35b4597ddb481c3a8f1/psycopg_c/psycopg_c/_psycopg/transform.pyx#L460-L467) for example.

Because most likely the Text loader is implemented in C you should be able to do everything without leaving the C layer (but we should test that we do the right thing if a Python loader is specified).

> A specific question: I see record.append(xxx) is compiled to __Pyx_PyList_Append, so does it need specific Cython calls or is that ok?

I personally tend to use the Python API function when I know what I am calling, so [`PyList_Append`](https://docs.python.org/3/c-api/list.html#c.PyList_Append).

psycopg_c/psycopg_c/types/composite.pyx
```diff
@@ -0,0 +1,169 @@
...
+        for i in range(len(record)):
+            if (f := record[i]) is not None:
+                value = cast(f)
+                Py_DECREF(f)
+                Py_INCREF(value)
+                PyTuple_SET_ITEM(record, i, value)
+        return record
+
+    def __dealloc__(self):
+        PyMem_Free(self.scratch)
```
dvarrazzo, line 38: Is `scratch` definitely set to NULL on creation?

psycopg_c/psycopg_c/types/composite.pyx
```diff
@@ -0,0 +1,169 @@
...
+
+    while buf < data + length:
+        record.append(_parse_record_token(&buf, data + length, scratch, sclen))
+        if buf[0] == b',':
+            buf += 1
+            if buf == data + length:
+                record.append(None)
+
+
+    return record
```
dvarrazzo, line 53: Here we are creating a list, appending objects, then converting it to a tuple.

I think it's worth to pre-scan the buffer and count the number of items, then directly allocate a tuple of the right size with `PyTuple_New(n)` and set the items there. In pre-scanning:

- nfields = 0
- the buffer must start with `(`
- if the next char is `"`, scan forward until you find another `"` not followed by a second `"`, then there must be `,` or `)`; nfields += 1
- otherwise scan forward until a `,` or `)`; nfields += 1
- `)` must only be at the end.
- return nfields

So it's similar to the loop in `_parse_record_token()`, but simpler because you don't need to unescape.

psycopg_c/psycopg_c/types/composite.pyx
```diff
@@ -0,0 +1,169 @@
...
+    cdef int num_escapes = 0
+
+    if has_quotes:
+        start += 1
+    cdef char *end = start
+
+    while end < bufend:
+        if (end[0] == b',' or end[0] == b')') and not quoted:
+            break
+        elif end +1 < bufend and ((end[0] == b'\\' and end[1] == b'\\') or (end[0] == b'"' and end[1] == b'"')):
```
dvarrazzo: Shorter expression, pruning earlier most of the false cases:
```suggestion
        elif (end[0] == b'\\' or end[0] == b'"') and end + 1 < bufend and end[1] == end[0]:
```
Might be also worth storing `const char *bufend1 = bufend - 1` before the loop, so you can compare `end < bufend1` in the loop.

dvarrazzo (changes requested):
Just quickly reviewed a few things I spotted. I can make a further review after the changes regarding finding the text loader.

psycopg_c/psycopg_c/types/composite.pyx
martinlehoux: I must say I don't know, I did not do something to ensure it. I took the pattern from array.pyx and I can't find an initialisation there

dvarrazzo:
Heads up: in #932 and in #1202 we have made some changes to the composite loaders and dumpers to allow [adapting generic objects](https://www.psycopg.org/psycopg3/docs/basic/pgtypes.html#composite-types-adaptation), not only sequences. The new features introduced (exposed by the `make_sequence` and `make_instance` parameters of `register_composite()`) will probably affect this PR too.

One of the changes included in the new features is the inheritance refactoring originally proposed here. That change is now on master, committed as 49575b8383daa5ce28908d2db5a584ce60aa7a87, and I have rebased this PR, therefore it included even if it doesn't figure as part of this PR anymore.

Another change made on master in the past days is the introduction of a Cython linter, therefore I have edited the changes in this PR to adhere to the linter style.

psycopg_c/psycopg_c/types/composite.pyx
dvarrazzo: I found it back: yes, [it is safe](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init):

>  Note that the object allocation in Python clears all fields and sets them to zero (or NULL).

PR #1178, opened 2025-10-06
ci: fix bad expansion of an array variable in build_libpq.sh

ubifred:
The `options` array variable have been introduced in `adb8336a392a71a3cb2ddf18bd9b17238a01a06e` but is used with `$options` which only expand the 1st value (for bash).

The full expansion have to be done with `${options[*]}`

Note that `-shared` and `-fPIC` options are still used as they are enabled by default by `./Configure`, but the `zlib` option is not taken into account (disabled by default)

### DIFF
--- a/tools/ci/build_libpq.sh
+++ b/tools/ci/build_libpq.sh
@@ -118,7 +118,7 @@ if [ "$ID" == "centos" ] || [ "$ID" == "macos" ]; then
         if [ -z "${MACOSX_ARCHITECTURE:-}" ]; then
             ./config ${options[*]}
         else
-            ./configure "darwin64-$MACOSX_ARCHITECTURE-cc" ${options[*]}
+            ./config "darwin64-$MACOSX_ARCHITECTURE-cc" ${options[*]}
         fi
 
         make -s depend


tools/ci/build_libpq.sh
```diff
@@ -116,7 +116,7 @@ if [ "$ID" == "centos" ] || [ "$ID" == "macos" ]; then
         options=(--prefix=${LIBPQ_BUILD_PREFIX} --openssldir=${LIBPQ_BUILD_PREFIX} \
             zlib -fPIC shared)
         if [ -z "${MACOSX_ARCHITECTURE:-}" ]; then
-            ./config $options
+            ./config ${options[*]}
```
dvarrazzo, line 119: I assume line 121 is broken too. Can we then do something like (untested, gotta check bash syntax):
```bash
if [[ "${MACOSX_ARCHITECTURE:-}" ]]; then
    options=("darwin64-$MACOSX_ARCHITECTURE-cc") + options
fi
./config ${options[*]}
```
Also isn't `./configure` at line 121 wrong and should be `./config`?

ubifred: Yes line 121 is also wrong.

I think `options=("darwin64-$MACOSX_ARCHITECTURE-cc") + options` is incorrect in bash and you have to use `+=`: 
`options+=("darwin64-$MACOSX_ARCHITECTURE-cc")` 

For `./configure` is does not exists in openssl source (`./Configure` exists, is darwin case-incesitive ?) so I think it's better to keep using `./config`

Sorry I did not made many tests, I encounter this issue while debugging another issue.

dvarrazzo: If we use += then `darwin64-$MACOSX_ARCHITECTURE-cc"` will become the last argument, not the first.

I am ok to keep the two calls on two separate lines. Life is too short for data structures in Bash.

Probably the line with the `./configure` is never hit in our CI so we never noticed. Darwin is most likely case sensitive, and `./config` actually is just a thin script calling `./Configure`.

tools/ci/build_libpq.sh
```diff
@@ -116,9 +116,9 @@ if [ "$ID" == "centos" ] || [ "$ID" == "macos" ]; then
...
         else
-            ./configure "darwin64-$MACOSX_ARCHITECTURE-cc" $options
+            ./configure "darwin64-$MACOSX_ARCHITECTURE-cc" ${options[*]}
```
dvarrazzo: For consistency with the other branch.

```suggestion
            ./config "darwin64-$MACOSX_ARCHITECTURE-cc" ${options[*]}
```

ubifred: seems fine :-)

PR #1182, opened 2025-10-10
Update DEFAULT_AUTO_FIELD setting in Django CI workflow

jacobtylerwalls:
Django 6.0 made adjustments to its tests to allow tests to be launched with the new default value for `DEFAULT_AUTO_FIELD`, which became `BigAutoField` in 6.0 (https://github.com/django/django/pull/19753)

Refs https://github.com/django/django-docker-box/pull/59 for an analogous change to Django's docker configuration for running tests.

Hoping this change is enough to make this workflow pass again. Thanks for testing against Django pre-releases!

### DIFF
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -135,8 +135,10 @@ jobs:
             pip_django: git+https://github.com/django/django.git#egg=Django
           - django_label: lts4
             pip_django: "'Django >= 4.2, < 4.3'"
+            default_auto_fields: 'DEFAULT_AUTO_FIELD = "django.db.models.AutoField"'
           - django_label: lts5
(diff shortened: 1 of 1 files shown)


dvarrazzo:
Thank you! Running the integration tests...

jacobtylerwalls:
Ok, maybe we need to try an explicit BigAutoField then. On mobile right now, can update the PR later today.

dvarrazzo:
Thank you for suggesting that change! It fixes indeed main, but it breaks lts4 and 5.

I've made setting that parameter optional and now [tests pass](https://github.com/psycopg/psycopg/actions/runs/18412628629).

So this changeset now fixes the 3rd party tests ci and can be merged as it is. If you have further improvement please let us know :slightly_smiling_face:

PR #1184, opened 2025-10-14
Added transaction attribute to record rollback/commit

### LINKED ISSUE #969: Store the commit/rollback state in the transaction object
After leaving a transaction context, we may store on the object whether the block was exited with a commit or a rollback.

> ```python
> with conn.transaction() as tx:
>    a_function(conn)  # call function that can raise Rollback()
> ```
> Does the transaction object `tx` save the rollback event, so I ca query it afterwards?

_Originally posted by @flashdagger in https://github.com/psycopg/psycopg/issues/967#issuecomment-2547096446_

immortalcodes:
## Done
- Added a transaction state to capture the state of transaction during execution

## Fixes
 #969

### DIFF
--- a/docs/api/objects.rst
+++ b/docs/api/objects.rst
@@ -247,7 +247,7 @@ Transaction-related objects
 
 See :ref:`transactions` for details about these objects.
 
-.. autoclass:: IsolationLevel
+.. autoclass:: IsolationLevel()
     :members:
 
     The value is usually used with the `Connection.isolation_level` property.
@@ -262,31 +262,53 @@ See :ref:`transactions` for details about these objects.
 
     .. autoattribute:: savepoint_name
     .. autoattribute:: connection
-    .. autoattribute:: status
+    .. attribute:: status
 
-        The current status of the transaction. Returns a `Transaction.Status`
-        enum value indicating whether the transaction is active, committed,
-        rolled back, or failed.
+        The current status of the transaction.
 
-        Possible values:
-
-        - `!NOT_STARTED`: Transaction created but not yet entered
-        - `!ACTIVE`: Transaction is currently active
-        - `!COMMITTED`: Transaction successfully committed
-        - `!ROLLED_BACK_EXPLICITLY`: Transaction explicitly rolled back
-        - `!ROLLED_BACK_WITH_ERROR`: Transaction rolled back due to an error
-        - `!FAILED`: Transaction failed due to connection failure
+        :type: `~psycopg.Transaction.Status`
 
         See :ref:`transaction-status` for more details and examples.
 
 
 .. autoclass:: AsyncTransaction()
 
-    .. autoattribute:: connection
-    .. autoattribute:: status
+    The class differs from `Transaction` for the following details:
 
-        The current status of the transaction. Returns a `AsyncTransaction.Status`
-        enum value. See `Transaction.status` for details.
+    .. autoattribute:: connection
+
+
+.. class:: psycopg.Transaction.Status()
+.. autoclass:: psycopg.AsyncTransaction.Status()
+
+    .. note: `:members:` doeesn't seem to work currently. Nested class problem?
+
+    .. attribute:: NON_STARTED
+
+        Transaction created but not yet entered.
+
+    .. attribute:: ACTIVE
+
+        Transaction currently active (inside the `!with` block).
+
+    .. attribute:: COMMITTED
+
+        Transaction successfully committed.
+
+    .. attribute:: ROLLED_BACK_EXPLICITLY
+
+        The transaction was explicitly rolled back (by raising
+        `~psycopg.Rollback` or calling `!transaction(force_rollback=True)`).
+
+    .. attribute:: ROLLED_BACK_WITH_ERROR
+
+        The transaction was rolled back due to an exception raised within the
+        transaction block.
+
+    .. attribute:: FAILED
+
+        The transaction failed due to a connection failure (e.g., the
+        connection was closed or became unavailable during the transaction).
 
 
 .. autoexception:: Rollback
--- a/docs/basic/transactions.rst
+++ b/docs/basic/transactions.rst
@@ -303,25 +303,16 @@ but not entirely committed yet.
 Transaction status
 ^^^^^^^^^^^^^^^^^^
 
-Each `Transaction` object exposes a `~Transaction.status` property that allows
-you to inspect the current state of the transaction. This can be useful for
+.. versionadded:: 3.3
+
+Each `!Transaction` object exposes a `~Transaction.status` property allowing
+to inspect the current state of the transaction. This can be useful for
 debugging, logging, or implementing custom transaction management logic.
 
-The `!status` property returns a `Transaction.Status` enum value that can be
-one of the following:
+The `!status` property is a `~psycopg.Transaction.Status` enum value:
+please check the enum documentation for the possible values.
 
-- `!NOT_STARTED`: The transaction has been created but not yet entered (before
-  the `!__enter__()` method is called).
-- `!ACTIVE`: The transaction is currently active (inside the `!with` block).
-- `!COMMITTED`: The transaction exited successfully and changes were committed.
-- `!ROLLED_BACK_EXPLICITLY`: The transaction was explicitly rolled back, either
-  by raising a `Rollback` exception or by using `!force_rollback=True`.
-- `!ROLLED_BACK_WITH_ERROR`: The transaction was rolled back due to an
-  exception raised within the transaction block.
-- `!FAILED`: The transaction failed due to a connection failure (e.g., the
-  connection was closed or became unavailable during the transaction).
-
-Here's an example of how to use the status property:
+Here are a few examples of how to use the `!status` property:
 
 .. code:: python
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -24,7 +24,7 @@ Psycopg 3.3.0 (unreleased)
 - Add `Cursor.set_result()` and `Cursor.results()` to move across the result
   sets of queries executed though `~Cursor.executemany()` or
   `~Cursor.execute()` with multiple statements (:tickets:`#1080, #1170`).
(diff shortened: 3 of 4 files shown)


dvarrazzo:
I don't like having a string constant to represent this state: it's open to problems such as case comparison.

I was thinking to have two booleans to represent exit with commit or with rollback; if both are false it would mean that the transaction is not finished yet. However we might want to capture more details, such as whether the rollback was explicit or implicitly caused by an exception.

So maybe it would be better to have a `TransactionStatus` enum whose states could be (for example)

- not_started
- active
- committed
- rolled_back_explicitly
- rolled_back_with_error

and a `status` attribute on the transaction object.

Note that there's already an enum `psycopg.pq.TransactionStatus` so maybe this is not a good name and we should figure out a different one. We could nest a `Status` class inside the `BaseTransaction` class for example.

psycopg/psycopg/transaction.py
```diff
@@ -85,7 +86,8 @@ def __repr__(self) -> str:
             status = "terminated"
 
         sp = f"{self.savepoint_name!r} " if self.savepoint_name else ""
-        return f"<{cls} {sp}({status}) {info} at 0x{id(self):x}>"
+        ew = self.exited_with
+        return f"<{cls} {sp}({status} with {ew}) {info} at 0x{id(self):x}>"
```
dvarrazzo: If `status` becomes an enum, this may become the likes of:
```
         return f"<{cls} {sp}({self.status.value}) {info} at 0x{id(self):x}>"
```
so the enum might be a str enum with somewhat-pretty values we would display here.

immortalcodes: Thanks for the suggestions, I will push the changes accordingly.

immortalcodes: Pushed the  changes

dvarrazzo: A few observations:

- `state` is a wrong name: the right name is `status`. The *state* is the internal configuration of an object, such as its name or its status, if the latter is stored in an attribute; the *status* refers to the outcome of an operation.
- the `Transaction` prefix is unneeded: this is an internal class, so users would have to refer to it as `Transaction.TransactionStatus`, which is redundant. `Status` (resulting in having to use it as `if tx.status == Transaction.Status.COMMITTED` or `if tx.status == tx.Status.COMMITED`) is much better, don't you think?
- Tests are failing (I haven't checked why, but probably because of the repr changed). Repr tests can be adjusted if we verify that the new output is what expected.
- There are no tests about how the state changes.
- Is `FAILED` status still needed? Is it observable in a way different than `ROLLED_BACK_WITH_ERROR`? I assume it is (is it the state before exiting the block?) but I'd like to clarify it (and having test proving it).
- We need documentation for the new attribute and enum.

I will review once tests showing the behaviour of the new feature are implemented and the full test suite passes.

Thank you very much 🙂

immortalcodes:
Hey @dvarrazzo ,
I guess it should be good to have another look, one test that is failing seems unrelated to my change, could you help me out there.
Thanks

docs/api/objects.rst
```diff
@@ -262,11 +262,31 @@ See :ref:`transactions` for details about these objects.
...
+
+        See :ref:`transaction-status` for more details and examples.
 
 
 .. autoclass:: AsyncTransaction()
 
     .. autoattribute:: connection
+    .. autoattribute:: status
+
+        The current status of the transaction. Returns a `AsyncTransaction.Status`
```
dvarrazzo: I wonder if we can document the enum itself.

Anyways leave it to me, because all the enums docs are currently broken, the idioms used to document the different entries stopped working some time in the past.

docs/basic/transactions.rst
```diff
@@ -297,6 +298,67 @@ but not entirely committed yet.
...
+
+.. code:: python
+
+    from psycopg import Rollback
+
+    # Successful transaction
+    with conn.transaction() as tx:
+        assert tx.status.name == "ACTIVE"
+        conn.execute("INSERT INTO data VALUES (%s)", ("Hello",))
+    assert tx.status.name == "COMMITTED"
```
dvarrazzo: Wouldn't it be better to suggest to access the enums by identity rather than by name?

```suggestion
    assert tx.status.name == tx.Status.COMMITTED
```

docs/basic/transactions.rst
```diff
@@ -297,6 +298,67 @@ but not entirely committed yet.
...
+- `!ROLLED_BACK_WITH_ERROR`: The transaction was rolled back due to an
+  exception raised within the transaction block.
+- `!FAILED`: The transaction failed due to a connection failure (e.g., the
+  connection was closed or became unavailable during the transaction).
+
+Here's an example of how to use the status property:
+
+.. code:: python
+
+    from psycopg import Rollback
```
dvarrazzo, line 319: Nice example, thank you :slightly_smiling_face:

psycopg/psycopg/transaction.py
```diff
@@ -62,6 +71,7 @@ def __init__(
         self._savepoint_name = savepoint_name or ""
         self.force_rollback = force_rollback
         self._entered = self._exited = False
```
dvarrazzo: Can we drop `_entered/_exited` anymore and use `status` only? Annotating where we use these names...

psycopg/psycopg/transaction.py
```diff
@@ -77,20 +87,14 @@ def savepoint_name(self) -> str | None:
...
-            status = "active"
-        else:
-            status = "terminated"
-
         sp = f"{self.savepoint_name!r} " if self.savepoint_name else ""
-        return f"<{cls} {sp}({status}) {info} at 0x{id(self):x}>"
+        return f"<{cls} {sp}({self.status.value}) {info} at 0x{id(self):x}>"
 
     def _enter_gen(self) -> PQGen[None]:
         if self._entered:
```
dvarrazzo:
```suggestion
        if self.status != self.Status.NOT_STARTED:
```

psycopg/psycopg/transaction.py
```diff
@@ -136,6 +141,12 @@ def _rollback_gen(self, exc_val: BaseException | None) -> PQGen[bool]:
 
         ex = self._pop_savepoint("rollback")
         self._exited = True
```
dvarrazzo: This can be dropped, and so the one in `_commit_gen`.
```suggestion
```
So I would say that we don't need these two attributes anymore and we can use the state only. Afaics there's no other use for these property, except in `__repr__()`, from where they have been removed already.

tests/test_transaction.py
```diff
@@ -523,6 +523,70 @@ def test_force_rollback_exception_exit(conn, svcconn):
     assert not inserted(svcconn)
 
 
+def test_transaction_status(conn_cls, dsn):
```
dvarrazzo, line 526: Thank you for the test :slightly_smiling_face:

tests/test_transaction_async.py
```diff
@@ -530,6 +530,70 @@ async def test_force_rollback_exception_exit(aconn, svcconn):
...
+async def test_transaction_status(aconn_cls, dsn):
+    aconn = await aconn_cls.connect(dsn)
+
+    """
+    The Transaction.status property ends up in committed state when no exceptions
+    are raised and force_rollback is False(default).
+    """
+    assert aconn.pgconn.transaction_status == pq.TransactionStatus.IDLE
+    async with aconn.transaction() as tx:
+        assert tx.status.name == "ACTIVE"
```
dvarrazzo: Can we use the enum identity here too please?
```suggestion
        assert tx.status is tx.Status.ACTIVE
```

tests/test_transaction.py
dvarrazzo, line 526: Can we also have a test to check that the status is reported correctly for nested transactions too?

dvarrazzo:
> Hey @dvarrazzo , I guess it should be good to have another look, one test that is failing seems unrelated to my change, could you help me out there. Thanks

I think it was just a fluke in a random test. Unfortunately I can't reproduce it: on linux:

```
$ pytest --randomly-seed=3627450784 "tests/test_adapt.py::test_random[Format.BINARY-s]"
platform linux -- Python 3.10.12, pytest-8.3.4, pluggy-1.5.0
...
========================= 1 passed in 0.08s =========================
```
so let's forget about it :zipper_mouth_face:

tests/test_transaction.py
immortalcodes: This is actually a bit tricky because there could be all sorts of permutations of outer and inner transactions, or even more than a single inner transaction.
For the sake of this PR, I have written a basic test that tests out the basic nested pattern.

dvarrazzo: Thank you, that tests looks good :slightly_smiling_face:

immortalcodes:
Hey @dvarrazzo , is the PR good for approval? Let me know the next steps.
Thanks

dvarrazzo:
Hello!

I asked if you could add a news entry. But no worries, I can also do it myself.

I'll try to render the docs and check if the enum renders ok. I'll squash the commits and merge this to release 3.3.

I think I can take over from here and probably I'll manage to get this merged today. Thank you very much!

docs/news.rst
```diff
@@ -21,6 +21,8 @@ Psycopg 3.3.0 (unreleased)
 - Add `Cursor.results()` to iterate over the result sets of the queries
   executed though `~Cursor.executemany()` or `~Cursor.execute()`
   (:ticket:`#1080`).
+- Add `Transaction.status` to report the status during and after a
+  `~Connection.transaction()` block (:ticket:`#969`).
```
immortalcodes: @dvarrazzo is that what you meant by news?

dvarrazzo: Yes

dvarrazzo:
Sorry for the delay!

The `Status` wasn't inheriting from `str`; it doesn't seem it was a problem, but I've fixed it.

I have cleaned up the documentation a bit. Running final tests before merging.

Thank you very much!

PR #1186, opened 2025-10-15
range type test: c_range + inner dumper

jerch:
Playground to illustrate the perf differences for range types:
- inner dumper - derive inner dumper from setting on range type
- c_range - implements a range dump function in cython

Perf differences:

```plain
# current master
bench_range_dumping: 0.04303 ±0.00089 (cv: 0.021, repeat: 50)

# inner_dumper
bench_range_dumping: 0.03689 ±0.00066 (cv: 0.018, repeat: 50)

# c_range
bench_range_dumping: 0.02805 ±0.00042 (cv: 0.015, repeat: 50)

# c_range + inner_dumper
bench_range_dumping: 0.01444 ±0.00027 (cv: 0.019, repeat: 50)
```

Perf tests are in `bench_range.py`, there is a second one with a full roundtrip to the database with copy.

Note: ~Currently the inner dumper is fixed and not yet derived from an inner oid setting, as I did not get `get_dumper_by_oid` working in cython (segfaulted).~ Fixed with [a9d89ba](https://github.com/psycopg/psycopg/pull/1186/commits/a9d89ba6af1450c29756e143e10dd8a1829a0e97).

### DIFF
--- a/psycopg/psycopg/types/range.py
+++ b/psycopg/psycopg/types/range.py
@@ -404,7 +404,7 @@ class RangeBinaryDumper(BaseRangeDumper):
             if _psycopg:
                 self._dump = _psycopg.dumper_by_oid_helper(self._tx, self.subtype_oid).dump
             else:
-                dump = self._tx.get_dumper_by_oid(self.subtype_oid, Format.BINARY).dump
+                self._dump = self._tx.get_dumper_by_oid(self.subtype_oid, Format.BINARY).dump
 
     def dump(self, obj: Range[Any]) -> Buffer | None:
         if (item := self._get_item(obj)) is not None:


jerch:
This last commit optimizes the c_range_dump function, which now completes in:
```text
bench_range_dumping: 0.01164 ±0.00051 (cv: 0.044, repeat: 50)
```

jerch:
Just saw that I have misread the `fail_dump` logic, so the c version does not quite the same yet :sweat_smile:

jerch:
@dvarrazzo The last commit is a suggestion for optimized bytearray handling in `_buffer_as_string_and_size` and setting the function inline. Both changes speed up the bytearray merging in cython by ~10%. Not sure if inline will create more issues at other caller places though.

PR #1187, opened 2025-10-16
Improve loading performance by transposing row-col loading order

### LINKED ISSUE #1155: Performance degradation compared to psycopg2

dvarrazzo:
This MR replaces #1163, which has the entire history of attempting pagination, fine-tuning the page size, and then figuring out that it wasn't needed. So it simply transposes the row-col load order.

Because the history of what happened in #1163 might be useful, we will keep that branch around ([here](https://github.com/dvarrazzo/psycopg/tree/fix/speedregression-fetchall), for example).

Close #1163
Close #1155

Attn. @jerch

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -20,6 +20,7 @@ Psycopg 3.2.11 (unreleased)
   the Python version (:ticket:`#1153`).
 - Don't raise warning, and don't leak resources, if a builtin function is used
   as JSON dumper/loader function (:ticket:`#1165`).
+- Improve performance of Python conversion on results loading (:ticket:`#1155`).
 
 
 Current release
--- a/psycopg_c/psycopg_c/_psycopg/transform.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/transform.pyx
@@ -443,36 +443,23 @@ cdef class Transformer:
         cdef object record  # not 'tuple' as it would check on assignment
 
         cdef object records = PyList_New(row1 - row0)
-        for row in range(row0, row1):
-            record = PyTuple_New(self._nfields)
-            Py_INCREF(record)
-            PyList_SET_ITEM(records, row - row0, record)
 
         cdef PyObject *loader  # borrowed RowLoader
-        cdef PyObject *brecord  # borrowed
+
         row_loaders = self._row_loaders  # avoid an incref/decref per item
 
-        for col in range(self._nfields):
-            loader = PyList_GET_ITEM(row_loaders, col)
-            if (<RowLoader>loader).cloader is not None:
-                for row in range(row0, row1):
-                    brecord = PyList_GET_ITEM(records, row - row0)
-                    attval = &(ires.tuples[row][col])
-                    if attval.len == -1:  # NULL_LEN
-                        pyval = None
-                    else:
+        for row in range(row0, row1):
+            record = PyTuple_New(self._nfields)
+
+            for col in range(self._nfields):
+                attval = &(ires.tuples[row][col])
+                if attval.len == -1:  # NULL_LEN
+                    pyval = None
+                else:
+                    loader = PyList_GET_ITEM(row_loaders, col)
+                    if (<RowLoader>loader).cloader is not None:
                         pyval = (<RowLoader>loader).cloader.cload(
                             attval.value, attval.len)
-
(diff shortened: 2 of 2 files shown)


dvarrazzo:
I can confirm the speedup using [this script](https://gist.github.com/dvarrazzo/c544aa848f30ce7ad9ba3302e0b86cce):

```
# Current 3.2.10 version 
$ PSYCOPG_IMPL=binary python scrap/bench-fetchall.py  -r 1000000 -m psycopg
creating table, cols: 10, rows: 1000000
testing psycopg[binary] 3.2.10
rows: 1000000 cols: 10: 0.8676 ±0.03862 (cv: 0.045, repeats: 10)

# Upcoming 3.2.11 version 
$ PSYCOPG_IMPL=c python scrap/bench-fetchall.py  -r 1000000 -m psycopg
creating table, cols: 10, rows: 1000000
testing psycopg[c] 3.2.11.dev1
rows: 1000000 cols: 10: 0.48332 ±0.00256 (cv: 0.005, repeats: 10)

# psycopg2
$ python scrap/bench-fetchall.py  -r 1000000 -m psycopg2
creating table, cols: 10, rows: 1000000
testing psycopg2 2.9.10 (dt dec pq3 ext lo64)
rows: 1000000 cols: 10: 0.52228 ±0.00479 (cv: 0.009, repeats: 10)
```

Loading is almost about 70% faster than the current psycopg release and about 10% faster than psycopg2.

Issue #1164, opened 2025-09-09
Introduce performance regression tests


dvarrazzo:
I think we are a bit in the dark about whether changes we introduce, or have introduced in the past, may result in performance regression.

It might be useful to have a script to trace and compare runtimes around:

- different version of psycopg 3
- latest psycopg2
- latest asyncpg

vs. different types of work (copy, select, insert) and different dataset sizes.

It would be useful to run such job periodically on Github workers and see if the noise is contained enough to make the comparisons useful (I assume, in the same run, we should run all the tests and compare the result, rather than compare changes over time).

I wonder if there is any Github integration to present with the results in a readable way, maybe a graph?

There is some basic micro-benchmark with different adapters in the [`tests/script`](https://github.com/psycopg/psycopg/tree/master/tests/scripts) directory.

References:

- #1155 
- #1146 
- #645

jerch:
> It would be useful to run such job periodically on Github workers and see if the noise is contained enough to make the comparisons useful (I assume, in the same run, we should run all the tests and compare the result, rather than compare changes over time).

In the past I wrote a benchmark lib for xterm.js, as we had the issue of not really knowing, what eats the precious CPU cycles during VT parsing. From my experience over there - github CI is not a good perf testground, as the system load of their big machines varies too much bringing in a huge uncertainty (even within a single test run it might spike randomly from naughty neighbors). We tried to put the perf tests into the CI chain to get reports from it, but that was horribly unreliable and unusable. So in the end we just went there with the "test perf-sensitive changes on your dev machine only" approach.

Doing only that for psycopg might not be the best idea, as almost every dev setup would only cover localhost setups. No clue, if github offers perf test machines with more stable system load promises, hmm.

asqui:
I recall that Wes McKinney had a micro-benchmark framework that would profile various usage scenarios for Pandas and display graphs for each one with changes over time (commit-by-commit).

Unfortunately this was a long time ago and I don't know offhand what ever came of that approach, whether it was spun-off as a separate project, and where/how the benchmarks were run to address the results variability concerns mentioned in this thread.

dlax:
There is [airspeed velocity](https://asv.readthedocs.io/en/latest/), which is used by in scientific python ecosystem, e.g.  https://numpy.org/doc/stable/benchmarking.html

jerch:
airspeed velocity looks promising, seems it has all nifty bells & whistles down to html summaries.

Sidenote on the uncertainty:
[...]

PR #1193, opened 2025-10-23
free value dumping in text without set_types

### LINKED ISSUE #1192: Regression in 3.2.11 regarding `COPY FROM` with values of different types
When trying to upgrade from psycopg 3.2.10 to 3.2.11, I encountered some errors around `cursor.copy`. I'm inserting values that are not of the same type, but can be implicitly converted by the database.

For example, this code worked in 3.2.10:

```py
import psycopg

with psycopg.connect(dbname='postgres', user='postgres') as conn:
	with conn.cursor() as cursor:
		cursor.execute("CREATE TEMPORARY TABLE t (a text)")

		with cursor.copy(f"COPY t FROM STDIN") as copy:
			for value in [1, '2']:
				copy.write_row((value,))
```

While in 3.2.11 there's an error trying to insert the second value (the one that actually matches the `text` type):

```
psycopg.DataError: integer expected, got 'str'
```

When we change the order of types to `['1', 2]`, the error is different:

```
TypeError: bad argument type for built-in operation
```

Similar errors appear with other types as well, e.g. when trying to copy values `[1, 2.0]` into a column of type `real` or `decimal`.

---

It seems that this change (using the type of the first value to decide how to handle all future values) has been introduced deliberately in #1158. Is this really the expected behavior?

The [documentation of `COPY`](https://www.psycopg.org/psycopg3/docs/basic/copy.html#writing-data-row-by-row) mentions:

> the Python values are adapted as they would be in normal querying.

But `INSERT` queries don't have this problem, e.g. this works fine:

```py
cursor.execute("INSERT INTO t VALUES (%s), (%s)", [1, '2'])
```

jerch:
Shall fix #1192.

@dvarrazzo I was only able to fix this for cython. The python dumpers are less strict in general. Question here would be, if the dumpers should act more similar between python & cython to get equal API conditions.

And second question would be, if `set_types` should be enforced for BINARY. It currently does that implicit pinning from first not-None values too (have to check again, if that still might trigger the OOM, if not, then a tiny bit of auto-detection seems reasonable to me).

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -43,8 +43,15 @@ Psycopg 3.3.0 (unreleased)
 - Drop support for Python 3.8 (:ticket:`#976`) and 3.9 (:ticket:`#1056`).
 
 
-Psycopg 3.2.11 (unreleased)
+Psycopg 3.2.12 (unreleased)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Allow copy to pass different types per column, as long as the database can
+  convert them. Regression introduced in 3.2.11 (:ticket:`#1192`).
+
+
+Psycopg 3.2.11
+^^^^^^^^^^^^^^
 
 - Fix spurious readiness flags in some of the wait functions (:ticket:`#1141`).
 - Fix high CPU usage using the ``wait_c`` function on Windows (:ticket:`#645`).


dvarrazzo:
Hello,

Ideally dumpers in Python and C should behave the same, but if it's not practical I am ok to keep the two to diverge.

If your test here is failing because of a difference in behaviour in the float dumper let's limit the test to the int/str ambiguity. That should tell us that we don't pin the type and should be enough to close the issue.

tests/test_copy.py
```diff
@@ -487,6 +487,53 @@ def test_copy_in_records_binary(conn, format):
     assert data == [(1, None, "hello"), (2, None, "world")]
 
 
+def test_copy_in_text_no_pinning(conn):
```
dvarrazzo, line 500: This test should be in `test_copy_async.py` and the sync counterpart should be generated by `tools/async_to_sync.py --docker`

jerch:
> Ideally dumpers in Python and C should behave the same, but if it's not practical I am ok to keep the two to diverge.

I think that is a goal worth hunting, but possibly not reasonable within this MR? Maybe address it by a new issue?

dvarrazzo:
Whether it's worth, I don't know. If the implementation of a Python dumper is `return str(ojb)` and it has to become more complex to match stricter requirements of the C code then maybe it's not worth.

Anyways, definitely it is not part of this MR.

dvarrazzo:
> This test should be in test_copy_async.py and the sync counterpart should be generated by tools/async_to_sync.py --docker

I noticed that the tests are running automatically in CI but not the lint job, which should flag the async/sync discrepancy. I've enabled them on master now: please rebase your branch. 

Thank you!

tests/test_copy.py
jerch: :eyes:  automatic async --> sync code gen? Thats awesome (and a bit frightening haha).

dvarrazzo: If it was AI-drive, it would... but it's the good old [visitor pattern](https://www.psycopg.org/articles/2024/09/23/async-to-sync/).

jerch:
> If your test here is failing because of a difference in behaviour in the float dumper let's limit the test to the int/str ambiguity. That should tell us that we don't pin the type and should be enough to close the issue.

Both accept str, but still I can make the test shorter to just str vs. int type input.

jerch:
Seems the async --> sync converter does not understand multiline strings? https://github.com/psycopg/psycopg/actions/runs/18754474914/job/53502686768?pr=1193#step:6:50

dvarrazzo:
> Seems the async --> sync converter does not understand multiline strings?

No, it doesn't understand implicit concatenation because the strings are put together by the tokenizer, so the AST only sees a single string. In order to see them the work should be done on a Concrete Syntax Tree.

If the problem is that the string becomes too long and triggers flake8 you can add a `per-file-ignores` entry in the `.flake8` file. See what's already there for similarity.

Alternatively you can use a proper multi-line string 
```python
"""
col1 serial primary key, col2 int, col3 int,
col4 double precision, col5 double precision,
"""
```
if it works with the syntax there, or explicit concatenation
```python
        "col1 serial primary key, col2 int, col3 int, "
        + "col4 double precision, col5 double precision",
```

jerch:
Well I shimmed it with a column def list instead.

jerch:
@dvarrazzo
FYI - I benchmarked pinned and unpinned text format with a full copy roundtrip to db with 5 integer colums:
```text
copy_unpinned: 0.03017 ±0.00713 (cv: 0.236, repeat: 20)
copy_pinned: 0.0286 ±0.00426 (cv: 0.149, repeat: 20)
Pinned is 5.52% faster.
```
The cv is really high which means there is high uncertainty. When repeating the perf test the speedup settles between -3% to +10%. In regards to #1164 this a known issue with macro-benchmarking where those macro components hide the details (here postgres and system IO). It is still the only way to tell the user, what perf difference he can expect from the high level APIs, which is prolly ~5% here. (On a sidenote - when I change the pinned version to BINARY the speedup settles between 5% - 20%.)

Edit: Oh it is much worse with the python version:
```text
copy_unpinned: 0.0972 ±0.00398 (cv: 0.041, repeat: 20)
copy_pinned: 0.05821 ±0.00783 (cv: 0.134, repeat: 20)
Pinned is 66.97% faster.
```

jerch:
@dvarrazzo In your post over here https://github.com/psycopg/psycopg/issues/1192#issuecomment-3436348335 you mentioned `executemany`. Are there other ends I should check, if the behavior is correct?

dvarrazzo:
Hello @jerch 

Yeah, the difference in performance is expected because we fetch a loader for every object, not for every column.

That's why I'm still moderately -1 on this fix.

About execuemany there is nothing I think we should check. It works for a reasonable input and if it barfs on unreasonable ones it is not our concern.

I think the expectation that you can ask postgres the same type flexibility of python is misplaced. Psycopg sits on the fence between the two so it had to make decisions. For me "don't send text to a int column" is a reasonable one...

dvarrazzo:
Hi @jerch 

MR looks ok. I have flattened your commits and I have fixed the test of the Python side, by creating a stricter version of the dumper for the test.

If we are happy with this we can merge and release.

jerch:
@dvarrazzo Wow nice idea with the stricter custom dumper. Good to go from my side, thanks!

Issue #1191, opened 2025-10-21
pessemization in parse_row_binary


jerch:
Hey @dvarrazzo me again 😸 

More accidentally I have found a pessemization in `parse_row_binary`. There you first populate a list with field buffers, to run the buffer to object conversation by a second iteration to populate the final tuple. By removing the second iteration and doing the conversion directly into the row tuple, things get at least 20% faster. I have only tested it with a single range column row, but my guess is, that this will gain further ground for wider rows.

My quickly hacked change is below, let me know if you want to see this addressed with an MR and some tests.

```cython
def parse_row_binary(data, tx: Transformer) -> tuple[Any, ...]:
    cdef unsigned char *ptr
    cdef Py_ssize_t bufsize
    _buffer_as_string_and_size(data, <char **>&ptr, &bufsize)
    cdef unsigned char *bufend = ptr + bufsize

    cdef uint16_t benfields
    memcpy(&benfields, ptr, sizeof(benfields))
    cdef int nfields = endian.be16toh(benfields)
    ptr += sizeof(benfields)

    cdef int col
    cdef int32_t belength
    cdef Py_ssize_t length

    out = PyTuple_New(nfields)
    cdef PyObject *loader  # borrowed RowLoader
    cdef Py_ssize_t size

    row_loaders = tx._row_loaders  # avoid an incref/decref per item
    if PyList_GET_SIZE(row_loaders) != nfields:
        raise e.ProgrammingError(
            f"cannot load sequence of {nfields} items:"
            f" {len(row_loaders)} loaders registered")

    for col in range(nfields):
        memcpy(&belength, ptr, sizeof(belength))
        ptr += sizeof(belength)
        if belength == _binary_null:
            Py_INCREF(None)
            PyTuple_SET_ITEM(out, col, None)
        else:
            length = endian.be32toh(belength)
            if ptr + length > bufend:
                raise e.DataError("bad copy data: length exceeding data")
            loader = PyList_GET_ITEM(row_loaders, col)
            if (<RowLoader>loader).cloader is not None:
                pyval = (<RowLoader>loader).cloader.cload(<const char *>ptr, length)
            else:
                field = PyMemoryView_FromMemory(<char*>ptr, length, PyBUF_READ)
                pyval = PyObject_CallFunctionObjArgs(
                    (<RowLoader>loader).loadfunc, <PyObject *>field, NULL)
            ptr += length
            Py_INCREF(pyval)
[...]

PR #1196, opened 2025-10-28
Refactor C API to use PyObject_Fast* functions

dvarrazzo:
[An error](https://github.com/psycopg/psycopg/actions/runs/18826813855) during a different development let me to discover the [`PySequence_Fast*()`](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast) family of functions.

Using these function is a more practical alternative to making sure that a sequence is a tuple or a list in order to call the appropriate API functions, and allow faster call for `len(seq)` and `seq[item]`.

Note: `seq[item]` returns a new reference (an `object`, in Cython), `PySequence_Fast_GET_ITEM(seq, item)` returns a borrowed reference (a `PyObject *`, in Cython). Therefore the casts around these objects are sort of swapped when passing the result as either Python object or PyObject reference.

@jerch thoughts?

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/transform.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/transform.pyx
@@ -364,9 +364,6 @@ cdef class Transformer:
         cdef list out = PyList_New(nparams)
         cdef PyObject *param
 
-        if self._none_oid < 0:
-            self._none_oid = self.adapters.get_dumper(NoneType, "s").oid
-
         dumpers = self._row_dumpers
         if dumpers:
             for i in range(nparams):
@@ -410,6 +407,8 @@ cdef class Transformer:
                 fmt = (<RowDumper>dumper_ptr).format
             else:
                 dumped = None
+                if self._none_oid < 0:
+                    self._none_oid = self.adapters.get_dumper(NoneType, "s").oid
                 oid = self._none_oid
                 fmt = PQ_TEXT
 


jerch:
Yes I was also intrigued to use it for one of the copy MRs, but was unsure of its python version coverage. From the docs it seems it was added to the stable ABI not before 3.10? Or does cython shim issues in that field away? (I am still a cython noob, maybe only API compat is needed with it wheels?)

dvarrazzo:
No, looks like they have been there [since 2.ancient](https://docs.python.org/2.6/c-api/sequence.html#PySequence_Fast_GET_ITEM)... It's the concept of "stable API" to be more modern.

jerch:
Yes they added those "stable ABI" notions quite recently to the docs (around 3.9 or 3.10). Not sure about its implications in regards to cython and the old API docs. The doc it links to notes, that the list below will stay stable within 3.x - so maybe it is just further hardening the API guarantees in the whole major 3 releases branch.

Offtopic:
And since you are on refactoring basic interfaces - in #1191 I replaced your buffer object creation with a simple memory view construction from a char pointer:
```cython
field = PyMemoryView_FromMemory(<char*>ptr, length, PyBUF_READ)
```
which is a lot faster than with the interim buffer object. Was there a reason for the buffer object in the first place I didnt grasp?

dvarrazzo:
The stabile API Is to guarantee some form of forward compatibility, but we build packages for every major version so it's not something we plan to make use of in the current artifacts production flow.

Regards the memory refactoring, it is in my pipeline to look at it, this branch is a smaller refactoring triggered by a bug to fix and extending the same pattern to where we have been too much ad-hoc.

jerch:
> The stabile API Is to guarantee some form of forward compatibility, but we build packages for every major version so it's not something we plan to make use of in the current artifacts production flow.

Oh I didnt check, if wheels get build for every python version separately, since this is the case it should just work with API compat yeah.

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -23,13 +24,8 @@ cdef int32_t _binary_null = -1
...
 cdef object _format_row_binary(object row, Transformer tx, bytearray out):
     """Convert a row of adapted data to the data to send for binary copy"""
-    cdef Py_ssize_t rowlen
-    if type(row) is list:
-        rowlen = PyList_GET_SIZE(row)
-    elif type(row) is tuple:
-        rowlen = PyTuple_GET_SIZE(row)
-    else:
-        rowlen = len(row)
+    cdef row_fast = PySequence_Fast(row, "'row' is not a valid sequence")
```
jerch, line 27: This might raise a TypeError if the row data is not a sequence or iterable. My question here would be, if this might raise on types, that worked before with the python level indexed access, e.g. do numpy arrays still work here? (or other custom sequence like types, that only partially implement the needed interfaces)

The fact that it now also supports iterables out-of-the-box is a win API-wise. I wonder though, how PySequence_Fast realizes the entry access on iterables, would it create a temp sequence from it to allow repeated access (thus generator values would get repeatable)?

dvarrazzo: It it wasn't a sequence it would have raised an error already. You wouldn't be able to call len on it.

PR #1197, opened 2025-10-28
Add riscv64 support for linux builds

mliezun:
Same as https://github.com/psycopg/psycopg2/pull/1813

### DIFF
--- a/.github/workflows/build-and-cache-libpq.yml
+++ b/.github/workflows/build-and-cache-libpq.yml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: [x86_64, ppc64le, aarch64]
+        arch: [x86_64, ppc64le, aarch64, riscv64]
         platform: [manylinux, musllinux]
 
     steps:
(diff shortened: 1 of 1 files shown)


dvarrazzo:
I can see that [the libpq build workflow run correctly](https://github.com/mliezun/psycopg/actions/runs/18874850540), but that [no binary packages build complete](https://github.com/mliezun/psycopg/actions/workflows/packages-bin.yml).

Could you please run a package build job manually so we can check that the artifacts match correctly? Then we will be able to merge this branch.

Thank you very much!

mliezun:
> I can see that [the libpq build workflow run correctly](https://github.com/mliezun/psycopg/actions/runs/18874850540), but that [no binary packages build complete](https://github.com/mliezun/psycopg/actions/workflows/packages-bin.yml).
> 
> Could you please run a package build job manually so we can check that the artifacts match correctly? Then we will be able to merge this branch.
> 
> Thank you very much!

I tried running from `master` here and the build is still failing, see here: https://github.com/mliezun/psycopg/actions/runs/18903013102/job/53954482532

That's a musllinux build with no cache, building from commit 90744c8ea2a1a54a99de417978c7630d53b7e06d

I see that the tests `"not slow and not flakey"` are failing after building the wheel. And the errors look like this:

```
      def _wait_listen(self):
          with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
              for i in range(20):
                  if 0 == sock.connect_ex((self.client_host, self.client_port)):
                      break
                  time.sleep(0.1)
              else:
  >               raise ValueError("the proxy didn't start listening in time")
  E               ValueError: the proxy didn't start listening in time
  
  /project/tests/fix_proxy.py:130: ValueError
```

Could it be that the Github Runners used in the official project have a different configuration than the ones that I'm using (free tier from github) and that's causing this errors?

dvarrazzo:
No, we use the same bog standard runners. I am giving it a go. There was something recently about these proxy tests but I can't remember now - I'll try a repro.

dvarrazzo:
Ok, the problem was the one solved in 2905e07f0fa2fec4dd67225950d728e1e8c66c6a. The commit was made in the maint-3.2 branch but wasn't cherry-picked to master.

I have added the commit to master and rebased your branch.

mliezun:
Thanks!

I ran the build package workflow again and it succeeded: https://github.com/mliezun/psycopg/actions/runs/18933990005

This is ready to merge ✅

PR #1202, opened 2025-11-03
Customize composite dumper

dvarrazzo:
The other side of #932: allow to specify a function to extract a sequence of values from an object in order to dump it as a composite.


dvarrazzo:
Merging this as it is. If anyone has different suggestions for this feature we have still time to refine it.

PR #1210, opened 2025-11-14
Fix AsyncCursor.copy concurrency

Zaczero:
Before this patch, using the same connection by multiple Tasks (with at least one performing COPY) would lead to:

`psycopg.OperationalError: sending query and params failed: another command is already in progress`

I believe that holding the lock for the entire duration of the COPY is the correct behavior.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -46,6 +46,13 @@ Psycopg 3.3.0 (unreleased)
 - Drop support for Python 3.8 (:ticket:`#976`) and 3.9 (:ticket:`#1056`).
 
 
+Psycopg 3.2.13 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Fix `Cursor.copy()` and `AsyncCursor.copy()` to hold the connection lock
+  for the entire operation, preventing concurrent access issues (:ticket:`#1210`).
+
+
 Current release
 ---------------
 
--- a/tests/test_copy_async.py
+++ b/tests/test_copy_async.py
@@ -17,7 +17,7 @@ from psycopg.types.hstore import register_hstore
 from psycopg.types.numeric import Int4
 
 from .utils import eur
-from .acompat import alist
+from .acompat import AEvent, alist, gather, spawn
 from ._test_copy import sample_binary  # noqa: F401
 from ._test_copy import AsyncFileWriter, ensure_table_async, py_to_raw
 from ._test_copy import sample_binary_rows, sample_records, sample_tabledef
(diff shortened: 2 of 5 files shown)


dvarrazzo:
Hello, thank you, that makes sense.

Are you able to add a test to show the problem and that the solution works? I would use a copy operation that writes two rows in the db with Event objects stopping the copy task:
- between enter and the first row
- between the first and the second row
- between the second row and exit
and verifying that tasks in other cursors spawned while the copy task is waiting behave correctly?

The test can be written in the `test_copy_async.py` module, using the `acompat.AEvent` and the `spawn/gather` functions, and can be converted to sync and threads by the `tools/async_to_sync.py` script.

Can you please also add a release note in the `docs/news.rst` with a comment about the fix?

Thank you very much!

docs/news.rst
```diff
@@ -45,6 +45,11 @@ Psycopg 3.3.0 (unreleased)
 
 - Drop support for Python 3.8 (:ticket:`#976`) and 3.9 (:ticket:`#1056`).
 
+.. rubric:: Bug fixes
```
dvarrazzo: We will release this in 3.2.13:

```suggestion

Psycopg 3.2.13 (unreleased)
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tests/test_copy_async.py
```diff
@@ -972,6 +972,105 @@ async def test_copy_table_across(aconn_cls, dsn, faker, mode):
...
+    # Verify that all workers completed AFTER copy exited
+    # This confirms the connection lock was held throughout the copy operation
+    copy_exit_idx = execution_log.index("exited_copy")
+    worker1_idx = execution_log.index("worker_1_completed")
+    worker2_idx = execution_log.index("worker_2_completed")
+    worker3_idx = execution_log.index("worker_3_completed")
+
+    assert copy_exit_idx < worker1_idx, "Worker 1 should complete after copy exits"
+    assert copy_exit_idx < worker2_idx, "Worker 2 should complete after copy exits"
+    assert copy_exit_idx < worker3_idx, "Worker 3 should complete after copy exits"
```
dvarrazzo: Pytest gives a meaningful error message when an assert fails. Maybe this would be enough to check?

```suggestion
    want = "exited_copy worker_1_completed worker_2_completed worker_3_completed".split()
    assert execution_log[:-4] == want
```

tests/test_copy_async.py
```diff
@@ -972,6 +972,105 @@ async def test_copy_table_across(aconn_cls, dsn, faker, mode):
...
+async def test_copy_concurrency(aconn):
+    """
+    Test that copy operations hold the connection lock for the entire operation.
+
+    This test verifies the fix for the concurrency issue where AsyncCursor.copy()
+    was not holding the connection lock throughout the copy context, allowing
+    concurrent operations to interfere.
+    """
+    cur = aconn.cursor()
+    await cur.execute("create temp table copy_concurrency_test (id int, data text)")
```
dvarrazzo: To avoid the risk of a typo and using nonlocal variables I propose to create cursors only where needed, in the workers.


```suggestion
    await aconn.execute("create temp table copy_concurrency_test (id int, data text)")
```

tests/test_copy_async.py
```diff
@@ -972,6 +972,105 @@ async def test_copy_table_across(aconn_cls, dsn, faker, mode):
...
+    wrote_first = AEvent()
+    wrote_second = AEvent()
+    can_proceed = AEvent()
+
+    # Track execution order to verify workers run after copy completes
+    execution_log = []
+
+    async def copy_task():
+        """Copy task that writes two rows with controlled pauses."""
+        async with cur.copy("copy copy_concurrency_test from stdin") as copy:
```
dvarrazzo, line 997: Creating the cursor locally
```suggestion
        cur = aconn.cursor()
        async with cur.copy("copy copy_concurrency_test from stdin") as copy:
```

tests/test_copy_async.py
```diff
@@ -972,6 +972,105 @@ async def test_copy_table_across(aconn_cls, dsn, faker, mode):
...
+    t_worker3 = spawn(worker_task, (3,))
+
+    # Allow copy to exit
+    can_proceed.set()
+
+    # Wait for all tasks to complete
+    await gather(t_copy, t_worker1, t_worker2, t_worker3)
+
+    # Verify the data was written correctly
+    await cur.execute("select * from copy_concurrency_test order by id")
```
dvarrazzo: Creating a cursor on the fly, after workers worked.

```suggestion
    cur = await aconn.execute("select * from copy_concurrency_test order by id")
```

tests/test_copy_async.py
Zaczero: I simplified it to this:

```py
assert execution_log == [
    "entered_copy",
    "wrote_row_1",
    "wrote_row_2",
    "exited_copy",
    "worker_completed",
    "worker_completed",
    "worker_completed",
]
```

The test shouldn't verify the order of worker completions, just that they finish after the copy.

dvarrazzo: That's perfectly good, thank you. It is what I would have actually written if I had written the test myself :)

tests/test_copy_async.py
```diff
@@ -17,11 +17,19 @@
...
+    AsyncFileWriter,
+    ensure_table_async,
+    py_to_raw,
+    sample_binary_rows,
+    sample_records,
+    sample_tabledef,
+    sample_text,
+    sample_values,
+    special_chars,
+)
```
dvarrazzo: This change is spurious, isn't it?

I tried to re-run isort, as well as upgrading and running it again, and it doesn't seem generated my side. Where does it come from? It's part of 68f004e6df0c04f8dcb96f57ae946e1d3d909bab

Zaczero:
I would prefer you to simply squash the PR during merge, you can then also name the squashed commit appropriately. This avoids dealing with history rewrites and works just the same (is cleaner).

~~As per isort, ...~~

Edit: I just saw the other comment. I'm on mobile so GitHub ergonomics are bad.

dvarrazzo:
> I would prefer you to simply squash the PR during merge, you can then also name the squashed commit appropriately. This avoids dealing with history rewrites and works just the same (is cleaner).

I will commit this with a merge commit, even if eventually it will be a single commit. GitHub interface doesn't have a squash and merge with merge commit option. No problem to do it myself on my desktop though.

> I simply assumed the existing isort config would sort the imports correctly to match project style.

Yes, that's the case: isort shouldn't have changed that file using the project config, and it didn't get changed in your previous ones.

You can leave it to me and I'll make these changes myself prior to merge. Thabk you for your contribution 🙂

tests/test_copy_async.py
Zaczero: My best guess is that something went wrong during my attempts to set up .venv with `uv` (I didn't have libpq installed). I also initially forgot that psycopg has custom isort formatter and just run the defaults. There are a few places where it could have happened but I don't really know at this point.

Zaczero:
> GitHub interface doesn't have a squash and merge with merge commit option.





Is that not it?

dvarrazzo:
> Is that not it?

No, not quite the same. You cannot merge it this way:
```
(.venv) piro@sherekhan:~/dev/psycopg3$ git lo -3 --graph
*   13c1d704 (HEAD -> master, origin/master, origin/HEAD) Merge pull request #1210
|\  
| * 193e9d92 (zaczero/Zaczero-patch-1) fix; keep the lock for the entire duration of copy
|/  
* b9d533be (zaczero/master) chore(deps): bump the actions group with 2 updates
```
I've done it. Thank you very much!

Issue #1095, opened 2025-05-19
Support free-threaded python


wlach:
Currently psyopg doesn't support free threaded builds of Python (i.e. running without the gil). From https://github.com/psycopg/psycopg/issues/889#issuecomment-2409002394 it's not clear when/if this will happen, but filing this ticket so we can track the issue (I have some investigations and follow-ups I'll post in a comment).

wlach:
At the PyCon 2025 I attempted to get tests working with free-threaded Python (3.13.3). I didn't get super far but putting this down in case it's helpful to someone else.

What I did:

* Set up a test environment locally
* Relaxed the Cython dependency to allow Cython 3.1 (required to enable free-threading)
* Specified that the extension module was freethreading compatible
* Installed [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) and tried running the tests via `py.test --parallel-threads=2`. This is recommended by https://py-free-threading.github.io/debugging/#pytest-plugins-to-discover-concurrency-issues

What I found:

* I had some difficulty running some of the tests locally (I think due to issues with my environment rather than any free threading bugs)
* There was some use of global state in `test_break_attempts` which failed because it used global state (`sys.stderr`) and threading.
* Running tests in parallel (i.e. `--parallel-threads=N` where N>=2) causes many failures. I'm pretty sure this is just because the tests currently assume serial access to the database. Fixing this is more than I can take on today.

laggron42:
Are there news on this? Looks like this hasn't moved since May, and now that 3.14 is right around the corner with free-threaded out of experimental, this would be great to have

dvarrazzo:
Hello,

I am not entirely knowledgeable about free-threaded Python and what is the scope of this task.

What constitutes free-threaded support? I understand there isn't a tag for wheel release yet. Is the goal to just pass tests in a free-threaded Python image?

I understand that making the tests parallel could be a good indication of correct free-threading support. Is there an expectation here that the test suite would be made parallel? This is likely a major task.

lysnikolaou:
Hey @dvarrazzo!

I'm planning to take a closer look at free-threaded support for `psycopg` over the next few weeks and wanted to confirm that this is actually helpful and useful for you.

I've recently worked on `sqlalchemy` as well and the plan in general is the following:
- Clearly define guarantees for what or cannot be shared across multiple threads (e.g. SQLAlchemy allows sharing an `Engine` object, but not a `Connection` object)
[...]

PR #1211, opened 2025-11-17
Add more multithreaded tests for Connection and Cursor and enable free-threaded Python CI

lysnikolaou:
I've added some more multithreaded tests for the `Connection` and `Cursor` classes after going through the docs and trying to figure out what the thread safety guarantees are for each one.

I've also enabled CI for free-threaded Python and skipped the tests that rely on behavior that's not expected under the free-threaded built. Please let me know if I can do anything that'll make the review of this and follow-up PRs easier.

Related to #1095.

### DIFF
--- a/tests/fix_gc.py
+++ b/tests/fix_gc.py
@@ -72,9 +72,7 @@ def fixture_gc():
 
     **Note:** This will skip tests on PyPy.
     """
-    if sys.implementation.name == "pypy" or bool(
-        sysconfig.get_config_var("Py_GIL_DISABLED")
-    ):
+    if sys.implementation.name == "pypy" or sysconfig.get_config_var("Py_GIL_DISABLED"):
         pytest.skip(reason="depends on refcount semantics")
     return GCFixture()
 
@@ -86,6 +84,6 @@ def gc_collect():
 
     **Note:** This will *not* skip tests on PyPy.
     """
-    if bool(sysconfig.get_config_var("Py_GIL_DISABLED")):
+    if sysconfig.get_config_var("Py_GIL_DISABLED"):
         pytest.skip(reason="depends on refcount semantics")
     return GCFixture.collect
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -206,6 +206,6 @@ def asyncio_run(coro: Any, *, debug: bool | None = None) -> Any:
 
 def skip_free_threaded(reason="unsafe under the free-threaded build"):
     return pytest.mark.skipif(
-        bool(sysconfig.get_config_var("Py_GIL_DISABLED")),
+        sysconfig.get_config_var("Py_GIL_DISABLED"),
         reason=reason,
     )


tests/pool/test_pool.py
```diff
@@ -861,6 +862,10 @@ def test_check_max_lifetime(dsn):
 
 
 @pytest.mark.slow
+@pytest.mark.skipif(
+    bool(sysconfig.get_config_var("Py_GIL_DISABLED")),
+    reason="timing not accurate under the free-threaded build",
+)
```
dvarrazzo: Can we wrap this decorator in a function in `tests/utils.py` so that this can become:

```suggestion
@skip_free_thread("timing not accurate under the free-threaded build")
```

tests/fix_gc.py
```diff
@@ -71,16 +72,18 @@ def fixture_gc():
...
 @pytest.fixture
-def gc_collect():
+def gc_collect(gc):
     """
     Provides a consistent way to run garbage collection.
 
     **Note:** This will *not* skip tests on PyPy.
     """
-    return GCFixture.collect
+    return gc.collect
```
dvarrazzo: If you make this change, then the message in the docstring:
```
    **Note:** This will *not* skip tests on PyPy.
```
will not be true anymore, right?

tests/test_concurrency.py
```diff
@@ -565,3 +568,169 @@ def run_process():
     assert proc.returncode != 0
     assert "KeyboardInterrupt" in stderr
     assert stdout == ""
+
+
+@pytest.mark.usefixtures("testctx")
+def test_concurrent_connection_insert(conn):
```
dvarrazzo: All these tests are probably slow (in relative time, i.e. they take more than 100ms). Please mark them with the `@pytest.mark.slow` decorator.

tests/test_concurrency.py
```diff
@@ -565,3 +568,169 @@ def run_process():
...
+
+
+def test_same_cursor_from_multiple_threads_no_crash(conn):
+    nthreads = 10
+    barrier = threading.Barrier(parties=nthreads)
+
+    cur = conn.cursor()
+
+    def worker():
+        """Multiple threads trying to use the same cursor"""
```
dvarrazzo: Cursors are not designed to be shared by thread, we can't guarantee any specific behaviour here.

What is the reason to have this test? Only as a basic check that things don't explore?

tests/test_concurrency.py
```diff
@@ -565,3 +568,169 @@ def run_process():
     assert proc.returncode != 0
     assert "KeyboardInterrupt" in stderr
     assert stdout == ""
+
```
dvarrazzo: This sets of tests is specific to prove that free-threading concurrency works as expected.

Can we create a new unit test file and put them there to have a place where to maintain them, add other tests, remember why that test was added, etc?

dvarrazzo (changes requested):
Thank you for this tests, it seems a good job. I've requested some changes but it's small stuff.

Do you have a way forward planned in order to introduce official free-threading support?

tests/fix_gc.py
lysnikolaou: That's correct, but it seems like GCFixture.collect also calls `gc.collect` which depends on GC semantics that may be different on PyPy. Not sure how these tests are succeeding on PyPy, probably I'm missing something.

I'll revert.

tests/test_concurrency.py
lysnikolaou: Yeah, that why all exceptions are ignored. It's basically to verify that there's no hard crashes. I contemplated whether I should add this or not but though we could discuss it here.

dvarrazzo: It's an ok test: let's keep it. Thank you!

tests/fix_gc.py
```diff
@@ -71,7 +72,9 @@ def fixture_gc():
 
     **Note:** This will skip tests on PyPy.
     """
-    if sys.implementation.name == "pypy":
+    if sys.implementation.name == "pypy" or bool(
```
dvarrazzo: nit: `bool()` is redundant in these contexts, as the expression is implicitly evaluated as a boolean.

This is just a small detail: I can fix these myself because, for the rest, the MR is good as it is and I am rebasing/squashing it 🙂

PR #1214, opened 2025-11-20
Handle CancelledError to avoid losing connections

### LINKED ISSUE #1123: Connections not properly discarding when terminated by database after using check method (AsyncConnectionPool)
**I'm noticing the following behavior with AsyncConnectionPool**
- Open pool (e.g. min size 1, max size 5)
- Pool grows to 2
- Requests work for a while (I can spam them and it handles it fine)
- Eventually, database terminates the connection for being idle or alive too long
- Very next request hangs for a while (due to pool deciding to grow to 3), eventually works, and all subsequent requests keep timing out/never get a connection

**This has been consistently reproducible for me. I have the pool implemented with a check method for the connection and I'm simply terminating the connections on the database via pg_terminate_backend**

Here are some example logs as well, for when the error occurs ("checking connection" and "Connection check failed" both come from inside my check method)
```
INFO:psycopg.pool:connection requested from 'pool-1'
checking connection
Connection check failed: terminating connection due to administrator command
INFO:psycopg.pool:connection failed check: Bad connection
INFO:psycopg.pool:not serving connection found broken
checking connection
Connection check failed: terminating connection due to administrator command
INFO:psycopg.pool:connection failed check: Bad connection
INFO:psycopg.pool:not serving connection found broken
INFO:psycopg.pool:growing pool 'pool-1' to 3
ERROR:root:Error querying database: couldn't get a connection after 30.00 sec
ERROR:root:stats in pid: 1863 are {'requests_num': 8, 'requests_queued': 2, 'connections_num': 2, 'connections_ms': 215, 'requests_wait_ms': 29147, 'usage_ms': 78, 'connections_lost': 2, 'requests_errors': 1, 'pool_min': 1, 'pool_max': 5, 'pool_size': 3, 'pool_available': 0, 'requests_waiting': 1}
```

It seems like something is not being cleaned up within the pool when a connection is terminated on the database side. I would expect the behavior of implementing the "check" method to handle this and discard the connection but also open a new spot in the pool (e.g. pool_size should shrink and yet it doesn't)

It seems to me like the code basically:

1. Checks the connection within .connection() --> getconn() --> _getconn_with_check_loop()
2. That fails and _put_conn is called which calls _return_connection
3. The connection state is reset. In this case, the state is unknown when I run into this issue. Nothing happens in reset (which I think is fine)
4. Then, in _return_connection, from_getconn is true and if conn.pgconn.transaction_status == TransactionStatus.UNKNOWN: is too
5. It then increments _CONNECTIONS_LOST and runs a task to AddConnection

Something in that set of steps seems to be failing to free up space in the pool, causing the behavior I'm experiencing. Let me know if you're unable to reproduce but it happens consistently when I try this and is preventing me from being able to use the pools properly!

### LINKED ISSUE #1208: AsyncConnectionPool: connections can leak on CancelledError or  (gRPC task cancellation)
We’re using `psycopg` + `psycopg_pool.AsyncConnectionPool` in a gRPC server (Python asyncio).
gRPC cancels handlers by raising `asyncio.CancelledError` in the task.

We’ve hit a situation where connections “leak” from the async pool under high load and frequent cancellations: the pool eventually stops handing out connections (timeouts), while some connections are later closed by the garbage collector instead of being cleanly returned to the pool.

The core problem is that `CancelledError` is a subclass of `BaseException`, not `Exception`, and isn’t caught by the places in psycopg / psycopg_pool that handle connection lifecycle.

In our case this consistently happens during the `check_connection` callback we pass to the pool.
When a gRPC request is cancelled, the `asyncio.CancelledError` is raised inside that callback, but since psycopg_pool internally catches only `Exception`, the cleanup code that would normally return the connection to the pool is skipped.
[pool_async.py#L255](https://github.com/psycopg/psycopg/blob/master/psycopg_pool/psycopg_pool/pool_async.py#L255)

As a result, the connection remains marked as “checked out” and is eventually closed later by the garbage collector.

Looking at the implementation of `AsyncConnectionPool`, it seems the same issue could also happen inside `getconn()` or `putconn()` if a `CancelledError` occurs while these methods are running — for example, during context manager exit.
[pool_async.py#L215](https://github.com/psycopg/psycopg/blob/master/psycopg_pool/psycopg_pool/pool_async.py#L215)

To prevent this class of leaks, it might make sense to explicitly handle `CancelledError` (for example, by catching `BaseException` and checking `isinstance(exc, asyncio.CancelledError)`), or to wrap critical sections such as connection acquisition and return in `asyncio.shield()` to ensure they complete even if the parent task is cancelled.

dvarrazzo:
This should solve the problem of connections lost from the pool when a check operation is cancelled.

Fix #1123 
Fix #1208

### DIFF
--- a/psycopg_pool/psycopg_pool/pool.py
+++ b/psycopg_pool/psycopg_pool/pool.py
@@ -15,7 +15,6 @@ from abc import ABC, abstractmethod
 from time import monotonic
 from types import TracebackType
 from typing import Any, Generic, Iterator, cast
-from asyncio import CancelledError
 from weakref import ref
 from contextlib import contextmanager
 
@@ -30,6 +29,8 @@ from .errors import PoolClosed, PoolTimeout, TooManyRequests
 from ._compat import Deque, Self
 from ._acompat import Condition, Event, Lock, Queue, Worker, current_thread_name
 from ._acompat import gather, sleep, spawn
+
+CLIENT_EXCEPTIONS = Exception
 
 logger = logging.getLogger("psycopg.pool")
 
@@ -210,7 +211,7 @@ class ConnectionPool(Generic[CT], BasePool):
             conn = self._getconn_unchecked(deadline - monotonic())
             try:
                 self._check_connection(conn)
-            except (Exception, CancelledError):
+            except CLIENT_EXCEPTIONS:
                 self._putconn(conn, from_getconn=True)
             else:
                 logger.info("connection given by %r", self.name)
@@ -248,7 +249,7 @@ class ConnectionPool(Generic[CT], BasePool):
         if not conn:
             try:
                 conn = pos.wait(timeout=timeout)
-            except BaseException:
+            except CLIENT_EXCEPTIONS:
                 self._stats[self._REQUESTS_ERRORS] += 1
                 raise
             finally:
@@ -284,7 +285,7 @@ class ConnectionPool(Generic[CT], BasePool):
             return
         try:
             self._check(conn)
-        except BaseException as e:
+        except CLIENT_EXCEPTIONS as e:
             logger.info("connection failed check: %s", e)
             raise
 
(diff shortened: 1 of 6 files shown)


dvarrazzo:
Merging and finally releasing a new pool package fixing the issue! Wheeee!!!

PR #1221, opened 2025-11-28
Mark external libpq C functions as noexcept nogil

lysnikolaou:
Add noexcept nogil declarations to libpq bindings and endian conversion functions. This improves things for when we start using critical sections inside Cython-defines types. It's also more accurate this way, instead of only marking the APIs that are explicitly used in a nogil block.

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/endian.pxd
+++ b/psycopg_c/psycopg_c/_psycopg/endian.pxd
@@ -147,17 +147,17 @@ cdef extern from * nogil:
 
 #endif
     """
-    cdef uint16_t htobe16(uint16_t host_16bits)
-    cdef uint16_t htole16(uint16_t host_16bits)
-    cdef uint16_t be16toh(uint16_t big_endian_16bits)
-    cdef uint16_t le16toh(uint16_t little_endian_16bits)
+    cdef uint16_t htobe16(uint16_t host_16bits) noexcept nogil
+    cdef uint16_t htole16(uint16_t host_16bits) noexcept nogil
+    cdef uint16_t be16toh(uint16_t big_endian_16bits) noexcept nogil
+    cdef uint16_t le16toh(uint16_t little_endian_16bits) noexcept nogil
 
-    cdef uint32_t htobe32(uint32_t host_32bits)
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Hello, thank you for this!

Could you please review if the notice receiver is properly set up and isolated? Key points are:

1) on PGconn creation we install a Cython callback

https://github.com/psycopg/psycopg/blob/e016fd8354190f03a1471e7b84564e32e4b7dc7d/psycopg_c/psycopg_c/pq/pgconn.pyx#L47-L55

2) The callback verifies if there is a Python callback and, if there is, calls it. As a consequence, it re-acquires the GIL.

https://github.com/psycopg/psycopg/blob/e016fd8354190f03a1471e7b84564e32e4b7dc7d/psycopg_c/psycopg_c/pq/pgconn.pyx#L752-L763

3) Usually a PGconn callback is registered on Connection creation, which allows to call more than one Python callbacks registered in a list:

https://github.com/psycopg/psycopg/blob/e016fd8354190f03a1471e7b84564e32e4b7dc7d/psycopg/psycopg/_connection_base.py#L340-L353

Maybe it's worth adding a multithread test exercising the notice handler heavily? One way to do it is to use the `do` statement to execute a snippet of PL/pgSQL raising a notice:

```python
In [1]: import psycopg

In [2]: conn = psycopg.connect()

In [4]: conn.add_notice_handler(lambda res: print(res.message_primary))

In [5]: conn.execute("""
   ...: do $$
   ...: begin
   ...:     raise notice 'hello world';
   ...: end
   ...: $$ language plpgsql;
   ...: """)
hello world
Out[5]: <psycopg.Cursor [COMMAND_OK] [INTRANS] (database=piro) at 0x7f3e29e09170>
```

lysnikolaou:
Sure, I'll check all of these. But should we do it in a follow-up PR? The `nogil noexcept` marking should not affect any of these at all. The callbacks can be reenabling the GIL. The APIs themselves (like `libpq.PQsetNoticeReceiver`) do not reenable the GIL however and do not interact with Python.

dvarrazzo:
Hmm... ok, I'll merge this as it is, thank you :slightly_smiling_face:

PR #1231, opened 2025-12-04
fix: adapters: avoid race condition when replacing class name with itself

### LINKED ISSUE #1230: adapters: race condition in get_dumper() when replacing class name with itself
We've noticed these at program startup when using multiple threads:

```
psycopg.ProgrammingError: cannot adapt type 'UUID' using placeholder '%s' (format: AUTO)
```

My guess is that the code in https://github.com/psycopg/psycopg/blob/6a24300ad2392e502a7c9d3c244ead6852890d90/psycopg/psycopg/_adapters_map.py#L218-L223 is not thread-safe, as `dmap[scls] = dmap.pop(fqn)` is not atomic.

To reproduce:

```
% mkdir /tmp/psycopg-adapter-get-dumper-race; cd /tmp/psycopg-adapter-get-dumper-race
% venv/bin/pip install --quiet --force ~/src/psycopg/psycopg ~/src/psycopg/psycopg_c 
% cat <<EOF > race_get_dumper_uuid.py
import sys
import threading
import uuid

import psycopg


def test_uuid_adapter() -> bool:
    conn = psycopg.connect()
    res = conn.execute("select gen_random_uuid() = %s", [uuid.uuid4()]).fetchone()[0]


if __name__ == "__main__":
    sys.setswitchinterval(0.00001)

    threads = [threading.Thread(target=test_uuid_adapter) for _ in range(5)]
    for t in threads:
        t.start()

    for t in threads:
        t.join()
EOF
% pg_virtualenv 
Creating new PostgreSQL cluster 18/regress ...
% while ./venv/bin/python race_get_dumper_uuid.py; do : ; done ; # wait a few seconds / minutes
Exception in thread Thread-4 (test_uuid_adapter):
  File "/usr/lib/python3.13/threading.py", line 1043, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "/usr/lib/python3.13/threading.py", line 994, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/psycopg-adapter-get-dumper-race/race_get_dumper_uuid.py", line 10, in test_uuid_adapter
    res = conn.execute("select gen_random_uuid() = %s", [uuid.uuid4()]).fetchone()[0]
          ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/psycopg-adapter-get-dumper-race/venv/lib/python3.13/site-packages/psycopg/connection.py", line 299, in execute
    raise ex.with_traceback(None)
psycopg.ProgrammingError: cannot adapt type 'UUID' using placeholder '%s' (format: AUTO)
```

(I'm using `sys.setswitchinterval` here to try and trigger the bug faster)

noirbee:
Another thread may be switched to between the dmap.pop(fqn) instruction and the dmap[scls] one, typically at program startup when multiple threads are making their "first" queries.

Close #1230

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -10,6 +10,12 @@
 Current release
 ---------------
 
+Psycopg 3.3.2
+^^^^^^^^^^^^^
+
+Fix race condition in adapters at startup (:ticket:`#1230`).
+
+
 Psycopg 3.3.1
 ^^^^^^^^^^^^^
 
--- a/psycopg/psycopg/_adapters_map.py
+++ b/psycopg/psycopg/_adapters_map.py
@@ -217,9 +217,10 @@ class AdaptersMap:
 
             # If the adapter is not found, look for its name as a string
             fqn = scls.__module__ + "." + scls.__qualname__
-            if fqn in dmap:
+            if (d := dmap.get(fqn)) is not None:
                 # Replace the class name with the class itself
-                d = dmap[scls] = dmap.pop(fqn)
+                dmap[scls] = d
+                dmap.pop(fqn, None)
                 return d
 
         format = PyFormat(format)


docs/news.rst
```diff
@@ -14,6 +14,7 @@ Psycopg 3.3.1
 ^^^^^^^^^^^^^
 
 Fix iteration on server-side cursors (:ticket:`#1226`).
+Fix race condition in adapters at startup (:ticket:`#1230`).
```
dvarrazzo: This will go in 3.3.2

noirbee: Obviously, that's my bad ! Thanks for the quick review; I moved this into its own new 3.3.2 section.

PR #1233, opened 2025-12-08
Use critical section to protect pgconn ptr

lysnikolaou:
This PR adds critical sections to the PGconn class so that race condition between calling into the libpq bindings and calling `close` which sets the `_pgconn_ptr` to NULL are eliminated.

Also add one high-level and one low-level tests that trigger TSAN warnings without this PR.

### DIFF
--- a/psycopg_c/psycopg_c/pq/pgconn.pyx
+++ b/psycopg_c/psycopg_c/pq/pgconn.pyx
@@ -288,10 +288,9 @@ cdef class PGconn:
     def exec_(self, const char *command) -> PGresult:
         cdef libpq.PGresult *pgresult
         cdef libpq.PGconn *pgconn_ptr
-        with cython.critical_section(self):
-            with nogil:
-                if (pgconn_ptr := self._pgconn_ptr) is not NULL:
-                    pgresult = libpq.PQexec(pgconn_ptr, command)
+        with cython.critical_section(self), nogil:
+            if (pgconn_ptr := self._pgconn_ptr) is not NULL:
+                pgresult = libpq.PQexec(pgconn_ptr, command)
         if pgconn_ptr is NULL:
             raise e.OperationalError("the connection is closed")
         if pgresult is NULL:
@@ -302,10 +301,9 @@ cdef class PGconn:
     def send_query(self, const char *command) -> None:
         cdef int rv
         cdef libpq.PGconn *pgconn_ptr
-        with cython.critical_section(self):
-            with nogil:
-                if (pgconn_ptr := self._pgconn_ptr) is not NULL:
-                    rv = libpq.PQsendQuery(pgconn_ptr, command)
+        with cython.critical_section(self), nogil:
+            if (pgconn_ptr := self._pgconn_ptr) is not NULL:
+                rv = libpq.PQsendQuery(pgconn_ptr, command)
         if pgconn_ptr is NULL:
             raise e.OperationalError("the connection is closed")
         if not rv:
@@ -330,12 +328,11 @@ cdef class PGconn:
 
         cdef libpq.PGconn *pgconn_ptr
         cdef libpq.PGresult *pgresult
-        with cython.critical_section(self):
-            with nogil:
-                if (pgconn_ptr := self._pgconn_ptr) is not NULL:
-                    pgresult = libpq.PQexecParams(
-                        pgconn_ptr, command, <int>cnparams, ctypes,
-                        <const char *const *>cvalues, clengths, cformats, result_format)
+        with cython.critical_section(self), nogil:
+            if (pgconn_ptr := self._pgconn_ptr) is not NULL:
+                pgresult = libpq.PQexecParams(
+                    pgconn_ptr, command, <int>cnparams, ctypes,
+                    <const char *const *>cvalues, clengths, cformats, result_format)
         _clear_query_params(ctypes, cvalues, clengths, cformats)
         if pgconn_ptr is NULL:
             raise e.OperationalError("the connection is closed")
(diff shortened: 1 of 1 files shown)


psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -502,73 +604,104 @@ cdef class PGconn:
...
                 f"setting nonblocking failed: {self.get_error_message()}")
 
     cpdef int flush(self) except -1:
-        if self._pgconn_ptr == NULL:
-            raise e.OperationalError("flushing failed: the connection is closed")
-        cdef int rv = libpq.PQflush(self._pgconn_ptr)
+        cdef int rv
+        with cython.critical_section(self):
+            if self._pgconn_ptr is NULL:
+                raise e.OperationalError("flushing failed: the connection is closed")
```
ngoldbaum: Shouldn't this happen outside the criitcal section like you've done elsewhere in this PR?

psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -502,73 +604,104 @@ cdef class PGconn:
...
+        with cython.critical_section(self):
+            rv = libpq.PQsetSingleRowMode(self._pgconn_ptr)
+        if not rv:
             raise e.OperationalError("setting single row mode failed")
 
     def set_chunked_rows_mode(self, size: int) -> None:
-        if not libpq.PQsetChunkedRowsMode(self._pgconn_ptr, size):
+        cdef int rv
+        with cython.critical_section(self):
+            rv = libpq.PQsetChunkedRowsMode(self._pgconn_ptr, size)
```
ngoldbaum: Sorry, I got confused an deleted some incorrect comments. Here's where the issue is with python overhead:

Image

psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -617,14 +755,17 @@ cdef class PGconn:
...
         stream = fdopen(fileno, b"w")
-        libpq.PQtrace(self._pgconn_ptr, stream)
+        with cython.critical_section(self):
+            libpq.PQtrace(self._pgconn_ptr, stream)
 
     def set_trace_flags(self, flags: Trace) -> None:
         _check_supported("PQsetTraceFlags", 140000)
-        libpq.PQsetTraceFlags(self._pgconn_ptr, flags)
+        with cython.critical_section(self):
+            libpq.PQsetTraceFlags(self._pgconn_ptr, flags)
```
ngoldbaum: here's another one:

Image

psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -86,173 +87,233 @@ cdef class PGconn:
...
+        return bool(_call_libpq_int(self, <conn_int_f>libpq.PQsslInUse))
 
     def exec_(self, const char *command) -> PGresult:
-        _ensure_pgconn(self)
         cdef libpq.PGresult *pgresult
-        with nogil:
-            pgresult = libpq.PQexec(self._pgconn_ptr, command)
+        cdef libpq.PGconn *pgconn_ptr
+        with cython.critical_section(self):
+            with nogil:
```
dvarrazzo: Does Cython support a syntax such as:
```python
        with cython.critical_section(self), nogil:
```
?

psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -423,58 +502,69 @@ cdef class PGconn:
...
 
     def close_prepared(self, const char *name) -> PGresult:
         _check_supported("PQclosePrepared", 170000)
-        _ensure_pgconn(self)
-        cdef libpq.PGresult *rv = libpq.PQclosePrepared(self._pgconn_ptr, name)
+        cdef libpq.PGresult *rv
+        cdef libpq.PGconn *pgconn_ptr
+        with cython.critical_section(self):
+            pgconn_ptr = self._pgconn_ptr
+            if pgconn_ptr is not NULL:
```
dvarrazzo: Does Cython support the walrus operator now?

```suggestion
            if (pgconn_ptr := self._pgconn_ptr) is not NULL:
```

psycopg_c/psycopg_c/pq/pgconn.pyx
```diff
@@ -502,73 +604,104 @@ cdef class PGconn:
...
         if not rv:
             raise e.OperationalError(
                 f"sending close prepared failed: {self.get_error_message()}"
             )
 
     def get_result(self) -> "PGresult" | None:
-        cdef libpq.PGresult *pgresult = libpq.PQgetResult(self._pgconn_ptr)
+        cdef libpq.PGresult *pgresult
+        with cython.critical_section(self):
+            pgresult = libpq.PQgetResult(self._pgconn_ptr)
```
dvarrazzo, line 599: In a few function you are not using the local pointer copy. Yes, it seems that in those functions we were not using `_ensure_pgconn()` previously, but I wonder if we shouldn't try to use a consistent pattern.

dvarrazzo:
You might want to rebase this MR on master to run the CI successfully (macos 13 runners were dropped).

lysnikolaou:
> Are the per-object locks getting released when the thread detaches from the interpreter?

Yes, the Cython docs also mention that and suggest not doing `with nogil` inside critical sections. I think there's two possibilities here:

1. Remove the `nogil` block and let the gil-enabled build run these with the GIL held. 
2. Document that calling `PGconn.finish` while there's queries currently executing is unsupported. @dvarrazzo will have a better idea if there's ever a usecase that needs this.

psycopg_c/psycopg_c/pq/pgconn.pyx
lysnikolaou: Not sure whether we should be doing additional work unless it's necessary. Are there specific APIs where a NULL connection pointer is okay while, in others, it's not?

dvarrazzo: Only at bug level. There were a couple of functions where a check for nullness was lacking, but I reported them upstream and got fixed.

So I think we are safe anyways. The calls that in some older libpq version were insecure are definitely protected by `_ensure_pgconn()`, so if those one are still protected after your refactoring we are ok.

ngoldbaum:
> Document that calling PGconn.finish while there's queries currently executing is unsupported. @dvarrazzo will have a better idea if there's ever a usecase that needs this.

Not just finish - all the spots where you're using `with.cython.critical_section(...), nogil` or spots where you're detacting from the runtime inside critical sections will release the locks in the critical section.

If it really isn't safe to call into the `libpq` API without any external synchronization, then releasing the GIL there is incorrect without some other synchronization in the extension. So the fix would be to add a per-connection lock (for example) and acquire that.

If it's safe to call into `libpq` with the GIL released in the sections where it's released right now, then I don't think adding critical sections will help. It adds overhead and doesn't add any safety plus it makes the code more complicated.

If the desire is only to prevent new issues that aren't possible on the GIL-enabled build, then I think also the correct thing is to remove the new critical sections you've added here that do `with nogil` inside.

lysnikolaou:
@dvarrazzo Friendly ping on this one.

dvarrazzo:
Hello @lysnikolaou! I sincerely apologise for the delay, it had very busy weeks.

Merged now, thank you very much!

Issue #446, opened 2022-11-30
Use mypyc to complie connection/cursor classes


dvarrazzo:
Profiling using py-spy is showing that psycopg is now nice and fast when it can stay down in the C level, but resurfacing back in Python has a visible impact.

Rewriting the whole classes in Cython would address the problem, but it's quite a major work.

[Mypyc](https://mypyc.readthedocs.io/) might fill the gap, making a good use of the extensive type annotations of the psycopg codebase.

CaselIT:
Note that cython can also compile pure python code for a moderate performance advantage. See for example [falcon](https://github.com/falconry/falcon)
I'm not sure how it compares with mypyc however wrt performance

tonycosentini:
I'm seeing this when profiling a workload switching from psycopg2 to psycopg3.

In particular, calling `description()` on cursor seems noticeably slower (this workload runs about 20% slower on psycopg3, although I'm not sure if it's all due to description calls).

Trace on 2.x (the timing numbers are not comparable, but description() on 3.x seems to be about 10x slower from my napkin math):
<img width="1060" height="237" alt="Image" src="https://github.com/user-attachments/assets/76de660d-8fd0-4409-bf9e-f625ba8ab1a1" />

Trace on 3.x:
<img width="1063" height="424" alt="Image" src="https://github.com/user-attachments/assets/083ff0a3-d543-424c-904f-d1b5e6150d73" />

dvarrazzo:
@tonycosentini There's also the case that psycopg 3 description is more informative. Also, running e.g. a 10ms query, what is the difference between the two `description` calls?

I don't remember if the `Column` attributes are calculated in a lazy way or eagerly on `__init__`. This might be optimized by leaving some more complex attributes to calcluate to the getattr, so that there would be saving if a user doesn't use them.

tonycosentini:
I'll try to isolate my example into a script that can reproduce it.

Here's what I saw with the only change being psycopg2 -> psycopg3 (and I guess also switching adapters in SQLAlchemy):
<img width="1779" height="685" alt="Image" src="https://github.com/user-attachments/assets/18550e4e-8938-4afd-9e83-600cbd0c7706" />

The workload in that screenshot is a bunch of small (like < 10ms roundtrip) reads on tables with a decent number of columns (like 50 or so).

I looked at `Column` a little it, currently it looks like most things are setup eagerly in `__init__`. I'll try to see if I can defer anything to get some gains.

PR #1240, opened 2025-12-26
Improve performance of BaseCursor.description().

tonycosentini:
This is opened after noticing [`description` is much slower in 3.x](https://github.com/psycopg/psycopg/issues/446#issuecomment-3688879410). I'm pretty unfamiliar with psycopg (and even more so with cython) so this might be more of a starting point for discussion.

## Changes

* Fetching the encoding (even though it's in a Python LRU cahce) was kind of expensive. Now it's cached in the cursor once its fetched once until a `reset()`.
* Cythonified the `Column` class so libpq can be called directly.
* Made the work to get names + types as lazy as possible.

## Benchmarking

Benchmark code (table has ~20 columns):
```
def benchmark_description(conn, iterations=10000):
    """Benchmark cursor.description access and return elapsed nanoseconds"""
    start_ns = time.perf_counter_ns()
    cur = conn.cursor()
    cur.execute("SELECT * FROM benchmark_table LIMIT 1")

    # Access description many times
    for _ in range(iterations):
        desc = cur.description

    cur.close()
    return time.perf_counter_ns() - start_ns
```

Rough timing from my machine:
* `psycopg2` (2.9.11) - 0.004170051 seconds (4170051 ns)
* `psycopg[c]` (3.3.2)  -  2.072374879 seconds (2072374879 ns)
* `psycopg[c]` (with this PR) - 0.395435509 seconds (395435509 ns)

## Testing

There are no API changes so the existing tests should be sufficient.


tonycosentini:
I'm closing this since while the performance change is noticeable - it's still orders of magnitude off from what psycopgp2 is doing. I don't think the the additional overhead/complexity is worth it for these small gains (but if anyone else wants to take it over, by all means go for it).

PR #1247, opened 2026-01-14
fix: retain pgconn on OperationalError

### LINKED ISSUE #1246: OperationalError.pgconn is None if password is needed
I use `OperationalError.pgconn.needs_password` in a script to prompt the user for his password if needed.  This no longer works as of 3.2.8 (specifically #1076) because attribute `pgconn` is always None.

Here's my reproducer:

```py
import psycopg

try:
    with psycopg.connect():
        pass
except psycopg.OperationalError as e:
    assert e.pgconn
    assert e.pgconn.needs_password
```

Running that script with a Postgres user who requires a password, but without providing that password, gives me this:

```
$ PGHOST=localhost PGPORT=5434 PGUSER=postgres python needs_password.py
  File "/home/ewie/src/psycopg-regress/needs_password.py", line 4, in <module>
    with psycopg.connect():
         ~~~~~~~~~~~~~~~^^
  File "/home/ewie/src/psycopg/psycopg/psycopg/connection.py", line 133, in connect
    raise enhanced_exception.with_traceback(None)
psycopg.OperationalError: connection failed: connection to server at "127.0.0.1", port 5434 failed: fe_sendauth: no password supplied
Multiple connection attempts failed. All failures were:
- host: None, port: None, hostaddr: '::1': connection failed: connection to server at "::1", port 5434 failed: fe_sendauth: no password supplied
- host: None, port: None, hostaddr: '127.0.0.1': connection failed: connection to server at "127.0.0.1", port 5434 failed: fe_sendauth: no password supplied

  File "/home/ewie/src/psycopg-regress/needs_password.py", line 7, in <module>
    assert e.pgconn
           ^^^^^^^^
AssertionError
```

ewie:
Fixes #1246.

When performing multiple connection attempts and all fail, a new `OperationalError` with the combined attempt errors is raised.  But the `pgconn` attribute from the original attempt errors is lost.  Fix this by including the `pgconn` from the last attempt error.

### DIFF
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -142,7 +142,8 @@ class AsyncConnection(BaseConnection[Row]):
             lines = [str(last_ex)]
             lines.append("Multiple connection attempts failed. All failures were:")
             lines.extend(f"- {descr}: {error}" for error, descr in conn_errors)
-            raise type(last_ex)("\n".join(lines)).with_traceback(None)
+            new_ex = type(last_ex)("\n".join(lines), pgconn=last_ex.pgconn)
+            raise new_ex.with_traceback(None)
 
         if (
             capabilities.has_used_gssapi()
--- a/tests/test_connection_async.py
+++ b/tests/test_connection_async.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import sys
 import time
 import logging
@@ -12,7 +13,7 @@ import psycopg
 from psycopg import errors as e
 from psycopg import pq
 from psycopg.rows import tuple_row
-from psycopg.conninfo import conninfo_to_dict, timeout_from_conninfo
+from psycopg.conninfo import conninfo_to_dict, make_conninfo, timeout_from_conninfo
 from psycopg._conninfo_utils import get_param
 
 from .acompat import asleep, skip_async, skip_sync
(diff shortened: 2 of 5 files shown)


psycopg/psycopg/connection_async.py
```diff
@@ -142,7 +142,9 @@ async def connect(
             lines = [str(last_ex)]
             lines.append("Multiple connection attempts failed. All failures were:")
             lines.extend(f"- {descr}: {error}" for error, descr in conn_errors)
-            raise type(last_ex)("\n".join(lines)).with_traceback(None)
+            raise type(last_ex)("\n".join(lines), pgconn=last_ex.pgconn).with_traceback(
+                None
+            )
```
dvarrazzo: Nit: it would be nicer on two lines:

```suggestion
            new_ex = type(last_ex)("\n".join(lines), pgconn=last_ex.pgconn)
            raise new_ex.with_traceback(None)
```

tests/test_connection_async.py
```diff
@@ -76,6 +76,27 @@ async def test_connect_error_multi_hosts_each_message_preserved(aconn_cls):
...
+        if not conn.pgconn.used_password:
+            pytest.skip("test connection needs no password")
+
+    with monkeypatch.context() as m:
+        try:
+            m.delenv("PGPASSWORD", raising=True)
+        except KeyError:
+            info = conninfo_to_dict(dsn)
+            del info["password"]  # should not raise per check above.
+            dsn = make_conninfo("", **info)
```
dvarrazzo: Good idea, but we need to make it a failing multi-attempt connection. We don't need to handle an explicit monkeypatch context because we don't need to access the pristine environment in the test after the block.

So my proposal is to drop the password and to specify the host twice:

```suggestion
    monkeypatch.delenv("PGPASSWORD", raising=False)
    info = conninfo_to_dict(dsn)
    info.pop("password", None)
    info["host"] = ",".join([str(info.get("host", os.environ.get("PGHOST", "")))] * 2)
    dsn = make_conninfo("", **info)
```

tests/test_connection_async.py
```diff
@@ -76,6 +76,27 @@ async def test_connect_error_multi_hosts_each_message_preserved(aconn_cls):
     assert any(expected_host2 in line and expected_error in line for line in msg_lines)
 
 
+async def test_operationalerror_pgconn(aconn_cls, dsn, monkeypatch):
```
dvarrazzo: The point of this test is to test with multiple attempts:

```suggestion
async def test_multi_attempt_error_pgconn(aconn_cls, dsn, monkeypatch):
```

psycopg/psycopg/connection_async.py
ewie: Done. Looks much nicer now.

tests/test_connection_async.py
ewie: Ah, it never occurred to me that I can specify the **same** host twice.

tests/test_connection_async.py
ewie: Done

ewie:
> Can you also please an entry to the `docs/news.rst` file?

Done with new section for 3.3.3.

> I am not sure if I am seeing this well in github interface: can you make sure to leave a blank space between the first and the second line in your commit message? I might be wrong but it seems it's missing.

No, the commit message was already formatted correctly, if you meant the standard blank line between subject and body.

ewie:
@dvarrazzo Thank you for reviewing and merging the fix.

dvarrazzo:
Thank you for finding the problem and helping to fix it!

PR #1250, opened 2026-01-26
Use `Literal` for `Range.bounds` parameter type hint

Viicos:
Given the added type ignores, this might not be worth it


dvarrazzo:
Yes, I don't see a reason to do this.  The bounds are a field computed from two values, if we have problems expressing `"["  + ")"` then this cannot go in a good direction.

PR #1256, opened 2026-02-10
Fix typo in tstrings error message and incorrect escape_identifier docs example

### LINKED ISSUE #421: > Yes, that example is wrong. The problem is that the correction is very ugly:
About the wrong example in escape_identifier: https://www.psycopg.org/psycopg3/docs/api/sql.html

> > Yes, that example is wrong. The problem is that the correction is very ugly:

> Maybe it should be removed then? Or replaced with a link to discussion #420 to attract more feedback?

Yes we should.

_Originally posted by @dlax in https://github.com/psycopg/psycopg/issues/410#issuecomment-1286586020_

veeceey:
## Summary

- **Fix error message in `_tstrings.py`**: The `ProgrammingError` raised when a non-Composable, non-Template value is used with the `:q` format specifier references the nonexistent class `sql.Composite`. This is corrected to `sql.Composable`, which is the actual base class defined in `psycopg.sql`.

- **Fix `docs/api/sql.rst` (closes #421)**: The documentation example for `escape_identifier()` passes a Python `str` to the function, but it actually requires `bytes` (a `Buffer`). As discussed in #421 and #410, the maintainers agreed the example should be removed since the corrected version is unwieldy. The section now mentions `escape_identifier()` exists as a low-level option but directs users to the `psycopg.sql` module as the recommended approach.

- **Fix typo in `conninfo.py`**: `re-exoprts` -> `re-exports`.

## Test plan

- [x] Verified no tests match on the exact error message string (tests use regex patterns like `"Template.*':q'"`)
- [x] Verified `sql.Composite` does not exist anywhere in the codebase (only `sql.Composable` and `types.composite.CompositeInfo`)
- [x] The docs change removes a code example that was provably incorrect (passing `str` to a `Buffer`-typed parameter)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/psycopg/psycopg/_tstrings.py
+++ b/psycopg/psycopg/_tstrings.py
@@ -83,7 +83,7 @@ class TemplateProcessor:
             elif fmt == FMT_SQL:
                 # It must have been processed already
                 raise e.ProgrammingError(
-                    "sql values must be sql.Composable, sql.SQL, or Template;"
+                    "sql values must be sql.Composed, sql.SQL, or Template;"
                     f" got {type(item.value).__qualname__}"
                     f" in {{{item.expression}:{fmt}}}"
                 )


psycopg/psycopg/_tstrings.py
```diff
@@ -83,7 +83,7 @@ def _process_template(self, t: Template) -> None:
             elif fmt == FMT_SQL:
                 # It must have been processed already
                 raise e.ProgrammingError(
-                    "sql values must be sql.Composite, sql.SQL, or Template;"
+                    "sql values must be sql.Composable, sql.SQL, or Template;"
```
dvarrazzo: No, it's really `Composite` here, it can't be `Composable`. You can't use a `sql.Identifier` as a query.

dvarrazzo: Ah sorry, it is [`Composed`](https://www.psycopg.org/psycopg3/docs/api/sql.html#psycopg.sql.Composed) 😅  "Composite" is a data type.

Let's fix it 💪 

```suggestion
                    "sql values must be sql.Composed, sql.SQL, or Template;"
```

veeceey: You're right, thank you for the correction! The class is indeed `sql.Composed`, not `sql.Composable` or `sql.Composite`. I'll apply your suggestion and push the fix.

dvarrazzo: Lovely, now I'm talking with claude.

Please amend the comment with a sarcastic comment such as "this MR was useful to fix a typo but is the clear demonstration of how much AI wastes people time in the hand of incompetent people pushing slop".

dvarrazzo: I've done it myself.

@veeceey next time you find a bug, open a bug report, not a merge request, if you don't know what you are doing and you need AI assistance. You waste other people's time and give a terrible image of yourself.

Thank you very much.

PR #1258, opened 2026-02-12
Fix typos

JohnnyDeuss:


docs/api/pool.rst
```diff
@@ -43,7 +43,7 @@ The `!ConnectionPool` class
 .. autoclass:: ConnectionPool
 
    This class implements a connection pool serving `~psycopg.Connection`
-   instances (or subclasses). The constructor has *alot* of arguments, but
+   instances (or subclasses). The constructor has *a lot* of arguments, but
```
dvarrazzo: This was a silly thing of mine but *alot* was written that way on purpose. However I guess time is passed and [the reference](https://hyperboleandahalf.blogspot.com/2010/04/alot-is-better-than-you-at-everything.html) is lost.

I assume "several" is more idiomatic?

```suggestion
   instances (or subclasses). The constructor has several arguments, but
```

can we also remove `alot` from [`pyproject.toml`](https://github.com/psycopg/psycopg/blob/09c89180f94606dc70475ed863e135f021a11038/pyproject.toml#L63
)?

dvarrazzo (changes requested):

docs/api/pool.rst
JohnnyDeuss: Haha, I definitely did not get the reference and appreciate the irony in it itself ending up in a pedantic typo PR. Heck, if it was intentional, I'd say leave it in. What's life without whimsy?

dvarrazzo: Ok, in this case let's remove this correction from the PR and I'll merge it :smile:

docs/basic/install.rst
```diff
@@ -73,13 +73,13 @@ summary of their differences; follow the links for more details.
    * - Installation Option
      - Command
      - Description
-     - Perfor­mance
+     - Performance
```
dvarrazzo: This correction should be removed too. I've spent several minutes squinting at the difference on github, failing to see anything. Then I looked at the source and there's a [soft hyphen](https://en.wikipedia.org/wiki/Soft_hyphen) there to make the column narrower, which is not displayed by the github webpacge.

Image

So please remove this change too.

dvarrazzo:
Merged, thank you!

PR #1263, opened 2026-02-20
Simplify local development with Taskfile

dutow:
This PR aims to simplify local development, especially for new developers

It contains two commits:
* The first makes `git commit` work without activating the venv, as it autodetects it, as I ran into this issue several times already
* The second adds a taskfile with basic commands for development related tasks (e.g. task setup, task test, tast docs:serve, etc)

What do you think?

## Make the pre-commit hooks work without the virtual env loaded
    
    Currently if somebody tries to do "git commit" with the hook set, but
    without the venv loaded it fails because it assumes everything is
    available in the path.
    
    This change makes it properly look up executables in the venv even in
    this case.
    
    It assumes that the virtual env is either in the .venv directory, or it
    is set using the VENV_DIR environment variable.

## Added a Taskfile with a short documentation
    
    This should make local development easier, especially for new
    contributors, who don't have to remember all the specific details with
    this.
    
    It's just task setup && task test, and that does everything.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,8 @@
+# Copy to .env to customize local development settings.
+# .env is gitignored and should not be committed.
+
+# Path to the Python virtual environment (default: .venv)
+#VENV_DIR=.venv
+
+# PostgreSQL connection string for tests (default: dbname=psycopg_test)
+#PSYCOPG_TEST_DSN=dbname=psycopg_test
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ __pycache__/
 /psycopg_binary/
 .vscode
 .venv
+.env
 .coverage
 htmlcov
 .idea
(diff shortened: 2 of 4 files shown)


dvarrazzo:
Hello

you cannot just come here, drop a new development workflow, _your_ favourite workflow, and ride towards the sunset, without even checking if this is welcome towards the people who have written the 99.999% of this project.

I find this pretty impolite and it is an unwelcome push.

Next time you want to change the way **other** people work, please write them beforehand to propose something.

dutow:
Hello

I asked this is a question - but I think it's easier to see how it behaves with actual code instead of as a theoretical question, that's why I opened it with this, and the question:

> What do you think?

And I did not drop a new development workflow, this is just a wrapper so that people don't have to execute many commands when you start working on the repo. Task is not my favorite workflow, it could be make or anything else - the point was just to make it a simple "setup" and "test" command, instead of executing many things separately.

But if you don't like it, that's fine, as I said this was meant as a question.

PR #1264, opened 2026-02-20
WIP: Add infrastructure for the test suite to start up its own Postgres instance

dutow:
Note: this is work in progress, but a requirement for proper OAuth testing. For an actual use of this code, see my oauth branch, which contains OAuth support for psycopg with a basic test suite: https://github.com/dutow/psycopg/tree/oauth

This PR is also missing parts related to #1263, which would make its usage even easier. (but these additional changes are present in the oauth branch)

I mainly created this PR as a discussion if you think this is a good approach or not - OAuth, and also some other PG19 features require custom server configurations, which is very difficult to test without the test suite managing the postgresql instance. (PG19: connection warnings support is already merged, requires specific system GUCs and users to actually do something)

And as I mentioned in the second commit message, I  think this have many advantages, even without that:

1. It doesn't require a separately running Postgres, together with the Taskfile this simplifies the test setup even more, contributors don't have to manage a local postgres instance
2. No worry about remaining test artifacts: even if somebody has a locally running Postgres instance, cleanup / custom configuration can be a worry either for analyzing test failures, or for the test suite changing the server with leftovers.
3. No permission/configuration questions: if the tests start the instance, the test suite clearly has admin access to it, so we can do anything with it
4. Which is especially important for PG18 OAuth, and some in progress PG19 features like the connection warnings, where we have to modify the server configuration in order to test it.

### DIFF
--- a/tests/fix_db.py
+++ b/tests/fix_db.py
@@ -25,7 +25,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--test-dsn",
         metavar="DSN",
-        default=os.environ.get("PSYCOPG_TEST_DSN"),
+        default=os.environ.get("PSYCOPG_TEST_DSN") or None,
         help=(
             "Connection string to run database tests requiring a connection"
             " [you can also use the PSYCOPG_TEST_DSN env var]."
@@ -46,8 +46,13 @@ def pytest_addoption(parser):
 
 
 def pytest_report_header(config):
-    if (dsn := config.getoption("--test-dsn")) is None:
-        return []
+    dsn = config.getoption("--test-dsn")
+    if dsn is None:
+        from .fix_pginstance import find_pg_binary
+
+        if find_pg_binary("pg_ctl") and find_pg_binary("initdb"):
+            return ["PostgreSQL: managed instance (auto-detected pg_ctl/initdb)"]
+        return ["PostgreSQL: not available (no --test-dsn, pg_ctl/initdb not found)"]
 
     try:
         with psycopg.connect(dsn, connect_timeout=10) as conn:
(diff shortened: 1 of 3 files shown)


tests/fix_db.py
```diff
@@ -25,7 +25,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--test-dsn",
         metavar="DSN",
-        default=os.environ.get("PSYCOPG_TEST_DSN"),
+        default=os.environ.get("PSYCOPG_TEST_DSN") or None,
```
dvarrazzo, line 28: TEST_DSN as an empty string is a valid connection string.

dvarrazzo:
Can't we just configure the server in CI and skip the test if we detect that the settings of the current server are not suitable?

dutow:
> Can't we just configure the server in CI and skip the test if we detect that the settings of the current server are not suitable?

Yes, but then you would have to skip local testing, or instruct contributors/interested users about following a very specific setup, I think that would be a lot more fragile.

I can rework the oauth test suite to work that way, but I think it's nice if a project is very easy to build and test even for somebody not 100% familiar with it.

psycopg/psycopg/pq/misc.py
```diff
@@ -51,6 +51,10 @@ class PGresAttDesc(NamedTuple):
 
 @cache
 def find_libpq_full_path() -> str | None:
+    # Allow explicit override via environment variable.
+    if libname := os.environ.get("LIBPQ"):
```
dvarrazzo, line 55: This can be done with `LD_LIBRARY_PATH`

kfox1111:
Any updates on this?

dvarrazzo:
I do understand the convenience of running a postgres on its own, but the load of work to accommodate all the possible ways things can go wrong is excessive. You are asking us to do all the admin work that every single user testing the system would have to do on their machines, and each one of them will open a ticket to us if we have forgotten to manage a weird use case. And that's to run the tests, which is not the core of our project.

I would rather prefer to see a similar feature implemented using docker, for example we could have a Dockerfile for every supported database/version, including all the extensions and configuration we may want for testing purposes, and add pytest command line support to decide which one to start/stop and the possible command line configuration for docker itself (for example to allow using podman instead of docker, to use a host directory as volume for debugging purposes, to leave the server running after the tests, etc).

PR #1268, opened 2026-02-27
perf: Speed up `_split_query`

akx:
This PR speeds up `_split_query()` (which, it seems, is called for all queries that involve any sort of placeholder or variable (at least once, given the LRU caches for `_query2pg_*`), so pretty hot in general) by 30% to 66% in a micro-benchmark that exercises only that function. (I can also add the pytest-benchmark file if it could be useful.)

The optimization journey, if you will, is in atomic commits, but can be squashed into a single commit for `master`.

PR #1269, opened 2026-02-27
perf: speed up Python UUID converters

akx:
Sibling of #1268 (from the same viztracer trace... 😄).

This speeds up the Python UUID dumpers by roughly 20-25% by the benchmark included. (This is significant since there is no Cython speedup for UUID dumping. (I tried implementing that, and it ended up only about 5% faster than these, so possibly not worth it.))

Very negligible speed-up (0-0.2%; could be measurement noise) for the loaders, but still avoiding an extra local.

Same story as in #1268 for `tests/benchmarks/`.

### DIFF
--- a/psycopg/psycopg/types/uuid.py
+++ b/psycopg/psycopg/types/uuid.py
@@ -25,14 +25,14 @@ class UUIDDumper(Dumper):
     oid = _oids.UUID_OID
 
     def dump(self, obj: uuid.UUID) -> Buffer | None:
-        return obj.hex.encode()
+        return b"%032x" % obj.int
 
 
 class UUIDBinaryDumper(UUIDDumper):
     format = Format.BINARY
 
     def dump(self, obj: uuid.UUID) -> Buffer | None:
-        return obj.bytes
+        return obj.int.to_bytes(16, "big")
 
 
 class UUIDLoader(Loader):
@@ -43,18 +43,14 @@ class UUIDLoader(Loader):
             from uuid import UUID
 
     def load(self, data: Buffer) -> uuid.UUID:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-        return UUID(data.decode())
+        return UUID((bytes(data) if isinstance(data, memoryview) else data).decode())
 
 
 class UUIDBinaryLoader(UUIDLoader):
     format = Format.BINARY
 
     def load(self, data: Buffer) -> uuid.UUID:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-        return UUID(bytes=data)
+        return UUID(bytes=(bytes(data) if isinstance(data, memoryview) else data))
 
 
 def register_default_adapters(context: AdaptContext) -> None:


.pre-commit-config.yaml
```diff
@@ -40,7 +40,7 @@ repos:
         language: system
         entry: mypy --pretty --follow-imports=silent
         files: \.py[i]?$
-        exclude: tests/test_tstring.py
+        exclude: (tests/test_tstring\.py|tests/benchmarks/)
```
dvarrazzo: And why do you want to exclude it?

Either the code in this project respects the quality standard or it doesn't belong to this project.

I don't believe we need to commit a benchmark for this change.

akx: I couldn't get it to easily behave with the speedup module maybe or maybe not being available. I can remove the benchmark code from this PR.

psycopg/psycopg/types/uuid.py
```diff
@@ -25,14 +25,14 @@ class UUIDDumper(Dumper):
...
-        return obj.hex.encode()
+        return b"%032x" % obj.int
 
 
 class UUIDBinaryDumper(UUIDDumper):
     format = Format.BINARY
 
     def dump(self, obj: uuid.UUID) -> Buffer | None:
-        return obj.bytes
+        return obj.int.to_bytes(16, "big")
```
dvarrazzo, line 35: How can be calling a method on an attribute faster than reading an attribute?

psycopg/psycopg/types/uuid.py
```diff
@@ -25,14 +25,14 @@ class UUIDDumper(Dumper):
     oid = _oids.UUID_OID
 
     def dump(self, obj: uuid.UUID) -> Buffer | None:
-        return obj.hex.encode()
+        return b"%032x" % obj.int
```
dvarrazzo, line 28: Cute. A micro-benchmark confirms:

```python
>>> timeit.timeit("obj.hex.encode()", "import uuid; obj = uuid.uuid4()")
0.1317441969877109

>>> timeit.timeit("b'%032x' % obj.int", "import uuid; obj = uuid.uuid4()")
0.1017561670159921
```

psycopg/psycopg/types/uuid.py
```diff
@@ -43,18 +43,14 @@ def __init__(self, oid: int, context: AdaptContext | None = None):
             from uuid import UUID
 
     def load(self, data: Buffer) -> uuid.UUID:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-        return UUID(data.decode())
+        return UUID((bytes(data) if isinstance(data, memoryview) else data).decode())
```
dvarrazzo, line 46: If you really want to make this any faster, why don't you use the Writebla UUID trick used by the C code?

psycopg/psycopg/types/uuid.py
dvarrazzo: Ah I see: `bytes` is a property:
```python
>>> timeit.timeit("obj.bytes", "import uuid; obj = uuid.uuid4()")
0.08018396701663733

>>> timeit.timeit("obj.int.to_bytes(16, 'big')", "import uuid; obj = uuid.uuid4()")
0.05676913500064984
```

psycopg/psycopg/types/uuid.py
akx: Because we'd still need to go from hex-possibly-with-dashes-within to int somehow and that's probably the slow part.

Btw, regarding that trick: why not just `u = __new__(...)` and `object.__setattr__(u, "int", ...)` (with a cached reference to `object.__setattr__`)?

psycopg/psycopg/types/uuid.py
akx: Yep. `UUID.hex` is `"%032x" % obj.int`, but since we want bytes, we can just as well do that "by hand".

psycopg/psycopg/types/uuid.py
dvarrazzo: I don't understand the trick. If you have something to propose go ahead.

akx: I mean, `uuid.UUID.__init__` [directly calls `object.__setattr__(self, ...)`](https://github.com/python/cpython/blob/41fa2dbc0ef5232efae42630119ba20a2efb3ad7/Lib/uuid.py#L264-L265) because `uuid.UUID.__setattr__` would complain. 

I was wondering if there was something more subtle about `psycopg_c`'s trick (https://github.com/psycopg/psycopg/pull/998 for context here) -- why bother with the writable class instead of doing the C equivalent of `uuid.UUID.__new__` + `object.__setattr(x)`.

dvarrazzo: Maybe: this seems to work:
```python
import uuid
obj = object.__new__(uuid.UUID)
object.__setattr__(obj, 'int', 0)
object.__setattr__(obj, 'is_safe', uuid.SafeUUID.unknown)
obj
UUID('00000000-0000-0000-0000-000000000000')
```

dvarrazzo: Implemented in #1272. WDYT @akx?

akx:
@dvarrazzo Would you like me to revert the `load` tricks from this and just keep the `dump` speedups so we can get this merged?

dvarrazzo:
> @dvarrazzo Would you like me to revert the load tricks from this and just keep the dump speedups so we can get this merged?

The changes in load are pretty unconsequential, we can keep them. I need a few days to close some works before I can address this but those changes are not blockers.

PR #1272, opened 2026-03-01
refactor(c): drop class replacement hack to make UUID writable

dvarrazzo:
From a suggestion by @akx Aarni Koskela, see
https://github.com/psycopg/psycopg/pull/1269#discussion_r2869270196


dvarrazzo:
Uhm... The original idea in #998 was actually what reimplemented here: see https://github.com/psycopg/psycopg/pull/998#issuecomment-2619228112

Looking at comments in the MR it seems that class replacement was an optimization.

We should benchmark the two solutions.

akx:
Huh 😄 If a benchmark proves that the `WritableUUID` approach is faster, a PR should at least add a comment in the source next to this so the next person wondering what's up (indeed, as I said in https://github.com/psycopg/psycopg/pull/1269#discussion_r2869270196, "something more subtle about it") knows...

PR #1286, opened 2026-03-27
perf: Do less `if X in Y: return Y[X]`

akx:
Use more Python 3.11+ zero-cost exception handling (and technically less race conditions, I guess!)

### DIFF
--- a/psycopg/psycopg/_adapters_map.py
+++ b/psycopg/psycopg/_adapters_map.py
@@ -212,8 +212,10 @@ class AdaptersMap:
 
         # Look for the right class, including looking at superclasses
         for scls in cls.__mro__:
-            if scls in dmap:
+            try:
                 return dmap[scls]
+            except KeyError:
+                pass
 
             # If the adapter is not found, look for its name as a string
             fqn = scls.__module__ + "." + scls.__qualname__
--- a/psycopg/psycopg/_capabilities.py
+++ b/psycopg/psycopg/_capabilities.py
@@ -98,9 +98,9 @@ class Capabilities:
 
         The expletive messages, are left to the user.
         """
-        if feature in self._cache:
+        try:
             msg = self._cache[feature]
-        else:
+        except KeyError:
             msg = self._get_unsupported_message(feature, want_version)
             self._cache[feature] = msg
 
--- a/psycopg/psycopg/types/composite.py
+++ b/psycopg/psycopg/types/composite.py
@@ -206,8 +206,10 @@ class RecordBinaryLoader(Loader):
         return tx.load_sequence(record)
 
     def _get_transformer(self, key: tuple[int, ...]) -> abc.Transformer:
-        if key in self._txs:
+        try:
             return self._txs[key]
+        except KeyError:
+            pass
 
         tx = Transformer(self._ctx)
         tx.set_loader_types([*key], self.format)


dvarrazzo:
I understand that's more 0 cost setup, not 0 cost handling. So It's mostly suitable when the "in" path is verifiably taken more often and the exception path is... exceptional.

- did you look at the code you change and verified that in every single occurrence this is the case?
- do you have a benchmark of the difference in performance? Is it relevant? Is it a regression?
- in 3.3.x (current master, probably it will be dropped when we will fork the 3.4 dev branch) we still support Python 3.10. Are your changes a regression on this version?

akx:
I didn't measure each of these code paths individually, no. The `try: ... except KeyError: ...` style is used occasionally throughout the codebase anyway.

A microbenchmark (code in details block below) seems to show the try-except (or  style (`v2`)  is consistently faster for a cache pattern (where we most of the time expect the thing to be there):

```
~/Desktop $ uv run --python=3.10 x-in.py
v1 3.0709988749586046
v2 2.499070749967359
~/Desktop $ uv run --python=3.11 x-in.py
v1 3.0146106659667566
v2 2.3874160000123084
~/Desktop $ uv run --python=3.12 x-in.py
v1 2.2464891669806093
v2 1.6922132920008153
~/Desktop $ uv run --python=3.13 x-in.py
v1 3.0302311669802293
v2 2.410757082980126
~/Desktop $ uv run --python=3.14 x-in.py
v1 2.0428653330309317
v2 1.4995768750086427
```


<details>

```python
import timeit

d = {}


def c(a):
    return 2 << a


def v1(a):
    if a in d:
        return d[a]
    else:
        d[a] = r = c(a)
        return r


def v2(a):
    try:
        return d[a]
    except KeyError:
        d[a] = r = c(a)
        return r


def clear_try(f):
    d.clear()
    return timeit.timeit(lambda: [f(x) for x in range(500)], number=100000)


assert [v1(x) for x in range(500)] == [v2(x) for x in range(500)]
print("v1", clear_try(v1))
print("v2", clear_try(v2))
```

</details>

However, good call on questioning this in all cases: FloatDumper, for the usual case where we're actually dumping floats most of the time and not infinities etc., suffers from this construction. I'll revert that bit.

```
-------------------------------- benchmark: 3 tests, 2 sources --------------------------------
Name (time in us)            0001_master OPS (Kops/s)  0002_keyerro OPS (Kops/s)  ΔOPS (Kops/s)
-----------------------------------------------------------------------------------------------
test_quote_all_special                         6.5016                     7.1571         +10.1%
test_quote_mostly_normal                       2.4346                     2.0018         -17.8%
test_quote_all_normal                          2.4004                     1.9472         -18.9%
-----------------------------------------------------------------------------------------------
```

<details>

```python
import math
import random

import pytest

from psycopg.types.numeric import FloatDumper

FLOATS_MOSTLY_NORMAL = (
    [random.uniform(-1e6, 1e6) for _ in range(970)]
    + [float("inf")] * 10
    + [float("-inf")] * 10
    + [float("nan")] * 10
)
random.shuffle(FLOATS_MOSTLY_NORMAL)

FLOATS_ALL_NORMAL = [random.uniform(-1e6, 1e6) for _ in range(1000)]
FLOATS_ALL_SPECIAL = [float("inf"), float("-inf"), float("nan")] * 334


@pytest.fixture
def dumper():
    return FloatDumper(float)


def bench_quote(benchmark, dumper, values):
    def run():
        q = dumper.quote
        for v in values:
            q(v)

    benchmark(run)


def test_quote_mostly_normal(benchmark, dumper):
    bench_quote(benchmark, dumper, FLOATS_MOSTLY_NORMAL)


def test_quote_all_normal(benchmark, dumper):
    bench_quote(benchmark, dumper, FLOATS_ALL_NORMAL)


def test_quote_all_special(benchmark, dumper):
    bench_quote(benchmark, dumper, FLOATS_ALL_SPECIAL)
```
</details>

psycopg/psycopg/types/numeric.py
```diff
@@ -68,6 +68,7 @@ def quote(self, obj: Any) -> Buffer:
         if not isinstance(value, bytes):
             value = bytes(value)
         if value in self._special:
+            # Path taken rarely; `try: except KeyError:` likely slower.
```
dvarrazzo: This comment is not needed. The fact that you AI agent proposed something that was not a good idea doesn't mean that someone with a brain would requires guidance to understand that a thing called "special" is more rare than a thing that is not special.

psycopg/psycopg/_conninfo_utils.py
```diff
@@ -70,8 +70,10 @@ def get_param(params: ConnMapping, name: str) -> str | None:
 
     The value may be also specified in a PG* env var.
     """
-    if name in params:
+    try:
```
dvarrazzo: This is not a rare case such as a cache miss. It is common for instance that `port` is not specified in a connection string.

dvarrazzo (changes requested):
Please use a commit message consistent with the convention used in this project (e.g. "perf: ask forgiveness rather than permission yadda yadda")

psycopg/psycopg/types/numeric.py
akx: This wasn't generated by an AI agent, thanks for asking 😄 (None of this PR was.)

I wrote this comment so other people who might be considering doing the same thing know that this _isn't_ a place to apply this.

psycopg/psycopg/_conninfo_utils.py
akx: This actually popped out to me in the diagram when I viztraced the test suite.

`get_param()` seemed to be surprisingly hot there.

psycopg/psycopg/types/numeric.py
dvarrazzo: If you haven't used an AI agent (which I don't consider a bad thing in itself, if the result are reviewed and not blindly accepted) then you have used `sed` and made carpet changes without understanding the code you are touching. Testament is the proposed change to code where performance is totally irrelevant such as the doc customization script, which is not even part of the released package.

I would prefer that, if someone proposes a change to this project, that they understand what they are changing.

In this context it is perfectly clear that this special case: it's written on the tin. I don't know how many other people will come across in the future with the wrong understanding of the Python 3.11 optimization (which optimises the try-branch only, not the catch-branch) and a will to replace that statement in the future, but trying to help them goes outside the scope of this project.

Consider this statement [a brown M&M](https://www.thesmokinggun.com/documents/crime/van-halens-legendary-mms-rider).

psycopg/psycopg/_conninfo_utils.py
dvarrazzo: This might depend on the `--test-dns` you used in your benchmark.

The `except` path here is not a rare path and becoming rarer with time such as in a cache miss. Whether an item is in the dict or not is just a fact depending on the connection string.

This code is also just used on connection only. If someone depends on micro-optimizations in this path they should have long being using a connection pool.

dvarrazzo:
I have picked only the cases where the non-except branch is the most common path, to make the most of the fast exception setup. Thank you very much!

PR #1305, opened 2026-05-13
feat(pool): add dedicated_connection() method

### LINKED ISSUE #1244: Feature request: `pool.dedicated_connection()`
Sometimes you have a pool with a nicely configured conninfo (or conninfo factory), a custom `connection_class` etc., but then need a long-lived connection that you do not want to be managed by the pool.

Currently, you need to duplicate your configuration.

It would be nice if a pool could give you an unmanaged connection, basically just reusing its configuration and serving as a factory.

Prior art: https://docs.rs/bb8/latest/bb8/struct.Pool.html#method.dedicated_connection

SAY-5:
Adds `ConnectionPool.dedicated_connection()` / `AsyncConnectionPool.dedicated_connection()`, which returns a connection configured with the pool's `conninfo`, `kwargs`, `connection_class`, and `configure` callback, but not managed by the pool: it is not counted in the pool size, not subject to `max_lifetime` or `max_idle`, and not returned to the pool on close. The caller is responsible for closing it.

Closes #1244, addressing the maintainer comment that this is a not-bad use case from the initial pool design phase.

### DIFF
--- a/docs/api/pool.rst
+++ b/docs/api/pool.rst
@@ -263,6 +263,9 @@ The `!ConnectionPool` class
 
    .. automethod:: getconn
    .. automethod:: putconn
+   .. automethod:: dedicated_connection
+
+      .. versionadded:: 3.3.2
 
 
 Pool exceptions
--- a/docs/api/pool.rst
+++ b/docs/api/pool.rst
@@ -265,7 +265,7 @@ The `!ConnectionPool` class
    .. automethod:: putconn
    .. automethod:: dedicated_connection
 
-      .. versionadded:: 3.3.2
+      .. versionadded:: 3.4.0
 
 
 Pool exceptions
--- a/docs/news_pool.rst
+++ b/docs/news_pool.rst
@@ -10,6 +10,13 @@
 Current release
 ---------------
 
+psycopg_pool 3.3.2 (unreleased)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Add `~ConnectionPool.dedicated_connection()` method to obtain a connection
+  configured like the pool's but not managed by it (:ticket:`#1244`).
+
+
 psycopg_pool 3.3.1
 ^^^^^^^^^^^^^^^^^^
 
--- a/docs/news_pool.rst
+++ b/docs/news_pool.rst
@@ -7,15 +7,18 @@
 ``psycopg_pool`` release notes
 ==============================
 
-Current release
+Future releases
 ---------------
 
-psycopg_pool 3.3.2 (unreleased)
+psycopg_pool 3.4.0 (unreleased)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Add `~ConnectionPool.dedicated_connection()` method to obtain a connection
   configured like the pool's but not managed by it (:ticket:`#1244`).
 
+
+Current release
+---------------
 
 psycopg_pool 3.3.1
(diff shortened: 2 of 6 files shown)


docs/news_pool.rst
```diff
@@ -10,6 +10,13 @@
 Current release
 ---------------
 
+psycopg_pool 3.3.2 (unreleased)
```
dvarrazzo: This is a new feature, so we cannot release it in 3.3.2, which would be a bugfix release. The feature can only be released in 3.4.0.

docs/api/pool.rst
```diff
@@ -263,6 +263,9 @@ The `!ConnectionPool` class
 
    .. automethod:: getconn
    .. automethod:: putconn
+   .. automethod:: dedicated_connection
+
+      .. versionadded:: 3.3.2
```
dvarrazzo: See other comment

psycopg_pool/psycopg_pool/pool.py
```diff
@@ -331,6 +331,32 @@ def putconn(self, conn: CT) -> None:
...
+        kwargs = self._resolve_kwargs()
+        conn = self.connection_class.connect(conninfo, **kwargs)
+        try:
+            if self._configure:
+                self._configure(conn)
+                if (status := conn.pgconn.transaction_status) != TransactionStatus.IDLE:
+                    sname = TransactionStatus(status).name
+                    raise e.ProgrammingError(
+                        f"connection left in status {sname} by configure function {self._configure}: discarded"
+                    )
```
dvarrazzo: This block seems exactly the same one from `_connect()`. I would put the common code in a `_do_configure()` private method (which would bail out early if `self._configure` is not set, otherwise call it and mange the error).

tests/pool/test_pool.py
```diff
@@ -201,6 +201,58 @@ def configure(conn):
...
+        p.wait(timeout=1.0)
+        stats_before = p.get_stats()
+        conn = p.dedicated_connection()
+        try:
+            assert getattr(conn, "_pool", None) is None
+            res = conn.execute("select 1")
+            assert res.fetchone() == (1,)
+        finally:
+            conn.close()
+        # Pool counters and size are unaffected.
```
dvarrazzo: Maybe the counters *should* be affected? I think that `_CONNECTIONS_NUM` and `_CONNECTIONS_ERRORS` are meaningful here.

tests/pool/test_pool.py
```diff
@@ -201,6 +201,58 @@ def configure(conn):
             assert res.fetchone()[0] == "on"
 
 
+def test_dedicated_connection(dsn):
+    with pool.ConnectionPool(dsn, min_size=1) as p:
+        p.wait(timeout=1.0)
+        stats_before = p.get_stats()
+        conn = p.dedicated_connection()
+        try:
```
dvarrazzo: This would also work in the tests, it's more idiomatic for the way we have written so far:

```suggestion
        with p.dedicated_connection() as conn:
```

which should also verify that a dedicated connection can be used well as context manager (which should be implicit, but you never know...)

dvarrazzo (changes requested):

docs/news_pool.rst
```diff
@@ -7,6 +7,16 @@
 ``psycopg_pool`` release notes
```
SAY-5: Updated to target 3.4.0 in both news_pool.rst and the api docs.

PR #1307, opened 2026-05-21
Discard prepared statement upon ALTER or DISCARD statements

dvarrazzo:
There is still the case that people may issue a DROP or ALTER into a prepared statement or something else obfuscating the status message, but at least now an explicit DISCARD can be used as escape hatch for whatever client/server state inconsistency.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,16 @@
 ``psycopg`` release notes
 =========================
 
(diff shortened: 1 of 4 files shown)


hlinnaka:
It'd be good to handle `DEALLOCATE ALL` as well.

PR #1308, opened 2026-05-22
Make libpq paths configurable through environment variables

m-kuhn:
I am a maintainer of [open-vcpkg/python-registry](https://github.com/open-vcpkg/python-registry). Recently, `vcpkg` moved to the meson build system for `libpq` and since, `pg_config` is no longer built as part of that port.
This pull request proposes to make it possible to build the psycopg c extension without `pg_config` and instead provide include and libdir via environment variables.

New variables:

`PSYCOPG_PG_INCLUDEDIR `: will have precedence for the includedir
`PSYCOPG_PG_LIBDIR `: will have precedence for the libdir
`PSYCOPG_PG_CONFIG`: path to `pg_config`, will have precedence over `pg_config` on `PATH`

Reference: https://github.com/open-vcpkg/python-registry/pull/240

### DIFF
--- a/docs/basic/install.rst
+++ b/docs/basic/install.rst
@@ -176,7 +176,13 @@ In order to perform a local installation you need some prerequisites:
 - a C compiler,
 - Python development headers (e.g. the ``python3-dev`` package).
 - PostgreSQL client development headers (e.g. the ``libpq-dev`` package).
-- The :program:`pg_config` program available in the :envvar:`PATH`.
+
+The build will need access to the include folder and lib folder of libpq.
+On most systems this can be determined through the :program:`pg_config` if it
+is in the :envvar:`PATH`.
+Alternatively the path to :program:`pg_config` can be either configured via the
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Have you seen the [`pg_config_vcpkg_stub`](https://github.com/psycopg/psycopg/blob/master/tools/ci/pg_config_vcpkg_stub/pg_config_vcpkg_stub/__init__.py)?

m-kuhn:
Yes, I've seen it. As far as I understand this needs to be installed first which can be avoided with this approach.

https://github.com/psycopg/psycopg/blob/57db5c86b71741a967001e62b784984115741525/tools/ci/wheel_win32_before_build.bat#L9

Another advantage of this patch is, that it's independent of the system (i.e. also usable for non-vcpkg installations).

If there are good reasons to stick to the vcpkg stub, I'm happy to close this.

dvarrazzo:
The stub works adequately as part of our build system and makes windows build similar to other sane systems where pg_config is installed.

I don't think this changeset is necessary.

Issue #71, opened 2021-08-29
Replication support


dvarrazzo:
Add support for streaming replication as in psycopg2

sassemlal:
Hi, 

If this still an issue, I could take a look and use your guidance to implement this.

dufferzafar:
@dvarrazzo Could you please also explain exactly what would need to be added to support replication directly in psycopg3? 

I'm taking a look at the code right now & having some pointers on exactly what needs to be done would be great!

dvarrazzo:
Replication support is not implemented in psycopg 3. It needs complete redesign and implementation. If you want to contribute its implementation you are welcome, but it is a major work.

dufferzafar:
Yeah, I realised that after looking at the code of pg2 & pg3. I guess it'll require a lot of interfacing with pq as well?

My current plan is to use something like [`wal2json`](https://github.com/eulerto/wal2json/) extension on the server, then create a slot and get_changes in the JSON format directly from the server. 

Or perhaps I won't do this in async mode and just use pg2 directly. I found pypgoutput which does client side decoding of pgoutput format to JSON: https://github.com/dgea005/pypgoutput

sergiors:
👀

RPG-fan:
## Working Implementation Available

I needed logical replication for a project and built a working solution. Since others seem to be stuck on this, I'm sharing it.

**Gist:** [https://gist.github.com/RPG-fan/b6d578e45712ae9467b05d6ac4e8dbc6]

### How it works

The implementation uses a "bridge" pattern:
1. Uses psycopg3 to establish connection and send START_REPLICATION
2. After entering COPY_BOTH mode, hands off to raw socket I/O
3. Reads/writes PostgreSQL wire protocol messages directly
4. Includes complete pgoutput binary protocol parser

### Why this approach

psycopg3's state machine gets confused trying to handle COPY protocol for replication. By taking direct control of the socket after COPY_BOTH is established, we bypass the state checking while still using psycopg3's solid connection foundation.

### Status

Running in production with 9 concurrent replication consumers. Handles BEGIN, COMMIT, RELATION, INSERT, UPDATE, DELETE, and TRUNCATE messages. Properly acknowledges LSN positions and handles keepalives.

### Usage

```
stream = LogicalReplicationStream(
    conn=conn,  # AsyncConnection with replication=database
    slot_name="my_slot",
    publication_name="my_pub"
)

async for raw_msg, decoded_msg in stream.start_replication():
[...]

PR #1311, opened 2026-05-30
POC: Implement Replication Connection Support

DylanYoung:
Here's a shot at implementing replication connection support (#71).  It has support for the full replication protocol and customizable decoders for the logical decoding payload in the XLogData message which is dependent on the output_plugin, including two builtin decoders: one for pgoutput (which allows custom logical row factories analogous to row factories for regular queries and uses the cursor's transformer to adapt to python types) and one for test_decoding (which should also work for any output plugin that delivers text in the server encoding).

I'm not entirely happy with the module layout right now.  It's a bit too nested for my taste and some of the deeper bits should probably be exposed at a higher level (i.e. in `psycopg.replication`). 

There's a temporary `replication/NEXT_STEPS.md` document that shows a few things that are still missing, primarily tests.  I'm sure most of the decoding pipeline could be implemented in C as well for a speedup (I've implemented a few small parts to start), but I wanted to discuss the direction before going ahead with that. The `consume*` functions are still untested, just to be aware if you're trying things out. 

Docs are a little rough, but should cover most of the functionality.   

I fixed a few small bugs I found along the way; those fixes are at the top of the branch (I wrote a regression test for one of them, let me know if you think the others warrant one).  There are also a couple of commits there marked "DISCUSS" that I needed to get tests and pre-commit hooks (mypy) working properly locally (perhaps you could advise on how you do this). 

Some of the more debatable parts of the implementation are still in separate commits for easier review.  There's also a commit + revert that I used to generate test data (I'd like to make this a tool if it makes sense to you to have unit tests for the decoding instead of just integration tests).

There are probably a few DISCUSS/FIXME/TODO comments scattered in the code or commit messages about things that either need polishing/finishing or that I wanted to discuss before implementing. 

Happy to finish this off if it looks like a reasonable direction to you, or revise as needed!

I'd love some feedback in particular on the various DISCUSS points and on the next steps of the C implementation.  Right now I'm thinking it might be best to just reimplement all of the message classes in Cython.


DylanYoung:
Looks like I've got a few compat issues with older python versions.  My apologies.  I forgot to test on older pythons, d'oh. I'll take a look at the CI when it's done and get those fixed up. 

EDIT: I don't have a local Python 3.10, but 3.11 to 3.14 are now working locally for me.  I also reverted the changes I made to get tests and pre-commit running locally as I thought they might be causing some of the failures.

tests/test_replication/test_base_replication_cursor_async.py
```diff
@@ -0,0 +1,814 @@
...
+    assert arepl_started_cur._last_received_lsn == msg_flush.data_start
+
+
+@parametrize_all_replication_cursors
+@parametrize_replication_connections
+async def test_read_message_no_start(arepl_cur, pytestconfig):
+    # TODO: pre-existing bug triggered by `--cov=psycopg.XXX` causes the raised
+    # OperationalError to be different from psycopg.errors.OperationalError
+    # when using the c module (and probably also the binary module).
+    coverage_args = pytestconfig.getoption("--cov")
```
DylanYoung, line 402: This was an interesting bug, not 100% on who owns it, but I suspect it's either coverage or pytest-cov.  Should probably be reported, though I'd like to come up with a minimal reproducer first.

tests/test_replication/fix_db_async.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@pytest.fixture
+def application_name():
+    return "psycopg_testing"
+
+
+@pytest.fixture
+async def aconn(aconn):
+    await aconn.set_autocommit(True)
```
DylanYoung, line 30: Hmmm... It looks like the test failures now are caused by this.  I thought this fixture was test-scoped, but maybe I misunderstood something.

DylanYoung: I'm investigating how to fix this in a simple manner.  Any advice welcome. Could be this isn't the issue, but that's my best hypothesis off-the-top after looking at the CI logs.  

I do also have local test failures on master. Is that expected?

dvarrazzo:
Hello,

thank you but I have already in progress some work to add replication support to Psycopg 3.

It is a bit too early to discuss about it, it's something that will be disclosed in the upcoming weeks. The interface is different from the one proposed here (which is inspired to psycopg2 replication support) and will be more advanced and much simpler to use.

So I don't think that it is useful to put more work in this MR.

DylanYoung:
@dvarrazzo  There are a bunch of small bug fixes at the top of the branch.  Would you like me to put up a separate PR for those?

chickeaterbanana:
> Hello,
> 
> thank you but I have already in progress some work to add replication support to Psycopg 3.
> 
> It is a bit too early to discuss about it, it's something that will be disclosed in the upcoming weeks. The interface is different from the one proposed here (which is inspired to psycopg2 replication support) and will be more advanced and much simpler to use.
> 
> So I don't think that it is useful to put more work in this MR.

Hey @dvarrazzo, any chance that now could be "upcoming weeks"? Maybe I've missed the announcement or hints and you can point me to the right place.

PR #1312, opened 2026-05-31
Various Bug fixes and minor improvements from Replication branch

DylanYoung:
As per title.  Let me know what you want or don't and I'm happy to make changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,6 +99,10 @@ will save you time and frustrations.
 
 .. __: https://pre-commit.com/
 
+Please follow the `conventional commits`__ specification for your commit messages.
+
+.. __: https://www.conventionalcommits.org/en/v1.0.0/
+
 
 Cross-compiling
 ---------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,6 +54,8 @@ templates_path = ["_templates"]
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv"]
 
+maximum_signature_line_length = 100
+
 # -- Options for HTML output -------------------------------------------------
 
 # The announcement may be in the website but not shipped with the docs
--- a/psycopg_c/psycopg_c/_psycopg.pyx
+++ b/psycopg_c/psycopg_c/_psycopg.pyx
@@ -14,6 +14,7 @@ from psycopg_c._psycopg cimport oids
 import logging
 
 from psycopg.pq import Format as _pq_Format
+from psycopg.pq import abc
 from psycopg._enums import PyFormat as _py_Format
 
 logger = logging.getLogger("psycopg")
--- a/tests/acompat.py
+++ b/tests/acompat.py
@@ -59,14 +59,13 @@ def spawn(f, args=None):
         return t
 
 
-def gather(*ts, return_exceptions=False, timeout=None):
+def gather(*ts, return_exceptions=False, timeout=10):
     """
     Equivalent to asyncio.gather or Thread.join()
     """
     if ts and inspect.isawaitable(ts[0]):
         rv: Any = asyncio.gather(*ts, return_exceptions=return_exceptions)
-        if timeout is None:
-            rv = asyncio.wait_for(rv, timeout)
+        rv = asyncio.wait_for(rv, timeout)
         return rv
     else:
         for t in ts:
--- a/tests/test_tstring.py
+++ b/tests/test_tstring.py
@@ -1,4 +1,5 @@
 from random import random
+from typing import Any
 
 import pytest
 
@@ -6,7 +7,7 @@ import psycopg
 from psycopg import sql
 from psycopg.pq import Format
 
-from .acompat import alist
+from .acompat import alist, skip_async
 from .test_sql import no_e
 from .test_adapt import make_dumper
 
@@ -199,6 +200,15 @@ async def test_sql_join(aconn):
     assert await cur.fetchone() == (0, 1, 2)
     assert cur.description[0].name == "foo"
     assert cur.description[2].name == "baz"
+
+
+@skip_async
+def test_sql_join_mixed_types():
+    ts: list[Any] = [t"{i} as {name:i}" for i, name in enumerate(("foo", "bar"))]
+    ts.append(sql.Literal(2))
+    for it in (ts, reversed(ts)):
+        with pytest.raises(TypeError, match="Template and Literal"):
+            fields = sql.SQL(",").join(it)
 
 
 async def test_copy(aconn):
--- a/psycopg/psycopg/_adapters_map.py
+++ b/psycopg/psycopg/_adapters_map.py
@@ -48,7 +48,7 @@ class AdaptersMap:
     already created are not affected if a wider scope (e.g. the global one) is
     changed.
 
-    The connections adapters are initialised using a global `!AdptersMap`
+    The connections adapters are initialised using a global `!AdaptersMap`
     template, exposed as `psycopg.adapters`: changing such mapping allows to
     customise the type mapping for every connections created afterwards.
 
--- a/psycopg/psycopg/pq/_debug.py
+++ b/psycopg/psycopg/pq/_debug.py
@@ -76,7 +76,7 @@ class PGconnDebug:
         return cls(debugging(PGconn.connect_start)(conninfo))
 
     @classmethod
-    def ping(self, conninfo: bytes) -> int:
+    def ping(cls, conninfo: bytes) -> int:
         return debugging(PGconn.ping)(conninfo)
 
 
--- a/psycopg/psycopg/pq/misc.py
+++ b/psycopg/psycopg/pq/misc.py
@@ -167,7 +167,7 @@ def connection_summary(pgconn: abc.PGconn) -> str:
         except ValueError:
             # It might happen if a new status on connection appears
             # before upgrading the ConnStatus enum.
-            status = f"status={pgconn.status} (unkndown)"
+            status = f"status={pgconn.status} (unknown)"
 
     if sparts := " ".join("%s=%s" % part for part in parts):
         sparts = f" ({sparts})"
--- a/psycopg/psycopg/sql.py
+++ b/psycopg/psycopg/sql.py
@@ -337,7 +337,7 @@ class SQL(Composable):
         cs = [first]
         for i in it:
             if isinstance(i, Template):
-                raise TypeError(f"can't mix Template and {type(i).__name__}")
+                raise TypeError(f"can't mix Template and {type(first).__name__}")
             cs.append(self)
             cs.append(i)
 
--- a/psycopg_c/psycopg_c/_psycopg/adapt.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/adapt.pyx
@@ -54,14 +54,14 @@ cdef class CDumper:
         """
         raise NotImplementedError()
 
-    def dump(self, obj) -> Buffer | None:
+    def dump(self, obj) -> abc.Buffer | None:
         """Return the Postgres representation of *obj* as Python array of bytes"""
         cdef rv = PyByteArray_FromStringAndSize("", 0)
         cdef Py_ssize_t length = self.cdump(obj, rv, 0)
         PyByteArray_Resize(rv, length)
         return rv
 
-    def quote(self, obj) -> Buffer:
+    def quote(self, obj) -> abc.Buffer:
         cdef char *ptr
         cdef char *ptr_out
         cdef Py_ssize_t length
--- a/psycopg_c/psycopg_c/_psycopg/generators.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/generators.pyx
@@ -10,7 +10,6 @@ from time import monotonic
 from collections import deque
 
 from psycopg import errors as e
-from psycopg.pq import abc
 from psycopg.abc import PipelineCommand, PQGen
 from psycopg._enums import Ready, Wait
 from psycopg._encodings import conninfo_encoding
--- a/psycopg_c/psycopg_c/types/bool.pyx
+++ b/psycopg_c/psycopg_c/types/bool.pyx
@@ -28,7 +28,7 @@ cdef class BoolDumper(CDumper):
 
         return 1
 
-    def quote(self, obj: bool) -> Buffer | None:
+    def quote(self, obj: bool) -> abc.Buffer | None:
         if obj is True:
             return b"true"
         elif obj is False:
--- a/psycopg_c/psycopg_c/types/numeric.pyx
+++ b/psycopg_c/psycopg_c/types/numeric.pyx
@@ -61,7 +61,7 @@ cdef class _IntDumper(CDumper):
     cdef Py_ssize_t cdump(self, obj, bytearray rv, Py_ssize_t offset) except -1:
         return dump_int_to_text(obj, rv, offset)
 
-    def quote(self, obj) -> Buffer | None:
+    def quote(self, obj) -> abc.Buffer | None:
         cdef Py_ssize_t length
 
         rv = PyByteArray_FromStringAndSize("", 0)
@@ -326,7 +326,7 @@ cdef class _FloatDumper(CDumper):
         PyMem_Free(out)
         return length
 
-    def quote(self, obj) -> Buffer | None:
+    def quote(self, obj) -> abc.Buffer | None:
         value = bytes(self.dump(obj))
         cdef PyObject *ptr = PyDict_GetItem(_special_float, value)
         if ptr != NULL:
@@ -436,7 +436,7 @@ cdef class DecimalDumper(CDumper):
     cdef Py_ssize_t cdump(self, obj, bytearray rv, Py_ssize_t offset) except -1:
         return dump_decimal_to_text(obj, rv, offset)
 
-    def quote(self, obj) -> Buffer | None:
+    def quote(self, obj) -> abc.Buffer | None:
         value = bytes(self.dump(obj))
         cdef PyObject *ptr = PyDict_GetItem(_special_decimal, value)
         if ptr != NULL:
--- a/psycopg_c/psycopg_c/types/string.pyx
+++ b/psycopg_c/psycopg_c/types/string.pyx
@@ -214,7 +214,7 @@ cdef class BytesDumper(CDumper):
         libpq.PQfreemem(out)
         return len_out
 
-    def quote(self, obj) -> Buffer:
+    def quote(self, obj) -> abc.Buffer:
         cdef char *ptr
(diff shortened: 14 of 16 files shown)


dvarrazzo:
Hello! Thank you. Maybe you can push them to your https://github.com/DylanYoung/psycopg/tree/replication branch so it will stay under MR #1311 and should we decide to instead act on it then everything is in the same place :)

DylanYoung:
@dvarrazzo  These fixes and minor enhancements are all present on the replication branch.  I created this one because I figure you'll still want at least the bug fixes even if you don't want the replication work.   I just found another small one, which I'll add to both branches as well.

tests/test_tstring.py
```diff
@@ -201,6 +202,14 @@ async def test_sql_join(aconn):
     assert cur.description[2].name == "baz"
 
 
+async def test_sql_join_mixed_types(aconn):
```
DylanYoung: TASK: Doesn't need to take aconn or be an async def.

tests/utils.py
```diff
@@ -138,9 +138,14 @@ def _match_version(self, got_tuple: tuple[int, ...]) -> bool:
             assert len(version_tuple) <= 2
             version_tuple = version_tuple[:1] + (0,) + version_tuple[1:]
 
-        op: Callable[[tuple[int, ...], tuple[int, ...]], bool]
-        op = getattr(operator, self._OP_NAMES[self.op])
-        return op(got_tuple, version_tuple)
+        if self.op == "==":
```
DylanYoung: In case this isn't clear, what this does is allow you to specify "==10" and have it apply to all "10.X" versions (or similarly "10.1" would apply to all "10.1.Y").

dvarrazzo: But that's not the right semantic. What you want to express is `>= 10, < 11`. Typically this is expressed with something like `~10`. In our tests we can just use the decorator twice with the two conditions, which will be the equivalent of ANDing them.

I wouldn't change the meaning of the `==` operator.

dvarrazzo:
> @dvarrazzo These fixes and minor enhancements are all present on the replication branch. I created this one because I figure you'll still want at least the bug fixes even if you don't want the replication work. I just found another small one, which I'll add to both branches as well.

Ah sorry I didn't understand. I appreciate, thank you. Reviewing...

psycopg_c/psycopg_c/_psycopg.pyx
```diff
@@ -14,6 +14,7 @@ from psycopg_c._psycopg cimport oids
 import logging
 
 from psycopg.pq import Format as _pq_Format
+from psycopg.abc import Buffer
```
dvarrazzo: As far as I have understood so far, no cython linter works well with the lexical includes, so I've been unable so far to have clean cython linting (my editor, at least, complains a lot).

I have seen that elsewhere we have used `abc.Buffer` and that `abc` is imported. So maybe we can fix the occurrences of the naked `Buffer` to use `abc.Buffer` instead:

```
$ git grep ' Buffer\b' ./psycopg_c/psycopg_c/
psycopg_c/psycopg_c/_psycopg/adapt.pyx:    def dump(self, obj) -> Buffer | None:
psycopg_c/psycopg_c/_psycopg/adapt.pyx:    def quote(self, obj) -> Buffer:
psycopg_c/psycopg_c/types/bool.pyx:    def quote(self, obj: bool) -> Buffer | None:
psycopg_c/psycopg_c/types/numeric.pyx:    def quote(self, obj) -> Buffer | None:
psycopg_c/psycopg_c/types/numeric.pyx:    def quote(self, obj) -> Buffer | None:
psycopg_c/psycopg_c/types/numeric.pyx:    def quote(self, obj) -> Buffer | None:
psycopg_c/psycopg_c/types/string.pyx:    def quote(self, obj) -> Buffer:
```

assuming that these are the warnings that you have been reported?

tests/acompat.py
```diff
@@ -65,7 +65,7 @@ def gather(*ts, return_exceptions=False, timeout=None):
     """
     if ts and inspect.isawaitable(ts[0]):
         rv: Any = asyncio.gather(*ts, return_exceptions=return_exceptions)
-        if timeout is None:
+        if timeout is not None:
```
dvarrazzo: Uhm... This was clearly an error. However, I see that both `asyncio.wait_for()` and `Thread.join()` take `timeout=None` as a mean to say "wait forever until completion". So maybe we can just drop the `if` and have a `return asyncio.wait_for(rv, timeout)`?

I also question the wisdom of using `timeout=None` as default (and a quick grep show that we mostly/always use the default). I don't think it's currently causes problems but this means that an error will result in the test pipelines blocking until the entire job times out (1-2 hours). Maybe a default of 10s would be a better choice, still leaving the option of passing None if some test require it. What do you think?

tests/test_tstring.py
dvarrazzo: I don't understand this comment. However this test doesn't use `aconn` at all, so it can have no argument, and doesn't even require to be async.

Ah, I think I see: all these tests (apart a couple of outliers, likely errors) are defined `async` because I wanted to use the async-to-sync script to generate the sync version. However at the time of writing this test module (which happened before Python 3.14 was released), some part of the infrastructure wasn't compatible with the template string syntax yet, so the idea was to start translating this module to sync once we would have moved the check pipeline to Python 3.14 (which happened only recently, in eeb99b21099185a84e73ec63a8d99ced53e8cc4e).

~So my proposal is that you introduce this test without the `aconn` parameter, which is not needed, and we can create the async/sync pair version of this file in a separate task (I can do it myself, no problem).~

---

Edit: Working on #1313 (which will wait for this MR to be merged to save you a complex rebasing) I noticed that if `aconn` is not specified then an error like "async def functions are not natively supported" is raised. I can't remember what is the magic that makes the `aconn` fixture magically marke the tests using it as `asyncio`. However after the async -> sync conversion this would mean that the test would run twice and it doesn't have any async/sync difference deserving so. Therefore we can mark it sync only:

```suggestion
@sync_only
def test_sql_join_mixed_types():
```

(requires `from .acompat import alist, skip_async` on top of the file)

docs/README.rst
```diff
@@ -10,3 +10,6 @@ Quick start::
     make serve
 
 and connect to <http://localhost:8000>.
+
+.. note::
+    Requires Python ≤ 3.12
```
dvarrazzo: Does it? I would have said >= 3.12 after the work related to #1262. Matter of fact `lint.yml` build the docs with Python 3.14.

dvarrazzo (changes requested):
Thank you for this cleanup.

You have used a commit convention in this branch that this project doesn't use: specifying the name of the file changed in the commit first line is not needed like since CVS times. Please use [the same style used by the project](https://www.conventionalcommits.org/en/v1.0.0/)

If it is too much of a chore, feel free to flatten all your changes to a single commit.

docs/README.rst
DylanYoung: This was before #1262 I think.  I'll take a look and either remove it entirely or update it.

psycopg_c/psycopg_c/_psycopg.pyx
DylanYoung: Yeah, I also haven't found a linter that works well with cython.  This was just for warnings generated by cython itself during compilation/installation.  I can update them to use `abc.Buffer` instead (That's what I had done initially, but when I noticed there were multiple locations to change, I thought it might be better to just import Buffer).  For some reason, the `Buffer | None` lines don't generate the warnings, but I'll update them too, to be consistent.

tests/acompat.py
DylanYoung: I think a default timeout would probably be wise.  I'll update it to 10s and see if it causes any problems.  I'll remove the `if` as well.

tests/test_tstring.py
DylanYoung: Sounds good.  I will add this change.  My intial comment was just a note to myself to remove the `aconn` fixture (sorry for the confusion).  I appreciate the suggestion to use `sync_only`, as I didn't realize this file would be run through async_to_sync.

tests/utils.py
DylanYoung: I'll happily remove this, but it might be worth noting that I don't think it actually will currently work with something like `==10` since the tuples to compare won't even be the same length;  I suspect it would simply fail to match any real version.   I (stupidly) didn't realize I could just use the decorator twice when I first wrote this.  I'll admit,  I think I would be surprised if semantically `==10` implied `==10.0`, but that might just be a difference of opinion.  It seems like a rare use-case to actually want to match an exact minor version, but if I did, I'd probably want to be explicit about it by writing `==10.0`.    I'll leave it for now and you can think on it and let me know if you want it removed.

DylanYoung:
> Thank you for this cleanup.
> 
> You have used a commit convention in this branch that this project doesn't use: specifying the name of the file changed in the commit first line is not needed like since CVS times. Please use [the same style used by the project](https://www.conventionalcommits.org/en/v1.0.0/)
> 
> If it is too much of a chore, feel free to flatten all your changes to a single commit.

I'll update the commits to use the proper convention, sorry I forgot to check the conventions used when I first started hacking.  Thanks for the link to the convention you're using.  Is that documented anywhere?  If not, I can add it to the HACKING section of the main README if you like. 

(I realize the filename isn't often used anymore, but I found it helpful to remember what was what when working on a large branch without having to pull up the full diff).

tests/acompat.py
DylanYoung: I've done this.  It looks fine on my end, but as I said in the replication branch, I have some pre-existing test failures, so we'll see how the CI does.

docs/README.rst
DylanYoung: I've removed this.  I've also added a `maximum_signature_line_length` in commit 3a7f58f133d0e1152ef12a9a9d039a1dd8c1338c now that you're on a version of auto-doc that supports it.  Happy to remove it if you don't like it, but I find it makes the API docs much easier to read.

psycopg_c/psycopg_c/_psycopg.pyx
DylanYoung: This is done.  I had to move the abc import from generators.pyx to _psycopg.pyx.  Otherwise I still got the warning for adapt.pyx:

`warning: psycopg_c/_psycopg/adapt.pyx:64:31: Unknown type declaration 'abc.Buffer' in annotation, ignoring`

tests/utils.py
DylanYoung: I've left this for now for you to decide whether you want something like `==14` to just fail due to a mismatched tuple-size, but as I said, I'm happy to remove it if you wish.   I don't think my change here actually changes the meaning of `==` for any cases that work currently.

tests/test_tstring.py
DylanYoung: This is done.

docs/conf.py
```diff
@@ -54,6 +54,8 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv"]
 
+maximum_signature_line_length = 100
```
DylanYoung, line 57: @dvarrazzo This is a new change enabled by your recent Sphinx upgrade.  I find it makes the API docs much more readable, but happy to remove it if you disagree (also happy to change the value).

DylanYoung:
RE 63be432f52d0618a2e338c71ae56bdee626b5b5a

Worth noting that there are also some stringified references to `Buffer` in the psycopg_c pq module.  I left them as they weren't generating warnings and I wasn't sure if I should change them.  You can find them by searching `psycopg_c` for `"Buffer"`

DylanYoung:
@dvarrazzo  I think that covers everything you mentioned.  Ready for another review.  Thanks!

docs/conf.py
dvarrazzo: Do you have a chance to show me a before/after difference?

tests/utils.py
dvarrazzo: As far as I am concerned, `==`, either implicit on explicit, has little to no use. Semi-open intervals (test this behaviour before this version, test this behaviour from this version on) I think cover 100% of the cases we make use of. Yes, nobody wants to test a behaviour that appeared in version 14 and disappeared in version 14.1, but also a "14 <= behaviour < 15" behaviour is rare (test a feature to be present in *precisely* one major version).

For me we can drop support for a naked number and always require an operator. I think that I picked the meaning of "no operator specified" to be "the user meant equals" because that's what we mean in intuitive parlance, but this is practically useless for practical versions testing. For me it seems that you are trying to save supporting a naked number without operator by giving it a somewhat more useful meaning. You really want to change the meaning of "no operator", not the meaning of ==.

I am against == starting to mean ~, but I am not opposed to drop the no operator specified. So, how about we drop the `?` from the op regexp in `parse()` and the support for op = None (as well as the `==` default) in the c'tor? We can do that and see what it breaks, but I doubt a single test uses that feature (that they use either `==` or no op implying `==`).

Keeping all the 6 operators in `_OP_MAP` is cheap, but I think that actually `==` and `!=` are never used, `<=` and `>` are possibly never used, `>=` and `<` are actually useful.

This test confirms (and a similar test on the `pg()` marker is no different):

```
ag --nonumber --nofilename '^@pytest\.mark\.libpq\(' ./tests | sort | uniq

@pytest.mark.libpq("< 10")
@pytest.mark.libpq(">= 10")
@pytest.mark.libpq("< 12")
@pytest.mark.libpq(">= 12")
@pytest.mark.libpq("< 14")
@pytest.mark.libpq(">= 14")
@pytest.mark.libpq("< 16")
@pytest.mark.libpq(">= 16")
@pytest.mark.libpq("< 17")
@pytest.mark.libpq(">= 17")
@pytest.mark.libpq("< 18")
@pytest.mark.libpq(">= 18")
```

So, I'm happy to do one of the following scrub levels:

1) make the operator mandatory
2) also drop support for `==` and `!=`
3) also drop support for `<=` and `>`

I am happy that you do 1. If you think there is value we could even do 2 or 3 but I think it doesn't hurt to leave them there, even just in order to copypaste or refactor the `VersionCheck` object in order to create something in a context where those other operators are actually useful.

Now that I typed the latter sentence I'm convinced that 2 and 3 are not a good idea. Ok for 1 then? (dropping the implicit `==`)?

dvarrazzo (changes requested):

dvarrazzo:
> Thanks for the link to the convention you're using. Is that documented anywhere? If not, I can add it to the HACKING section of the main README if you like.

It's a pretty common convention, but I think no, it's not explicit documented. Yes please, why don't you add a sentence with a link in the readme?

> (I realize the filename isn't often used anymore, but I found it helpful to remember what was what when working on a large branch without having to pull up the full diff).

I don't know if it might be of use but I use `git log --name-only` (which I have aliased as `git lf`) to see only the names of the files touched by a commit. You can add `--oneline` to see the first line of the message only to have an ugly but very compact representation.

docs/conf.py
DylanYoung: After:
Image

Before:
Image

tests/utils.py
DylanYoung: I agree that it's not particularly useful and the double decorator is fine if it is needed.  I'll change it to just do (1) for you.

docs/conf.py
dvarrazzo: That looks a nice improvements, thank you :+1:

DylanYoung:
> > Thanks for the link to the convention you're using. Is that documented anywhere? If not, I can add it to the HACKING section of the main README if you like.
> 
> It's a pretty common convention, but I think no, it's not explicit documented. Yes please, why don't you add a sentence with a link in the readme?

I've added a note about conventional commits to the project README. 



> > (I realize the filename isn't often used anymore, but I found it helpful to remember what was what when working on a large branch without having to pull up the full diff).
> 
> I don't know if it might be of use but I use `git log --name-only` (which I have aliased as `git lf`) to see only the names of the files touched by a commit. You can add `--oneline` to see the first line of the message only to have an ugly but very compact representation.

Thanks.  I use `--name-status` often to look at a specific commit, but don't often think to use it for my logs.  It's a helpful tip.  I use a custom oneline format for most of my logs, which is a bit prettier than the default: `"%C(Yellow)%h  %C(reset)%ad (%C(Green)%cr)%x09%C(Cyan)%aN:%x09%C(reset)%s%C(Magenta)%d"`.  It plays nice with `--name-only` option, which is nice.  Thanks!

tests/utils.py
DylanYoung: @dvarrazzo  I couldn't actually do (1) easily.  The crdb_skip decorator passes "skip" to the `VersionCheck.parse`.   I'm just going to remove any changes here.

DylanYoung:
Should be good for another review.

tests/utils.py
dvarrazzo: I so don't understand GitHub anymore. I don't see this comment of yours if I read the MR conversation but only if I follow a link from the mail. Which means it will be lost forever.

Meh. `</rant>`

I didn't get what the problem is, but this is not a problem on fire, just the removal of an unused feature, so we can leave it for another time, out of this MR.

README.rst
```diff
@@ -99,6 +99,10 @@ will save you time and frustrations.
 
 .. __: https://pre-commit.com/
 
+Commit messages follow the `conventional commits`__ specification.
```
dvarrazzo: In this readme we have used more a discursive style ("you should the this, you should do that") so let's keep that style instead of just a declaration ("things are this way"):

```suggestion
Please follow the `conventional commits`__ specification for your commit messages.
```

Thank you!

dvarrazzo (changes requested):
Asking for a small adjustment in the readme change, otherwise this is good, thank you!

tests/utils.py
DylanYoung: I'm not a big fan of the Github MR interface.  It seems very easy to lose things.   

This is the problem in tests/fix_crdb.py:

    elif mark.name == "crdb_skip":
        ...
        spec = _crdb_reason_version.get(reason, "skip")
    else:
        ...

    pred = VersionCheck.parse(spec)

So the string "skip" (which doesn't have an operator) gets passed as the `spec` to `VersionCheck.parse`.  It could be fixed either by making the regex slightly more complicated or maybe by changing the line above to:

`spec = _crdb_reason_version.get(reason, "skip ==")` 

But it requires a bit of testing and I think this MR is ready otherwise.

README.rst
DylanYoung: Done.

tests/utils.py
DylanYoung: [DELETED] (I misunderstood the code)

TLDR; the `spec` "skip" is passed to `VersionCheck.parse` to unconditionally skip a test.

DylanYoung:
I think this is probably ready for merge now.  Thanks!

dvarrazzo:
Merged, thank you very much!

PR #1316, opened 2026-06-06
Catalog types: xid, xid8, tid, cid, pg_lsn, int2vector, and oidvector

### LINKED ISSUE #320: Provide data type mapping for catalog types
Several types found in pg_catalog are not mapped to Python types.

We should introduce a `psycopg.types.catalog` module implementing mappings:

- `oid`: `uint32` (already implemented in `psycopg.types.numeric`).
- `cid`: `uint32`
- `int2vector`: similar to `smallint[]`, with space-separated text representation, 0-based array binary representation.
- `oidvector`: similar to `int[]`, see int2vector
- `tid`: pair `(block, offset)` (`uint32`, `uint16`)
- `xid`: `uint32`
- `xid8`: `uint64`
- `pg_lsn`: `uint64`

This query gives an overview of where the types are used:

select a.atttypid::regtype, relname, attname from pg_attribute a join pg_class c on a.attrelid = c.oid join pg_namespace s on s.oid = c.relnamespace where nspname = 'pg_catalog' order by a.atttypid::regtype::text, relname, attname;

DylanYoung:
Replaces #1285 with @dvarrazzo's feedback implemented.  
Closes #320 .  All types mentioned in the issue are loaded using wrappers that subclass `str`, providing backwards compatibility and eliminating the need for dumpers. 

I haven't done the docs yet, but am happy to do that once an OK on the design is given.


psycopg/psycopg/_adapters_map.py
```diff
@@ -189,6 +189,22 @@ def register_loader(self, oid: int | str, loader: type[Loader]) -> None:
         if self._register_loader_callback:
             self._register_loader_callback(oid, loader)
 
+    def unregister_loader(self, oid: int | str, format: pq.Format) -> type[Loader]:
```
DylanYoung: @dvarrazzo  This part is the only really contentious bit.  I added it to double-check backwards compatibility in the tests (commit 5f66d6efa19729ef7d05b8d7338d63e68d0f032f).  I'm happy to move it to the tests, but I thought it might be more generally useful.  If you think it is, I'm happy to document it and add a corresponding version for dumpers.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,244 @@
...
+
+    result = cur.execute(f"select %{fmt_in.value}::{typname}[]", ([val],)).fetchone()[0]
+    assert cur.pgresult.fformat(0) == fmt_out
+    assert cur.pgresult.ftype(0) == builtins[typname].array_oid
+    assert isinstance(result[0], str)
+    assert isinstance(result[0], cls) is with_loaders
+    assert result[0] == val
+
+
+@pytest.mark.parametrize("cls", [CID, XID, XID8])
```
DylanYoung: Since the dumpers haven't changed, these quote tests can probably be removed.

DylanYoung:
The pool-integration test failures seems to be a wider issue.  Since the pool-integration tests are testing against older versions of psycopg, some of the modules are not available and test collection fails (even thought the tests are not relevant to the `-m pool`).  I'd suggest maybe passing `--continue-on-collection-errors` to pytest for those tests (or maybe there's a less global option to bypass specific test files during collection?).  

The other failing test seems to be an ordering fluke... maybe it should be modified to use set comparisons?  Or maybe it indicates something wrong with the `_dns` module.  I'm not 100%, but it's unrelated to this work either way.

psycopg/psycopg/types/_catalog.py
```diff
@@ -0,0 +1,118 @@
+"""
```
dvarrazzo: I would rather have a single `psycopg.types.catalog` public module rather than all these small modules, one per type.

Or even, if a single module became too large to handle, to make `catalog.py` the public one we consider public interface, and single types ones or smaller groups private implementation ones (e.g. `_lsn.py`).

psycopg/psycopg/types/_catalog.py
```diff
@@ -0,0 +1,118 @@
...
+from ..abc import Buffer
+from ..adapt import Loader
+from .._compat import Self, TypeVar
+from .._struct import unpack_len, unpack_uint4, unpack_uint8
+
+
+class _StrSubclass(str):
+    @classmethod
+    @abstractmethod
+    def from_buffer(cls, val: Buffer) -> Self: ...
```
dvarrazzo: I would prefer for these objects to not knowing anything about the postgres format. The code going here (in the concrete sublcasses) should rather be in the loaders.

psycopg/psycopg/types/_catalog.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}('{self}')"
```
dvarrazzo: This doesn't roundrip correctly. The following should:

```suggestion
        return f"{type(self).__name__}({super().__repr__()})"
```

psycopg/psycopg/types/lsn.py
```diff
@@ -0,0 +1,48 @@
+from ..abc import AdaptContext, Buffer
+from .._compat import Self
+from ._catalog import _Int8IntStrBinaryLoader, _IntStr, _StrSubclassLoader
+
+
+class LSN(_IntStr):
```
dvarrazzo: For the work I'm doing on the replication, I have created an LSN class. It has properties different than this one, for example it compares numerically with others and supports a few operators. It might be a better implementation, or anyways it is worth a comparison.

psycopg/psycopg/types/_catalog.py
```diff
@@ -0,0 +1,118 @@
...
+        obj.value = value
+
+        return obj
+
+    @classmethod
+    def from_int(cls, val: int) -> Self:
+        obj = cls(val)
+        obj.value = val
+
+        return obj
```
dvarrazzo: I think that this implementation will result in broken objects if you call `CID(10)` or `CID("10")`: they will not have the `value` attribute.

The customary way to subclass immutable type is to override `__new__()`. It would also be useful to have the numeric value in a read-only property and to implement `__slots__`.

However I am not sure how much these types deserve an integer syntax. What type of arithmetic should they support?

psycopg/psycopg/_struct.py
```diff
@@ -36,8 +37,10 @@ def __call__(self, data: Buffer, start: int | None) -> tuple[int]: ...
 unpack_int4 = cast(UnpackInt, struct.Struct("!i").unpack)
 unpack_uint4 = cast(UnpackInt, struct.Struct("!I").unpack)
 unpack_int8 = cast(UnpackInt, struct.Struct("!q").unpack)
+unpack_uint8 = cast(UnpackInt, struct.Struct("!Q").unpack)
 unpack_float4 = cast(UnpackFloat, struct.Struct("!f").unpack)
 unpack_float8 = cast(UnpackFloat, struct.Struct("!d").unpack)
+unpack_uint4_uint2 = cast(Unpack2Int, struct.Struct("!IH").unpack)
```
dvarrazzo: I think this is not so much of a "base type" and it would be better to move it close to the TID object.

psycopg/psycopg/_adapters_map.py
dvarrazzo: I don't think it's needed: we can simulate backwards compatibility by registering the unknown loader for example. Maybe even creating a totally empty adapters map could be a way to simulate it.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+
+@pytest.mark.parametrize(
+    "cls", [CID, XID, pytest.param(XID8, marks=pytest.mark.pg(">=13"))]
+)
+def test_quote_int_catalog_types(conn, cls):
+    tx = Transformer()
+    val = cls.from_int(12)
+    typname = cls.__name__.lower()
+
+    assert tx.get_dumper(val, PyFormat.TEXT).quote(val) == b"'12'"
```
dvarrazzo: Instead of using a low-level object such a Transformer you can make this test using sql.SQL or a server-side cursor and verify that `'12'` is in the resulting query.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+
+    cur = conn.cursor()
+    cur.execute(sql.SQL("select {v}::" + typname).format(v=sql.Literal(val)))
```
dvarrazzo: Why SQL + string composition? Just one or the other:

```suggestion
    cur.execute(sql.SQL("select {}::{}").format(val, sql.Identifier(typname)))
```

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+@pytest.mark.parametrize("with_loaders", [True, False])
+def test_roundtrip_lsn(conn, val, fmt_in, fmt_out, with_loaders):
+    cur = conn.cursor(binary=fmt_out)
+    typname = "pg_lsn"
+
+    if not with_loaders:
+        unregister_loaders(cur.adapters, typname, fmt_out)
+
+    result = cur.execute(f"select %{fmt_in.value}::{typname}", (val,)).fetchone()[0]
+    assert cur.pgresult.fformat(0) == fmt_out
```
dvarrazzo: I don't think it's a relevant check. Maybe leave it in but I think it's redundant.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+    assert cur.pgresult.ftype(0) == builtins[typname].oid
```
dvarrazzo: This is probably better read from `cur.description`, which is the public interface.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+    cur = conn.cursor()
+    cur.execute(sql.SQL("select {v}::" + typname).format(v=sql.Literal(val)))
+    assert isinstance(val, cls)
+    assert cur.fetchone()[0] == val
+
+
+@pytest.mark.parametrize("fmt_in", PyFormat)
+@pytest.mark.parametrize("fmt_out", pq.Format)
+@pytest.mark.parametrize("val", ["0/0", LSN.from_int(2**64 - 1)])
+@pytest.mark.parametrize("with_loaders", [True, False])
```
dvarrazzo: If we really want to test compatibility with previous behaviour I would leave these tests as a separate one, instead of a parametrization of this one, which interfers badly with binary.

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+
+    result = cur.execute(f"select %{fmt_in.value}::{typname}[]", ([val],)).fetchone()[0]
+    assert cur.pgresult.fformat(0) == fmt_out
+    assert cur.pgresult.ftype(0) == builtins[typname].array_oid
+    assert isinstance(result[0], str)
+    assert isinstance(result[0], LSN) is with_loaders
+    assert result[0] == val
+
+
+def test_quote_lsn(conn):
```
dvarrazzo: Maybe the quoting tests can be munged into a single test parametrized with a vector of inputs (type, input value, expected output value)?

tests/types/test_catalog.py
```diff
@@ -0,0 +1,250 @@
...
+        assert int2vector == "12 2 0"
+
+    @pytest.mark.parametrize(
+        "oidvector",
+        [OidVector.from_list([12, 2, 0]), OidVector.from_buffer(b"12 2 0")],
+    )
+    def test_oidvector(self, oidvector):
+        assert oidvector.value == [12, 2, 0]
+        assert repr(oidvector) == "OidVector('12 2 0')"
+        assert oidvector == "12 2 0"
```
dvarrazzo, line 514: Can we have a `test_system_columns` test to fetch the hidden columns of a table, fetch a record, and verify that they are all of the type expected? E.g. starting from (`person` is just a test table):

```
piro=# select array_to_string(array_agg(attname), ', ') from pg_attribute where attrelid = 'person'::regclass and attnum < 1;
            array_to_string             
----------------------------------------
 tableoid, cmax, xmax, cmin, xmin, ctid
(1 row)
```

dvarrazzo:
> The pool-integration test failures seems to be a wider issue.

We work around the issue in several test modules simply catching `ImportError`, see here for example:
https://github.com/psycopg/psycopg/blob/85413678ae3577337f510a96d9e4f253a53668b0/tests/test_capabilities.py#L7-L11

DylanYoung:
> > The pool-integration test failures seems to be a wider issue.
> 
> We work around the issue in several test modules simply catching `ImportError`, see here for example:
> 
> https://github.com/psycopg/psycopg/blob/85413678ae3577337f510a96d9e4f253a53668b0/tests/test_capabilities.py#L7-L11

Great!  Will add that.

dvarrazzo (changes requested):
Hello,

as explained in #1322, I wanted to share the implementation of one of the catalog objects. We can confront the two and establish some common guidelines for all the object, so maybe you can review it.

psycopg/psycopg/types/_catalog.py
DylanYoung: I think that will be less efficient and/or require more code duplication, but I'll take a look.  Worth noting that this is just the text representation, not used for binary.

psycopg/psycopg/types/_catalog.py
DylanYoung: I think it's the same result, but I'm happy to use super instead.  I'll make that change.

psycopg/psycopg/types/lsn.py
DylanYoung: Happy to add the magic methods.  Definitely an oversight on my part.  I wasn't sure what ops you wanted these objects to support.

DylanYoung: Should I add the operators to the common base class?   I don't think it would hurt to have the other IntStrs support them.

psycopg/psycopg/types/_catalog.py
DylanYoung: happy to change this.  I was just following the pattern from other types.

psycopg/psycopg/_adapters_map.py
DylanYoung: Sounds good.  Just thought it might be useful if someone wanted the unknown loader for something with a default loader, but I didn't think of registering it manually.

psycopg/psycopg/_struct.py
DylanYoung: Happy to do that (it felt weird to put here).  I just put here because of your comment on the prior PR about not importing struct.

tests/types/test_catalog.py
DylanYoung: Can do.  I'm still getting used to how things are quoted by the sql module.

tests/types/test_catalog.py
DylanYoung: Can remove it.  I adapted these from the boolean roundtrip tests.

tests/types/test_catalog.py
DylanYoung: can do.  Do you think we need to test backwards compatibility?  I was really just making sure for my own sanity :p

tests/types/test_catalog.py
DylanYoung: can do.  I adapted this from another test.  Would you like other uses updated as well?

tests/types/test_catalog.py
DylanYoung: Can do.  Do you think the quoting tests are even necessary if we don't register new text format dumpers?

tests/types/test_catalog.py
DylanYoung: Can do.

psycopg/psycopg/types/_catalog.py
DylanYoung: @dvarrazzo   I just checked and this is from an older implementation; the current version uses a cached_property for value, so it's lazily computed when using the default contructor.  Sorry! Github didn't notify me that you'd reviewed this already for some reason (I thought this review just came in).

DylanYoung:
@dvarrazzo Thanks for the great feedback.  I've left a few questions/comments.  I'll probably get to this tomorrow.

psycopg/psycopg/types/_catalog.py
dvarrazzo: It's the same result until the string contains a single quote...

psycopg/psycopg/_struct.py
dvarrazzo: I know: the rationale is to have generic functions in `_struct` but a (uint4, uint2) pair seems a pretty specific one. For example if you look at `numeric.py` you'll see both `_struct` and `struct` used, the former to use generic `unpack_float8` types of function, the latter for things specific of numeric structures that we don't expect to reuse (so e.g. `_unpack_numeric_head()` is defined in that module).

psycopg/psycopg/types/_catalog.py
dvarrazzo: I apologise, it might have been my fault too. I started the review 3-4 days ago but had to suspend it because busy with personal reasons and in order to prepare the LSN MR to show you as an example and to compare with.

psycopg/psycopg/types/_catalog.py
dvarrazzo: Yes, I am aware it is somewhat arbitrary. The `psycopg.types` package exposes modules containing adapters related group of types, such as datetime or network objects. I see "the catalog objects" as a family of objects that have sense to group together, but I'm aware it's a pretty bikesheddable opinion.

At the moment it doesn't seem that this module would become massive, not larger than numeric or dates module, so probably we can get away with implementing everything into a single `catalog.py`, which would simplify imports and common base classes.

tests/types/test_catalog.py
dvarrazzo: I see this pattern used in libpq function tests, which were among the first written, and then in numeric and bool types, which were probably among the first adapters implemented. So I assume that these were pretty ancient tests of a time in which we were just moving from the libpq wrapping layer to creating the high level adapters, so they made sense at time.

I think the test suite is full of patters that made sense at time because of the lack of certain structures and which now are no more idiomatic: maybe for example the bool adapter was implemented before cursor.description or maybe now we give for granted that the type and format of the result are the correct one because those objects are more battle-tested than at the time of writing those tests.

I don't know how much is worth cleaning up the historical tests to align them to more current pattern: probably massive and open-ended. My comment is only in the context of this MR and seeing this pattern in the context of the current state of the codebase.

tests/types/test_catalog.py
dvarrazzo: I don't think they are strictly necessary, no, but they are a nice touch for testing in-depth. If they don't complicate massively the test suite I would keep them, but they are more on the nice-to-have side of things.

tests/types/test_catalog.py
dvarrazzo: It's a sensible thing to test, but in a way we are also sort of implicitly testing it in other ways. I think that having this parameter complicates this test more than reducing it, but maybe the test is simple elsewhere.

FYI, the thing I was suggesting, to test with an empty map, might go for example this way:

```python
# Create an adapters map which knows all the types but not how to adapt them.
# I'm not sure we even need the types, nor the binary unknown as we want to test text only.
>>> empty_map = psycopg.adapt.AdaptersMap(types=psycopg.adapters.types)
>>> empty_map.adapters.register_loader(0, psycopg.adapters.get_loader(0, psycopg.pq.Format.BINARY))
>>> empty_map.adapters.register_loader(0, psycopg.adapters.get_loader(0, psycopg.pq.Format.TEXT))

# Create a connection using this specific map
>>> conn = psycopg.connect("", context=empty_map)

# We can see that it doesn't know how to adapt something that by default we can.
>>> conn.execute("select now()").fetchone()[0]
'2026-06-11 14:04:00.582025+01'
```

psycopg/psycopg/types/_catalog.py
DylanYoung: Sounds good to me!  I don't like all the tiny files anyways :p

psycopg/psycopg/types/_catalog.py
DylanYoung: @dvarrazzo  🤦 .  Of course.  I'll throw that in the tests too.

tests/types/test_catalog.py
DylanYoung: @dvarrazzo  Sounds good.

psycopg/psycopg/types/_catalog.py
dvarrazzo: I don't think a test is needed: a round tripping repr is just a convenience, not a dependable interface 🙂

DylanYoung:
> > The pool-integration test failures seems to be a wider issue.
> 
> We work around the issue in several test modules simply catching `ImportError`, see here for example:
> 
> https://github.com/psycopg/psycopg/blob/85413678ae3577337f510a96d9e4f253a53668b0/tests/test_capabilities.py#L7-L11

FYI: I had to do a bit more than just catch the import error, since the classes are used at the module level.

psycopg/psycopg/types/_catalog.py
DylanYoung: Plus, now that I think of it, none of these objects should have quotes in them. 

This is fixed now in any case :)

psycopg/psycopg/types/lsn.py
DylanYoung: I've added these to the IntStr base class as I think they all make sense as operations on the other int strs (CID, XID, and XID8).  The other classes have the ordering operators implemented as well. 

EDIT: I've also added the ability to subtract an int from an IntStr and get an IntStr back (subject to underflow checks).

psycopg/psycopg/types/_catalog.py
DylanYoung: @dvarrazzo  I couldn't use cached_property anymore after adding __slots__, but I use `__getattr__` now to populate `value` if it hasn't been filled in and have unified the interface a bit.  I think it should be nice and fast to construct them now.  Let me know what you think when you get a chance.

tests/types/test_catalog.py
DylanYoung: I've moved these to separate tests.

FYI Setting up the empty map was a bit more complicated than that.  I needed to add the string dumper for all pyformats as well and register the arrays.

tests/types/test_catalog.py
DylanYoung: I have kept them and parametrized them with a vector as you suggested.

tests/types/test_catalog.py
DylanYoung: I haven't done this yet, but I plan to.

psycopg/psycopg/types/_catalog.py
DylanYoung: I was wrong.  This was easy to get rid of and really much better without it (basically this just duplicated the regular constructor with an encoding: `str(b: bytes, encoding=...)`) .  Thanks for the suggestion!

psycopg/psycopg/types/_catalog.py
DylanYoung: I haven't done this yet, but I plan to.

DylanYoung:
@dvarrazzo This isn't quite done, yet, but in a state for another review when you're ready.  I've kept all the changes in the last four commits for now so it's easier to review. 

What's left:

1. ~~Refactoring to a single catalog.py file~~
2. ~~testing the system columns~~
3. sorting out the details of the new magic methods
4. ~~testing the new magic methods~~
5. docs

DylanYoung:
[Interesting CI failures](https://github.com/psycopg/psycopg/actions/runs/27391991458) using that `empty_map`.  Looks like pure python is failing, but C isn't (sorry I forgot to check before pushing).  Perhaps a bug to fix.... I'll investigate. 

EDIT: Figured it out.  The python code was grabbing the NoneType dumper even when the params didn't contain None.  I've fixed that (ac16cb7a) and thrown in a small micro-optimization (00d7e164).

DylanYoung:
@dvarrazzo  OK.  I think everything that you had mentioned is done now except the docs.   I may move that bug I found over to a different branch and create an actual regression test for it.  I still haven't cleaned up the commits in case you wanted to look at the individual changes since the last review.   I think there are ~~two~~ a few primary outstanding issues to discuss:

1. `__new__` versus lazy loading with `__getattr__`.  The benefit of `__new__` is immediate overflow errors when constructing the objects manually.  The benefit of `__getattr__` is faster loading from the database (at the cost of a tiny hit the first time the object needs to access `value` and delayed overflow checks for manually constructed instances).  
2. Exactly which magic methods to support and their semantics. 
3. Docs.  You want these in a new section under "Adapting other PostgreSQL types"?  Do you want that section at the very end? 
4. I just had one other thought: I think we could probably get away with just having one XID class with some sort of flag to indicate whether it includes the epoch or not.  It would be more pythonic, at the cost of a small performance hit.

tests/types/test_catalog.py
DylanYoung: Done. I also added tests for the other types not included in system columns.

psycopg/psycopg/types/_catalog.py
DylanYoung: This is done.

dvarrazzo:
Hello, thank you for the follow-up! I'll be away for a few days; I'll come back to review this and the others MR you are proposing on the next few days.

Thank you very much:)

PR #1321, opened 2026-06-08
Minor performance, mypy upgrade, and other small fixes

### LINKED ISSUE #1224: Check why all workflows run on tag now
Has something changed in the github actions spec?

Image

DylanYoung:
Branch includes the following:

- avoid copying memoryview to bytes where unnecessary
- update README.md HACKING section to suggest skipping tests that might fail locally (subprocess, proxy, timing)
- add explicit tags-ignore to workflows (may fix #1224: the docs still say it shouldn't be necessary, but it shouldn't hurt even if it doesn't help)
- upgrade mypy to version 2.1.0 (it's much faster and handles bytes properly by default; all typing tests and linting still pass locally)
- fixes one sometimes-failing test.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,12 @@ Now hack away! You can run the tests using::
     export PSYCOPG_TEST_DSN="dbname=psycopg_test"
     pytest
 
+If some of the tests fail on your local host, it may be helpful to exclude the
+`proxy`, `subprocess`, and/or `timing` tests to get a clean test run, for
+example::
+
+     pytest  -m 'not proxy and not timing'
+
 The project includes some `pre-commit`__ hooks to check that the code is valid
 according to the project coding convention. Please make sure to install them
 by running::
--- a/psycopg/psycopg/types/array.py
+++ b/psycopg/psycopg/types/array.py
@@ -172,8 +172,6 @@ class ListDumper(BaseListDumper):
                         tokens.append(b"NULL")
                     else:
                         if needs_quotes(ad):
-                            if not isinstance(ad, bytes):
-                                ad = bytes(ad)
                             ad = b'"' + self._re_esc.sub(rb"\\\1", ad) + b'"'
                         tokens.append(ad)
                 else:
--- a/psycopg/psycopg/types/net.py
+++ b/psycopg/psycopg/types/net.py
@@ -6,6 +6,7 @@ Adapters for network types.
 
 from __future__ import annotations
 
+import re
 from typing import TYPE_CHECKING, TypeAlias
 from collections.abc import Callable
 
@@ -36,6 +37,7 @@ PGSQL_AF_INET = 2
 PGSQL_AF_INET6 = 3
 IPV4_PREFIXLEN = 32
 IPV6_PREFIXLEN = 128
+_SLASH_RE = re.compile(b"/")
 
 
 class _LazyIpaddress:
@@ -126,13 +128,10 @@ class _LazyIpaddressLoader(Loader, _LazyIpaddress):
 
 class InetLoader(_LazyIpaddressLoader):
     def load(self, data: Buffer) -> Address | Interface:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-
-        if b"/" in data:
-            return ip_interface(data.decode())
+        if _SLASH_RE.search(data):
+            return ip_interface(str(data, "utf-8"))
         else:
-            return ip_address(data.decode())
+            return ip_address(str(data, "utf-8"))
 
 
 class InetBinaryLoader(_LazyIpaddressLoader):
@@ -158,10 +157,7 @@ class InetBinaryLoader(_LazyIpaddressLoader):
 
 class CidrLoader(_LazyIpaddressLoader):
     def load(self, data: Buffer) -> Network:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-
-        return ip_network(data.decode())
+        return ip_network(str(data, "utf-8"))
 
 
 class CidrBinaryLoader(_LazyIpaddressLoader):
--- a/psycopg/psycopg/types/net.py
+++ b/psycopg/psycopg/types/net.py
@@ -174,8 +174,6 @@ class CidrBinaryLoader(_LazyIpaddressLoader):
         else:
             return IPv6Network((packed, prefix))
 
-        return ip_network(data.decode())
-
 
 def register_default_adapters(context: AdaptContext) -> None:
     adapters = context.adapters
--- a/psycopg/psycopg/types/numeric.py
+++ b/psycopg/psycopg/types/numeric.py
@@ -267,9 +267,7 @@ class Float8BinaryLoader(Loader):
 
 class NumericLoader(Loader):
     def load(self, data: Buffer) -> Decimal:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-        return Decimal(data.decode())
+        return Decimal(str(data, "utf-8"))
 
 
 DEC_DIGITS = 4  # decimal digits per Postgres "digit"
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -11,6 +11,8 @@ on:
     paths-ignore:
       - "docs/*"
       - "tools/*"
+    tags-ignore:
+      - "**"
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,6 +5,8 @@ on:
     branches:
       # This should match the DOC3_BRANCH value in the psycopg-website Makefile
       - master
+    tags-ignore:
+      - "**"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,6 +6,8 @@ on:
     # on.<push|pull_request>.<branches|tags> GitHub Actions docs.
     branches:
       - "**"
+    tags-ignore:
+      - "**"
   pull_request:
   schedule:
     - cron: '18 6 * * *'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,8 @@ on:
     # on.<push|pull_request>.<branches|tags> GitHub Actions docs.
     branches:
       - "**"
+    tags-ignore:
+      - "**"
     paths-ignore:
       - README.rst
       - "docs/*"
--- a/psycopg/psycopg/types/string.py
+++ b/psycopg/psycopg/types/string.py
@@ -112,9 +112,7 @@ class TextLoader(Loader):
 
     def load(self, data: Buffer) -> bytes | str:
         if self._encoding:
-            if isinstance(data, memoryview):
-                data = bytes(data)
-            return data.decode(self._encoding)
+            return str(data, self._encoding)
         else:
             # return bytes for SQL_ASCII db
             if not isinstance(data, bytes):
--- a/psycopg/psycopg/types/uuid.py
+++ b/psycopg/psycopg/types/uuid.py
@@ -43,7 +43,7 @@ class UUIDLoader(Loader):
             from uuid import UUID
 
     def load(self, data: Buffer) -> uuid.UUID:
-        return UUID((bytes(data) if isinstance(data, memoryview) else data).decode())
+        return UUID(str(data, "utf-8"))
 
 
 class UUIDBinaryLoader(UUIDLoader):
--- a/psycopg/pyproject.toml
+++ b/psycopg/pyproject.toml
@@ -71,7 +71,7 @@ pool = [
 test = [
     "anyio >= 4.0",
     # Mypy >= 1.19 depends on librt, not available for PyPy.
-    "mypy >= 1.19.0; implementation_name != \"pypy\"",
+    "mypy >= 2.1.0; implementation_name != \"pypy\"",
     "pproxy >= 2.7",
     "pytest >= 6.2.5",
     "pytest-cov >= 3.0",
@@ -86,7 +86,7 @@ dev = [
     "flake8 >= 4.0",
     "isort[colors] >= 6.0",
     "isort-psycopg",
-    "mypy >= 1.19.0",
+    "mypy >= 2.1.0",
     "pre-commit >= 4.0.1",
     "types-setuptools >= 57.4",
     "types-shapely >= 2.0",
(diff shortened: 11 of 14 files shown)


DylanYoung:
@dvarrazzo  I've seen this test fail twice in CI now: `tests/test_dns_srv.py::test_srv_async`.  I'm not 100% sure if it's just a problem with the test itself or a problem with the `_dns` module.  I can take a crack at fixing it if it's just the test that's wrong (e.g. by comparing a set of (host, port) tuples instead of a string).  I just don't want to fix the test if it's a real issue :p 

(the other transient failure might be fixable by setting DOCKER_CLIENT_TIMEOUT, but I'm not sure if it's worth it; depends how often you see it)

EDIT (link to CI run): 
https://github.com/psycopg/psycopg/pull/1321/checks?sha=9f5f5d6ab3167504372426eabb013fb06816e73c

(I tried to skip the checks on the fixup commit I just pushed, but ~~github doesn't seem to respect the trailer mentioned [here](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) anymore *sigh*~~ I screwed up using `--cleanup=verbatim` 🤦)

psycopg/psycopg/types/array.py
```diff
@@ -172,8 +172,6 @@ def dump_list(obj: list[Any]) -> None:
                         tokens.append(b"NULL")
                     else:
                         if needs_quotes(ad):
-                            if not isinstance(ad, bytes):
-                                ad = bytes(ad)
                             ad = b'"' + self._re_esc.sub(rb"\\\1", ad) + b'"'
```
dvarrazzo, line 175: Thank you: I checked and indeed `Pattern.sub()` supports all the variant of `Buffer`:

```python
>>> re.compile(b'A').sub(b'X', bytearray(b'ABC'))
b'XBC'

>>> re.compile(b'A').sub(b'X', memoryview(b'ABC'))
b'XBC'
```

psycopg/psycopg/types/numeric.py
```diff
@@ -267,9 +267,7 @@ def load(self, data: Buffer) -> float:
 
 class NumericLoader(Loader):
     def load(self, data: Buffer) -> Decimal:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-        return Decimal(data.decode())
+        return Decimal(str(data, "utf-8"))
```
dvarrazzo, line 270: This type of change seems worthwile too, thank you.

psycopg/psycopg/types/net.py
```diff
@@ -126,13 +126,10 @@ def __init__(self, oid: int, context: AdaptContext | None = None):
 
 class InetLoader(_LazyIpaddressLoader):
     def load(self, data: Buffer) -> Address | Interface:
-        if isinstance(data, memoryview):
-            data = bytes(data)
-
-        if b"/" in data:
-            return ip_interface(data.decode())
+        if ord(b"/") in data:
```
dvarrazzo: This tripped me. I made a test and I understand that `bytes in memoryview` returns false. This requires a comment.

Another observation: we can make a module env var `_SASH_INT = ord(b"/")` to save the call here which is really a constant. Similar to:

https://github.com/psycopg/psycopg/blob/a6d5134282f5fd532cffeb9bed4694265650b233/psycopg/psycopg/types/multirange.py#L336-L338

I have also played a bit with microbenchmarks. `int in bytes` and `int in bytearray` is about 5x faster than `bytes in bytes` and `bytes in bytearray` (likely the more complex substring search kicks in and there's no shortcut for the 1 byte needle):

```
In [59]: data = b'10.20.30.40/50'

In [61]: %timeit _SLASH_INT in data
22.5 ns ± 0.743 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [63]: %timeit b"/" in data
113 ns ± 0.342 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

However `int in memoryview` is slower, in the order f `bytes in bytes`:

```
In [64]: data = memoryview(b'10.20.30.40/50')

In [65]: %timeit _SLASH_INT in data
135 ns ± 1.26 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

For memoryview, using a regex turns out to be actually ~2x faster.

```
In [75]: re_slash = re.compile(b"/")

In [76]: %timeit re_slash.search(data)
74.4 ns ± 0.454 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

So maybe it's worth using a regexp, as we likely end up with a memoryview (at least with the C transformer)?

Note: above tests are with Python 3.10. Maybe it's worth repeating them with more modern Pythons, 3.13 or 3.14.

tests/pool/test_pool.py
```diff
@@ -478,7 +478,7 @@ def worker(n):
 
         sleep(1)
 
-    assert results == [(4, 4), (4, 3), (3, 2), (2, 2), (2, 2)]
+    assert results[:5] == [(4, 4), (4, 3), (3, 2), (2, 2), (2, 2)]
```
dvarrazzo: In which way this is flakey for you? I don't remember it being so.

README.rst
```diff
@@ -88,6 +88,11 @@ Now hack away! You can run the tests using::
     export PSYCOPG_TEST_DSN="dbname=psycopg_test"
     pytest
 
+Some of the tests might fail on your local host. It may be helpful to exclude the
+`proxy`, `subprocess`, and/or `timing` tests to get a clean test run::
+
+     pytest  -m 'not proxy and not subprocess and not timing'
```
dvarrazzo: I am not sure about this statement. These markers mostly exist to make the tests pass on Github Actions, where the Windows and the macOS runners seem particularly slow, probably overcommitted.

What are the tests that you are finding flakey? What hardware are you locally using?

dvarrazzo (changes requested):
Thank you, I would say that these changes are worth applying

Thank you for teaching me the use of the `str()` function 😅 

Could we please review what to do around the net loader, and could you please rebase and squash the commit and two fixups pertaining to mypy into a single commit?

README.rst
DylanYoung: I'm developing on my old intel mac.  Some of the timing tests work after my wait_async work, but some still fail.  I'm not sure what the issue is with subprocess, but I suspect maybe some system integrity protection issues as they only fail with the c implementation. 

The proxy issues are super wierd, they pass if I insert a pdb statement :/

psycopg/psycopg/types/net.py
DylanYoung: I'll do some testing. Thanks!

DylanYoung:
> Thank you, I would say that these changes are worth applying
> 
> Thank you for teaching me the use of the `str()` function 😅
> 
> Could we please review what to do around the net loader, and could you please rebase and squash the commit and two fixups pertaining to mypy into a single commit?

will do.  Should get to that tonight!

tests/pool/test_pool.py
DylanYoung: This is probably related to the timing failures on my machine.  It sometimes returns 6 results iirc.

README.rst
DylanYoung: It took me a while to figure out how to get a clean test run, so I think some kind of note would be helpful, but I'm happy to reword it.  Maybe be more explicit about trying all the tests before using these flags? And maybe use only one of the flags in the example?

dvarrazzo: I think it is worth suggesting to skip these tests in case someone's computer is slow or a certain pattern of failure is seen (largely because I don't use Windows or macOS regularly so maybe some subtleties in these tests which might depend on platform implementation behaviours are not correctly accounted for).

I propose a less strict wording, something along the lines of:

> Certain tests might be not reliable if your machine is particularly slow or if some part of the testing infrastructure doesn't work correctly on your platform.
>
> Tests that should be skipped for a certain reason have the proper marker, for example:
>
> - `timing`: tests depending on somewhat precise timing
> - `flakey`: tests inherently unreliable
> - `proxy`: tests involving setting up a connection proxy to connect to the test database (e.g. to disrupt the connection)
> 
> Other markers exists: you can get a full list using `pytest --markers`.
>
> If certain groups of tests fail regularly on your machine you can use `pytest -m` with an expression to avoid them, for example `pytest -m "not proxy and not timing"`

dvarrazzo: > The proxy issues are super wierd, they pass if I insert a pdb statement :/

The Heisenbug! It might be a timing issue?

psycopg/psycopg/types/net.py
DylanYoung: Looks to me like `re` is the clear winner for this use case.  My string is bigger than yours so the numbers are a bit different. 
`re` is about O(300 nsec) for bytes and O(10 nsec) for memoryview.
`b"/" in bytes_o` is about O(300 nsec) as well, so no real degradation for bytes. 
`SLASH_INT in memory_o` is O(10 usec), so unacceptably slow.  

I do wonder why the regex so much slower on bytes and the `ord` is so much slower on memoryview, but it is what it is. 

The numbers are slightly different, but consistent between 3.11 and 3.14. 

I'lll make this change.

README.rst
DylanYoung: I've updated the wording and example here as I suggested above to try to make it clear that this should be used only for the specific tests that are failing.

psycopg/psycopg/types/net.py
DylanYoung: This is done.

DylanYoung:
@dvarrazzo  This should be ready for another review!

DylanYoung:
Hold on.  CI must have updated cython-lint.  I'll cherry-pick those fixes from my other branch. 

EDIT: done.

pjonsson:
> * upgrade mypy to version 2.1.0 (it's much faster and handles bytes properly by default; all typing tests and linting still pass locally)

I don't know if it's already done elsewhere, but if you set the environment variable `MYPY_NUM_WORKERS` to the  number of cores of the machine it's running on you should get another performance boost with MyPy 2.x.

tests/pool/test_pool.py
dvarrazzo: The test is marked `timing` so I would rather it is skipped with other timing-based tests on your machine rather than changed this way. The original test asks that exactly 5 items are in the list and your change could mask the presence of a spurious 6th one.

dvarrazzo:
@DylanYoung I am rebasing this branch on master. Two commits can be dropped now as on master already (the ones related to cython-linter), and I would like to drop the flakey pool test.

I am unable to push in this branch. Can you enable me to do so?

dvarrazzo:
Thank you @pjonsson it's worth taking a look.

The time spent in CI is negligible. If we can speed up the command line runs that would be useful. For example, at the moment mypy only runs on the files changed, because `pass_filenames: false` (which is the advised way to use it - see https://github.com/python/mypy/issues/13916) is noticeably slower.

The problem is to know the number of cores in a machine. Also the feature is not necessarily stable.

```
$ mypy --version && mypy --help | grep -A2 num-workers
mypy 2.1.0 (compiled: yes)
  -n NUM_WORKERS, --num-workers NUM_WORKERS
                            Number of separate mypy worker processes
                            (experimental)
```

we could try dropping a `--num-workers 4` in pre-commit config and stop passing filenames to see if the performance is acceptable (at least not noticeably slower than before)

DylanYoung:
@dvarrazzo I've done the rebase and dropped the commits you mentioned.  I've also given you push access in case you need to make any more changes.

DylanYoung:
> > * upgrade mypy to version 2.1.0 (it's much faster and handles bytes properly by default; all typing tests and linting still pass locally)
> 
> I don't know if it's already done elsewhere, but if you set the environment variable `MYPY_NUM_WORKERS` to the number of cores of the machine it's running on you should get another performance boost with MyPy 2.x.

Nice idea.  It runs pretty fast right now (compared to before), but I'll experiment with what @dvarrazzo  suggested locally and see if it's feasible to change the pass_filenames argument.

README.rst
DylanYoung: > > The proxy issues are super wierd, they pass if I insert a pdb statement :/
> 
> The Heisenbug! It might be a timing issue?

😆 

It might be a timing issue, except I tried introducing other timing changes and couldn't get them working; only the pdb trace reliably got them working 🤷‍♂️

DylanYoung:
@dvarrazzo After a quick experiment, what I would recommend is something like this:

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 3f2fd0ce..2bbd6049 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@
 # inplace removing the `exclude` entries in order to run linting of those files
 # too in CI.

+default_install_hook_types: [pre-push, pre-commit]
 repos:
   - repo: local
     hooks:
@@ -42,11 +43,21 @@ repos:

       - id: mypy
         name: mypy
+        stages: [pre-commit, manual]
         language: system
         entry: mypy --pretty --follow-imports=silent
         files: \.py[i]?$
         exclude: tests/test_tstring(_async)?.py

+      - id: mypy-all
+        name: mypy
+        stages: [pre-push, manual]
+        language: system
+        entry: mypy --pretty --follow-imports=silent --num-workers=4
+        files: \.py[i]?$
+        exclude: tests/test_tstring(_async)?.py
+        pass_filenames: false
+
       - id: cython-lint
         name: cython-lint
         language: system
```

Rationale:  It's nice to have the commit hooks run fast and they do that now with this upgrade.  Passing num_workers seems to slow them down (I suspect the process overhead isn't worth it for `pass_filenames: true`).   But a pre-push hook can be a bit slower and would help prevent spurious CI problems.   Happy to throw up a PR if you agree. 

CC @pjonsson

dvarrazzo:
I assume that `pre-push` would run before push on the client? I am not sure I like the idea of mypy slowing down both commit and pushing.

DylanYoung:
@dvarrazzo It runs before push with my example change, but you can just change `default_install_hook_types` to `[pre-commit]` and it will be the same as before, but with the option to run on push if the user installs the hook manually, which I do as many of my commits are created with git-absorb, which doesn't run pre-commit hooks.

dvarrazzo:
I think that's too many choices. Let's leave this for future times, when mypy parallelism becomes more convenient

DylanYoung:
> I think that's too many choices. Let's leave this for future times, when mypy parallelism becomes more convenient

I think maybe you misunderstood my last comment.  You could also just leave off `default_install_hook_types` as pre-commit only is the default, so nothing would really change for the developer unless they wanted it to, and you could leave it undocumented.  But I don't mind either way.  If you don't do it, my pushes are faster :)

dvarrazzo:
Ok, finally closed, thank you very much! On to the other MRs, thank you very much for your work!

PR #1322, opened 2026-06-11
Add pg_lsn adapter

dvarrazzo:
This adapters comes from a work in progress from another project but it will be useful in psycopg too.

Because a few MR are already addressing adding catalog type (#1285, #1316), I am sharing here an example. We may compare this approach with others' and create an uniform family of objects.

Missing: docs

cc @DylanYoung.


DylanYoung:
@dvarrazzo  Looks nice.  I'll update my PR with the missing magic methods and the binary dumper.  Do you think the text dumper is necessary to add as well since the regular text text dumper will handle it?  I think the class constructors I have will be slightly more performant, but happy to update if you prefer the overriding `__new__` approach.

dvarrazzo:
@DylanYoung 

> I'll update my PR with the missing magic methods

Please note that not necessarily all catalog types need implementing complete arithmetic. I haven't looked into the semantics of each of the types in your MR and I don't have it fresh on top of my head. Maybe not all those types represent an Euclidean space like LSNs do (i.e. supporting LSN - LSN -> int, LSN + int -> LSN operators).

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+    def __int__(self) -> int:
+        return self._value
+
+    def __bool__(self) -> bool:
+        return bool(self._value)
+
+    def __add__(self, other: int) -> Lsn:  # type: ignore[override]
+        return Lsn(int(self) + other)
+
+    def __sub__(self, other: Lsn) -> int:
```
DylanYoung, line 57: @dvarrazzo Is there a reason you want to subtract two LSNs?  Seems like it would be more useful to subtract an int from an LSN and get an LSN back (and do an overflow check of course), but maybe I'm missing some use-case for this.

DylanYoung: I guess I can understand this now as a means of calculating the distance between two LSNs.  I still think what I suggested may be useful in some cases, so I may add that to my implementation as well.

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class LsnBinaryLoader(Loader):
+    format = Format.BINARY
+
+    def load(self, data: Buffer) -> Lsn:
+        return Lsn(unpack_uint8(data)[0])
+
+
+class LsnDumper(Dumper):
```
DylanYoung, line 93: Are the dumpers necessary?  I suppose they provide a small performance improvement?

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __ge__(self, other: object) -> bool:
+        if isinstance(other, str):
+            return int(self) >= int(Lsn(other))
+        return NotImplemented
+
+
+class LsnLoader(Loader):
+    def load(self, data: Buffer) -> Lsn:
+        return Lsn(bytes(data).decode())
```
DylanYoung, line 83: unnecessary bytes copy (use str with encoding arg directly)

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+    def __bool__(self) -> bool:
+        return bool(self._value)
+
+    def __add__(self, other: int) -> Lsn:  # type: ignore[override]
+        return Lsn(int(self) + other)
+
+    def __sub__(self, other: Lsn) -> int:
+        return int(self) - int(other)
+
+    def __lt__(self, other: object) -> bool:
```
DylanYoung, line 60: It would be more efficient to access `_value` directly in these methods rather than going though '__int__`

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+        lsn._value = value
+        return lsn
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({super().__repr__()})"
+
+    def __int__(self) -> int:
+        return self._value
+
+    def __bool__(self) -> bool:
```
DylanYoung, line 51: I implemented this in mine as well, but I'm not convinced this is actually useful.  Why should the first log segment number have any special meaning here?  Or do LSNs start at 1?

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+    Subclass of ``str`` so existing code that receives ``pg_lsn`` as a plain
+    string continues to work. Ordering operators use integer comparison, which
+    is correct for LSNs (string ordering is wrong when the hex segments differ
+    in digit count, e.g. ``"F/0"`` vs ``"10/0"``).
+    """
+
+    __slots__ = ("_value",)
+
+    _value: int
+
```
DylanYoung, line 28: Using __new__ provides a lot of benefits, but if the goal is loading speed, take a look at my approach using __getattr__.  I'd especially consider `__getattr__` if you're planning on decoding the logical message stream LSNs as `Lsn` classes, as you want that to be pretty fast.

psycopg/psycopg/types/catalog.py
```diff
@@ -0,0 +1,113 @@
...
+    Subclass of ``str`` so existing code that receives ``pg_lsn`` as a plain
+    string continues to work. Ordering operators use integer comparison, which
+    is correct for LSNs (string ordering is wrong when the hex segments differ
+    in digit count, e.g. ``"F/0"`` vs ``"10/0"``).
+    """
+
+    __slots__ = ("_value",)
+
+    _value: int
```
DylanYoung, line 27: Is there a good reason for value to be private?

tests/fix_crdb.py
```diff
@@ -119,6 +119,7 @@ def crdb_scs_off(*args):
     "severity_nonlocalized": 81794,
     "stored procedure": 1751,
     "to_regtype": None,
+    "pg_current_wal_lsn": None,
```
DylanYoung, line 122: crdb doesn't support the pg_lsn type at all, does it?  Or any other catalog types for that matter, right?  Because of that I just marked the whole test file, and added "catalog types" here.

psycopg/psycopg/crdb/_types.py
```diff
@@ -151,7 +153,9 @@ def register_crdb_types(types: TypesRegistry) -> None:
         TypeInfo("numeric", 1700, 1231, typemod=NumericTypeModifier),
         TypeInfo("oid", 26, 1028),
         TypeInfo("oidvector", 30, 1013),
+        TypeInfo("pg_lsn", 3220, 3221),
```
DylanYoung, line 156: Oh, this is new.  Disregard my last comment about marking the entire file with crdb_skip.   I'll have to think about marking my tests differently.  I didn't realize crdb had an oidvector either.

tests/types/test_lsn.py
```diff
@@ -0,0 +1,135 @@
...
+    lsn_f = Lsn("F/0")
+    lsn_10 = Lsn("10/0")
+    assert lsn_f < lsn_10
+    assert lsn_10 > lsn_f
+    assert lsn_f <= lsn_10
+    assert lsn_f <= lsn_f
+    assert lsn_10 >= lsn_10
+
+
+def test_equality() -> None:
```
DylanYoung, line 87: No support for int equality?

psycopg/psycopg/types/catalog.py
DylanYoung: Ah, nevermind.  0 is InvalidXLogRecPtr.  This is sensible.

tests/fix_crdb.py
dvarrazzo: It supports the type, which seems to behave well, but it doesn't have a `pg_current_wal_lsn()` for a whole E2E test. I haven't found another suitable function, however I think it's not necessary and this test is not different than a roundtrip test that we perform elsewhere.

DylanYoung: Agreed.

PR #1323, opened 2026-06-11
Cython Implementations for RecordLoader, RecordBinaryLoader, _CompositeLoader, _CompositeBinaryLoader 

DylanYoung:
Replaces #1175. Implements #1174.

Using the same benchmark mentioned from #1175 (with 150 rows fetched 1000 times):

### C Module:

1.  Row Binary: 0.5739
2. Row Text: 0.4093
3. Composite Binary: 0.6924
4.  Composite Text: 0.6378

### Pure Python:

1. Row Binary: 2.2821
2. Row Text: 2.4856
3. Composite Binary: 2.2862
4. Composite Text: 3.2623
 
The most debatable bit is how the composite loaders base classes are plugged in on the python side.  I created a new `abc.CompositeLoader` for the return types of `_make_loader` and `_make_binary_loader` and had to use `types.make_class` instead of `type` as mypy was complaining (not sure why it wasn't before).  Happy to change that as directed.  There's also a small improvement on the python side in 0220c88bf728f97c1ebbaec258cc9bc659dd3e43, which I'm happy to remove if you find it objectionable.  Worth noting that this work also exhibits the pessimization mentioned in #1191, but I have an idea for tackling that all at once, which I'll push when I get to it (it affects row binary, composite binary, and composite text). 

I've left commit e3f178a2a7d83755fa48b3be86288215db265d46 separate for now, as it might be more contentious, but it can be squashed in if approved. 

I still need to add the type stubs (totally forgot until writing this).


DylanYoung:
The one failed CI here is just a docker fetch timeout.

PR #1324, opened 2026-06-13
Faster C Load Sequence

### LINKED ISSUE #1191: pessemization in parse_row_binary

DylanYoung:
Fixes #1191. 

Also encapsulates the API for use in other work (e.g. replication and #1323 or perf follow-up). 

I haven't done a proper benchmark, but I updated the `tests/scripts/copytest.py` to check some basic numbers.  Looks like about a 20% to 35% speedup.   Interestingly, there seems to be an unrelated pessimization in the async code path that doesn't appear on the "COPY IN" side of things.... 

EDIT:  I think the async pessimization is entirely due to the `wait_async` method.  I suspect it could be written in Cython for a healthy speedup in all things async. 

Question:  Why does the `Copy` class take a binary argument? It doesn't work unless the query itself specifies format binary, in which case it can be inferred (as it is by default).  Is the hope that someday there will be support for setting it on the protocol level?   

#### SYNC
0.757028 sec --> 0.498472 sec
#### SYNC BINARY
0.710254 sec --> 0.444449 sec
#### ASYNC
1.110847 sec --> 0.887244 sec
#### ASYNC BINARY
1.069361 sec  --> 0.834796 sec

#### SYNC
0.873483 sec --> 0.532475 sec
#### SYNC BINARY
0.767204 sec --> 0.477692 sec
#### ASYNC
1.135291 sec --> 0.854667 sec
#### ASYNC BINARY
1.165879 sec --> 0.853913 sec

### DIFF
--- a/psycopg_c/psycopg_c/_psycopg/copy.pyx
+++ b/psycopg_c/psycopg_c/_psycopg/copy.pyx
@@ -7,13 +7,12 @@ C optimised functions for the copy system.
 
 from libc.stdint cimport int32_t, uint16_t, uint32_t
 from libc.string cimport memcpy
+from cpython.tuple cimport PyTuple_New, PyTuple_SET_ITEM
 from cpython.sequence cimport PySequence_Fast, PySequence_Fast_GET_ITEM
 from cpython.sequence cimport PySequence_Fast_GET_SIZE
-from cpython.bytearray cimport PyByteArray_AS_STRING, PyByteArray_FromStringAndSize
-from cpython.bytearray cimport PyByteArray_GET_SIZE, PyByteArray_Resize
-from cpython.memoryview cimport PyMemoryView_FromObject
+from cpython.bytearray cimport PyByteArray_AS_STRING, PyByteArray_GET_SIZE
+from cpython.bytearray cimport PyByteArray_Resize
 
-from psycopg_c.pq cimport ViewBuffer
 from psycopg_c._psycopg cimport endian
 
 from psycopg import errors as e
@@ -244,8 +243,10 @@ def parse_row_binary(data, tx: Transformer) -> tuple[Any, ...]:
     memcpy(&benfields, ptr, sizeof(benfields))
     cdef int nfields = endian.be16toh(benfields)
     ptr += sizeof(benfields)
-    cdef list row = PyList_New(nfields)
 
+    tx._c_check_num_loaders(nfields)
+
+    cdef tuple row = PyTuple_New(nfields)
     cdef int col
     cdef int32_t belength
     cdef Py_ssize_t length
@@ -259,14 +260,13 @@ def parse_row_binary(data, tx: Transformer) -> tuple[Any, ...]:
             length = endian.be32toh(belength)
             if ptr + length > bufend:
                 raise e.DataError("bad copy data: length exceeding data")
-            field = PyMemoryView_FromObject(
-                ViewBuffer._from_buffer(data, ptr, length))
+            field = tx._c_load_item(ptr, length, col)
             ptr += length
 
         Py_INCREF(field)
-        PyList_SET_ITEM(row, col, field)
+        PyTuple_SET_ITEM(row, col, field)
 
-    return tx.load_sequence(row)
+    return row
 
 
 def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
@@ -276,11 +276,15 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
 
     # politely assume that the number of fields will be what in the result
     cdef int nfields = tx._nfields
-    cdef list row = PyList_New(nfields)
 
+    tx._c_check_num_loaders(nfields)
+
+    cdef tuple row = PyTuple_New(nfields)
     cdef unsigned char *fend
     cdef unsigned char *rowend = fstart + size
     cdef unsigned char *src
+    cdef unsigned char *scratch = NULL
+    cdef size_t sclen = 0
     cdef unsigned char *tgt
     cdef int col
     cdef int num_bs
@@ -312,15 +316,18 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
         # Is this a field with no backslash?
         elif num_bs == 0:
             # Nothing to unescape: we don't need a copy
-            field = PyMemoryView_FromObject(
-                ViewBuffer._from_buffer(data, fstart, fend - fstart))
+            field = tx._c_load_item(fstart, fend - fstart, col)
 
         # This is a field containing backslashes
         else:
             # We need a copy of the buffer to unescape
-            field = PyByteArray_FromStringAndSize("", 0)
-            PyByteArray_Resize(field, fend - fstart - num_bs)
-            tgt = <unsigned char *>PyByteArray_AS_STRING(field)
+            if <unsigned int>(fend - fstart - num_bs) > sclen:
+                # Above condition will always be true on the first run
+                # since sclen is 0. After that, we only extend scratch
+                # if it's too small.
+                sclen = fend - fstart - num_bs
+                scratch = <unsigned char *>PyMem_Realloc(scratch, sclen)
+            tgt = scratch
             src = fstart
             while (src < fend):
                 if src[0] != b'\\':
@@ -330,15 +337,18 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
                     tgt[0] = copy_unescape_lut[src[0]]
                 src += 1
                 tgt += 1
+            field = tx._c_load_item(scratch, fend - fstart - num_bs, col)
 
         Py_INCREF(field)
-        PyList_SET_ITEM(row, col, field)
+        PyTuple_SET_ITEM(row, col, field)
 
         # Start of the field
         fstart = fend + 1
 
-    # Convert the array of buffers into Python objects
-    return tx.load_sequence(row)
+    if sclen > 0:
+        PyMem_Free(scratch)
+
+    return row
 
 
 cdef extern from *:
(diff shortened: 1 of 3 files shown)


psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -312,15 +316,15 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
...
         # This is a field containing backslashes
         else:
             # We need a copy of the buffer to unescape
-            field = PyByteArray_FromStringAndSize("", 0)
-            PyByteArray_Resize(field, fend - fstart - num_bs)
-            tgt = <unsigned char *>PyByteArray_AS_STRING(field)
+            if fend - fstart - num_bs > sclen:
+                sclen = fend - fstart - num_bs
+                scratch = <unsigned char *>PyMem_Realloc(scratch, sclen)
+            tgt = scratch
```
jerch: This code looks wrong to me - `scratch` gets initialized as NULL, but only set to a valid pointer behind this condition. The block afterwards uses it in any case as `tgt`.

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -312,15 +316,15 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
         # Is this a field with no backslash?
         elif num_bs == 0:
             # Nothing to unescape: we don't need a copy
-            field = PyMemoryView_FromObject(
-                ViewBuffer._from_buffer(data, fstart, fend - fstart))
+            field = tx._c_load_item(fstart, fend - fstart, col)
```
jerch, line 319: Is this replacement correct? `_c_load_item` does a lot more?

psycopg_c/psycopg_c/_psycopg/copy.pyx
```diff
@@ -330,15 +334,18 @@ def parse_row_text(data, tx: Transformer) -> tuple[Any, ...]:
                     tgt[0] = copy_unescape_lut[src[0]]
                 src += 1
                 tgt += 1
+            field = tx._c_load_item(scratch, fend - fstart - num_bs, col)
```
jerch, line 340: Why is `_c_load_item` now needed here?

psycopg_c/psycopg_c/_psycopg/transform.pyx
jerch, line 535: Is this function now obsolete? Haven't checked the caller sides. Could also be, that @dvarrazzo wants it for API purposes.

psycopg_c/psycopg_c/_psycopg/transform.pyx
```diff
@@ -564,6 +563,33 @@ cdef class Transformer:
...
+            raise e.ProgrammingError(
+                f"cannot load item at index {index}:"
+                f" {len(self._row_loaders)} loaders registered")
+        loader = <RowLoader>PyList_GET_ITEM(row_loaders, index)
+        if (<RowLoader>loader).cloader is not None:
+            pyval = loader.cloader.cload(<char *>ptr, size)
+        else:
+            item = PyMemoryView_FromObject(
+                ViewBuffer._from_buffer(self, ptr, size)
+            )
```
jerch, line 584: Did you try the `PyMemoryView_FromMemory` approach and it failed? I think it would give additional performance, but it could be that it does not cover all type edge cases / is less versatile.

jerch (commented):
@DylanYoung Left a few comments, see below. I think the `scratch` pointer handling is faulty, but maybe I have missed a crucial detail.

jerch:
On a sidenote - I am surprised that with `set_types` the runtime is worse. Without `set_types` the loader allocation happens on-the-fly, but with a few indirections. I expected the `set_types` tests to be faster - could be a hint that `set_types` itself does something overly explicit?

psycopg_c/psycopg_c/_psycopg/transform.pyx
DylanYoung: I just tried it, and didn't see an improvement, but I don't have a proper benchmark running yet, so it's possible there's some small gain.  I'm not sure it's worth the cost though.  I worry that some loader might be holding onto the memoryview...  I may revisit it at a later date, though I suspect there are lower-lieing fruit.

DylanYoung:
> On a sidenote - I am surprised that with `set_types` the runtime is worse. Without `set_types` the loader allocation happens on-the-fly, but with a few indirections. I expected the `set_types` tests to be faster - could be a hint that `set_types` itself does something overly explicit?

I suspect this is just a relic of my ad-hoc benchmarking.  What really surprises me is the performance degradation of async.  I might investigate that. 

EDIT: I think the async pessimization is entirely due to the `wait_async` function.  I suspect it could be written in Cython for a healthy speedup in all things async.

psycopg_c/psycopg_c/_psycopg/transform.pyx
DylanYoung: It's still used on the python end (e.g. in the composite and record loaders).  I tried replacing it there, but the extra tuple allocation + c-level load-sequence wins, though just marginally.   Once those loaders are written in C, it could be removed, but I suspect it may still be useful for people implementing new loaders in pure-python, so I probably wouldn't remove the API even then.

psycopg_c/psycopg_c/_psycopg/copy.pyx
DylanYoung: `sclen` is initialized to zero, so on the first pass, this will always allocate memory, then it only allocates more memory if the next string is larger than the largest string seen so far.  `scratch` is initialized as `NULL` so that the first allocation is a malloc, rather than a realloc.

psycopg_c/psycopg_c/_psycopg/copy.pyx
DylanYoung: _c_load_item encapsulates all the code you added.  it replaces `tx.load_sequence(row)`

psycopg_c/psycopg_c/_psycopg/copy.pyx
DylanYoung: Because `_c_load_item` also replaces the work done in `tx.load_sequence(row)`

psycopg_c/psycopg_c/_psycopg/transform.pyx
DylanYoung: If you wanted to pull my code and try it out, I'd be happy to see if there's a measurable difference.

psycopg_c/psycopg_c/_psycopg/copy.pyx
DylanYoung: It's the same logic as in array.pyx and my other PR for composite/record loaders.

DylanYoung:
Latest push fixes a clang warning, adds come more niceties to `tests/scripts/copytest.py`, and removes a check on the index being valid (as the caller of _c_load_item should have previously verified this, e.g. by calling _c_check_num_loaders) .  

I think this last will allow us to use it in `load_row` and `load_rows` without a regression.  I haven't done this yet, because it's a hugely global change and I'd like to check with @dvarrazzo as to whether there's a good way to run a performance regression.  I do have a commit making that change on my local branch.

psycopg_c/psycopg_c/_psycopg/copy.pyx
jerch: Could you add a comment that the condition is always true at first run?

psycopg_c/psycopg_c/_psycopg/transform.pyx
jerch: Yes loaders are hookable with custom impls, so this has to stay for that purpose.

psycopg_c/psycopg_c/_psycopg/transform.pyx
jerch: I am a bit short of time currently, so will prolly not get to test it within the next days.

psycopg_c/psycopg_c/_psycopg/copy.pyx
DylanYoung: Will do!

PR #1326, opened 2026-06-14
Fix Dumping non-None values with no NoneType dumper registered in Python implementation

### LINKED ISSUE #1325: AdaptersMap with no NoneType Loader fails to load non-none objects in Python Implementation
Does not effect C implementation. 

Reproducer:

    str_map = adapt.AdaptersMap()
    str_map.adapters.register_loader(
        INVALID_OID,
        adapters.get_loader(INVALID_OID, pq.Format.TEXT),  # type: ignore
    )
    for fmt in PyFormat:
        str_map.adapters.register_dumper(str, adapters.get_dumper(str, fmt))

    conn = psycopg.connect(dsn, context=str_map)

    with str_conn.cursor() as cur:
        # shouldn't raise, but does
        cur.execute(f"SELECT %{fmt_in.value}", ["not_none"])

    conn.close()

DylanYoung:
Fixes #1325. 

Should ideally be merged before #1316.  Not sure how far back it should be backported, but I can check if you like.  If I should target a maintenance branch instead of master, let me know.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -15,6 +15,8 @@ Psycopg 3.3.5 (unreleased)
 
 - Discard prepared statements upon :sql:`ALTER *` or `DISCARD *`
   (:ticket:`#1307`).
+- Fix `!ProgrammingError` when dumping non-`!None` values with
+  no `!NoneType` dumper registered in python implementation (:ticket:`#1325`).
 - Fix `!wait_selector` wait function to not raise `!KeyError`
   (:ticket:`#1327`).
 
--- a/psycopg/psycopg/_py_transformer.py
+++ b/psycopg/psycopg/_py_transformer.py
@@ -65,6 +65,7 @@ class Transformer(AdaptContext):
 
     def __init__(self, context: AdaptContext | None = None):
         self._pgresult = self.types = self.formats = None
+        self._none_oid = -1
 
         # WARNING: don't store context, or you'll create a loop with the Cursor
         if context:
@@ -187,12 +188,16 @@ class Transformer(AdaptContext):
                     out[i] = self._row_dumpers[i].dump(param)
             return out
 
-        types = [self._get_none_oid()] * nparams
+        types = [-1] * nparams
         pqformats = [TEXT] * nparams
 
         for i in range(nparams):
             if (param := params[i]) is None:
+                if self._none_oid < 0:
+                    self._none_oid = self._get_none_oid()
+                types[i] = self._none_oid
                 continue
+
             dumper = self.get_dumper(param, formats[i])
             out[i] = dumper.dump(param)
             types[i] = dumper.oid
@@ -267,16 +272,9 @@ class Transformer(AdaptContext):
 
     def _get_none_oid(self) -> int:
         try:
-            return self._none_oid
-        except AttributeError:
-            pass
-
-        try:
-            rv = self._none_oid = self._adapters.get_dumper(NoneType, PY_TEXT).oid
+            return self._adapters.get_dumper(NoneType, PY_TEXT).oid
         except KeyError:
             raise e.InterfaceError("None dumper not found")
-
-        return rv
 
     def get_dumper_by_oid(self, oid: int, format: pq.Format) -> abc.Dumper:
         """
--- a/psycopg/psycopg/_py_transformer.py
+++ b/psycopg/psycopg/_py_transformer.py
@@ -193,9 +193,7 @@ class Transformer(AdaptContext):
 
         for i in range(nparams):
             if (param := params[i]) is None:
-                if self._none_oid < 0:
-                    self._none_oid = self._get_none_oid()
-                types[i] = self._none_oid
+                types[i] = self._get_none_oid()
                 continue
 
             dumper = self.get_dumper(param, formats[i])
@@ -271,10 +269,12 @@ class Transformer(AdaptContext):
             return dumper
 
     def _get_none_oid(self) -> int:
-        try:
-            return self._adapters.get_dumper(NoneType, PY_TEXT).oid
-        except KeyError:
-            raise e.InterfaceError("None dumper not found")
+        if self._none_oid < 0:
+            try:
+                self._none_oid = self._adapters.get_dumper(NoneType, PY_TEXT).oid
+            except KeyError:
+                raise e.InterfaceError("None dumper not found")
+        return self._none_oid
 
     def get_dumper_by_oid(self, oid: int, format: pq.Format) -> abc.Dumper:
         """
--- a/tests/test_adapt.py
+++ b/tests/test_adapt.py
@@ -59,6 +59,15 @@ def test_quote_none(data, result, global_adapters):
     assert dumper.quote(data) == result
 
 
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_no_none_dumper_registered(dsn, fmt_in):
+    with psycopg.connect(dsn, context=make_str_map()) as conn:
+        with conn.cursor() as cur:
+            cur.execute(f"SELECT %{fmt_in.value}", ["not_none"])
+            with pytest.raises(e.ProgrammingError, match="NoneType"):
+                cur.execute(f"SELECT %{fmt_in.value}", [None])
+
+
 def test_register_dumper_by_class(conn):
     dumper = make_dumper("x")
     assert conn.adapters.get_dumper(MyStr, PyFormat.TEXT) is not dumper
@@ -561,3 +570,17 @@ def make_bin_loader(suffix):
     cls = make_loader(suffix)
     cls.format = pq.Format.BINARY
     return cls
+
+
+def make_str_map():
+    from psycopg._oids import INVALID_OID
+
+    # Construct a minimal AdaptersMap to no none dumper bug
+    str_map = psycopg.adapt.AdaptersMap()
+    invalid_oid_loader = psycopg.adapters.get_loader(INVALID_OID, pq.Format.TEXT)
+    assert invalid_oid_loader is not None
+    str_map.adapters.register_loader(INVALID_OID, invalid_oid_loader)
+    str_map.register_dumper(str, StrDumper)
+    str_map.register_dumper(str, StrBinaryDumper)
+
+    return str_map


psycopg/psycopg/_py_transformer.py
```diff
@@ -187,18 +187,26 @@ def dump_sequence(
...
+                types[i] = none_oid
                 continue
+
             dumper = self.get_dumper(param, formats[i])
             out[i] = dumper.dump(param)
             types[i] = dumper.oid
             pqformats[i] = dumper.format
 
-        self.types = tuple(types)
+        self.types = tuple(types)  # type: ignore
```
DylanYoung: This is all non-None at this point, but MyPy isn't smart enough to know that.

dvarrazzo:
Why does the issue not affects c? Are the two implementations  different in any way that doesn't make sense?

DylanYoung:
@dvarrazzo The python version pre-populates the none-type oids into a list, whereas the c version does it lazily when it encounters the first None, like my patch does for python. 

See https://github.com/psycopg/psycopg/blob/85413678ae3577337f510a96d9e4f253a53668b0/psycopg_c/psycopg_c/_psycopg/transform.pyx#L411-L415

psycopg/psycopg/_py_transformer.py
```diff
@@ -187,18 +187,26 @@ def dump_sequence(
...
+        types = [none_oid] * nparams
         pqformats = [TEXT] * nparams
 
         for i in range(nparams):
             if (param := params[i]) is None:
+                if nones_filled:
+                    continue
+                if none_oid is None:
+                    none_oid = self._get_none_oid()
+                types[i] = none_oid
```
dvarrazzo: on't understand what the `none_filled` is for. The C algorithm doesn't use it, it just lazy-loads the oid:

```suggestion
                if self._none_oid is None:
                    self._none_oid = self._get_none_oid()
                types[i] = self._none_oid
```

and we don't seem to need the `none_oid` local var either.

psycopg/psycopg/_py_transformer.py
```diff
@@ -61,10 +61,10 @@ class Transformer(AdaptContext):
 
     _adapters: AdaptersMap
     _pgresult: PGresult | None
-    _none_oid: int
+    _none_oid: int | None
```
dvarrazzo: We can use -1 as default for _none_oid, so we don't even need to trick Mypy downstream

tests/types/test_none.py
```diff
@@ -1,5 +1,42 @@
...
+
+
+@pytest.fixture
+def str_map():
+    # Construct a minimal AdaptersMap to no none dumper bug
+    str_map = adapt.AdaptersMap()
+    str_map.adapters.register_loader(
+        INVALID_OID,
+        adapters.get_loader(INVALID_OID, pq.Format.TEXT),  # type: ignore
+    )
```
dvarrazzo: Why the `type: ignore`? If it's to handle a possible `None` return we can assert it.

```suggestion
    assert (loader := adapters.get_loader(INVALID_OID, pq.Format.TEXT))
    str_map.adapters.register_loader(INVALID_OID, loader)
```

tests/types/test_none.py
```diff
@@ -9,3 +46,11 @@ def test_quote_none(conn):
     cur = conn.cursor()
     cur.execute(sql.SQL("select {v}").format(v=sql.Literal(None)))
     assert cur.fetchone()[0] is None
+
+
+@pytest.mark.parametrize("fmt_in", PyFormat)
+def test_no_none_dumper_registered(str_conn, fmt_in):
+    with str_conn.cursor() as cur:
```
dvarrazzo: I would say that creating a fixture for this type of connection is overkilling. I would create a connection in the test taking the dns as parameter. The other details are not relevant here, for example we don't make these tests depend on the version.


```suggestion
def test_no_none_dumper_registered(conn_cls, dns, str_map, fmt_in):
    with conn_cls.connect(dns, context=str_map) as conn:
        with str_conn.cursor() as cur:
```

TBH I don't even think you need `str_map` as a fixture: it's not plenty reused and it doesn't depend on any environmental injection. You can just put this code in this test or in a local function if we want to reuse it.

dvarrazzo (changes requested):
Thank you very much!

I think that the test you propose goes in the Transformer tests, not in the None type dumping.

psycopg/psycopg/_py_transformer.py
DylanYoung: They're just small optimizations.  In the typical case, self._none_oid will already be populated and so types will already be populated with the none_oid, so this bypasses filling `types` one-by-one in the usual case.   In the C code, `self._none_oid` is (I think) a simple memory access.  In the python case, it's an attribute lookup, so using a local variable should be a bit faster.

psycopg/psycopg/_py_transformer.py
DylanYoung: Can do.  It's slightly more expensive to check, but only marginally I think.

tests/types/test_none.py
DylanYoung: will do!

tests/types/test_none.py
DylanYoung: Can do!

psycopg/psycopg/_py_transformer.py
DylanYoung: I'll take a second look at the c code though.  It's possible it could be made a bit faster by pre-filling the none_oid. 

EDIT: the c-code is fine as it initializes `types` to an uninitialized tuple of the proper length.

DylanYoung:
> Thank you very much!
> 
> I think that the test you propose goes in the Transformer tests, not in the None type dumping.

Will do!

psycopg/psycopg/_py_transformer.py
DylanYoung: I'll do this in a separate commit, so you can decide.  On my machine, the cost is about 7 nsec (from 14.1 to 21.3) to switch from an `is None` to `< 0`  check.    Not critical, but it's also a bit more pythonic to check `None`, so I'll let you decide.

tests/types/test_none.py
DylanYoung: Done.

tests/types/test_none.py
DylanYoung: Done

psycopg/psycopg/_py_transformer.py
DylanYoung: This is in commit fb4d7c8ad132ec07adf109ad5ad09542da09242a.  Happy to squash it in to the first commit if you want it.

psycopg/psycopg/_py_transformer.py
DylanYoung: I've left in these optimizations as I think they're worth-while, but let me know if you think not.  Just to clarify the savings/costs:

1. The local variable saves two attribute lookups (replacing them with locals lookups) per None encountered.
2. `nones_filled` saves a `__setitem__` call and  `if is None` (or `if <=0`) for each `None` encountered on every load except the first and costs an extra `if bool` check on the first load and an `if is None` (or `if >= 0`) on each call to `load_sequence`.  

Each of these operations is roughly the same order of magnitude (about 14 to 35 ns on my machine), attribute access being the slowest.

dvarrazzo: No, I don't think these optimizations are worthwhile. They obfuscate the code to save a few opcodes per parameter. There is very clearly a faster path, the C code, we don't need to squeeze this level of performance from the Python implementation.

psycopg/psycopg/_py_transformer.py
dvarrazzo: Yes please, let's use -1 as sentinel, not None.

DylanYoung:
@dvarrazzo  I just pushed up an alternative option in commit 30707a08b0dc33e114658f9ceaf6a1ae58e1480e.  It just sets the oid when initializing the transformer in both c and python.   Let me know if you like that better.   The only downside I can see is that it costs a bit even if the user is never planning to pass `None`.

psycopg/psycopg/_py_transformer.py
DylanYoung: Sure.  It looks like PyPy might be defunct soonish, so I suppose it's not worth optimizing for.  I don't think it's obfuscated at all, but I guess that's a matter of opinion.  I just pushed an alternative option that might be more to your liking.   Take a look and if you don't like it, I'll make these final changes.

docs/news.rst
```diff
@@ -15,6 +15,8 @@ Psycopg 3.3.5 (unreleased)
 
 - Discard prepared statements upon :sql:`ALTER *` or `DISCARD *`
   (:ticket:`#1307`).
+- Fix `~errors.ProgrammingError` when dumping non-`None` values with
```
dvarrazzo: Does this render correctly in Sphinx or does it require an escape?

```suggestion
- Fix `~errors.ProgrammingError` when dumping non-\ `None` values with
```

dvarrazzo:
Yes, I don't think I would want to look up for the None oid on Transformer constructor as there are many queries where None is not used as parameter.

So I think we can rebase these branch by squashing the FIXUP and dropping the ALT, then I can merge :slightly_smiling_face: 

I'm actually a bit confused by the ALT and now the diff is not the state I would like to merge it. I'll have a final review after your rebasing, but if things are as expected this is good to go.

docs/news.rst
DylanYoung: I'll check.

DylanYoung:
> Yes, I don't think I would want to look up for the None oid on Transformer constructor as there are many queries where None is not used as parameter.
> 
> So I think we can rebase these branch by squashing the FIXUP and dropping the ALT, then I can merge 🙂
> 
> I'm actually a bit confused by the ALT and now the diff is not the state I would like to merge it. I'll have a final review after your rebasing, but if things are as expected this is good to go.

Sounds good.  I'll get to that tonight.

docs/news.rst
DylanYoung: Looks fine to me, links to the python docs too: 
Screen Shot 2026-06-22 at 10 24 01 PM

DylanYoung:
@dvarrazzo  This one should be good to go I think.

docs/news.rst
dvarrazzo: Thank you for checking. However linking to None in the docs is noise. It is strange that you don't get a link to `ProgrammingError` but that would be noise too: I don't see a need to cross-reference either of these objects from the release notes.

```suggestion
- Fix `!ProgrammingError` when dumping non-`!None` values with
```

tests/test_adapt.py
```diff
@@ -16,6 +16,13 @@
 from psycopg.types.array import ListBinaryDumper, ListDumper
 from psycopg.types.string import StrBinaryDumper, StrDumper
 
+try:
+    from psycopg._oids import INVALID_OID
+except ImportError:
+    # allow importing on old versions of psycopg for
+    # psycopg_pool compatibility tests.
+    from psycopg.postgres import INVALID_OID  # type: ignore
+
```
dvarrazzo: If we have this complication we can just `from psycopg._oids import INVALID_OID` in the body of `make_str_map()`.

psycopg/psycopg/_py_transformer.py
```diff
@@ -187,12 +188,16 @@ def dump_sequence(
...
 
-        types = [self._get_none_oid()] * nparams
+        types = [-1] * nparams
         pqformats = [TEXT] * nparams
 
         for i in range(nparams):
             if (param := params[i]) is None:
+                if self._none_oid < 0:
+                    self._none_oid = self._get_none_oid()
+                types[i] = self._none_oid
```
dvarrazzo: It doesn't make sense that it is the loop to check for the cache. if there is a `_get_none_oid()` method then it should be the one who should check for the cache.

```suggestion
                types[i] = self._get_none_oid()
```

psycopg/psycopg/_py_transformer.py
```diff
@@ -267,17 +272,10 @@ def get_dumper(self, obj: Any, format: PyFormat) -> abc.Dumper:
 
     def _get_none_oid(self) -> int:
         try:
-            return self._none_oid
-        except AttributeError:
-            pass
-
-        try:
-            rv = self._none_oid = self._adapters.get_dumper(NoneType, PY_TEXT).oid
+            return self._adapters.get_dumper(NoneType, PY_TEXT).oid
```
dvarrazzo: Related to this other comment, I don't think this change makes sense. There is a method, it has access to the state of the instance: this method should manage the cache.

I think that the code was perfectly fine previously, it was only the placement of when it was called to cause the bug. Maybe trying and failing on AttributeError is not particularly elegant and we can replace it with the `self._none_oid = -1` but function + cache not handled by the function is just wrong.

if we want to drop the `AttributeError` thing let's just change the test on top of the function with:

```python
if self._none_oid >= 0:
    return self._none_oid
```

and the rest doesn't need any change.

dvarrazzo (changes requested):

tests/test_adapt.py
DylanYoung: Can do.

psycopg/psycopg/_py_transformer.py
DylanYoung: Are you sure?  That adds an extra function call to every None item.  I don't really think of it as a cache.  I think it is the none oid for the transformer.  It just happens to be lazily populated.

psycopg/psycopg/_py_transformer.py
DylanYoung: Ok, I can make that change, but for the record, I think it's not the best idea (see comment above).   The performance hit will be relatively isolated at least.

DylanYoung: ~~Oh I know.  I'll just make it a cached_property and that will address both concerns I think.~~

Would have been nice, but cached_property doesn't work with slots.

docs/news.rst
DylanYoung: Done

psycopg/psycopg/_py_transformer.py
DylanYoung: I've done this in latest fixup commit and given you push access to the branch.  If you want it, you can run `git rebase master --autosquash`.  If not `git reset HEAD~ --hard`.

tests/test_adapt.py
DylanYoung: Done

dvarrazzo:
Rebasing to merge

PR #1331, opened 2026-06-20
Improve Performance of wait_async

DylanYoung:
When working on #1324 I noticed a significant pessimization in `wait_async` relative to the sync `wait` functions.   This is the first of two pieces of work addressing this.  The branch also applies the same logic to `wait_conn_async` even though performance there is less critical. 

Results from the script in commit ~~69412cef1c134ecca3c38804be2a68fdf605939c~~ f647b9d6281dfb1a99fb0e31d43e5a0f81f8c1d2 (borrowed from #1324).

### Copy Out 200,000 records
`python tests/scripts/copytest.py --dsn=$PSYCOPG_TEST_DSN --nrecs=200000   --no-copy-in --copy-out  --repeat=40 --async`

#### Python 3.11 (~21% improvement)
1.  This Branch:  0.872787 sec [0.889793 +/- 0.006508 from 38 results]
2. Master:  1.099157 sec [1.114730 +/- 0.005760 from 39 results]

#### Python 3.14 (~21% improvement)
1. This Branch:  1.044676 sec [1.056462 +/- 0.006708 from 40 results]
2. Master:  1.258763 sec [1.276973 +/- 0.006159 from 39 results]

EDIT: More benchmarks

### Execute Insert 20,000 records
`python tests/scripts/copytest.py --dsn=$PSYCOPG_TEST_DSN --nrecs=20000   --no-copy-in --insert  --repeat=10 --no-executemany --async`  

#### Python 3.11 (~15%)
1.  This Branch:  3.396047 sec [3.424639 +/- 0.023496 from 10 results]
2. Master:  4.007168 sec [4.284256 +/- 0.407844 from 10 results]

#### Python 3.14 (~18%)
1. This Branch:  3.564128 sec [3.610474 +/- 0.051486 from 10 results]
2. Master:  4.318846 sec [4.513517 +/- 0.109055 from 10 results]

### Execute Many Insert 20,000 records
`python tests/scripts/copytest.py --dsn=$PSYCOPG_TEST_DSN --nrecs=20000   --no-copy-in --insert  --repeat=10  --async`

#### Python 3.11 (~28%)
1.  This Branch:  2.867469 sec [2.881319 +/- 0.016169 from 10 results]
2. Master:  3.967285 sec [3.989584 +/- 0.022576 from 10 results]

#### Python 3.14 (~14%)
1. This Branch: 3.066811 sec [3.080809 +/- 0.012988 from 10 results]
2. Master:  3.517950 sec [3.645877 +/- 0.121882 from 10 results]

### Execute Select 20,000 records
`python tests/scripts/copytest.py --dsn=$PSYCOPG_TEST_DSN --nrecs=20000   --no-copy-in --select  --repeat=10 --no-executemany --async`

#### Python 3.11 (~15%)
1.  This Branch:  2.855694 sec [2.886564 +/- 0.032648 from 10 results]
2. Master:  3.359361 sec [3.401692 +/- 0.027498 from 10 results]

#### Python 3.14 (~11%)
1. This Branch: 2.993697 sec [3.020909 +/- 0.061160 from 10 results]
2. Master:  3.373374 sec [3.491772 +/- 0.111900 from 10 results]

### Execute Many Select 20,000 records
`python tests/scripts/copytest.py --dsn=$PSYCOPG_TEST_DSN --nrecs=20000   --no-copy-in --select  --repeat=10  --async`

#### Python 3.11 (~29%)
1.  This Branch:  2.863128 sec [2.875833 +/- 0.017448 from 10 results]
2. Master:  4.044594 sec [4.078965 +/- 0.022460 from 10 results]

#### Python 3.14 (~12%)
1. This Branch: 3.051476 sec [3.063282 +/- 0.014864 from 10 results]
2. Master:  3.467364 sec [3.582415 +/- 0.132430 from 10 results]

### DIFF
--- a/psycopg/psycopg/generators.py
+++ b/psycopg/psycopg/generators.py
@@ -74,27 +74,28 @@ def _connect(conninfo: str, *, timeout: float = 0.0) -> PQGenConn[PGconn]:
 
     conn = pq.PGconn.connect_start(conninfo.encode())
     logger.debug("connection started: %s", conn)
+    if conn.status == BAD:
+        encoding = conninfo_encoding(conninfo)
+        raise e.OperationalError(
+            f"connection is bad: {conn.get_error_message(encoding)}", pgconn=conn
+        )
+    wait = WAIT_W
     while True:
-        if conn.status == BAD:
-            encoding = conninfo_encoding(conninfo)
-            raise e.OperationalError(
-                f"connection is bad: {conn.get_error_message(encoding)}", pgconn=conn
-            )
+        socket = conn.socket
+
+        while not (yield socket, wait):
+            if deadline and monotonic() > deadline:
+                raise e.ConnectionTimeout("connection timeout expired")
 
         status = conn.connect_poll()
         logger.debug("connection polled: %s", conn)
 
-        if status == POLL_READING or status == POLL_WRITING:
-            wait = WAIT_R if status == POLL_READING else WAIT_W
-            while True:
-                ready = yield conn.socket, wait
-                if deadline and monotonic() > deadline:
-                    raise e.ConnectionTimeout("connection timeout expired")
-                if ready:
-                    break
-
-        elif status == POLL_OK:
+        if status == POLL_OK:
             break
+        elif status == POLL_READING:
+            wait = WAIT_R
+        elif status == POLL_WRITING:
+            wait = WAIT_W
         elif status == POLL_FAILED:
             encoding = conninfo_encoding(conninfo)
             raise e.OperationalError(
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -15,7 +15,7 @@ import sys
 import select
 import logging
 import selectors
-from asyncio import Event, TimeoutError, get_event_loop, wait_for
+from asyncio import Event, TimeoutError, get_running_loop, wait_for
 from selectors import DefaultSelector
 
 from . import errors as e
@@ -145,7 +145,7 @@ async def wait_async(gen: PQGen[RV], fileno: int, interval: float = 0.0) -> RV:
     # Use an event to block and restart after the fd state changes.
     # Not sure this is the best implementation but it's a start.
     ev = Event()
-    loop = get_event_loop()
+    loop = get_running_loop()
     ready: int
     s: Wait
 
@@ -206,7 +206,7 @@ async def wait_conn_async(gen: PQGenConn[RV], interval: float = 0.0) -> RV:
     # Use an event to block and restart after the fd state changes.
     # Not sure this is the best implementation but it's a start.
     ev = Event()
-    loop = get_event_loop()
+    loop = get_running_loop()
     ready: Ready
     s: Wait
 
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -15,7 +15,7 @@ import sys
 import select
 import logging
 import selectors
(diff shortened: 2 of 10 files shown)


dvarrazzo:
Hello,

this seems to be a good improvement: as you could see from the comment you dropped the original design was a first implementation and didn't change significantly since 2020, because there be dragons.

However the test run has two tests hanging until CI timeout, one of which seemed clearly hanging on an async test, and both on Python 3.10. So, judging from this fist run there is something unreliable happening there. I propose we test it a bit more. I am re-running the tests to verify whether it was a fluke in the runners but I think this is not a smell to ignore.

If we cannot find the issue we could leave both the implementation and use the older one on Python 3.10 only, and leave it in until we drop support for this version.

DylanYoung:
@dvarrazzo I'll take a look.  I suspect this may be fixed by the fix cancel branch.   Can you link me to the failing CI run?  The one here is passing.

dvarrazzo:
They were 2 jobs in the first CI run, but I run them again and they passed. I'm running the whole job a few other time, once already passed. I think it was a fluke.

dvarrazzo:
Uhm... happening again [here](https://github.com/psycopg/psycopg/actions/runs/28072802434/job/83892946754?pr=1331) and [here](https://github.com/psycopg/psycopg/actions/runs/28072802434/job/83892946727?pr=1331), again with Python 3.10

DylanYoung:
> here

Dang.  I'll take a look tonight. I saw this before and i thought I fixed it.  Probably something subtle with the timing. 

I think I have an idea what the problem might be.

EDIT: I think I found the fix.  It's not really related to this work, but I'll push it up here and in a separate PR shortly.  Basically, it's the same issue as with the cancel generator, but this time affecting the connect generator.  It's revealed by this work because it makes the wait_async_conn faster (similar to how I found the cancel bug).

DylanYoung:
@dvarrazzo I think this is similar to the cancel issue, and manifests similarly (just on different tests).  I've opened #1339  and #1338 to fix it.  I've also included the commits here so you can check if it fixes the hangs you're seeing (I still haven't been able to replicate them).

dvarrazzo:
Hello @DylanYoung 

Thank you for this work, and for the patience while it sat.

Your diagnosis was right: `wait_async()` was doing considerably more work than the sync path, and your benchmarks reproduced here. The result is being prepared in #1416, with credits to you in the commit message.

The implementation there ended up different from yours. Blocking on a bare `Future` rather than an `Event` awaited through `wait_for()` is where the gain comes from: `wait_for()` wraps every single wait in a `Task` until Python 3.12, together with deferring the event loop lookup, which is what makes `copy()` faster. Keeping the file descriptor registered for the generator's lifetime, as your branch does, I measured to be worth nothing on its own, and slightly worse when paired with the `Event`.

What I did not take is the `await sleep(0)` polling. It is faster still on a busy socket, but it spins the event loop: a connection sitting idle burns a whole core, which no throughput benchmark shows. It may also be behind the 3.10 hangs nobody could reproduce, since a loop spinning at a zero timeout changes the task interleaving everywhere.

The benchmark script is part of #1416, so an alternative implementation can be measured against the current one with `--wait-async module:function`; the full cross-platform comparison lives [in the `wait-async-perf-bench` branch of my fork](https://github.com/dvarrazzo/psycopg/actions?query=branch%3Await-async-perf-bench+workflow%3A%22Wait%20benchmark%22).

Separately: #1341 is merged, and #1339 is still wanted: it is independent of this and I will review it on its own.

Closing in favour of #1416.

PR #1333, opened 2026-06-22
Fix cancel generator

### LINKED ISSUE #1335: Connection.cancel hangs (with timeout=0) or raises CancellationTimeout when it shouldn't
Noticed that `test_cancel` sometimes hung.  It's because the cancel generator wasn't actually waiting on the correct state of the socket for two reasons:

1) it wasn't actually checking the ready value sent by wait
2) it wasn't waiting for write readiness before the first poll

See https://www.postgresql.org/docs/17/libpq-cancel.html#LIBPQ-PQCANCELSTART for (2)

Fix should probably be backported to all maintained major releases.

DylanYoung:
Fixes #1335.  Noticed that `test_cancel` sometimes hung.  It's because the cancel generator wasn't actually waiting on the correct state of the socket for two reasons:

1) it wasn't actually checking the ready value sent by wait
2) it wasn't waiting for write readiness before the first poll

See https://www.postgresql.org/docs/17/libpq-cancel.html#LIBPQ-PQCANCELSTART for (2)

Should probably be backported to all maintained major releases. 

This fixes both issues.  The existing cancel test covers it, but it's possible a test that more reliably fails could be written by wrapping the cancel generator and checking the values sent yielded (though I'm not 100%). 

I haven't filed an issue yet, but I'll try to do that tomorrow.   ~~The news item is kind of sparse.  Let me know what you'd like to see there.~~

EDIT: lint CI was complaining about an unrelated file and an unrelated part of generators.pyx, so I fixed that and in the process fixed another maybe-bug.  Interestingly, my local wasn't complaining, so maybe I need to update some of the linting tools.

### DIFF
--- a/psycopg/psycopg/generators.py
+++ b/psycopg/psycopg/generators.py
@@ -112,16 +112,20 @@ def _connect(conninfo: str, *, timeout: float = 0.0) -> PQGenConn[PGconn]:
 
 def _cancel(cancel_conn: PGcancelConn, *, timeout: float = 0.0) -> PQGenConn[None]:
     deadline = monotonic() + timeout if timeout else 0.0
+    wait = WAIT_W
+
     while True:
-        if deadline and monotonic() > deadline:
-            raise e.CancellationTimeout("cancellation timeout expired")
+        socket = cancel_conn.socket
+        while not (yield socket, wait):
+            if deadline and monotonic() > deadline:
+                raise e.CancellationTimeout("cancellation timeout expired")
 
         if (status := cancel_conn.poll()) == POLL_OK:
             break
         elif status == POLL_READING:
-            yield cancel_conn.socket, WAIT_R
+            wait = WAIT_R
         elif status == POLL_WRITING:
-            yield cancel_conn.socket, WAIT_W
+            wait = WAIT_W
         elif status == POLL_FAILED:
             raise e.OperationalError(
                 f"cancellation failed: {cancel_conn.get_error_message()}"
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -19,6 +19,8 @@ Psycopg 3.3.5 (unreleased)
   no `!NoneType` dumper registered in python implementation (:ticket:`#1325`).
 - Fix `!wait_selector` wait function to not raise `!KeyError`
   (:ticket:`#1327`).
+- Fix `Connection.cancel_safe()` sometimes hanging or raising
+  `errors.CancellationTimeout` when it should have succeeded (:ticket:`#1335`).
 
 
 Current release
--- a/psycopg/psycopg/waiting.py
+++ b/psycopg/psycopg/waiting.py
@@ -292,13 +292,14 @@ def wait_select(gen: PQGen[RV], fileno: int, interval: float = 0.0) -> RV:
 
 
 if hasattr(selectors, "EpollSelector"):
-    _epoll_evmasks = {
-        WAIT_R: select.EPOLLONESHOT | select.EPOLLIN,
-        WAIT_W: select.EPOLLONESHOT | select.EPOLLOUT,
-        WAIT_RW: select.EPOLLONESHOT | select.EPOLLIN | select.EPOLLOUT,
-    }
+    _epoll_evmasks = [
+        0,  # unused
+        select.EPOLLONESHOT | select.EPOLLIN,  # WAIT_R
+        select.EPOLLONESHOT | select.EPOLLOUT,  # WAIT_W
+        select.EPOLLONESHOT | select.EPOLLIN | select.EPOLLOUT,  # WAIT_RW
+    ]
 else:
-    _epoll_evmasks = {}
+    _epoll_evmasks = []
 
 
 def wait_epoll(gen: PQGen[RV], fileno: int, interval: float = 0.0) -> RV:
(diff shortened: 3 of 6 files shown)


DylanYoung:
Something's wrong with the fix. Closing this temporarily while I sort it out.

DylanYoung:
Reopening as CI is passing now.  I was too tired when I made the fix last night ;)

DylanYoung:
Rebased on master to remove those lint fixes.  

Question, would you be amenable to adding `venv*` to .gitignore instead of just venv?  I keep multiple venvs with different interpreter/psycopg versions and it would make my tooling integration a bit easier?  Going to push that in a commit so you can merge if you don't have any changes requested.  Happy to remove it if you don't want it :)   I could move to a subdirectory structure, but then I'd have to recreate all the envs.

DylanYoung:
@dvarrazzo   I suspect you should look at this one next.  `cancel_safe` was very broken before.

dvarrazzo:
I have cherry-picked your `.venv*` changed to master. Trying to rebase this branch but I don't have permission either.

If you can enable permission to push I will merge it from here, otherwise I'll have to merge from command line.

dvarrazzo:
Do you have a reliably failing test? Tests pass on my machine and I'm running a more complete set of tests [here](https://github.com/dvarrazzo/psycopg/actions/runs/28320258292) but I haven't seen a failure yet.

psycopg/psycopg/generators.py
```diff
@@ -112,16 +112,20 @@ def _connect(conninfo: str, *, timeout: float = 0.0) -> PQGenConn[PGconn]:
 
 def _cancel(cancel_conn: PGcancelConn, *, timeout: float = 0.0) -> PQGenConn[None]:
     deadline = monotonic() + timeout if timeout else 0.0
+    wait = WAIT_W
+
     while True:
-        if deadline and monotonic() > deadline:
-            raise e.CancellationTimeout("cancellation timeout expired")
+        socket = cancel_conn.socket
+        while not (yield socket, wait):
```
dvarrazzo, line 119: Here you are meaning that the result is `Ready.NONE`, right? It would be better to make it explicit instead of relying on its value being 0.

dvarrazzo:
I am not super, super convinced that the timeout only results when completeness state of the socket is neither read nor write. Especially in the face of a connection being lost.

I also read your interpretation of the docs differently. [The docs](https://www.postgresql.org/docs/17/libpq-cancel.html#LIBPQ-PQCANCELSTART) say:

> On the first iteration, i.e., if you have yet to call `PQcancelPoll(cancelConn)`, behave as if it last returned `PGRES_POLLING_WRITING`

So the docs seem to say explicitly "assume you can write". You seem to think, in your point 2, that instead we should wait for the socket to be write-ready.

DylanYoung:
> I am not super, super convinced that the timeout only results when completeness state of the socket is neither read nor write. Especially in the face of a connection being lost.
> 
> I also read your interpretation of the docs differently. [The docs](https://www.postgresql.org/docs/17/libpq-cancel.html#LIBPQ-PQCANCELSTART) say:
> 
> > On the first iteration, i.e., if you have yet to call `PQcancelPoll(cancelConn)`, behave as if it last returned `PGRES_POLLING_WRITING`
> 
> So the docs seem to say explicitly "assume you can write". You seem to think, in your point 2, that instead we should wait for the socket to be write-ready.

The docs say if it has last returned `PGRES_POLLING_WRITING`, then we're supposed to wait for write readiness. (i.e. `PGRES_POLLING_WRITING` doesn't indicate write readiness; it tells us what we're supposed to wait for next).

I think you're correct that the TImeout can still happen in other cases.  This just fixes the case where it happens and shouldn't have.

psycopg/psycopg/generators.py
DylanYoung: I can change this, though I think this is how most of the other generators behave (maybe I'm misremembering).

DylanYoung:
> Do you have a reliably failing test? Tests pass on my machine and I'm running a more complete set of tests [here](https://github.com/dvarrazzo/psycopg/actions/runs/28320258292) but I haven't seen a failure yet.

I think a reliable failure is quite difficult to produce, as it all comes down to timing.  If you have any ideas though....  One of my branches with faster `waits` did cause it to fail much more frequently.  Let me see if I can dig that up. 

EDIT: see my latest reply

psycopg/psycopg/generators.py
DylanYoung: I'v left this for now, because this is what all the generators do (READY_NONE isn't even imported in that file), and many of the `wait` function but I'll give you rebase permissions and you can change it if you like.

DylanYoung:
> Do you have a reliably failing test? Tests pass on my machine and I'm running a more complete set of tests [here](https://github.com/dvarrazzo/psycopg/actions/runs/28320258292) but I haven't seen a failure yet.

The latest commit includes a performance enhancement that causes `tests/test_generators.py::test_cancel` to fail almost every time on my machine without this fix.  Give it a try.

dvarrazzo:
Thanks, and apologies for the delay. Same plan as #1339: I would like this in 3.3.7, not in 3.3.6, which is about to go out with the Python 3.15 support.

It needs a rebase too: #1411 changed how the cancellation deadline is handled in `_cancel()`, in both implementations, so the branch conflicts there. We should also check that the original problem still shows on current master.

In case you get to rebase this the commit "perf(py): minor improvement to wait_poll and wait_epoll" maybe belongs to its own PR. Turning the evmask dicts into lists is unrelated to the cancel generator, and I would rather review the fix by itself.

PR #1339, opened 2026-06-29
Fix connect generator

### LINKED ISSUE #1338: Connection.connect hangs (with timeout=0) or raises ConnectionTimeout when it shouldn't
Similar to #1335 

As per the [docs](https://www.postgresql.org/docs/current/libpq-connect.html):

> On the first iteration, i.e., if you have yet to call PQconnectPoll, behave as if it last returned PGRES_POLLING_WRITING.

Which means 

> Conversely, if PQconnectPoll(conn) last returned PGRES_POLLING_WRITING, wait until the socket is ready to write, then call PQconnectPoll(conn) again.

The current implementation does not wait at all before calling `PQconnectPoll`. 

Like the issue with cancel, this is very hard to detect without very fast and responsive wait functions, and even then can be sporadic.

DylanYoung:
Fixes #1338 

Also some minor performance boost:

1.  only get the socket after calling poll (not everytime the wait func sends a reply)
2.  don't double check the value of poll
3.  don't check the connection status in the loop (poll handles this)

for 3, see the [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PQCONNECTSTARTPARAMS):

> At any time during connection, the status of the connection can be checked by calling [PQstatus](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSTATUS). If this call returns CONNECTION_BAD, then the connection procedure has failed; if the call returns CONNECTION_OK, then the connection is ready. Both of these states are equally detectable from the return value of PQconnectPoll, described above.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -19,6 +19,8 @@ Psycopg 3.3.5 (unreleased)
   no `!NoneType` dumper registered in python implementation (:ticket:`#1325`).
 - Fix `!wait_selector` wait function to not raise `!KeyError`
   (:ticket:`#1327`).
+- Fix `Connection.connect()` sometimes hanging or raising
+  `errors.CancellationTimeout` when it should have succeeded (:ticket:`#1338`).
 
 
 Current release
--- a/psycopg/psycopg/generators.py
+++ b/psycopg/psycopg/generators.py
@@ -74,27 +74,28 @@ def _connect(conninfo: str, *, timeout: float = 0.0) -> PQGenConn[PGconn]:
 
     conn = pq.PGconn.connect_start(conninfo.encode())
     logger.debug("connection started: %s", conn)
+    if conn.status == BAD:
+        encoding = conninfo_encoding(conninfo)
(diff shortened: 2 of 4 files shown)


dvarrazzo:
Thanks for this, and sorry it sat so long. I am about to release 3.3.6, which is mostly about Python 3.15 support, so I would rather not touch the connection path in it. I would like to finalize this PR for 3.3.7.

It needs a rebase first. master has moved since June: #1411 reworked where the connection deadline is computed and checked, and your branch conflicts in `_connect()` in both the Python and the C implementation. I can take care of the rebase after 3.3.6 release, in case you moved on with your life and don't have time for this.

One thing that would be useful is if you could confirm that the problem in #1338 still reproduces as you described. The code around it has changed, even though I don't believe the fix is in.

I have cherry-picked "test: add reasonable connect_timeout to most tests" into #1416, with your authorship, because it is worth having on its own, so that commit can be dropped from here.

PR #1340, opened 2026-06-29
fix: return none if there is no value to yield from.

giscardff:
I did create the following issue, this patch is to fix it
https://github.com/psycopg/psycopg/issues/1337

### DIFF
--- a/psycopg/psycopg/_connection_base.py
+++ b/psycopg/psycopg/_connection_base.py
@@ -477,8 +477,9 @@ class BaseConnection(Generic[Row]):
         else:
             self.pgconn.send_query_params(command, None, result_format=result_format)
 
-        result: PGresult = (yield from generators.execute(self.pgconn))[-1]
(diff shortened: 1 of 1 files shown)


dvarrazzo:
We had more conversation about it in #1337.

After calling `_send_query()`, with whatever input, given a working FE-BE implementation, we expect one or more results. Even if we send no command, we would get a single result of type EMPTY_QUERY.

The only case in which we get no result is that something else is using the connection concurrently, no other cause that I can see. This I think it is happening in #1337 because of a broken runtime in which the `asyncio.Lock` is not doing its job.

I will change the code a bit here to avoid to get blamed for an `IndexError` and change the code to raise an `InteralError` if the results list is empty. I don't want to assume all right and return None if there is no result in the list because the condition is totally unexpected on a working system.

dvarrazzo:
Replacing with #1344

PR #1342, opened 2026-07-01
test: make async test_leak less flakey

DylanYoung:
This one failed very occasionally, but with my new work on #1331 (which I'll get up tomorrow), it fails in CI quite a lot, especially on windows. 

This fixes the flakiness.

### DIFF
--- a/tests/test_cursor_async.py
+++ b/tests/test_cursor_async.py
@@ -9,6 +9,9 @@ from psycopg import errors as e
 from psycopg import pq, rows
 from psycopg.adapt import PyFormat
 
+if True:  # ASYNC
+    from asyncio import sleep
+
 
 async def test_default_cursor(aconn):
     cur = aconn.cursor()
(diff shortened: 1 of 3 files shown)


dvarrazzo:
Thanks for spotting this, but I don't think we need it any more.

The `await sleep(0)` compensates for the wait function keeping the file descriptor registered for the whole life of the generator: the callback handle outlives `work()`, so `gc.count()` sees objects which are about to be freed. The implementation being merged in #1416 registers and removes the descriptor around each wait, so that handle is gone before the count is taken.

With that in place the three `test_leak` tests pass consistently here: 156 tests, three runs in a row, no failures.

I'm also wary of the change on its own merits. A `sleep(0)` between `work()` and `gc.count()` gives the loop a chance to release objects inside the very tests whose job is to notice when something is not released.

If these start failing on Windows once #1416 is in, let's reopen it: that would be a real signal, rather than a workaround for something we can't see.

PR #1353, opened 2026-07-08
Unwrap nested Binary() in the binary dumpers (#1352)

### LINKED ISSUE #1352: `psycopg.Binary` wrapping another `Binary` instance raises unhandled low-level `TypeError`
`BinaryTextDumper.dump()` unwraps one level of `Binary` wrapping via `obj.obj`. If `obj.obj` is itself a `Binary` instance, the unwrapped value is passed directly into the underlying C escaping routine, which fails with a low-level `TypeError` referencing an internal psycopg class rather than any documented error.

**Minimal reproduction:**
```python
from psycopg.dbapi20 import Binary, BinaryTextDumper

inner = Binary(b"hello")
outer = Binary(inner)
dumper = BinaryTextDumper(Binary)
dumper.dump(outer)
```

**Expected:** a clear, documented error, or correct handling of the nested case.
**Actual:**
```
TypeError: bytes or buffer expected, got <class 'psycopg.dbapi20.Binary'>
```

apoorvdarshan:
## Summary

Fixes #1352. `BinaryTextDumper.dump()` / `BinaryBinaryDumper.dump()` unwrap only one level of `Binary()` (via `obj.obj`). When `obj.obj` is itself a `Binary`, the inner wrapper reaches the underlying escaping routine:

```python
from psycopg.dbapi20 import Binary, BinaryTextDumper
BinaryTextDumper(Binary).dump(Binary(Binary(b"hello")))
# TypeError: cannot convert 'Binary' object to bytes
```

The binary-format dumper doesn't raise but silently returns the inner `Binary` object instead of the bytes.

## Fix

Unwrap **every** level of `Binary()` nesting before dumping the buffer:

```python
while isinstance(obj, Binary):
    obj = obj.obj
return super().dump(obj)
```

## Verification

- Added `tests/test_adapt.py::test_dump_nested_binary` (parametrized over both dumpers): `dump(Binary(Binary(Binary(b"hello")))) == dump(b"hello")`.
- Verified directly: on `master`, `BinaryTextDumper` raises `TypeError` and `BinaryBinaryDumper` returns a wrong `Binary` value on nested input; with this change both return `b"hello"`, identical to the unwrapped and single-wrapped cases.

*(The full suite needs a live PostgreSQL; the fix and its regression were verified via direct dumper calls, which require no connection.)*

---

### DIFF
--- a/psycopg/psycopg/dbapi20.py
+++ b/psycopg/psycopg/dbapi20.py
@@ -79,18 +79,18 @@ class Binary:
 
 class BinaryBinaryDumper(BytesBinaryDumper):
     def dump(self, obj: Buffer | Binary) -> Buffer | None:
-        if isinstance(obj, Binary):
-            return super().dump(obj.obj)
-        else:
-            return super().dump(obj)
+        # Unwrap any level of Binary() nesting before dumping the buffer.
+        while isinstance(obj, Binary):
+            obj = obj.obj
+        return super().dump(obj)
 
 
 class BinaryTextDumper(BytesDumper):
     def dump(self, obj: Buffer | Binary) -> Buffer | None:
-        if isinstance(obj, Binary):
-            return super().dump(obj.obj)
-        else:
-            return super().dump(obj)
+        # Unwrap any level of Binary() nesting before dumping the buffer.
+        while isinstance(obj, Binary):
+            obj = obj.obj
+        return super().dump(obj)
 
 
 def Date(year: int, month: int, day: int) -> dt.date:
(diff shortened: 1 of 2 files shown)


DylanYoung:
@apoorvdarshan Does it make  more sense to do this in the `Binary` constructor?

i.e. 

```
class Binary:
    def __init__(self, obj: Any):
        self.obj = obj
```

becomes:

```
class Binary:
    def __new__(cls, obj: Any):
        if isinstance(obj, Binary):
            return obj
        self = object.__new__(cls)
        self.obj = obj
        return self
```

dvarrazzo:
The `Binary` object is only provided for DBAPI compliance and I don't expect it to be used by any meaningful code. Not interested in fixing it.

PR #1354, opened 2026-07-08
namedtuple_row: rename duplicate column names instead of crashing (#1348)

### LINKED ISSUE #1348: `psycopg.rows.namedtuple_row` raises unhandled `ValueError` on duplicate column names
`namedtuple_row` raises `ValueError: Encountered duplicate field name` when the query result has two columns sharing the same name — a common and legitimate SQL pattern (e.g. `SELECT f.*, b.*` in a join, where both sides have a same-named column). `dict_row`, `class_row`, and `kwargs_row` all degrade gracefully on the identical input (silently keeping the last value), but `namedtuple_row` does not. Confirmed against a live PostgreSQL server (not just a mocked cursor).

**Root cause:**
```python
@functools.lru_cache(512)
def _make_nt(enc: str, *names: bytes) -> type[NamedTuple]:
    snames = tuple(_as_python_identifier(n.decode(enc)) for n in names)
    return namedtuple("Row", snames)  # rename=False (default)
```
Python's `namedtuple()` raises `ValueError` on duplicate field names unless `rename=True` is passed. `_make_nt` does not pass `rename=True`.

**Minimum reproduction:**
```python
import psycopg
from psycopg.rows import namedtuple_row

conn = psycopg.connect("host=localhost dbname=postgres user=postgres password=postgres", autocommit=True)
conn.execute("CREATE TABLE foo (id int PRIMARY KEY)")
conn.execute("CREATE TABLE bar (id int PRIMARY KEY, foo_id int REFERENCES foo(id))")
conn.execute("INSERT INTO foo (id) VALUES (1)")
conn.execute("INSERT INTO bar (id, foo_id) VALUES (10, 1)")

cur = conn.cursor(row_factory=namedtuple_row)
cur.execute("SELECT f.id, b.id FROM foo f JOIN bar b ON f.id = b.foo_id")
cur.fetchone()
```

**Expected behavior:** a `Row` namedtuple instance, degrading gracefully (analogous to `dict_row`'s behavior on the same input).
**Actual behavior:**
```
  File ".../psycopg/rows.py", line 136, in namedtuple_row
    nt = _make_nt(cursor._encoding, *(res.fname(i) for i in range(nfields)))
  File ".../psycopg/rows.py", line 145, in _make_nt
    return namedtuple("Row", snames)
  File "/usr/lib/python3.12/collections/__init__.py", line 412, in namedtuple
    raise ValueError(f'Encountered duplicate field name: {name!r}')
ValueError: Encountered duplicate field name: 'id'
```

**System info:**
```
python: 3.12
PostgreSQL server: 16 (Docker)
OS: Linux
```

apoorvdarshan:
## Summary

Fixes #1348. `namedtuple_row` raises `ValueError: Encountered duplicate field name` when a result has two columns sharing a name — a common, legitimate pattern such as `SELECT a.*, b.*` on a join where both sides have a same-named column:

```python
# with a cursor whose result has duplicate column names:
cur.row_factory = namedtuple_row
cur.execute("SELECT 1 AS id, 2 AS id")
cur.fetchone()   # ValueError: Encountered duplicate field name: 'id'
```

`dict_row`, `class_row`, and `kwargs_row` all handle the identical input gracefully; only `namedtuple_row` crashes.

## Fix

Pass `rename=True` to `namedtuple()` in `_make_nt`, so duplicate (and otherwise invalid) names are renamed to positional `_N` names — consistent with the docstring's note that field names are taken "with some mangling to deal with invalid names". All columns remain accessible (by position for the renamed ones).

`_make_nt("utf-8", b"id", b"name", b"id")._fields == ("id", "name", "_2")`.

## Verification

- Added `tests/test_rows.py::test_make_nt_duplicate_columns` (no DB needed; exercises `_make_nt` directly). It fails on `master` (`ValueError`) and passes here.
- Verified directly: on `master`, `_make_nt` with a duplicate name raises `ValueError`; with this change it returns fields `('id', 'name', '_2')`, and normal / invalid-identifier inputs are unchanged.
- `black` / `flake8` clean; `mypy` clean on `rows.py`.

*(The full suite needs a live PostgreSQL; the fix and its regression were verified via direct `_make_nt` calls, which require no connection.)*

---

### DIFF
--- a/psycopg/psycopg/rows.py
+++ b/psycopg/psycopg/rows.py
@@ -142,7 +142,9 @@ def namedtuple_row(cursor: BaseCursor[Any, Any]) -> RowMaker[NamedTuple]:
 @functools.lru_cache(512)
 def _make_nt(enc: str, *names: bytes) -> type[NamedTuple]:
     snames = tuple(_as_python_identifier(n.decode(enc)) for n in names)
-    return namedtuple("Row", snames)  # type: ignore[return-value]
+    # ``rename`` renames duplicate columns (e.g. ``SELECT a.*, b.*`` on a join)
+    # to positional ``_N`` names instead of raising ``ValueError``.
+    return namedtuple("Row", snames, rename=True)  # type: ignore[return-value]
 
 
 def class_row(cls: type[T]) -> BaseRowFactory[T]:
--- a/tests/test_rows.py
+++ b/tests/test_rows.py
@@ -68,6 +68,15 @@ def test_namedtuple_row(conn):
     assert r3.f_eur == 5
 
 
+def test_make_nt_duplicate_columns():
+    # gh-1348: duplicate column names (e.g. `SELECT a.*, b.*` on a join) must
+    # not raise; the duplicates are renamed to positional _N names.
+    rows._make_nt.cache_clear()
+    nt = rows._make_nt("utf-8", b"id", b"name", b"id")
+    assert nt._fields == ("id", "name", "_2")
+    assert tuple(nt("a", "b", "c")) == ("a", "b", "c")
+
+
 def test_class_row(conn):
     cur = conn.cursor(row_factory=rows.class_row(Person))
     cur.execute("select 'John' as first, 'Doe' as last")


DylanYoung:
Are you sure `dict_row` and `class_row` are handling this gracefully?  On a quick glance, it looks like they're just silently overwriting the value with the last seen value?  I think a hard error would probably be better.   I think there's a design question as to how this should behave, and then all row factories should be made consistent with the design.

apoorvdarshan:
Fair point, and I agree there's a real design question here.

To be precise about the current behavior: `dict_row`/`kwargs_row` do silently keep the last value (dict semantics), `class_row` keeps the last via `setattr`, and only `namedtuple_row` hard-crashes — so the factories aren't consistent today, and this PR was aimed narrowly at the crash reported in #1348.

`rename=True` is the standard `collections.namedtuple` mechanism for exactly this situation (duplicate/invalid field names → positional `_N`), so it keeps all columns accessible by position without losing data, which is why I went that way.

That said, I don't have a strong opinion on the destination — "silently keep last" is arguably the worse behavior, and a consistent **hard error across all row factories** is a perfectly reasonable design. That's a broader, intentional change though, so I'd defer to @dvarrazzo on the direction. If a hard error (or some other unified behavior) is preferred, I'm happy to rework this PR — or split it — to make all the factories consistent rather than just unbreaking `namedtuple_row`.

DylanYoung:
@apoorvdarshan  I appreciate your point that with your current change each factory is now following the semantics of its type, which is also a fine design choice.  Maybe it should be documented if that ends up being the decision.

psycopg/psycopg/rows.py
```diff
@@ -142,7 +142,9 @@ def namedtuple_row(cursor: BaseCursor[Any, Any]) -> RowMaker[NamedTuple]:
 @functools.lru_cache(512)
 def _make_nt(enc: str, *names: bytes) -> type[NamedTuple]:
     snames = tuple(_as_python_identifier(n.decode(enc)) for n in names)
-    return namedtuple("Row", snames)  # type: ignore[return-value]
+    # ``rename`` renames duplicate columns (e.g. ``SELECT a.*, b.*`` on a join)
+    # to positional ``_N`` names instead of raising ``ValueError``.
```
dvarrazzo, line 146: Raising an exception is better.

dvarrazzo:
Thank you, but I'm not interested in a silent renaming. If an user wants to obtain a namedtuple from "select 1 as foo, 2 as foo" they deserve an exception.

PR #1355, opened 2026-07-08
Detect recursive lists in the array dumpers' dump path (#1350)

### LINKED ISSUE #1350: `psycopg.types.array.ListDumper.dump` raises unhandled `RecursionError` on a self-referential list, when the dumper's `oid` is already known
`psycopg.types.array.ListDumper.dump` raises unhandled `RecursionError` on a self-referential list, when the dumper's `oid` is already known

**Description:**
`ListDumper` has cycle-detection logic elsewhere in `BaseListDumper` (via `_flatiter`, which raises `DataError("cannot dump a recursive list")`), but this path is only reached through `get_key()`, which is skipped whenever the dumper's `oid` is already set — such as when the dumper is reused via `Transformer.set_dumper_types()` (e.g. in a `COPY` block with `copy.set_types()`), bypassing the `get_key()`/`upgrade()` step where cycle detection lives. For such dumpers, calling `dump()` directly on a self-referential list does not go through any cycle check.

**Minimal reproduction:**
```python
from psycopg.types.array import ListDumper

dumper = ListDumper(list)
dumper.oid = 1007

a = []
a.append(a)
dumper.dump(a)
```

**Expected:** `psycopg.DataError: cannot dump a recursive list` (the same error raised elsewhere in the module for a self-referential list).
**Actual:** `RecursionError: maximum recursion depth exceeded`

apoorvdarshan:
## Summary

Fixes #1350. A self-referential list raises `RecursionError` (instead of the module's `DataError: cannot dump a recursive list`) when the list dumper's `oid` is already set:

```python
from psycopg.types.array import ListDumper
d = ListDumper(list)
d.oid = 1007          # e.g. reused via COPY set_types()
a = []; a.append(a)
d.dump(a)             # RecursionError
```

## Root cause

Cycle detection lives in `BaseListDumper._flatiter`, which is only reached through `get_key()`/`upgrade()` (the dumper-preparation step). When the dumper already has an `oid` — e.g. reused via `Transformer.set_dumper_types()` in a `COPY` block — that step is skipped, so `dump()` recurses directly over the list with no cycle check.

## Fix

Track visited list `id()`s in the recursive helpers of both list dumpers — `ListDumper.dump()`'s `dump_list()` and `ListBinaryDumper.dump()`'s `calc_dims()` — raising the same `DataError("cannot dump a recursive list")` as `_flatiter`. This keeps the pre-set-`oid` path consistent with the normal path (which already rejects recursive/shared lists).

## Verification

- Added `tests/types/test_array.py::test_dump_recursive_list_with_oid` (both `PyFormat`s, no DB needed).
- Verified directly: on `master`, both the text and binary dumpers raise `RecursionError`; with this change both raise `DataError`. Normal nested lists (`[[1],[2]]`, `[1,2,3]`, `[]`) still dump correctly.
- `black` / `flake8` / `mypy` clean on the changed files.

*(The full suite needs a live PostgreSQL; the fix and its regression were verified via direct dumper calls, which require no connection.)*

---

### DIFF
--- a/psycopg/psycopg/types/array.py
+++ b/psycopg/psycopg/types/array.py
@@ -158,7 +158,11 @@ class ListDumper(BaseListDumper):
         tokens: list[Buffer] = []
         needs_quotes = _get_needs_quotes_regexp(self.delimiter).search
 
-        def dump_list(obj: list[Any]) -> None:
+        def dump_list(obj: list[Any], seen: set[int]) -> None:
+            if id(obj) in seen:
+                raise e.DataError("cannot dump a recursive list")
+            seen.add(id(obj))
+
             if not obj:
                 tokens.append(b"{}")
                 return
@@ -166,7 +170,7 @@ class ListDumper(BaseListDumper):
             tokens.append(b"{")
             for item in obj:
                 if isinstance(item, list):
-                    dump_list(item)
+                    dump_list(item, seen)
                 elif item is not None:
                     if (ad := self._dump_item(item)) is None:
                         tokens.append(b"NULL")
@@ -181,7 +185,7 @@ class ListDumper(BaseListDumper):
 
             tokens[-1] = b"}"
 
-        dump_list(obj)
+        dump_list(obj, set())
 
         return b"".join(tokens)
 
@@ -249,14 +253,17 @@ class ListBinaryDumper(BaseListDumper):
         dims: list[int] = []
         hasnull = 0
 
-        def calc_dims(L: list[Any]) -> None:
+        def calc_dims(L: list[Any], seen: set[int]) -> None:
             if isinstance(L, self.cls):
+                if id(L) in seen:
+                    raise e.DataError("cannot dump a recursive list")
+                seen.add(id(L))
                 if not L:
                     raise e.DataError("lists cannot contain empty lists")
                 dims.append(len(L))
-                calc_dims(L[0])
+                calc_dims(L[0], seen)
 
-        calc_dims(obj)
+        calc_dims(obj, set())
 
         def dump_list(L: list[Any], dim: int) -> None:
             nonlocal hasnull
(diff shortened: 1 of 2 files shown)


DylanYoung:
Any idea what the cost is of this change?

apoorvdarshan:
Good question — I benchmarked it (compiled `psycopg_c`, dumping `int8[]` arrays through the dumpers directly, min of several runs).

The added work is a per-list `id()` membership test + set insert, so it scales with the **number of list objects**, not the number of elements:

| array | text `master` → this PR | binary |
|---|---|---|
| realistic 2D `100×100` (10k ints, 101 lists) | 5.15 → 5.09 ms (within noise) | ~unchanged |
| flat 2000 ints (1 list) | ~unchanged | ~unchanged |
| pathological `2000×[1]` (2001 one-element lists) | 1.33 → 1.56 ms (**+~17%**) | ~unchanged |

So:

- **Binary dumper: no measurable cost** in any shape — its check lives in `calc_dims`, which only walks the `L[0]` chain (`O(depth)`, ~2 lists for a 2-D array), not every sub-list.
- **Text dumper: no measurable cost on realistic arrays.** The only measurable hit is a degenerate array made almost entirely of tiny sub-lists, where the per-list bookkeeping dominates the (trivial) per-element serialization. Even there it's a constant factor, not a complexity change.

It's also the same kind of tracking the existing `_flatiter` cycle check already does on the normal (non-preset-`oid`) path, so it stays consistent with that.

If the worst case is a concern I'm happy to gate it (e.g. only start tracking once a nested list is actually encountered), but it didn't seem worth complicating a safety check for.

dvarrazzo:
Not interested in fixing it: too narrow as use case and an exception is raised anyways, no big deal that it is not `DataError`.

PR #1364, opened 2026-07-12
Fix client encoding aliases

### LINKED ISSUE #1363: psycopg fails when PostgreSQL reports client_encoding as UNICODE
### Summary

psycopg 3.3.4 raises `NotSupportedError: codec not available in Python: 'UNICODE'` during connection initialization.

### Environment

* psycopg: 3.3.4
* PostgreSQL server: 12.3
* Python: 3.11.9
* SQLAlchemy: 2.0.51

### Reproducer

```python
import psycopg

conn = psycopg.connect(
    host="...",
    dbname="...",
    user="...",
    password="...",
)

print(conn.info.encoding)
```

### Actual behavior

Connection fails with:

```text
psycopg.NotSupportedError: codec not available in Python: 'UNICODE'
```

### Investigation

The server reports:

```sql
SHOW client_encoding;
```

Result:

```text
UTF8
```

However:

```python
conn.pgconn.parameter_status(b"client_encoding")
```

returns:

```text
b'UNICODE'
```

The failure happens because `psycopg._encodings.py_codecs` does not contain a mapping for:

```python
b"UNICODE"
```

Adding the alias below makes the connection work:

```python
import psycopg._encodings

psycopg._encodings.py_codecs[b"UNICODE"] = "utf-8"
```

After applying the alias, `conn.info.encoding` works correctly and the connection is usable.

### Question

Should `UNICODE` be treated as an alias for UTF8 in psycopg's encoding map?

DylanYoung:
Fix #1363 

- Add the missing aliases from the [PostgreSQL charset table](https://www.postgresql.org/docs/current/multibyte.html#CHARSET-TABLE)
- update one test to fail previously 
- ~~add a new test to check the aliases are registered correctly~~

Probably needs a news item?

### DIFF
--- a/psycopg/psycopg/_encodings.py
+++ b/psycopg/psycopg/_encodings.py
@@ -22,61 +22,65 @@ OK = ConnStatus.OK
 
 
 _py_codecs = {
-    "BIG5": "big5",
-    "EUC_CN": "gb2312",
-    "EUC_JIS_2004": "euc_jis_2004",
-    "EUC_JP": "euc_jp",
-    "EUC_KR": "euc_kr",
-    # "EUC_TW": not available in Python
-    "GB18030": "gb18030",
-    "GBK": "gbk",
-    "ISO_8859_5": "iso8859-5",
-    "ISO_8859_6": "iso8859-6",
-    "ISO_8859_7": "iso8859-7",
-    "ISO_8859_8": "iso8859-8",
-    "JOHAB": "johab",
-    "KOI8R": "koi8-r",
-    "KOI8U": "koi8-u",
-    "LATIN1": "iso8859-1",
-    "LATIN10": "iso8859-16",
-    "LATIN2": "iso8859-2",
-    "LATIN3": "iso8859-3",
-    "LATIN4": "iso8859-4",
-    "LATIN5": "iso8859-9",
-    "LATIN6": "iso8859-10",
-    "LATIN7": "iso8859-13",
-    "LATIN8": "iso8859-14",
-    "LATIN9": "iso8859-15",
-    # "MULE_INTERNAL": not available in Python
-    "SHIFT_JIS_2004": "shift_jis_2004",
-    "SJIS": "shift_jis",
+    ("BIG5", "WIN950", "Windows950"): "big5",
+    ("EUC_CN",): "gb2312",
+    ("EUC_JIS_2004",): "euc_jis_2004",
+    ("EUC_JP",): "euc_jp",
+    ("EUC_KR",): "euc_kr",
+    # ("EUC_TW",): not available in Python
+    ("GB18030",): "gb18030",
+    ("GBK", "WIN936", "Windows936"): "gbk",
+    ("ISO_8859_5",): "iso8859-5",
+    ("ISO_8859_6",): "iso8859-6",
+    ("ISO_8859_7",): "iso8859-7",
+    ("ISO_8859_8",): "iso8859-8",
+    ("JOHAB",): "johab",
+    ("KOI8R", "KOI8"): "koi8-r",
+    ("KOI8U",): "koi8-u",
+    ("LATIN1", "ISO88591"): "iso8859-1",
+    ("LATIN10", "ISO885916"): "iso8859-16",
+    ("LATIN2", "ISO88592"): "iso8859-2",
+    ("LATIN3", "ISO88593"): "iso8859-3",
+    ("LATIN4", "ISO88594"): "iso8859-4",
+    ("LATIN5", "ISO88599"): "iso8859-9",
+    ("LATIN6", "ISO885910"): "iso8859-10",
+    ("LATIN7", "ISO885913"): "iso8859-13",
+    ("LATIN8", "ISO885914"): "iso8859-14",
+    ("LATIN9", "ISO885915"): "iso8859-15",
+    # ("MULE_INTERNAL",): not available in Python
+    ("SHIFT_JIS_2004",): "shift_jis_2004",
+    ("SJIS", "Mskanji", "ShiftJIS", "WIN932", "Windows932"): "shift_jis",
     # this actually means no encoding, see PostgreSQL docs
     # it is special-cased by the text loader.
-    "SQL_ASCII": "ascii",
-    "UHC": "cp949",
-    "UTF8": "utf-8",
-    "WIN1250": "cp1250",
-    "WIN1251": "cp1251",
-    "WIN1252": "cp1252",
-    "WIN1253": "cp1253",
-    "WIN1254": "cp1254",
-    "WIN1255": "cp1255",
-    "WIN1256": "cp1256",
-    "WIN1257": "cp1257",
-    "WIN1258": "cp1258",
-    "WIN866": "cp866",
-    "WIN874": "cp874",
+    ("SQL_ASCII",): "ascii",
+    ("UHC", "WIN949", "Windows949"): "cp949",
+    ("UTF8", "Unicode"): "utf-8",
+    ("WIN1250",): "cp1250",
+    ("WIN1251", "WIN"): "cp1251",
+    ("WIN1252",): "cp1252",
+    ("WIN1253",): "cp1253",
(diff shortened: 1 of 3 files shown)


psycopg/psycopg/_encodings.py
```diff
@@ -67,9 +67,33 @@
     "WIN866": "cp866",
     "WIN874": "cp874",
 }
+_pg_aliases = {
```
dvarrazzo: This map looks kinda backwards: it maps pg name -> list of aliases and then it must be double-iterated, and twice resolved: alias -> pg name -> py name.

Wouldn't it be just simpler to have an alias -> py name, like `_py_codecs`?

```python
_py_aliases = {
    "WIN950", "big5",
    "Windows950": "big5",
    "KOI8": "koi8-r",
    ...
}
```

and join the two maps together in the Pg->Py direction, whilst using the `_py_codecs` map only for the Py->Pg direction?

psycopg/psycopg/_encodings.py
```diff
@@ -67,9 +67,33 @@
...
+    "WIN866": ["ALT"],
+    "WIN1251": ["WIN"],
+    "WIN1258": ["ABC", "TCVN", "TCVN5712", "VSCII"],
+}
 
 py_codecs: dict[bytes, str] = {}
 py_codecs.update((k.encode(), v) for k, v in _py_codecs.items())
+for orig, aliases in _pg_aliases.items():
+    _py_codec = _py_codecs[orig]
+    py_codecs.update((alias.encode(), _py_codec) for alias in aliases)
```
dvarrazzo: Instead of this the above might be be changed to the following maybe (untested)?

```python
py_codecs = {k.upper().encode(): v for k, v in {**_py_codecs, **_py_aliases}.items()}
```

`upper()` would normalize aliases that are shown in lowercase in the aliases column. Creating an empty dict seems unneeded.

tests/test_encodings.py
```diff
@@ -38,6 +38,18 @@ def test_pg2py(pyenc, pgenc):
     assert encodings.pg2pyenc(pgenc.encode()) == pyenc
 
 
+@pytest.mark.parametrize(
+    "pgenc, aliases",
+    list(encodings._pg_aliases.items()),
+)
+def test_pg_aliases(pgenc, aliases):
```
dvarrazzo: I don't think this test is useful: it just tests a map against itself. Let's just test test that connection using an alias works as expected.

tests/test_encodings.py
```diff
@@ -49,6 +61,8 @@ def test_pg2py_missing(pgenc):
     [
         ("", "utf-8"),
         ("user=foo, dbname=bar", "utf-8"),
+        ("user=foo, dbname=bar, client_encoding=Unicode", "utf-8"),
```
dvarrazzo: Let's test that the aliases with lowercase work. Also note that the connection string is actually invalid: the `,` are spurious. The test will parse it as "user=foo,". No big deal but let's fix it.

```suggestion
        ("user=foo dbname=bar client_encoding=Unicode", "utf-8"),
        ("user=foo dbname=bar client_encoding=UNICODE", "utf-8"),
```

dvarrazzo (changes requested):
So weird: I never noticed the aliases column, but it seems by git history in the postgres repo that it had always been there...

psycopg/psycopg/_encodings.py
DylanYoung: We could do that.  I thought this might be easier to maintain as it matches the documentation more closely.   Happy to change it over though.

DylanYoung: It's also a little bit more compact this way, but it's your call!

tests/test_encodings.py
DylanYoung: Can remove.  What it does is verify that the map is implemented correctly and used correctly to update py_codecs (and not contaminate _py_codecs), but it certainly isn't strictly necessary.

tests/test_encodings.py
DylanYoung: will do.

psycopg/psycopg/_encodings.py
DylanYoung: Seems reasonable to me.  See my comments above on why you may prefer the other approach.  But happy to make the change.

psycopg/psycopg/_encodings.py
DylanYoung: @dvarrazzo I just pushed up another option as well.  It's slightly more complicated, but it keeps the current structure for easy cross-reference with the docs and initializes `py_codecs` in one pass.

tests/test_encodings.py
DylanYoung: This is removed.

tests/test_encodings.py
DylanYoung: this is done, but I did it with mskanji instead of unicode.  unicode isn't a great test case because the result is the same as the default.

tests/pool/test_pool_async.py
DylanYoung, line 1: @dvarrazzo This was yet another linting error, caused by MyPy 2.3.0 I think.  I'm not sure this is the best solution.  Let me know if you want something else.

tests/pool/test_pool.py
```diff
@@ -529,14 +529,7 @@ def test_reconnect_failure(proxy, async_cb):
...
 
-    if async_cb:
-
-        def failed(pool):
-            assert pool.name == "this-one"
-            nonlocal t1
-            t1 = time()
-
-    else:
+    if not async_cb:
```
dvarrazzo: This doesn't belong to this MR.

psycopg/psycopg/_encodings.py
dvarrazzo: Your other option is unreadable, sorry, but totally no.

dvarrazzo: If we wanted to maintain an easy visual mapping from the docs table to the encodings we could have a `dict[tuple[str, ...], str]` mapping the base name and eventual aliases to a Python name:

```python
_py_codecs {
    ("BIG5", "WIN950", "Windows950"): "big5",
    ("EUC_CN",): "gb2312",
    ...
}
```

To create the backwards map we would only map back the first item: `{v: k[0] for k, v in _py_codecs.items()}`.

tests/test_encodings.py
dvarrazzo: Let's keep "unicode" because it's been the subject of some bug reports in the past, but please do add a test with mskanji if you think it's comprehensive - I agree.

tests/pool/test_pool.py
DylanYoung: No you're right.  I can move it to a new MR.

psycopg/psycopg/_encodings.py
DylanYoung: That seems reasonable.  I'll do that.

DylanYoung: This is done.

tests/pool/test_pool.py
DylanYoung: I've moved this to a new MR.  I do think it's worth accepting little fixes like this in existing MRs.  It saves a lot of overhead for contributors over time, and allows CI to pass.  As long as it isn't holding up a merge of course.   On my teams, we typically put all style/linting fixes at the top of a branch and then they can be merged separately or together easily.

psycopg/psycopg/_encodings.py
```diff
@@ -22,61 +22,65 @@
...
+    ("WIN1258", "ABC", "TCVN", "TCVN5712", "VSCII"): "cp1258",
+    ("WIN866", "ALT"): "cp866",
+    ("WIN874",): "cp874",
 }
 
-py_codecs: dict[bytes, str] = {}
-py_codecs.update((k.encode(), v) for k, v in _py_codecs.items())
+py_codecs = dict(
+    (k.upper().encode(), v) for aliases, v in _py_codecs.items() for k in aliases
+)
```
dvarrazzo: This can be a dict comprehension. Fixing it myself: my review took too long.

PR #1365, opened 2026-07-12
fix: reject unterminated text copy rows

### LINKED ISSUE #1361: _parse_row_text silently drops the last byte of a field when the data is missing its trailing newline (pure-Python implementation)
### Description

`psycopg._copy_base._parse_row_text` (pure-Python implementation) assumes
its input always ends with a `\n` row terminator (as real COPY TEXT
protocol data always does) and unconditionally strips the last byte of
the last field (`fields[-1] = fields[-1][:-1]`) to remove it. If the data
is missing that terminator, this silently discards the actual last byte
of real data instead of raising any error.

### Reproduction (no PostgreSQL server required)

    from psycopg._copy_base import _parse_row_text

    tx = Transformer()

    print(_parse_row_text(b"hello", tx))
    # -> ('hell',)   -- silently lost the trailing 'o', no error

### Expected behavior

A documented `psycopg.DataError` when text COPY data does not end with
the expected `\n` terminator.

### Actual behavior

Returns a value with its last byte silently dropped, no exception
raised.

### Environment

oiahoon:
## Summary

- reject text COPY rows without the required newline terminator in the pure-Python parser
- raise the same `DataError` and message already used by the C implementation
- add a database-independent regression test and an unreleased news entry

The direct C parser requires result-backed field metadata, so constructing a standalone Transformer only with `set_loader_types()` fails before it can exercise row parsing. With normal COPY result metadata, the C implementation already reports `bad copy data: field delimiter not found` for this malformed input.

Fixes #1361.

## Verification

- focused parser regression: 1 passed
- Python implementation COPY suite against PostgreSQL 16 UTF-8: 225 passed
- C implementation COPY suite against PostgreSQL 16 UTF-8: 224 passed
- all changed-file pre-commit hooks
- complete async-to-sync regeneration check
- psycopg sdist and wheel build
- `git diff --check`

### DIFF
--- a/psycopg/psycopg/_copy_base.py
+++ b/psycopg/psycopg/_copy_base.py
@@ -350,6 +350,8 @@ def _format_row_binary(row: Sequence[Any], tx: Transformer, out: bytearray) -> N
 def _parse_row_text(data: Buffer, tx: Transformer) -> tuple[Any, ...]:
     if not isinstance(data, bytes):
         data = bytes(data)
+    if not data.endswith(b"\n"):
+        raise e.DataError("bad copy data: field delimiter not found")
     fields = data.split(b"\t")
     fields[-1] = fields[-1][:-1]  # drop \n
     row = [None if f == b"\\N" else _load_re.sub(_load_sub, f) for f in fields]


dvarrazzo:
Thank you. I've rebased the MR to fix the conflict in the news file.

dvarrazzo:
Actually: just noticed that this MR adds a new test file. This is not necessary: I will move your new test into an existing test suite.

dvarrazzo:
No, I've checked better the test and, as much as the fix is correct, the test is not. It tests only the Python path, not the C path, which has a different behaviour and fails with a different error (`psycopg.ProgrammingError: cannot load sequence of 0 items: 1 loaders registered`).

These functions are internal ones, and the functions calling them always create a valid input for that function, therefore this is not an issue that gets exposed to the external interface.

I will include the change proposed, but we don't need a news entry because it's not a public issue and we don't need a new test, we can just verify that there is no regression.

PR #1367, opened 2026-07-14
dev: fix mypy linting issue

DylanYoung:
- caused by upgrade to mypy 2.3.0

### DIFF
--- a/tests/pool/test_pool.py
+++ b/tests/pool/test_pool.py
@@ -529,14 +529,7 @@ def test_reconnect_failure(proxy, async_cb):
 
     t1 = None
 
-    if async_cb:
-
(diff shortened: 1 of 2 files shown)


DylanYoung:
The functions could also be given different names and passed based on the async_cb fixture, but this seemed simpler and doesn't have any runtime overhead.

dvarrazzo:
I actually prefer a version with separate names. Talking about runtime overhead of a single if is out of place in a test function marked slow, taking several seconds to run. Fixed in 376415ad3bcf02081e1f2c8f5679729fe8d82cb6

DylanYoung:
I prefer the version with multiple names too :)

PR #1370, opened 2026-07-15
fix(copy): handle truncated binary row headers

### LINKED ISSUE #1359: _parse_row_binary raises a raw struct.error instead of DataError on truncated binary COPY data (pure-Python implementation)
### Description

`BinaryFormatter.parse_row` already validates malformed binary COPY input,
raising `psycopg.DataError` for a bad/missing signature. But a truncated
buffer — one too short for the 2-byte field-count header, or for a field's
4-byte length prefix — escapes as a raw `struct.error` from
`struct.unpack_from` instead, in the pure-Python `_parse_row_binary`
(used when no `psycopg[c]`/`psycopg[binary]` acceleration is installed).

Same class of malformed input, different failure mode.

### Reproduction

    import struct
    from psycopg._copy_base import _parse_row_binary

    tx = Transformer()

    # A single truncated byte -- not even enough to read the 2-byte
    # field-count header.
    _parse_row_binary(b"\x00", tx)

### Expected behavior

A documented `psycopg.DataError`, consistent with how
`BinaryFormatter.parse_row` already handles a bad/missing binary
signature.

### Actual behavior

    struct.error: unpack_from requires a buffer of at least 2 bytes for
    unpacking 2 bytes at offset 0 (actual buffer size is 1)

The same happens for any cut inside the first field's 4-byte length
prefix (e.g. a buffer of 2-5 bytes), always with a `struct.error` from
`struct.unpack_from`, never a `DataError`.

### Environment

Sanjays2402:
`_parse_row_binary()` let `struct.error` escape when binary COPY data ended inside the field-count or field-length header. Convert those unpacking failures to `DataError` and cover both truncation points.

Tested with `PSYCOPG_IMPL=python pytest -q tests/test_copy_base.py`.

Closes #1359


tests/test_copy_base.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+from psycopg import errors as e
+from psycopg import pq
+from psycopg.adapt import Transformer
+from psycopg._copy_base import _parse_row_binary
```
DylanYoung: You should import `parse_row_binary` so the test covers both the python and the c implementation.  You'll also need to resolve the psycopg-pool CI failures.  Those tests run against an old version of psycopg where `_copy_base` doesn't seem to have existed yet (you could just run the import in a `try` block, catch the `ImportError`, and pass).

DylanYoung: After taking a quick look, I can see the c-code actually behaves differently, so you'll need to test and iron out the differences.  Off-the-top, it looks like the c-code needs two additional checks for when ptr + X goes past bufend.

Sanjays2402: You were right on both counts, and the C side was the more interesting half.

`parse_row_binary()` in `copy.pyx` memcpy's the 2-byte field count and each 4-byte field length straight out of the buffer with no comparison against `bufend`, so a header truncated at either point reads past the end. The `ptr + length > bufend` check further down only guards the field *payload*, not the headers. So I added the two missing checks there.

On the Python side I kept the `struct.error` conversion but also added the `length exceeding data` check the C version already had, so the two now behave the same on a field length that overruns the buffer. Both raise `DataError("bad copy data: truncated row header")` for a truncated header, matching the existing `bad copy data:` prefix instead of the wording I'd invented.

The test file now parametrizes over both implementations, covers empty input, truncation at the field-count and field-length boundaries, an overrunning length, and a well-formed row. The `_copy_base` import is in a `try`/`except ImportError` so the psycopg-pool job against older psycopg skips rather than failing at collection.

Verified locally against a build of `psycopg_c` from this branch: 12 passed. Reverting just the `copy.pyx` checks makes the `empty-c` and `field-count-c` cases fail, so the C coverage is real.

dvarrazzo:
Hello,

apologies for the slow reply to this PR. As a consequence, because I don't want to waste your time, I've made some changes myself:

1) I've dropped the C/Python parametrization. The test suite either runs the whole C code or the whole Python code, according to the PSYCOPG_IMPL it is run on. Therefore explicit parametrization was not needed.

2) Some parsing tests were already added to the test suite in #1383. However those tests needlessly run twice, in sync and async code. Because a test module just for C copy parsing is being added here, I've moved those tests here too.

3) I have differentiated the error message according to where it happens, so now there are two checks in two points, rather than a try/except, in C as it was proposed in Python.

PR #1373, opened 2026-07-17
Fix missing f-string prefix in two DataError messages

### LINKED ISSUE #1372: DataError messages leak the literal `{...}` placeholder (missing f-string prefix)
Two `DataError` messages in the pure-Python implementation are plain string literals that contain `{...}` placeholders but are missing the `f` prefix, so the literal placeholder text is shown to the user instead of the offending value.

**1. `psycopg/types/datetime.py:692`** — `_get_timestamp_load_error()`:

```python
return DataError("timestamp too small (before year 1): {s!r}")
```

The two sibling branches immediately below it are correctly f-strings (`f"timestamp too large (after year 10K): {s!r}"`, `f"can't parse timestamp {s!r}: ..."`), which is what makes the missing prefix stand out.

**2. `psycopg/types/json.py:280`** — `JsonbBinaryLoader.load()`:

```python
raise DataError("unknown jsonb binary format: {data[0]}")
```

### Reproduction (pure-Python, no server needed)

```python
from psycopg.types.datetime import _get_timestamp_load_error
from psycopg.types.json import JsonbBinaryLoader
from psycopg import _oids

print(_get_timestamp_load_error(None, b"0001-01-01 00:00:00 BC"))
# -> timestamp too small (before year 1): {s!r}

try:
    JsonbBinaryLoader(_oids.JSONB_OID, None).load(b"\x02{}")
except Exception as e:
    print(e)
# -> unknown jsonb binary format: {data[0]}
```

### Expected

```
timestamp too small (before year 1): '0001-01-01 00:00:00 BC'
unknown jsonb binary format: 2
```

A repo-wide grep confirms these are the only two occurrences of the pattern (every other `{...}`-in-a-plain-string case is an intentional `sql.SQL(...).format(...)` template). I'll open a PR adding the missing `f` prefix to both.

winklemad:
Two `DataError` messages were built from plain string literals containing `{...}` placeholders but missing the `f` prefix, so the literal placeholder text was shown to the user instead of the offending value.

- `psycopg/types/datetime.py` — `_get_timestamp_load_error()`:
  `"timestamp too small (before year 1): {s!r}"`
- `psycopg/types/json.py` — `JsonbBinaryLoader.load()`:
  `"unknown jsonb binary format: {data[0]}"`

The two sibling branches right next to the datetime message are already f-strings (`f"timestamp too large (after year 10K): {s!r}"`, `f"can't parse timestamp {s!r}: ..."`), which is what makes the omission stand out.

### Before

```
timestamp too small (before year 1): {s!r}
unknown jsonb binary format: {data[0]}
```

### After

```
timestamp too small (before year 1): '0001-01-01 00:00:00 BC'
unknown jsonb binary format: 2
```

A repo-wide grep confirms these are the only two occurrences of the pattern; every other `{...}`-in-a-plain-string case is an intentional `sql.SQL(...).format(...)` template.

Fixes #1372

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -19,6 +19,9 @@ Psycopg 3.3.5 (unreleased)
   no `!NoneType` dumper registered in python implementation (:ticket:`#1325`).
 - Fix `!wait_selector` wait function to not raise `!KeyError`
   (:ticket:`#1327`).
+- Fix `!DataError` messages leaking the literal ``{...}`` placeholder instead
+  of the offending value when loading a pre-year-1 :sql:`timestamp` or a
+  malformed binary :sql:`jsonb` value (:ticket:`#1372`).
 
 
 Current release
--- a/psycopg/psycopg/types/datetime.py
+++ b/psycopg/psycopg/types/datetime.py
@@ -689,7 +689,7 @@ def _get_timestamp_load_error(
             return len(s.split()[-1]) > 4  # year is last token
 
     if s == "-infinity" or s.endswith("BC"):
-        return DataError("timestamp too small (before year 1): {s!r}")
+        return DataError(f"timestamp too small (before year 1): {s!r}")
     elif s == "infinity" or is_overflow(s):
         return DataError(f"timestamp too large (after year 10K): {s!r}")
     else:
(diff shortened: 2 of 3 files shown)


dvarrazzo:
Merged, thank you!

PR #1375, opened 2026-07-25
fix: raise ProgrammingError for infinite connect_timeout

### LINKED ISSUE #1346: `psycopg.conninfo.timeout_from_conninfo` raises unhandled `OverflowError` instead of `ProgrammingError` at `connect_timeout='inf'`
`timeout_from_conninfo({"connect_timeout": "inf"})` raises `OverflowError` instead of `ProgrammingError`. Every other invalid `connect_timeout` value (non-numeric strings, and even `"nan"`) correctly raises `ProgrammingError`, since `int(float(value))` failing with `ValueError` is caught by the existing `except` clause. This value is plausibly reachable through the `PGCONNECT_TIMEOUT` environment variable or a `connect_timeout=inf` conninfo string, since neither path performs type or numeric validation before reaching this function; the exact call path from `psycopg.connect()` down to this function was not traced end-to-end.

**Root cause:**
`float("inf")` succeeds and returns `inf`, but `int(float("inf"))` raises `OverflowError`, a distinct exception class not covered by `timeout_from_conninfo`'s `except ValueError:` clause. The structurally identical `"nan"` case is handled correctly because `int(float("nan"))` raises `ValueError`, not `OverflowError`.

**Minimum reproduction:**
```python
from psycopg.conninfo import timeout_from_conninfo

timeout_from_conninfo({"connect_timeout": "nan"})   # comportamento correto
timeout_from_conninfo({"connect_timeout": "inf"})   # BUG
```

**Expected behavior:** `psycopg.ProgrammingError: bad value for connect_timeout: 'inf'` (mesmo padrão do caso `nan`, linha 140)
**Actual behavior:**
```
File ".../psycopg/conninfo.py", line 138, in timeout_from_conninfo
    timeout = int(float(value))
OverflowError: cannot convert float infinity to integer
```

**System info:**
```
python: 3.12
OS: Linux
```

Sanjays2402:
Closes #1346

`timeout_from_conninfo()` wraps `int(float(value))` in a `try` that only catches `ValueError`, but `float("inf")` succeeds and `int(inf)` raises `OverflowError`, so `connect_timeout=inf` escaped as a raw `OverflowError` instead of the `ProgrammingError` every other invalid value (including the structurally identical `"nan"`) already raises. Catching `OverflowError` alongside `ValueError` routes it through the existing error path.

Regression test parametrizes the non-numeric, `nan` and infinite spellings; the three infinite cases fail with `OverflowError` without the source change and pass with it.

### DIFF
--- a/psycopg/psycopg/conninfo.py
+++ b/psycopg/psycopg/conninfo.py
@@ -136,7 +136,7 @@ def timeout_from_conninfo(params: ConnDict) -> int:
         value = _DEFAULT_CONNECT_TIMEOUT
     try:
         timeout = int(float(value))
-    except (ValueError, OverflowError):
+    except Exception:
         raise e.ProgrammingError(f"bad value for connect_timeout: {value!r}") from None
 
     if timeout <= 0:
--- a/psycopg/psycopg/conninfo.py
+++ b/psycopg/psycopg/conninfo.py
@@ -136,7 +136,7 @@ def timeout_from_conninfo(params: ConnDict) -> int:
         value = _DEFAULT_CONNECT_TIMEOUT
     try:
         timeout = int(float(value))
-    except ValueError:
+    except (ValueError, OverflowError):
         raise e.ProgrammingError(f"bad value for connect_timeout: {value!r}") from None
 
     if timeout <= 0:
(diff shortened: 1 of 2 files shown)


dvarrazzo:
Thank you for this fix, but I'd rather extend it to catch any exception, not strictly the two you proposed.

PR #1376, opened 2026-07-25
Fix IndexError in _as_python_identifier on empty string

### LINKED ISSUE #1347: `psycopg._encodings._as_python_identifier` raises `IndexError` on empty string input
`_as_python_identifier("")` raises `IndexError` instead of producing a valid Python identifier, which is the function's own stated purpose ("Reduce a string to a valid Python identifier"). This is a defect in the function's own contract, reproducible directly and unconditionally. Whether PostgreSQL itself can ever surface an empty column name to `namedtuple_row` (the function's only current caller) was not confirmed one way or the other; this report is scoped to the function-level defect regardless of that open question.

**Root cause:**
```python
def _as_python_identifier(s: str, prefix: str = "f") -> str:
    if not s.isidentifier():        # "".isidentifier() is False -> enters block
        if s[0] in "1234567890":    # s[0] on an empty string raises IndexError
            ...
```
`"".isidentifier()` is `False`, so execution enters the branch that accesses `s[0]` without first checking that `s` is non-empty.

**Minimal reproduction (confirmed, real execution):**
```python
from psycopg._encodings import _as_python_identifier

_as_python_identifier("")
```

**Expected:** a valid, non-empty Python identifier, consistent with how the rest of the function already handles other invalid inputs (names starting with a digit, names with special characters).
**Actual:**
```
  File "test.py", line 3, in <module>
    _as_python_identifier("")
  File ".../psycopg/_encodings.py", line 141, in _as_python_identifier
    if s[0] in "1234567890":
       ~^^^
IndexError: string index out of range
```

**System info:**
```
python: 3.12
OS: Linux
```

Sanjays2402:
Closes #1347

`_as_python_identifier("")` raised `IndexError`: `"".isidentifier()` is `False`, so the branch reading `s[0]` was entered on an empty string. An empty name now falls back to the `prefix`, matching how the function already handles names starting with a digit or an underscore.

The new parametrized case fails with the reported `IndexError` on `master` and passes with the fix.

### DIFF
--- a/tests/test_encodings.py
+++ b/tests/test_encodings.py
@@ -56,3 +56,19 @@ def test_pg2py_missing(pgenc):
 )
 def test_conninfo_encoding(conninfo, pyenc):
     assert encodings.conninfo_encoding(conninfo) == pyenc
+
+
+@pytest.mark.parametrize(
+    "s, want",
+    [
+        ("", "f"),
+        ("foo", "foo"),
+        ("1foo", "f1foo"),
+        ("_foo", "f_foo"),
+        ("foo bar", "foo_bar"),
+    ],
+)
+def test_as_python_identifier(s, want):
+    got = encodings._as_python_identifier(s)
+    assert got == want
+    assert got.isidentifier()
--- a/psycopg/psycopg/_encodings.py
+++ b/psycopg/psycopg/_encodings.py
@@ -138,7 +138,9 @@ def _as_python_identifier(s: str, prefix: str = "f") -> str:
     the first letter is an '_'.
     """
     if not s.isidentifier():
-        if s[0] in "1234567890":
+        if not s:
+            s = prefix
+        elif s[0] in "1234567890":
             s = prefix + s
         if not s.isidentifier():
             s = _re_clean.sub("_", s)


tests/test_encodings.py
```diff
@@ -56,3 +56,19 @@ def test_pg2py_missing(pgenc):
...
+    [
+        ("", "f"),
+        ("foo", "foo"),
+        ("1foo", "f1foo"),
+        ("_foo", "f_foo"),
+        ("foo bar", "foo_bar"),
+    ],
+)
+def test_as_python_identifier(s, want):
+    got = encodings._as_python_identifier(s)
```
dvarrazzo, line 72: We don't test internal functions. Either you produce the issue with a public facing interface, or the bug doesn't exist.

Sanjays2402: You're right, and I couldn't produce it from the public interface. I went looking on PostgreSQL 18 and the server rejects a zero-length identifier everywhere the name could come from: `select 1 as ""`, `create type t as ("" int)`, `alter table ... rename column ... to ""`, and a `values (1) v("")` alias all fail with `zero-length delimited identifier`. Unnamed expression columns come back as `?column?`, not empty. Both call sites (`rows._make_nt` via `res.fname()`, and `composite._nt_from_info` via the catalog) are fed by that same identifier space, so there is no query I can write that reaches `_as_python_identifier("")`.

The only way I got an empty `attname` in front of a cursor was hand-editing `pg_attribute`, which isn't an interface anyone should be defended against.

So the crash is only reachable by calling the private helper directly, which is exactly what you're saying doesn't count. Happy to close this — say the word, or if you'd rather keep the two-line guard purely as defensiveness against a future caller I can drop the test and leave just the `if not s:` branch.

PR #1377, opened 2026-07-25
fix(rows): raise DataError on duplicate column names in namedtuple_row

### LINKED ISSUE #1348: `psycopg.rows.namedtuple_row` raises unhandled `ValueError` on duplicate column names

Sanjays2402:
Closes #1348

`namedtuple_row` let a raw `ValueError` from `collections.namedtuple` escape when a result set had two columns with the same name (e.g. `SELECT f.id, b.id` across a join). `_make_nt` now catches it and raises `DataError` instead, per your comment on the issue that converting the error is acceptable but allowing the duplicate is not.

Added `test_make_nt_duplicate_names` (fails with `ValueError` without the change, passes with it) and a news entry.

### DIFF
--- a/tests/test_rows.py
+++ b/tests/test_rows.py
@@ -68,9 +68,10 @@ def test_namedtuple_row(conn):
     assert r3.f_eur == 5
 
 
-def test_make_nt_duplicate_names():
+def test_namedtuple_row_duplicate_names(conn):
+    cur = conn.cursor(row_factory=rows.namedtuple_row)
     with pytest.raises(psycopg.DataError):
-        rows._make_nt("utf-8", b"id", b"id")
+        cur.execute("select 1 as id, 2 as id").fetchall()
 
 
 def test_class_row(conn):
--- a/tests/test_rows.py
+++ b/tests/test_rows.py
@@ -68,6 +68,11 @@ def test_namedtuple_row(conn):
     assert r3.f_eur == 5
 
 
+def test_make_nt_duplicate_names():
+    with pytest.raises(psycopg.DataError):
+        rows._make_nt("utf-8", b"id", b"id")
+
+
 def test_class_row(conn):
     cur = conn.cursor(row_factory=rows.class_row(Person))
     cur.execute("select 'John' as first, 'Doe' as last")
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -22,6 +22,8 @@ Psycopg 3.3.5 (unreleased)
 - Fix `!DataError` messages leaking the literal ``{...}`` placeholder instead
   of the offending value when loading a pre-year-1 :sql:`timestamp` or a
   malformed binary :sql:`jsonb` value (:ticket:`#1372`).
+- Raise `!DataError` instead of `!ValueError` when `~psycopg.rows.namedtuple_row`
+  receives duplicate column names (:ticket:`#1348`).
 - Fix building C extension with Cython 3.3.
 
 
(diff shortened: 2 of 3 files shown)


tests/test_rows.py
```diff
@@ -68,6 +68,11 @@ def test_namedtuple_row(conn):
     assert r3.f_eur == 5
 
 
+def test_make_nt_duplicate_names():
+    with pytest.raises(psycopg.DataError):
+        rows._make_nt("utf-8", b"id", b"id")
```
dvarrazzo: Don't test internal functions. Test the function from the public interface.

Sanjays2402: Fixed in 73e6453. The test now goes through the cursor:

```python
def test_namedtuple_row_duplicate_names(conn):
    cur = conn.cursor(row_factory=rows.namedtuple_row)
    with pytest.raises(psycopg.DataError):
        cur.execute("select 1 as id, 2 as id").fetchall()
```

Confirmed it is not vacuous: reverting the `rows.py` change makes it fail with the raw `ValueError: Encountered duplicate field name: 'id'` from `collections.namedtuple`, which is the escape this PR closes.

`tests/test_rows.py` is 19 passed against PostgreSQL 16.14. I did not run the rest of the suite.

Issue #1351, opened 2026-07-08
`psycopg.Timestamp()` and `psycopg.TimestampFromTicks()` produce timezone-inconsistent `datetime` objects


BERTopic:
`Timestamp()` and `TimestampFromTicks()` are the two DB-API 2.0 constructors for building a timestamp value. `Timestamp()` returns a naive `datetime` (`tzinfo=None`), while `TimestampFromTicks()` returns a timezone-aware `datetime` (`tzinfo=UTC`). Since psycopg documents that naive `datetime` objects are dumped as PostgreSQL `timestamp` and aware ones as `timestamptz`, these two sibling constructors cause otherwise-equivalent values to be adapted to two different PostgreSQL column types depending on which constructor built them.

Note: ticket #1058 already aligned `DateFromTicks`, `TimeFromTicks`, and `TimestampFromTicks` to consistently return UTC-aware values among themselves. This report concerns a different axis: the mismatch between the ticks-based constructors (aware) and their non-ticks sibling constructors `Date()`/`Time()`/`Timestamp()` (naive), which #1058 did not address.

**Minimal reproduction:**
```python
from psycopg.dbapi20 import Timestamp, TimestampFromTicks
import time

t1 = Timestamp(2024, 1, 1, 12, 0, 0)
t2 = TimestampFromTicks(time.time())
print(t1.tzinfo)
print(t2.tzinfo)
```

**Expected:** both constructors return `datetime` objects with the same timezone-awareness.
**Actual:**
```
None
UTC
```

**System info:** `psycopg: 3.3.4, impl=binary, libpq=180000`, `python: 3.12`, `OS: Linux`

dvarrazzo:
The dbapi objects are legacy objects, they predate the datetime objects that were introduced in Python 2.3 or 2.4. New code doesn't use it, and I don't think it is wise to change the object returned for old code.

Therefore I think fixing this issue will be of no use for anybody, and create problem in the unlikely case that some code is still using it. I prefer to leave the things as they are.

PR #1378, opened 2026-07-26
fix: Fix  to return a timezone-aware datetime

sureshdsk:
Fixes : 
- `Timestamp` to return a timezone-aware datetime (in UTC), consistent with `TimestampFromTicks`, so that both constructors are adapted to the same ( issue #1351  )

@dvarrazzo please take a look at this and review.


dvarrazzo:
I prefer not merge this PR for the reasons explained in #1351

PR #1379, opened 2026-07-27
fix: handle empty string in _as_python_identifier (#1347)

### LINKED ISSUE #1347: `psycopg._encodings._as_python_identifier` raises `IndexError` on empty string input

ahmed5145:
## Summary
- `_as_python_identifier("")` raised `IndexError` because it indexed `s[0]` after `"".isidentifier()` returned `False`.
- Guard the digit-prefix path with `s and ...`, and treat an empty result like other non-field-safe names so it becomes `prefix` (default `"f"`), consistent with existing sanitization (`"_" -> "f_"`, digit-leading names get `prefix`).
- Add regression tests in `tests/test_encodings.py`.

Empty input maps to `prefix` (not `f_`) because there is no character content to preserve; `"_"` still becomes `f_` by keeping the underscore.

Fixes #1347

## Test plan
- [x] `python -P -c "import psycopg; from psycopg._encodings import _as_python_identifier; r=_as_python_identifier(''); print(repr(r), r.isidentifier())"` → `'f' True`
- [x] encoding unit tests green (24 passed)
- [ ] CI on the PR

### DIFF
--- a/psycopg/psycopg/_encodings.py
+++ b/psycopg/psycopg/_encodings.py
@@ -135,15 +135,15 @@ def _as_python_identifier(s: str, prefix: str = "f") -> str:
     Reduce a string to a valid Python identifier.
 
     Replace all non-valid chars with '_' and prefix the value with `!prefix` if
-    the first letter is an '_'.
+    the string is empty or the first letter is an '_'.
     """
     if not s.isidentifier():
-        if s[0] in "1234567890":
+        if s and s[0] in "1234567890":
             s = prefix + s
         if not s.isidentifier():
             s = _re_clean.sub("_", s)
     # namedtuple fields cannot start with underscore. So...
-    if s[0] == "_":
+    if not s or s[0] == "_":
         s = prefix + s
     return s
 
--- a/tests/test_encodings.py
+++ b/tests/test_encodings.py
@@ -56,3 +56,26 @@ def test_pg2py_missing(pgenc):
 )
 def test_conninfo_encoding(conninfo, pyenc):
     assert encodings.conninfo_encoding(conninfo) == pyenc
+
+
+@pytest.mark.parametrize(
+    "s, expected",
+    [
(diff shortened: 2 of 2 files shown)


ahmed5145:
The `macos-14 (c, 3.11)` failure in `tests/test_generators.py::test_cancel` looks unrelated to this change (only `_as_python_identifier` / encoding tests). It appears to be a flaky cancel/libpq connection drop. I don’t have permission to re-run the job

ahmed5145:
Closed as per maintainer request

PR #1381, opened 2026-08-02
fix: allow binary dump of nested empty arrays

### LINKED ISSUE #1349: `psycopg.types.array` binary array dumper raises `DataError` for nested empty lists that the text dumper accepts, when the dumper's OID is already known
When a `ListDumper`/`ListBinaryDumper` instance already has its `oid` set (as is the case for dumpers of a registered array type, or when reused via `Transformer.set_dumper_types()`, e.g. in a `COPY ... (FORMAT BINARY)` block with `copy.set_types()`), `.dump()` is called directly without going through the `get_key()`/`upgrade()` type-detection step. In this scenario, `ListDumper.dump()` accepts a list containing nested empty lists (e.g. `[[], []]`) without error, while `ListBinaryDumper.dump()` raises `DataError` for the identical Python value.

**Minimal reproduction:**
```python
from psycopg.types.array import ListDumper, ListBinaryDumper

ListDumper(list).dump([[], []])
ListBinaryDumper(list).dump([[], []])
```

**Expected:** consistent behavior between text and binary dumpers for the same Python value.
**Actual:**
```
text dump: b'{{},{}}'
binary dump raised: DataError: lists cannot contain empty lists
```

deepakganesh78:
Fixes #1349.

## Reproduction
`ListDumper(list).dump([[], []])` returns `b'{{},{}}'`, but `ListBinaryDumper(list).dump([[], []])` raised `DataError: lists cannot contain empty lists` before this change. This is reproducible without a PostgreSQL server.

## Root cause
The binary array dumper calculated dimensions by descending into the first element of each nested list, but treated an empty nested list as invalid even when all sibling nested lists had the same empty shape. The binary array loader also assumed every dimension length was non-zero when re-nesting decoded items.

## Fix
Allow dimension calculation to stop at empty nested lists, producing a zero-length inner dimension for the binary array representation. Also teach the binary array loader to reconstruct nested empty arrays from zero-element dimensions, and add a regression test for the generated binary payload and round trip through the loader.

## Compatibility
Existing invalid ragged arrays, such as arrays mixing non-empty and empty nested lists, still raise `DataError` because their dimensions remain inconsistent. Top-level empty arrays are unchanged.

## Validation
- `python -m pytest ..\tests\types\test_array.py ..\tests\test_adapt.py -q` from the `psycopg` package directory with `PSYCOPG_IMPL=binary`: 48 passed, 173 skipped.
- `python -m black --check --diff --quiet psycopg\psycopg\types\array.py tests\types\test_array.py`
- `python -m flake8 psycopg\psycopg\types\array.py tests\types\test_array.py`
- `python -m isort --check --diff psycopg\psycopg\types\array.py tests\types\test_array.py`
- Skips are expected in this environment because no PostgreSQL server/`--test-dsn` is available; I did not run the server-dependent integration suite.
- Verified the new regression test fails on the previous implementation with `DataError: lists cannot contain empty lists`.

### DIFF
--- a/tests/types/test_array.py
+++ b/tests/types/test_array.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import gc
+import struct
 from math import prod
 from typing import Any
 from decimal import Decimal
@@ -10,10 +11,11 @@ import pytest
 import psycopg
 import psycopg.types.numeric
 from psycopg import pq, sql
+from psycopg._oids import TEXT_OID
 from psycopg.adapt import Dumper, PyFormat, Transformer
 from psycopg.types import TypeInfo
 from psycopg.postgres import types as builtins
-from psycopg.types.array import register_array
+from psycopg.types.array import ListBinaryDumper, _load_binary, register_array
 
 from ..test_adapt import StrNoneBinaryDumper, StrNoneDumper
 
@@ -126,6 +128,19 @@ def test_bad_binary_array(input):
     tx = Transformer()
     with pytest.raises(psycopg.DataError):
         tx.get_dumper(input, PyFormat.BINARY).dump(input)
+
+
+def test_dump_binary_nested_empty_array():
+    got = ListBinaryDumper(list).dump([[], []])
+    want = b"".join(
+        [
+            struct.pack("!III", 2, 0, TEXT_OID),
+            struct.pack("!II", 2, 1),
+            struct.pack("!II", 0, 1),
+        ]
+    )
+    assert got == want
+    assert _load_binary(got, Transformer()) == [[], []]
 
 
 @pytest.mark.crdb_skip("nested array")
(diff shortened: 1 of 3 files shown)


tests/types/test_array.py
```diff
@@ -128,6 +130,19 @@ def test_bad_binary_array(input):
...
+        [
+            struct.pack("!III", 2, 0, TEXT_OID),
+            struct.pack("!II", 2, 1),
+            struct.pack("!II", 0, 1),
+        ]
+    )
+    assert got == want
+    assert _load_binary(got, Transformer()) == [[], []]
+
+
```
dvarrazzo, line 145: This is not a suitable test. Please demonstrate that you can dump this object to a database column and retrieve it back.

dvarrazzo (changes requested):

PR #1383, opened 2026-08-16
fix: raise DataError when binary copy field length exceeds data

### LINKED ISSUE #1360: _parse_row_binary silently truncates a field's value instead of raising, when the declared length exceeds the available data (pure-Python implementation)
### Description

When a binary COPY field declares a length (in its 4-byte length prefix)
greater than the number of bytes actually remaining in the buffer,
`psycopg._copy_base._parse_row_binary` (pure-Python implementation) does
not validate this and simply slices out whatever bytes are available
(Python slicing doesn't raise for an out-of-range stop index). This
silently returns a shorter, corrupted value instead of raising any error
at all.

### Reproduction

    import struct
    from psycopg._copy_base import _parse_row_binary

    tx = Transformer()

    # Field declares length 10, but only 2 bytes ("ab") are actually present.
    data = struct.pack("!h", 1) + struct.pack("!i", 10) + b"ab"
    print(_parse_row_binary(data, tx))
    # -> ('ab',)   -- no error, silently truncated

### Expected behavior

A documented `psycopg.DataError` when a field's declared length does not
match the data actually available in the buffer.

### Actual behavior

Returns `('ab',)` with no exception whatsoever — the caller has no way
to know the row was truncated/corrupted.

### Environment

chjnett:
Closes #1360.

## Problem

The pure-Python `_parse_row_binary` slices `data[pos : pos + length]` without
checking that the declared `length` fits in the remaining buffer. Python slicing
doesn't raise for an out-of-range stop, so a field declaring more bytes than are
present is silently returned truncated instead of raising.

The C implementation already validates this and raises
`DataError("bad copy data: length exceeding data")`, so the two implementations
diverge on the same malformed input.

## Change

Add the same bounds check to `_parse_row_binary`, raising
`DataError("bad copy data: length exceeding data")` when `pos + length > len(data)`.

## Reproduction

```python
from psycopg._copy_base import _parse_row_binary

tx = Transformer()
# One field declaring length 10, but only two bytes ("ab") present.
data = bytes.fromhex("00010000000a6162")
_parse_row_binary(data, tx)
# before: returned ('ab',) silently
# after:  psycopg.errors.DataError: bad copy data: length exceeding data
```

## Testing

Added `test_copy_binary_parse_length_exceeding_data` (in both the sync and async
test modules) covering the malformed case. I ran this unit test against the
pure-Python implementation and it passes; the full suite requires a running
PostgreSQL, which I did not have available, so the DB-backed copy tests were not
run locally.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -24,6 +24,8 @@ Psycopg 3.3.5 (unreleased)
   malformed binary :sql:`jsonb` value (:ticket:`#1372`).
 - Raise `!DataError` instead of `!ValueError` when `~psycopg.rows.namedtuple_row`
   receives duplicate column names (:ticket:`#1348`).
+- Raise `!DataError` on inconsistent copy data length in pure Python,
+  consistently with the C implementation (:ticket:`#1360`).
 - Fix building C extension with Cython 3.3.
 
 
--- a/psycopg/psycopg/_copy_base.py
+++ b/psycopg/psycopg/_copy_base.py
@@ -366,6 +366,8 @@ def _parse_row_binary(data: Buffer, tx: Transformer) -> tuple[Any, ...]:
         length = _unpack_int4(data, pos)[0]
         pos += 4
         if length >= 0:
+            if pos + length > len(data):
+                raise e.DataError("bad copy data: length exceeding data")
             row.append(data[pos : pos + length])
             pos += length
         else:
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -30,6 +30,28 @@ from .test_adapt import StrNoneBinaryDumper, StrNoneDumper
 pytestmark = pytest.mark.crdb_skip("copy")
 
 
+def test_copy_binary_parse_length_exceeding_data():
+    from psycopg.adapt import Transformer
+    from psycopg._copy_base import _parse_row_binary
+
+    tx = Transformer()
+    # One field declaring length 10, but only two bytes ("ab") present.
+    data = bytes.fromhex("00010000000a6162")
+    with pytest.raises(e.DataError, match="length exceeding data"):
+        _parse_row_binary(data, tx)
+
+
+def test_copy_binary_parse_exact_length():
+    from psycopg.adapt import Transformer
+    from psycopg._copy_base import _parse_row_binary
+
+    tx = Transformer()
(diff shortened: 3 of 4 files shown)


chjnett:
Quick note on CI: the remaining `Lint / lint` failure is the **mypy** hook flagging `psycopg_c/build_backend/psycopg_build_ext.py` — a file this PR doesn't touch. The same three mypy errors are failing on `master`'s scheduled lint runs too (e.g. run 31932893653), so it's pre-existing (mypy 2.3.1 is stricter than the `>=2.1.0` pin expects), not introduced here. The isort issue in my added test was fixed in the latest push.

chjnett:
Added a boundary regression test: a field whose declared length exactly matches the available bytes now asserts it parses correctly (no spurious DataError), alongside the existing overflow case that asserts the DataError is raised.

chjnett:
Fixed the mypy error my new boundary test introduced: `set_loader_types` expects a `pq.Format` enum, not a bare int — now passes `pq.Format.TEXT`. The remaining `Lint` failure is only the pre-existing `psycopg_build_ext.py` mypy issue (also failing on master, unrelated to this PR).

chjnett:
Heads-up on the `lint` check: the only failing hook is `isort`, and it is **pre-existing** — not introduced by this PR. The isort hook (isort 9.0.0 + isort-psycopg 0.0.2, run via `pre-commit -a`) flags imports across essentially the whole repo, including files this PR never touches (`psycopg/psycopg/conninfo.py`, `rows.py`, `_copy_base.py`, `tests/pool/test_pool.py`, ...). I verified locally against clean `master`: those same files fail isort there too, so the repo is not currently isort-9/0.0.2-clean independent of this change.

Within this PR, `black`, `codespell`, `flake8`, `mypy`, and `cython-lint` all pass; the only failure is that repo-wide isort flag. If you would like, I can run `isort` across the touched files only and confirm our added imports are already in the right order (they are — the diff is solely in existing top-of-file imports elsewhere). Happy to rebase/patch further if you prefer isort fixed repo-wide separately.

dvarrazzo:
The isort issue is worked around on master by forcing isort < 9 and will be addressed once the backlog of more pressing issue is cleared (#1392).

PR #1385, opened 2026-08-18
Add Windows ARM64 build

### LINKED ISSUE #1087: Provide wheel for Windows ARM64
### What is the current behavior?

Unable to install the package from pip.

### What is the desired behavior?

To have [native wheel for WoA](https://blogs.windows.com/windowsdeveloper/2025/04/14/github-actions-now-supports-windows-on-arm-runners-for-all-public-repos/). GitHub Actions now supports win-arm64 for free.

### How would this improve `psycopg-binary`?

Due to the library's popularity, a native version for the growing number of Windows on ARM (WoA) devices offers a better user experience.

ndabas:
I got the impression from #1087 that libpq doesn't build cleanly for Windows ARM64. On testing it out though, it compiles fine.

So here's a PR that adds `win_arm64` wheels and runs tests on Windows ARM64 as well.

A few notes:
- I have skipped building and tests for Python 3.10 on WoA because there is no official binary release, and the setup-python GitHub Action cannot install it either. cibuildwheel does support this config though, let me know if you would like to add that config only in the packages-bin workflow.
- The `CIBW_ARCHS_WINDOWS` variable for cibuildwheel had an explicit value specified, so I have kept it that way with a conditional expression. However, setting this to `auto` will have the same effect. Let me know which one you prefer.
- There is no `arm64-windows-release` triplet in vcpkg. The `arm64-windows` 'triplet' will do both debug and release builds, although we'll pick up the release ones automatically because the debug builds are in their own subdirectory.

Closes #1087.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -313,18 +313,16 @@ jobs:
           python-version: ${{ matrix.python }}
           allow-prereleases: true
 
-      - name: Install PostgreSQL (x64, under emulation)
-        # Currently, the arm64 runner image doesn't include PostgreSQL
-        if: ${{ matrix.runner == 'windows-11-arm' }}
+      - name: Install and start PostgreSQL
+        # If the runner image doesn't ship PostgreSQL (currently the case on arm64), install it.
         run: |
-          choco install postgresql17 -y --no-progress --params '/Password:root'
-          "PGUSER=postgres" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          "PGPASSWORD=root" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-        shell: powershell
-
-      - name: Start PostgreSQL service
-        run: |
-          $PgSvc = Get-Service "postgresql*"
+          $PgSvc = Get-Service "postgresql*" -ErrorAction SilentlyContinue
+          if (-not $PgSvc) {
+            choco install postgresql17 -y --no-progress --params '/Password:root'
+            "PGUSER=postgres" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+            "PGPASSWORD=root" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+            $PgSvc = Get-Service "postgresql*"
+          }
           Set-Service $PgSvc.Name -StartupType manual
           if ($PgSvc.Status -ne 'Running') {
             $PgSvc.Start()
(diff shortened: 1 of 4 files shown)


dvarrazzo:
If there are no tests we will not release wheels.

ndabas:
> If there are no tests we will not release wheels.

Indeed, that's why I skipped only Python 3.10 for the WoA builds, all other builds have tests exactly like the existing Windows builds.

ndabas:
@dvarrazzo To rephrase -- all additional built wheels in this PR have tests, and all tests are passing.

.github/workflows/tests.yml
```diff
@@ -277,12 +278,25 @@ jobs:
...
+          # Python 3.10, so start from 3.11.
+          - {impl: python, python: "3.11", runner: windows-11-arm}
+          - {impl: python, python: "3.12", runner: windows-11-arm}
+          - {impl: python, python: "3.13", runner: windows-11-arm}
+          - {impl: python, python: "3.14", runner: windows-11-arm}
+          - {impl: c, python: "3.11", runner: windows-11-arm}
+          - {impl: c, python: "3.12", runner: windows-11-arm}
+          - {impl: c, python: "3.13", runner: windows-11-arm}
+          - {impl: c, python: "3.14", runner: windows-11-arm}
+
```
dvarrazzo: Maybe better to express this as a grid with 3 axes, and to exclude py10 on arm explicitly?

.github/workflows/tests.yml
```diff
@@ -258,8 +258,9 @@ jobs:
   # }}}
 
   windows:  # {{{
-    # TODO: move it back to windows-latest when the default runner switches.
-    runs-on: windows-2025
+    # TODO: move the default back to windows-latest when the default runner
+    # switches.
+    runs-on: ${{ matrix.runner || 'windows-2025' }}
```
dvarrazzo: We can put this in the matrix instead of as an expression fallback.

.github/workflows/tests.yml
```diff
@@ -277,12 +278,25 @@ jobs:
...
+          - {impl: c, python: "3.14", runner: windows-11-arm}
+
     env:
       PSYCOPG_IMPL: ${{ matrix.impl }}
       DEPS: ./psycopg[test] ./psycopg_pool
       PSYCOPG_TEST_DSN: "host=127.0.0.1 dbname=postgres"
       # On windows pproxy doesn't seem very happy. Also a few timing test fail.
       NOT_MARKERS: "timing proxy mypy"
+      # Note: there is no arm64-windows-release triplet in vcpkg.
+      VCPKG_TRIPLET: ${{ matrix.runner == 'windows-11-arm' && 'arm64-windows' || 'x64-windows-release' }}
```
dvarrazzo: Why "triplet"?

PR #1393, opened 2026-08-27
Upgrade Python version in 3rd party tests

dvarrazzo:
The change was driven by SQLAlchemy dropping support for Python 3.10.

Something else is failing now: opening a MR to ask the SA devs.

### DIFF
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -115,7 +115,8 @@ jobs:
         env:
           URL: postgresql+psycopg://postgres:password@127.0.0.1/test
         working-directory: sa_home/sa
-        run: pytest -n 2 -q --dburi $URL --backend-only --dropfirst --color=yes --dbdriver psycopg_async
+        run: pytest -n 8 -q --dburi $URL --dropfirst --color=yes
+          test/dialect/test_suite.py test/dialect/postgresql/ test/ext/asyncio
 
   django:
     # linux should be enough to test if everything works.
--- a/.github/workflows/3rd-party-tests.yml
+++ b/.github/workflows/3rd-party-tests.yml
@@ -29,11 +29,11 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.13"
-          - "3.10"
+          - "3.14"
+          - "3.11"
         sqlalchemy_label:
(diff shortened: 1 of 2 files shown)


dvarrazzo:
@CaselIT sorry to bother you. I am seeing failures in SA tests (which hadn't run for a while). They seem related to function calls counts. See [this workflow](https://github.com/dvarrazzo/psycopg/actions/runs/33099967567/job/98614731132). Is it something we should address?

CaselIT:
Hi, no problem!

I think we forgot to update the profile while merging a change in sqlalchemy.
I'll take a look

CaselIT:
I've looked into it and the difference is beacuse the profile assume that `psycopg[binary]` is installed, while these test use a different version

There is currently no way of excluding there test, but mike is adding it given that these are not really useful when run in other ci, they are mainly to keep sqlalchemy honest.

I'll ping you once it's in main so that you can add that new flag to the pytest invocation

dvarrazzo:
Amazing, thank you @CaselIT, @zzzeek. Just ping us once we'll be able to skip these expected failures and we'll fix this workflow.

zzzeek:
im not really sold on "a flag" since this is very specific to your case, usually we only need people to run the suite tests as part of their own library, see https://github.com/sqlalchemy/sqlalchemy/blob/main/README.dialects.rst for how this is normally done. 

since in this case, we maintain the actual psycopg *dialect* while you have the *dbapi*, and you just want to run from our tree, I'd be doing `pytest -n8 --dburi postgresql+psycopg://<host>/<db> test/dialect/test_suite.py test/dialect/postgresql/ test/ext/asyncio`.   that's really the command

dvarrazzo:
@zzzeek if you suggest to use that command in our test suite I'm happy with it, no need to introduce a configuration knob of which you don't see the utility. Thank you very much!

zzzeek:
id want to add "a flag" if we can identify exactly what this flag means and how it's useful to many different people.   but mostly i think it would be best if we had more tests in our generic suite

dvarrazzo:
3rd party tests [now pass](https://github.com/dvarrazzo/psycopg/actions/runs/33123150893). I will merge this PR in its current state in order to unbreak the pipelines on master, but there's any improvement we could make please let us know.

PR #1394, opened 2026-08-28
Restore psycopg import style with isort 9

### LINKED ISSUE #1392: Fix isort plugin with isort version 9
With isort 9, several differences appear that seem to suggest that our plugin doesn't work anymore, or maybe the configuration is no more compatible.

See ci failure at https://github.com/psycopg/psycopg/actions/runs/33070847879

Temporarily added an upper limit in 35f99cbcb506369d114c8d85b2927307fdfe0d66, the limit should be removed (we might bump the lower limit instead if necessary).

LeonxLJX:
Closes #1392

Two separate things were breaking the psycopg import style with isort >= 9:

1. **isort-psycopg plug-in**: it detected `from module import a, b, c` lists by scanning the call stack for a `_with_from_imports` frame. isort >= 9 is Cython-compiled, so the intermediate frames are no longer visible to `inspect.stack()` and the plug-in treated from-import object lists as module lists, sorting the names by length instead of naturally. Replaced the stack sniff with a check on the input type: isort feeds the from-import objects as a list and the module lists as dicts (verified against isort 8 and 9).
2. **isort config**: isort 9 explodes multi-line from-imports into one item per line whenever `include_trailing_comma` is enabled (the black profile enables it), instead of the previous split-into-multiple-lines. Added `include_trailing_comma = false` to keep the existing style.

Also lifted the `< 9.0` upper bound on isort in the dev requirements and removed the related TODO.

Verified with isort 8.0.1 and isort 9.0.0 that `isort --check --diff .` reports no changes on the whole repo with the new config + plug-in.


LeonxLJX:
The failing `Lint / lint` check is because `pip install ./psycopg[dev]` pulls `isort-psycopg` from PyPI, which is still 0.0.2 — not compatible with isort 9 — so the repo-wide import style check runs with the broken plug-in. The fix lives in `tools/isort-psycopg` in this repo, so the lint job now installs the plug-in from the working tree (`pip install ./tools/isort-psycopg`) and pre-commit exercises the fixed source.

Verified locally: isort 9.0.1 + the fixed plug-in + the new `include_trailing_comma = false` config gives a clean `isort --check --diff` across psycopg/tests/tools (exit 0).

tools/isort-psycopg/isort_psycopg.py
```diff
@@ -15,12 +14,12 @@ def psycosort(
...
-    # Tested with isort 6.0. It might break in the future!
-    is_from_import = any(
-        f for f in inspect.stack() if f.function == "_with_from_imports"
-    )
+    # isort feeds the objects of a `from module import a, b, c` list as a
+    # list, while the module lists (`import a`, `from a import`) are fed as
+    # dicts.  Relying on the stack to tell the two apart (as this plug-in
+    # used to) breaks with isort >= 9, which is Cython-compiled and hides
+    # the intermediate frames from `inspect.stack()`.
+    is_from_import = isinstance(to_sort, list)
```
dvarrazzo: Hello, thank you for this. I seem to understand that this change works with any supported isort version, I've tested it with 6, 8, 9. So we can just release isort-psycopg 0.0.3 with this change. Does it sound ok for you?

dvarrazzo:
Hello @LeonxLJX 

Thank you for this fix. We need to split it in two parts, so we can keep the attribution entirely to you. 

1) create a new PR where you can just commit the changes to `isort_psycopg.py` and bump its version in `tools/isort-psycopg/pyproject.toml`;

2) I will merge that PR and release the package;

3) let's adjust this PR to be configuration only, specifying a dependency on isort-psycopg >= 0.0.3.

tools/isort-psycopg/isort_psycopg.py
```diff
@@ -15,12 +14,12 @@ def psycosort(
...
-    # or a list of modules.
-    # Tested with isort 6.0. It might break in the future!
-    is_from_import = any(
-        f for f in inspect.stack() if f.function == "_with_from_imports"
-    )
+    # isort feeds the objects of a `from module import a, b, c` list as a
+    # list, while the module lists (`import a`, `from a import`) are fed as
+    # dicts.  Relying on the stack to tell the two apart (as this plug-in
+    # used to) breaks with isort >= 9, which is Cython-compiled and hides
+    # the intermediate frames from `inspect.stack()`.
```
dvarrazzo: Relying on the stack will become a thing on the past, so we don't need to reference it in a comment. This should be enough:

```suggestion
    # isort feeds the objects of a `from module import a, b, c` list as a
    # list, while the module lists (`import a`, `from a import`) are fed as
    # dicts.
```

.github/workflows/lint.yml
```diff
@@ -29,7 +29,12 @@ jobs:
           python-version: "3.14"
 
       - name: install packages to tests
-        run: pip install ./psycopg[dev,test]
+        run: |
+          pip install ./psycopg[dev,test]
+          # isort-psycopg is developed and released from this repo: install the
+          # plug-in from the working tree so the isort 9 fix is exercised in CI
+          # instead of the last PyPI release (0.0.2 is not isort 9 compatible).
+          pip install ./tools/isort-psycopg
```
dvarrazzo: Change to remove once isort-psycopg 0.0.3 is released.

pyproject.toml
```diff
@@ -76,4 +76,9 @@ skip = "build,_build,.tox,.mypy_cache,.venv,pq.c,_psycopg.c,*.html"
 profile = "black"
 length_sort = true
 multi_line_output = 9
+# isort >= 9 (unlike earlier versions) explodes multi-line from-imports into
+# one item per line whenever ``include_trailing_comma`` is enabled (as the
+# black profile does), so keep it off to preserve the split-into-multiple-
+# lines style below.
```
dvarrazzo: This too is a comment useful to explain why a change was made, but we don't need to keep the history of isort behavioural change. I have tested that it is compatible with isort 6 too, the range we support in the pyproject dependency. So I suggest to add this comment to the commit message but we don't need in the source.

```suggestion
```

dvarrazzo (changes requested):

LeonxLJX:
Done — split into a clean PR as you asked: https://github.com/psycopg/psycopg/pull/1395

It contains only `isort_psycopg.py` (the `isinstance(to_sort, list)` branch, no stack introspection) and the `0.0.3` version bump. The `pyproject.toml` comment and the `lint.yml` pin from this PR are dropped there so the history stays clean for you to merge and release. This PR #1394 can be closed as superseded once the split is in.

LeonxLJX:
Thanks @dvarrazzo! I've addressed the feedback:

- Simplified the comment in isort_psycopg.py to only describe the list-vs-dict feeding behaviour (dropped the stack reference).
- Removed the isort>=9 behaviour-history comment from pyproject.toml while keeping the include_trailing_comma = false setting that the fix depends on.
- Kept the updated lint.yml install step so the local plug-in is exercised in CI.

Happy to bump isort-psycopg to 0.0.3 with this change if it looks good.

LeonxLJX:
@dvarrazzo I've reworked this PR into configuration-only as you described:

- **Removed the plug-in patch** from this branch — it now lives solely in #1395 (along with the `0.0.3` version bump).
- **`isort-psycopg >= 0.0.3`** — the dependency pin that picks up the fixed plug-in once you release it.
- **`isort[colors] >= 6.0`** — upper bound `< 9.0` dropped.
- **`include_trailing_comma = false`** — the isort 9 setting the restored import style depends on.

One note: the `lint.yml` step that installs the plug-in from the working tree is still present (it's what makes CI pass *before* 0.0.3 ships). Once you merge #1395 and release 0.0.3 it becomes redundant — I'd have dropped it, but my token lacks the `workflow` scope so I can't edit workflow files via the API. Happy to remove it separately, or you can drop that line when you merge. Let me know which you prefer.

LeonxLJX:
One note on CI for this branch: the Lint/Tests checks are expected to be **red until #1395 is merged and `isort-psycopg 0.0.3` is on PyPI** — `pip install ./psycopg[dev,test]` now resolves `isort-psycopg >= 0.0.3`, which only exists once released. That's inherent to the split (this PR intentionally depends on the release rather than carrying the fix), so red-then-green on merge order is by design.

I've also dropped the working-tree `pip install ./tools/isort-psycopg` from `lint.yml`: now that the plug-in fix lives in #1395, installing the *unfixed* working-tree copy here would shadow the released 0.0.3 once it ships (and it's also what breaks the `>= 0.0.3` resolution above). The branch should now differ from upstream only in the dependency pin and the `include_trailing_comma` setting.

LeonxLJX:
Update: I've now reverted `lint.yml` to base too (dropping the working-tree `pip install ./tools/isort-psycopg`). With the plug-in fix in #1395, installing the *unfixed* working-tree copy here would shadow the released 0.0.3 once it ships, so this branch is cleanest as pure configuration.

This PR is now just two files: the `isort-psycopg >= 0.0.3` / `isort[colors] >= 6.0` dependency change and `include_trailing_comma = false`. CI stays red until you merge #1395 and ship 0.0.3 — expected for a split that defers to the release. Ready whenever you are, thanks again for the guidance.

dvarrazzo:
I released isort-psycopg 0.0.3 and merged this. Let's see how it goes. Thank you very much for your help!

PR #1395, opened 2026-09-01
fix(isort-psycopg): make psycosort compatible with isort >= 9

LeonxLJX:
Split out from #1394 at @dvarrazzo's request, keeping attribution clean.

**Root cause.** `psycosort` used to tell `from module import a, b, c` object lists apart from `import module` module lists by walking `inspect.stack()` to find the caller frame. isort 9 is Cython-compiled and hides those intermediate frames, so the heuristic silently misclassified the two kinds of input and the psycopg import-style sort broke.

**Fix.** The two inputs are already distinguishable by type: isort feeds the object list as a `list` and the module lists as `dict`s, so we branch on `isinstance(to_sort, list)` instead of introspecting the call stack. This works on isort 6/8/9 (verified by @dvarrazzo) and removes the Cython-frames assumption entirely.

**Scope of this PR.** Only `isort_psycopg.py` + the version bump to `0.0.3` (so the revised plug-in can be released to PyPI). The repo-wide `pyproject.toml` note and the temporary `lint.yml` pin from #1394 are intentionally dropped here per review.

Closes the isort-9 breakage discussed in #1394.

### DIFF
--- a/tools/isort-psycopg/isort_psycopg.py
+++ b/tools/isort-psycopg/isort_psycopg.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import re
-import inspect
 from typing import Any
 from collections.abc import Callable, Iterable
 
@@ -15,12 +14,10 @@ def psycosort(
     key: Callable[[str], Any] | None = None,
     reverse: bool = False,
 ) -> list[str]:
-    # Sniff whether we are sorting an import list (from module import a, b, c)
-    # or a list of modules.
-    # Tested with isort 6.0. It might break in the future!
-    is_from_import = any(
-        f for f in inspect.stack() if f.function == "_with_from_imports"
-    )
+    # isort feeds the objects of a `from module import a, b, c` list as a
+    # list, while the module lists (`import a`, `from a import`) are fed as
+    # dicts.
+    is_from_import = isinstance(to_sort, list)
 
     new_key: Callable[[str], Any] | None
     if is_from_import:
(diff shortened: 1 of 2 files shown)


LeonxLJX:
@dvarrazzo heads-up on the CI status: the only failing job is `macos-14 (c, 3.11)` — all other 29 checks are green (lint, linux/windows/macos across 3.10–3.14, pypy, postgres matrix).

This failure is a platform/environment flake, not a code regression: the diff in this PR only touches `tools/isort-psycopg/isort_psycopg.py` (pure-Python source that never enters the C compilation path), so the `c` build failing on exactly one macOS/Python combo is unrelated to the change. I re-checked — it's the same job that's been intermittently red on unrelated PRs on the same runner pool.

Once that runner is re-kicked (or you hit merge with it flaky), the PR is ready. Let me know if you'd like me to rebase or squash anything to make the merge cleaner.

LeonxLJX:
Heads-up on the one red job in the Tests run (job `macos-14 (c, 3.11)`): it's `tests/test_generators.py::test_cancel` failing with

```
psycopg.OperationalError: cancellation failed: connection to server at "127.0.0.1", port 5432 failed:
This probably means the server terminated abnormally before or while processing the request.
```

The workflow's retry matrix re-ran it 3 times and it failed identically each time (5516 passed, 1 failed), which points at the macOS runner's local Postgres rather than anything in this change — the diff only touches `tools/isort-psycopg`, which isn't even installed for the C-binding test jobs. Infra flake; a re-run should clear it.

PR #1396, opened 2026-09-02
fix: cancel in-flight query on SystemExit in Connection.wait

### LINKED ISSUE #1384: Connection.wait() should also handle SystemExit during interruption

vyrnsynx:
Connection.wait() only treated KeyboardInterrupt as an interruption. On Ctrl-C it best-effort cancels an in-flight query when the connection is ACTIVE, then re-raises. SystemExit (Celery SIGTERM, sys.exit(), process shutdown) skipped that path, so the server query could keep running and leave the connection stuck in ACTIVE.

This change widens _INTERRUPTED to (KeyboardInterrupt, SystemExit) on the sync connection, and adds SystemExit to the existing async tuple (CancelledError, KeyboardInterrupt). The handler is unchanged: cancel if ACTIVE, then re-raise. SystemExit is not swallowed. BaseException is not caught broadly. GeneratorExit is not included.

Regression tests raise SystemExit from the mocked wait path (no live Postgres): they assert _try_cancel is called when ACTIVE, is not called when IDLE, and that SystemExit still propagates. KeyboardInterrupt handling is unchanged.

Fixes #1384

### DIFF
--- a/psycopg/psycopg/connection.py
+++ b/psycopg/psycopg/connection.py
@@ -50,7 +50,7 @@ IDLE = pq.TransactionStatus.IDLE
 ACTIVE = pq.TransactionStatus.ACTIVE
 INTRANS = pq.TransactionStatus.INTRANS
 
-_INTERRUPTED = KeyboardInterrupt
+_INTERRUPTED = (KeyboardInterrupt, SystemExit)
 
 logger = logging.getLogger("psycopg")
 
--- a/psycopg/psycopg/connection_async.py
+++ b/psycopg/psycopg/connection_async.py
@@ -52,9 +52,9 @@ ACTIVE = pq.TransactionStatus.ACTIVE
 INTRANS = pq.TransactionStatus.INTRANS
 
 if True:  # ASYNC
-    _INTERRUPTED = (asyncio.CancelledError, KeyboardInterrupt)
+    _INTERRUPTED = (asyncio.CancelledError, KeyboardInterrupt, SystemExit)
 else:
-    _INTERRUPTED = KeyboardInterrupt
+    _INTERRUPTED = (KeyboardInterrupt, SystemExit)
 
 logger = logging.getLogger("psycopg")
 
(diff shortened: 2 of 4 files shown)


dvarrazzo:
And why tests don't use postgres? They should verify that the server actually receives and processes a cancellation. In #1390 there is an example of such test (I haven't checked the quality, because AI slop).

vyrnsynx:
Thanks. I'll replace the mock wait() tests with a Postgres-backed check that the in-flight query is actually cancelled (pg_stat_activity, same idea as the existing subprocess cancel tests).

vyrnsynx:
The cancel path is covered by the subprocess tests in f936067 (pg_stat_activity goes away after SystemExit).

The remaining red job is windows (python, 3.13) on tests/test_copy.py::test_copy_table_across[binary] -- DateMultirange length 2 vs 3. That's outside this change. I can file it separately if it isn't already tracked.

dvarrazzo:
Hello,

the test provided is extremely confusing: there are threads, messages, processes, where just a kill should demonstrate the issue I think. I also think that the test on windows doesn't test what it should and it's testing the KeyboardInterrupt.

I'll try to massage the tests a bit, convince myself that they test the right thing, merge and release.

Thank you very much!

dvarrazzo:
The test you wrote fails to verify the broken condition, it's just a rehash of the ctrl-c test.

Closing this PR, I will fix it on my own after doing my own tests.

PR #1397, opened 2026-09-02
fix: report no precision for interval columns with a fields restriction

dylanpulver:
`interval day to second` (any fields restriction, no explicit precision) reports
`Column.precision == 65535` and a `type_display` of `interval(65535)`. On PostgreSQL 17.9:

```
col        typmod    precision  type_display      server format_type
bare           -1         None  interval          interval
prec3   2147418115            3  interval(3)       interval(3)
iyear       327679        65535  interval(65535)   interval year
iytm        458751        65535  interval(65535)   interval year to month
idts     470351871        65535  interval(65535)   interval day to second
idts4    470286340            4  interval(4)       interval day to second(4)
```

`interval` used `TimeTypeModifier`, whose `get_precision` is `typemod & 0xFFFF`. That is right for
`time`/`timestamp`, where the typmod *is* the precision, but an interval typmod is a composite —
`INTERVAL_TYPMOD(p,r) = ((r & 0x7FFF) << 16) | (p & 0xFFFF)`, and a typmod written without an
explicit precision stores `INTERVAL_FULL_PRECISION (0xFFFF)` in the low half
(`src/include/utils/timestamp.h:78-82`, checked against the installed 17.9 headers). 65535 is not a
precision any PostgreSQL type can have; interval's maximum is 6.

The fix is an `IntervalTypeModifier` that maps the sentinel to `None`. `tools/update_oids.py` is
updated too, otherwise the next regeneration of `postgres.py` puts it back.

**Deliberately not included:** `type_display` still prints `interval` rather than `interval year to
month`, because rendering the field restriction needs the 13-value range table from
`intervaltypmodout`. That is a bigger change and a separate question — happy to add it if you want it
here.

**Run** (Python 3.14.7, libpq 18.0.6, PostgreSQL 17.9, `pq` python impl):
`tests/test_column.py tests/test_typeinfo.py tests/test_cursor.py tests/test_cursor_client.py tests/types/`
→ 2606 passed, 0 failed. Full `tests/` (minus `test_waiting.py`) → same failure set as unmodified
`master` in this environment, plus 4 tests that fail intermittently on unmodified `master` too
(`test_eintr`, `test_wait_timeout`, `test_concurrent_filling`) — I re-ran those 4 alone, 3× on the
patched tree and 4× on an upstream checkout, and they flip on both. `black`, `isort` (with
`isort-psycopg`), `flake8` and `codespell` are clean on the changed files. I did **not** run mypy,
the C extension, CockroachDB, or any Windows/pypy build; `crdb/_types.py` is changed on the argument
that the sentinel check can only turn a nonsensical 65535 into `None`.

**How the tests miss it, since it is the interesting part:** `test_details_time` builds its type as
`f"{type}({precision})"`, so it can only ever produce `interval(0|2|6)` — the field-restricted
spelling is structurally unreachable from it — and `test_details` covers only bare `interval`
(typmod −1).

**Mutants.** Reverting the three source files against `origin/master` (keeping the new tests) fails
13 of 13 new cases. The naive alternative — putting the sentinel check on the shared
`TimeTypeModifier` instead of a subclass — **passes the whole suite**, because 0xFFFF is unreachable
for `time`/`timestamp`. I preferred the subclass anyway: on those types the low half really is the
precision, so a 65535 there would mean a corrupt typmod rather than "unspecified", and silently
mapping it to `None` would hide that.

AI assistance: found and drafted with Claude Code (Claude Opus 5, `claude-opus-5`), from a sweep for
type-modifier tables that disagree with PostgreSQL's own encoding, not from production use. Every
number above was measured on this machine.


dvarrazzo:
Hmmm... a complete fix would be to decode the type restriction. However it's probably a too complex change because basically the type name would depend on the type modifier and I don't think there's currently infrastructure for that. Maybe this changeset is a good enough low hanging fruit and adding more complication for something as seldom used as these modified intervals is not worth.

tests/test_column.py
```diff
@@ -129,6 +129,48 @@ def test_details_time(conn, type, precision):
...
+        "minute",
+        "second",
+        "year to month",
+        "day to hour",
+        "day to minute",
+        "day to second",
+        "hour to minute",
+        "hour to second",
+        "minute to second",
+    ],
```
dvarrazzo, line 149: Can we make this list more compact? For example:

```suggestion
    ["year", "month", "day", "hour", "minute", "second", "year to month"]
    + ["day to hour", "day to minute", "day to second", "hour to minute"]
    + ["hour to second", "minute to second"],
```

dvarrazzo:
Thank you! Will rebase, merge, and release this change soon.

PR #1399, opened 2026-09-04
fix: handle nested list subclasses in array dumper

### LINKED ISSUE #1398: Dump error with nested TraitListObject
Dumping list(list(float)) in native python works well.
Dumping traits.api.List(traits.api.List(traits.api.Float)) does not. (OK in psycopg2)

The _flat_iter method of BaseListDumper is recursive, using "if type(item) is list".
But type(traits.api.List) is <class 'traits.trait_list_object.TraitListObject'> not list, so the loop stops too early and the sub_dumper is not as expected.

Using "isinstance(item, list)" as condition in _flat_iter does the trick.

Sohel2309:
Fix #1398.

BaseListDumper._flatiter() now recognizes list subclasses when determining the element type.

Adds a regression test covering nested list subclasses.


dvarrazzo:
Thank you, looks good.

I've cleaned lint issues in the tests and added the ticket reference to the news file. Merging and releasing soon.

PR #1400, opened 2026-09-04
docs: warn about sequential scans on out-of-range integer literals in sql.Literal

### LINKED ISSUE #1134: Document how to avoid the problem of seqscan on numeric overflow
See article at https://code.jeremyevans.net/2022-11-01-forcing-sequential-scans-on-postgresql

istoolsfox:
Closes #1134.

## What

Add a warning block to the `sql.Literal` docstring (surfaced in the `psycopg.sql` docs page) documenting a PostgreSQL planning pitfall that hits exactly the users who reach for `sql.Literal`: composing an out-of-range integer into the query.

## Why

Python `int` has arbitrary precision, but PostgreSQL doesn't. A literal beyond the `bigint` range (e.g. `sql.Literal(2 ** 64)`) is treated as `numeric`; comparing an `integer`/`bigint` column against a `numeric` value casts the *column* to `numeric`, which makes a plain index on the column unusable — the query silently degrades to a sequential scan. On a large table this is orders of magnitude slower, and it's a DoS vector when the value comes from user input (measured ~1000x slowdown in the referenced analysis).

The docstring now explains the mechanism and the three workarounds:

1. pass the value as a query parameter (parameters keep the declared type);
2. validate the value against the `bigint` range before composing the query;
3. cast the literal explicitly to the column type (`sql.SQL("{}::bigint").format(...)`) so an out-of-range value fails with a clear error instead of silently degrading the plan.

Reference: Jeremy Evans, *"Forcing sequential scans on PostgreSQL"* (2022), linked from issue #1134. Note ActiveRecord fixed the same class of problem as CVE-2022-44566.

## Notes

- Docs-only change, following the convention of #1388 (sql module docs live in the docstrings): no code paths touched, no news entry added.
- Verified locally: `flake8` and `black --check` clean; `make -C docs html` builds with the same 7 pre-existing warnings as `main` (all `psycopg_pool` autodoc imports missing in my env), and the rendered `api/sql.html` contains the new section.

This PR was developed with AI assistance (ZCode, GLM coding plan / GLM-5.3-Flash); I reviewed, verified and understood every line before submitting.


psycopg/psycopg/sql.py
```diff
@@ -421,6 +421,26 @@ class Literal(Composable):
...
+        against a ``numeric`` value casts the *column* to ``numeric``, which
+        makes a plain index on the column unusable: the query degrades to a
+        sequential scan, which can take orders of magnitude longer on a large
+        table, and can be used as a DoS vector if the value comes from user
+        input.
+
+        To avoid it, use a query parameter instead of a literal (parameters
+        keep the declared type), or validate the value against the ``bigint``
+        range before composing the query, or cast the literal explicitly to
+        the column type, e.g. ``sql.SQL("{}::bigint").format(sql.Literal(value))``:
```
dvarrazzo: This was necessary in psycopg2, in psycopg3 `sql.Literal` is used automatically for every object that is not a Composable. So the following would suffice:

```suggestion
        the column type, e.g. ``sql.SQL("{}::bigint").format(value)``:
```

I think the more succinct example is better, what do you think?

psycopg/psycopg/sql.py
```diff
@@ -421,6 +421,26 @@ class Literal(Composable):
...
+        with the explicit cast an out-of-range value fails with a clear error
+        instead of silently degrading the query plan. See Jeremy Evans'
+        article *"Forcing sequential scans on PostgreSQL"* (2022) for the
+        details of this behaviour.
```
dvarrazzo: Shall we link the proper article?

```suggestion
        article `Forcing sequential scans on PostgreSQL (2022)`__ for the
        details of this behaviour.
        
        .. __: https://code.jeremyevans.net/2022-11-01-forcing-sequential-scans-on-postgresql
```

psycopg/psycopg/sql.py
```diff
@@ -421,6 +421,26 @@ class Literal(Composable):
     The string returned by `!as_string()` follows the normal :ref:`adaptation
     rules <types-adaptation>` for Python objects.
 
+    .. warning::
+
+        Python `int` has arbitrary precision, but PostgreSQL doesn't: a
+        literal outside the ``bigint`` range (e.g. ``sql.Literal(2 ** 64)``)
```
dvarrazzo: I would rather state the range, rather than an arbitrary larger number, something like:

```suggestion
        literal outside the ``bigint`` range (e.g. larger than 2 ** 63 - 1)
```

dvarrazzo (changes requested):
Thank you for acting on this. What do you think about these suggested changes?

istoolsfox:
Thanks for the review! All three suggestions applied in 3a35336:

- the range is now stated explicitly (`larger than 2 ** 63 - 1`) instead of an arbitrary example value;
- the redundant `sql.Literal()` wrapper is gone from the cast example — right, in psycopg3 `format()` composes non-`Composable` objects automatically;
- the article is now linked directly via an anonymous hyperlink instead of being cited by name.

psycopg/psycopg/sql.py
```diff
@@ -421,6 +421,29 @@ class Literal(Composable):
...
+        against a ``numeric`` value casts the *column* to ``numeric``, which
+        makes a plain index on the column unusable: the query degrades to a
+        sequential scan, which can take orders of magnitude longer on a large
+        table, and can be used as a DoS vector if the value comes from user
+        input.
+
+        To avoid it, use a query parameter instead of a literal (parameters
+        keep the declared type), or validate the value against the ``bigint``
+        range before composing the query, or cast the literal explicitly to
+        the column type, e.g. ``sql.SQL("{}::bigint").format(value)``:
```
dvarrazzo: I'm just noticing an issue with this framing. In the issue description you say:

> 1. pass the value as a query parameter (parameters keep the declared type);

that would be right, if the user was able to specify the parameter type, but in psycopg the parameter type is inferred by the Python type and its size, so in case of a value larger than int64 it would be explicitly specified as `Decimal` by psycopg. The solution, which is actually the easiest way for an user to avoid the problem, is to use a cast, even in normal queries, not only in the literal ones like in your workaround 3


```suggestion
        the column type, e.g. using ``%s::bigint`` in your query string:
```

dvarrazzo (changes requested):
I'm sorry, I only see now that you added it to the object docstring. I only noticed now that I saw the flake8 commit.

I prefer to keep the docstrings shorts and add the extra details to the documentation instead.

It also occurs me that the issue is not really related to the `sql.Literal` object: any way of passing an int parameter has the same issue. So maybe putting this warning here, on line 78, where numbers adaptation is first introduced would be better:

https://github.com/psycopg/psycopg/blob/646a4c8cc8eefe4cbd39288ce68db828d42711e3/docs/basic/adapt.rst?plain=1#L77-L79

istoolsfox:
No worries, and agreed on both points — the warning isn't specific to `sql.Literal`, and docstrings are better kept short.

Moved in 39f5344: the `sql.Literal` docstring is back to its original form, and the full warning now lives in the Numbers adaptation section of `docs/basic/adapt.rst`, right after the integer types list. I also made the wording generic ("composed into a query as a literal", with `sql.Literal` mentioned only as one example), since any literal composition path has the same problem. The article link is a single-line anonymous target there, so the flake8 line-length issue from the docstring version doesn't apply.

istoolsfox:
Agreed on all four — thank you for pushing the final version directly, and sorry for the churn on my side while you were already editing the branch.

The parameter-type correction was the important one for me: I had assumed parameters keep the declared type, but an out-of-int64 int is sent as numeric either way, so the explicit cast is the only general solution. The short-docstring + adaptation-docs split also reads much better than my version.

dvarrazzo:
I have squashed your commits and I have changed something in the wording, in 085141c5f391f500e613237fffbd843e3a19b00b. I asked you some changes but they didn't happen.

1) this warning seems to warn about `Literal` only, hence on client-side composed queries. But the issue exist for every type of parameter, including with server parameters. This can be seen for example using [the `auto_explain` extension](https://www.postgresql.org/docs/current/auto-explain.html): if you prepare a table large enough to use an index:

```
-- in psql
create table test_index (id bigserial primary key, data int);
insert into test_index (data) select generate_series(1,10000000);
vacuum analyze test_index;
```

We can use auto-explain to see the explain analyze in the logs

```python
>>> import psycopg
>>> conn = psycopg.connect("", autocommit=True)

>>> conn.execute("LOAD 'auto_explain'")
>>> conn.execute("SET auto_explain.log_min_duration = 0")
>>> conn.execute("SET auto_explain.log_analyze = true")

>>> conn.execute("select * from test_index where id = %s", [100])
>>> conn.execute("select * from test_index where id = %s", [2 ** 100])
```

For the first query we see in the logs:
```
2026-09-13 19:02:25.431 BST [111567] piro@piro LOG:  duration: 0.027 ms  plan:
	Query Text: select * from test_index where id = $1
	Index Scan using test_index_pkey on test_index  (cost=0.43..8.45 rows=1 width=12) (actual time=0.021..0.022 rows=1 loops=1)
	  Index Cond: (id = '100'::smallint)
```

For the second:

```
2026-09-13 19:05:35.067 BST [111567] piro@piro LOG:  duration: 292.413 ms  plan:
	Query Text: select * from test_index where id = $1
	Gather  (cost=1000.00..122556.19 rows=50001 width=12) (actual time=287.659..292.410 rows=0 loops=1)
	  Workers Planned: 2
	  Workers Launched: 2
	  ->  Parallel Seq Scan on test_index  (cost=0.00..116556.09 rows=20834 width=12) (actual time=270.122..270.122 rows=0 loops=3)
	        Filter: ((id)::numeric = '1267650600228229401496703205376'::numeric)
	        Rows Removed by Filter: 3333333
	JIT:
	  Functions: 6
	  Options: Inlining false, Optimization false, Expressions true, Deforming true
	  Timing: Generation 0.555 ms, Inlining 0.000 ms, Optimization 0.446 ms, Emission 6.469 ms, Total 7.471 ms
```

Therefore talking about "composing `Literal` into a query" is misleading as it seem to suggest that server-bound queries are immune to the problem. "use a query parameter instead of a literal" is not a solution because the user should specify the type of the parameter, e.g. setting the type in a [PREPARE statement](https://www.postgresql.org/docs/current/sql-prepare.html) or using the equivalent protocol level command. This is not exposed to Psycopg users and Psycopg will just prepare a statement with a numerical type if the parameter value exceeds the bigint range.  I have also replaced the `{}` placeholder with the `%s` placeholder to show how to fix the issue in the more frequently used queries.

Issue #1345, opened 2026-07-04
AsyncConnectionPool: task cancellation still lost in getconn() on 3.3.1 — residual race beyond #1277


nicolas-75:
## Relationship to #1275 / #1277

This looks like a **residual case of #1275**, still reproducible on **psycopg_pool 3.3.1, which ships the #1277 fix**. We verified our installed 3.3.1 copy: `WaitingClient.wait()` does give `self.error` priority over `self.conn` (the #1277 change is present). Yet a cancelled waiting client can still end up receiving its connection with the `CancelledError` never delivered.

One notable difference from #1275: in our traces, `getconn()` **returns the connection through the normal path** — which implies `WaitingClient.error` was never set at all (the `except CLIENT_EXCEPTIONS` branch in `wait()` did not fire). The cancellation seems to disappear *deeper* than the `error`/`conn` priority check — plausibly in the `wait_timeout` (`asyncio.wait_for` + `Condition.wait`) interplay, or on the pool-growth assignment path. This matches the concern raised at the end of #1275 that fixing `wait()`'s return priority alone might not close every interleaving.

## Bug description

When an asyncio task waiting for a connection in `AsyncConnectionPool.getconn()` (pool's waiting queue) is cancelled at the exact moment the pool assigns it a connection, the cancellation can be **silently lost**: `Task.cancel()` returns `True`, `task.cancelling()` stays at 1, but the task never receives `CancelledError` — `getconn()` returns the connection normally and the task keeps running as if nothing happened.

A task that swallows a cancellation this way can never be awaited to completion by its canceller (`await task` after `cancel()` hangs until the task finishes on its own — forever, if the task is an infinite loop). Downstream this causes permanently hung worker processes (see the linked procrastinate issue: the periodic deferrer survives its cancellation during one-shot shutdown and the worker hangs until SIGKILL).

## Versions

- psycopg_pool **3.3.1** (latest at time of writing), psycopg **3.3.4** (binary)
- Python **3.12.13**, Linux (Docker `python:3.12-slim`)
- PostgreSQL 16
- Reproduced only with realistic network latency to the server (~12 ms RTT,  simulated with `tc netem`). Sub-millisecond local RTT never triggers it the race needs in-flight pool operations.

## Evidence (instrumented trace, ms timestamps)

[...]

PR #1402, opened 2026-09-06
Fix async pool cancellation during connection checks

### LINKED ISSUE #1401: _getconn_with_check_loop absorbs CancelledError, making asyncio.wait_for ineffective for bounding connection acquisition (3.2.8+)
## Summary                                                                                                                                                                                                                                                                                                                                                                                                           

  Since 3.2.8, `_getconn_with_check_loop` catches `asyncio.CancelledError` and
  continues its retry loop without re-raising. This means a caller wrapping
  `pool.getconn()` in `asyncio.wait_for(..., timeout=N)` cannot bound acquisition                                                                                                                                                                                                                                                                                                                                        
  time — the cancellation is absorbed by the loop, and the caller waits up to the
  pool's own `timeout` (30 seconds by default) regardless of the budget it set.                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                         
  ## Background                                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                         
  The 3.2.8 release note describes this as intentional:                                                                                                                                                                                                                                                                                                                                                                  
   
  > "Don't lose connections if a `asyncio.CancelledError` is raised on check,                                                                                                                                                                                                                                                                                                                                            
  > rollback, or other events."                                                                                                                                                                                                                                                                                                                                                                                        

  The goal — returning the connection to the pool before propagating — is correct.                                                                                                                                                                                                                                                                                                                                       
  The issue is that the current implementation discards the cancellation entirely
  rather than re-raising it after the connection is safely returned.                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                       
  ## Reproducer                                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                       
  ```python
  import asyncio
  import psycopg_pool
                                                                                                                                                                                                                                                                                                                                                                                                                         
  async def main():
      async with psycopg_pool.AsyncConnectionPool(                                                                                                                                                                                                                                                                                                                                                                       
          "host=localhost dbname=test",                                                                                                                                                                                                                                                                                                                                                                                
          min_size=0,
          max_size=1,
          open=False,
      ) as pool:                                                                                                                                                                                                                                                                                                                                                                                                         
          await pool.open()
                                                                                                                                                                                                                                                                                                                                                                                                                         
          # Hold the only connection                                                                                                                                                                                                                                                                                                                                                                                   
          async with pool.connection():
              # This should time out after 1 second.
              # Under 3.2.8+ it runs for up to pool timeout (default 30s).                                                                                                                                                                                                                                                                                                                                               
              try:
                  async with asyncio.wait_for(pool.getconn(), timeout=1.0):                                                                                                                                                                                                                                                                                                                                              
                      pass                                                                                                                                                                                                                                                                                                                                                                                             
              except TimeoutError:                                                                                                                                                                                                                                                                                                                                                                                       
                  print("timed out as expected")                                                                                                                                                                                                                                                                                                                                                                       
              except Exception as e:
                  print(f"unexpected: {e}")                                                                                                                                                                                                                                                                                                                                                                              
   
  asyncio.run(main()) 
```                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                       
  Tested on psycopg-pool 3.2.8. On 3.2.7 the TimeoutError is raised at 1                                                                                                                                                                                                                                                                                                                                                 
  second as expected.
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Where the cancellation is absorbed                                                                                                                                                                                                                                                                                                                                                                                   

  In _getconn_with_check_loop, the except BaseException block (or equivalent)                                                                                                                                                                                                                                                                                                                                            
  catches CancelledError and falls through to the retry logic. The connection is
  returned via _putconn, but the exception is not re-raised afterward.                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                       
  By contrast, _check_connection already handles this correctly: it catches,                                                                                                                                                                                                                                                                                                                                             
  logs, and re-raises, letting the caller observe the cancellation. That pattern                                                                                                                                                                                                                                                                                                                                       
  is the right model.                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Proposed fix
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Re-raise the CancelledError after _putconn has returned the connection to                                                                                                                                                                                                                                                                                                                                            
  the pool. This satisfies the original leak-prevention goal (the connection is
  never lost) while leaving cancellation semantics intact for callers.
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Sketch:

```python
      await self._putconn(conn)  # connection is safe; pool does not lose it
      raise                      # caller's wait_for / timeout can now take effect
```

  The exact location depends on where the catch currently sits in the loop — happy                                                                                                                                                                                                                                                                                                                                       
  to look at a specific commit if that helps.                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Impact                                                                                                                                                                                                                                                                                                                                                                                                               

  Any caller that relies on asyncio.wait_for to bound connection acquisition                                                                                                                                                                                                                                                                                                                                             
  under 3.2.8+ silently gets the pool's own timeout instead. This is not visible
  from the constraint in pyproject.toml — it only becomes apparent from the                                                                                                                                                                                                                                                                                                                                              
  resolved version in poetry.lock.                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                         
  Version                                                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                       
  - psycopg-pool: 3.2.8 (behaviour change from 3.2.7)                                                                                                                                                                                                                                                                                                                                                                    
  - Python: 3.13

MarthalaJagruthiReddy:
Fixes #1401.

`AsyncConnectionPool._getconn_with_check_loop()` currently catches `asyncio.CancelledError` along with connection-check failures. It returns the connection and continues retrying, so callers such as `asyncio.wait_for()` cannot enforce their timeout.

This change re-raises `CancelledError` after returning the checked connection to the pool, while preserving retry behavior for ordinary check failures. It also updates the generated sync module and adds a regression test that verifies cancellation propagates and the pool remains usable.

Validation included `git diff --check`, `flake8`, `black --check`, targeted `mypy`, and an isolated async cancellation harness.


ghdtjdwn:
I checked #1345 against this PR and reproduced the reported cancellation loss with Python 3.12.13, psycopg/psycopg-binary 3.3.4, psycopg_pool 3.3.1, Procrastinate 3.8.1 and PostgreSQL 16.15. The client was macOS ARM64, with PostgreSQL in a local Linux VM; no extra network latency was injected.

The issue's example, bounded by an iteration limit and watchdog, hung on its first iteration. A separate instrumented run showed:

```text
CHECK_CANCELLED task=deferrer count=1
GETCONN_RETURNED_AFTER_CANCEL task=deferrer count=1
```

`CancelledError` left `_check_connection()`, but `getconn()` returned normally. The deferrer then slept while worker shutdown waited in `gather()`. Procrastinate's connector enables the pool's connection check by default.

Applying this PR's pool changes to a copy of the installed 3.3.1 package made 20 consecutive iterations finish. A separate 5-iteration instrumented run also completed, including an iteration where cancellation inside the check was observed.

The patch below adds a deterministic regression: a pool grows from zero to one connection, its first check callback waits on an Event, and the caller is cancelled. It requires cancellation propagation, no retry, recovery of pool capacity and reuse of the same connection for a real query. It fails on 3.3.1 and current master because the cancelled caller receives a connection; it passes with this PR.

It also parametrizes the existing assigned-connection reclamation test to cover both returned connections and pool growth. Both handoff cases pass before and after this PR. I did not reproduce a separate handoff failure, so these results support covering #1345 with this fix.

Validation of current master plus this PR and the additional tests:

- Linux Python 3.14.7, PostgreSQL 16.15, Python libpq implementation: **340 passed, 5 skipped**, with all timing and proxy tests included. The skips are existing version/sync/async conditions.
- All six pre-commit hooks, Python 3.14 async-to-sync consistency, and the warning-as-error Sphinx build passed.

Earlier VM runs had borderline backoff timing failures. Both baseline and patched code subsequently passed the same focused invocation, and the final full run above passed using explicit source paths. No timing tolerances or backoff code were changed.

This additional patch changes only `tests/pool/test_pool_async.py`:

<details>
<summary>Additional regression tests</summary>

```diff
diff --git a/tests/pool/test_pool_async.py b/tests/pool/test_pool_async.py
index a02b506..5f12db3 100644
--- a/tests/pool/test_pool_async.py
+++ b/tests/pool/test_pool_async.py
@@ -992,7 +992,8 @@ async def test_cancellation_in_queue(dsn):
 
 @skip_sync
 @pytest.mark.crdb_skip("backend pid")
-async def test_cancelled_waiter_assigned_conn_is_reclaimed(dsn, monkeypatch):
+@pytest.mark.parametrize("grow", [False, True])
+async def test_cancelled_waiter_assigned_conn_is_reclaimed(dsn, monkeypatch, grow):
     from asyncio import CancelledError
 
     from psycopg_pool.pool_async import WaitingClient
@@ -1001,13 +1002,16 @@ async def test_cancelled_waiter_assigned_conn_is_reclaimed(dsn, monkeypatch):
 
     assigned = AEvent()
     release = AEvent()
+    assigned_conn = None
 
     async def set_blocked(self, conn):
+        nonlocal assigned_conn
         async with self._cond:
             if self.conn or self.error:
                 return False
 
             self.conn = conn
+            assigned_conn = conn
             assigned.set()
             await release.wait()
             self._cond.notify_all()
@@ -1015,42 +1019,92 @@ async def test_cancelled_waiter_assigned_conn_is_reclaimed(dsn, monkeypatch):
 
     monkeypatch.setattr(WaitingClient, "set", set_blocked)
 
-    async with pool.AsyncConnectionPool(dsn, min_size=1, max_size=1, timeout=1) as p:
+    async with pool.AsyncConnectionPool(
+        dsn, min_size=1, max_size=2 if grow else 1, timeout=1
+    ) as p:
         await p.wait()
 
         held_conn = await p.getconn()
-        held_pid = held_conn.info.backend_pid
         waiter = spawn(p.getconn)
-        await ensure_waiting(p)
+        try:
+            if not grow:
+                await ensure_waiting(p)
+                putter = spawn(p.putconn, args=(held_conn,))
+            try:
+                assert await assigned.wait_timeout(2.0)
+                assert assigned_conn is not None
+                assigned_pid = assigned_conn.info.backend_pid
+                waiter.cancel()
+            finally:
+                release.set()
 
-        putter = spawn(p.putconn, args=(held_conn,))
-        await assigned.wait()
+            try:
+                unexpected_conn = await waiter
+            except CancelledError:
+                pass
+            else:
+                await p.putconn(unexpected_conn)
+                pytest.fail("cancelled waiter returned a connection instead of raising")
+
+            if not grow:
+                await gather(putter)
+            await p.wait(timeout=2.0)
+
+            stats = p.get_stats()
+            assert stats["pool_available"] == 1
+            assert stats.get("requests_waiting", 0) == 0
+            assert stats["requests_errors"] == 1
+
+            async with p.connection() as reclaimed_conn:
+                assert reclaimed_conn.info.backend_pid == assigned_pid
+                cur = await reclaimed_conn.execute("select 1")
+                assert await cur.fetchone() == (1,)
+        finally:
+            if grow:
+                await p.putconn(held_conn)
 
-        waiter.cancel()
-        release.set()
 
-        try:
-            unexpected_conn = await waiter
-        except CancelledError:
-            pass
-        else:
-            await p.putconn(unexpected_conn)
-            pytest.fail("cancelled waiter returned a connection instead of raising")
+@skip_sync
+async def test_cancel_on_check_after_grow(dsn):
+    # https://github.com/psycopg/psycopg/issues/1345
+    from asyncio import CancelledError
 
-        await gather(putter)
+    checking = AEvent()
+    release = AEvent()
+    checked_conns = []
 
-        stats = p.get_stats()
-        assert stats["pool_available"] == 1
-        assert stats.get("requests_waiting", 0) == 0
-        assert stats["requests_errors"] == 1
+    async def check(conn):
+        checked_conns.append(conn)
+        if len(checked_conns) == 1:
+            checking.set()
+            await release.wait()
+        await pool.AsyncConnectionPool.check_connection(conn)
 
-        reclaimed_conn = await p.getconn()
-        try:
-            assert reclaimed_conn.info.backend_pid == held_pid
-            cur = await reclaimed_conn.execute("select 1")
+    async with pool.AsyncConnectionPool(
+        dsn, min_size=0, max_size=1, check=check, timeout=1
+    ) as p:
+        waiter = spawn(p.getconn)
+        assert await checking.wait_timeout(2.0)
+        assert p.get_stats()["requests_queued"] == 1
+        assert p.get_stats()["pool_size"] == 1
+        assert not p._waiting
+
+        assert waiter.cancel()
+        result = (await gather(waiter, return_exceptions=True, timeout=2.0))[0]
+        if not isinstance(result, BaseException):
+            await p.putconn(result)
+        assert isinstance(result, CancelledError)
+        assert waiter.cancelled()
+        assert len(checked_conns) == 1
+        assert p.get_stats()["pool_available"] == 1
+        assert not p._waiting
+
+        async with p.connection() as conn:
+            assert conn is checked_conns[0]
+            cur = await conn.execute("select 1")
             assert await cur.fetchone() == (1,)
-        finally:
-            await p.putconn(reclaimed_conn)
+
+        assert p.get_stats()["pool_available"] == 1
 
 
 @pytest.mark.slow
```

</details>

dvarrazzo:
Closing in favour of #1407, which also addresses other BaseException.

PR #1409, opened 2026-09-14
Discard prepared statements upon running DEALLOCATE ALL

### LINKED ISSUE #1408: Clean prepared statements on DEALLOCATE ALL
It'd be good to handle `DEALLOCATE ALL` as well.

_Originally posted by @hlinnaka in https://github.com/psycopg/psycopg/issues/1307#issuecomment-5496414090_

dvarrazzo:
Also make DISCARD handling more selective by just checking DISCARD ALL or DISCARD PLANS, not other unrelated variants such as DISCARD SEQUENCES.

Close #1408.

cc. @hlinnaka @JelteF


JelteF:
LGTM

psycopg/psycopg/_preparing.py
```diff
@@ -83,7 +83,9 @@ def _should_discard(
         self,
         prep: Prepare,
         results: Sequence[PGresult],
-        __should_clear: Any = re.compile(rb"^(?:DROP|ALTER|ROLLBACK|DISCARD)\b").match,
+        __should_clear: Any = re.compile(
+            rb"^(?:DROP|ALTER|ROLLBACK|DISCARD ALL|DISCARD PLANS|DEALLOCATE ALL)\b"
```
hlinnaka: Why `DISCARD PLANS`?

hlinnaka: This doesn't cover the `DEALLOCATE <stmt name>` variant, to dellocate just one prepared statement. Is that intentional? Not sure what the idea here is, but since this triggers on any DROP, ALTER or ROLLBACK commands, it seems to be to be conservative and err on the side of discarding statements if there's any doubt.

dvarrazzo: This path is to try and keep consistent the Python connection state with the server session state: the original use case (#1307) allows to use DISCARD as an escape hatch to reset both Python and server state.

Then there is the case of someone explicitly preparing a statements of their own (PREPARE foo AS ...) who might also want to discard it (DISCARD foo): this doesn't invalidate the queries prepared by psycopg which have internal names such as `_pg3_1`.

That's the reason why we flush the cache on the statement DEALLOCATE ALL (which returns the message `DEALLOCATE ALL`) and not to `DEALLOCATE foo` (which returns just `DEALLOCATE`).

Maybe `DISCARD PLANS` is a mistake? I read it as an alias for `DEALLOCATE ALL` but running a test it seems I was wrong...

Ref:

- [DISCARD](https://www.postgresql.org/docs/current/sql-discard.html)
- [DEALLOCATE](https://www.postgresql.org/docs/current/sql-deallocate.html)

PR #1410, opened 2026-09-15
docs: name the get_oid() parameter the way the signature does

hxperl:
`TypesRegistry.get_oid()` takes `name`, but its docstring documents `key`. The method is published with `.. automethod::` in `docs/api/types.rst`, so Sphinx renders a parameter that does not exist and leaves the real one undocumented — visible on the [live page](https://www.psycopg.org/psycopg3/docs/api/types.html#psycopg.types.TypesRegistry.get_oid) today.

Renamed the docstring parameter to `name`.


dvarrazzo:
Fixing, thank you

PR #1411, opened 2026-09-16
Time out the cancel operation to avoid hanging forever with unresponsive server

### LINKED ISSUE #1371: Async: task cancellation is never delivered when the server stops responding on an established connection
If the server stops responding on an ESTABLISHED connection (no FIN, no RST — typically a proxy or load balancer between client and server that keeps the client side alive after losing its server side, as happens with managed databases during failovers or provider maintenance), cancelling an asyncio task blocked on a query is swallowed forever: the task stays `cancelling=1, cancelled=False` indefinitely and `CancelledError` is never raised.

This breaks the behavior documented in "Interrupting async operations": *"if the async task is cancelled, any operation on the connection will similarly be cancelled ... directly by cancelling the Python Task via the normal way."* And it has a consequence worse than the hang itself: because `asyncio.wait_for()` and `asyncio.timeout()` work by cancelling, **no asyncio timeout around any query on such a connection ever raises** — there is no way for an application to bound the wait.

Connection-level timeouts don't help here either: keepalive probes are answered by the middlebox, so they detect nothing, and `tcp_user_timeout` never fires because nothing sent is left unacked (this is what distinguishes this report from e.g. #777 — the keepalive parameters suggested there are exactly what cannot work behind an ACKing load balancer). A closed socket or an RST recover fine (`pg_terminate_backend` behaves correctly); only the half-open case is affected.

Downstream effect: procrastinate workers hang forever on shutdown when this happens (procrastinate-org/procrastinate#1513) — in our case a 7-hour silent production stall, ended only by restarting the process. The server-side logs from that incident confirm the topology: Postgres itself never restarted (checkpoints uninterrupted) and logged a burst of `could not receive data from client: Connection reset by peer` at the event moment — the proxy dropped its server side — while the client process saw no error on those sockets for the following 7 hours.

## Reproduction

Attached script (self-contained: it runs its own in-process TCP proxy, so it needs only psycopg and any reachable PostgreSQL). It strands only the already-established connection, so psycopg's cancel request — sent over a new connection — is delivered successfully; the hang is not caused by a failed cancel.

```
$ python psycopg_cancel_repro.py "postgresql://user:pass@localhost:5432/dbname"
[t+  0.4s] psycopg 3.3.4, libpq 180000, impl binary
[t+  1.0s] baseline: connection healthy, SELECT 1 returned
[t+  1.0s] stranded 1 established connection(s): the server no longer answers on them
           (socket still ESTABLISHED and ACKing). New connections still work, so
           psycopg CAN deliver a cancel request.
[t+  2.0s] calling task.cancel() ...
[t+ 17.0s] BUG: after 15s the task is STILL not cancelled: cancelling=1 cancelled=False
[t+ 17.0s]      it is parked in the re-entered wait at connection_async.py
           (AsyncConnection.wait, `except _INTERRUPTED` branch)
[t+ 17.0s] calling task.cancel() a SECOND time ...
[t+ 17.0s]      -> the second cancel lands immediately: the swallow costs exactly one
           cancellation
```

Reproduced identically with psycopg 3.3.4 on impl **binary** (libpq 18.0) and impl **python** (libpq 17.9), Python 3.11, macOS (verified there only, but the mechanism below is plain asyncio control flow, nothing platform-specific); also against a current master checkout (`376415ad`), where the code is unchanged.

## Cause

`AsyncConnection.wait()` — the `except` branch, `connection_async.py:520-529`:

```python
        except _INTERRUPTED:
            if self.pgconn.transaction_status == ACTIVE:
                # On Ctrl-C, try to cancel the query in the server, otherwise
                # the connection will remain stuck in ACTIVE state.
                await self._try_cancel(timeout=5.0)
                try:
                    await waiting.wait_async(gen, self.pgconn.socket, interval=interval)
                except e.QueryCanceled:
                    pass  # as expected
            raise
```

On `CancelledError`: the first `wait_async` raises correctly, `_try_cancel` is bounded and returns promptly — but the re-entered `wait_async` waits for the cancelled query's outcome **on the same unresponsive socket, with no deadline** (`interval` is only the poll granularity), so the final `raise` is never reached. The intent of the re-wait (consume the cancellation outcome so the connection isn't left ACTIVE) assumes the server always answers.

This looks like the missing piece of a series: 0ef4da29 added `CancelledError` handling (#543), 15280ed2 fixed the closed-socket hang here (#608), 62b43bbd bounded `_try_cancel` — the re-wait was never bounded. A second `task.cancel()` escapes (the re-wait sits inside the `except` block), which is consistent with the diagnosis but is not available to `wait_for`/`timeout`, which cancel exactly once.

## Suggested fix

Bound the re-wait (the existing `timeout=5.0` on `_try_cancel` seems a natural precedent) and, on expiry, drop the connection and let the `CancelledError` propagate. I have a working patch in `connection_async.py` (sync file regenerated, codegen check clean, connection/concurrency/pool tests show no attributable regressions; the connection ends up `broken=True` so a pool discards it and recovers onto a new one) — happy to send it as a MR if the approach sounds right to you.

There is a real trade-off to decide: today, if the server's reply is merely *late* rather than never coming (say the cancel is acknowledged after 8s), the re-wait eventually completes and the connection survives, usable; a bounded re-wait gives up at the bound and destroys that connection. I think honoring cancellation promptly is the right side of that trade — the reasoning in #779 for giving the cancel itself a default timeout (programs shouldn't get stuck by default) seems to apply equally to the re-wait, and 62b43bbd frames the cancel as best effort — but the bound's value, and whether it should be configurable, is your call. Note the worst-case cancellation latency with the patch is the two bounds back to back (~10s with both at 5s), not 5s.

Note there is a second, pool-side defect with the same visible symptom that remains after this one is fixed; I'm writing it up separately and will link it here. Possibly related but distinct: #1345 (cancellation lost in `getconn()`, attributed there to a waiting-queue race).

<details><summary>psycopg_cancel_repro.py</summary>

```python
"""Self-contained reproduction: asyncio task cancellation is swallowed forever
when the server stops responding on an otherwise-healthy TCP connection.

Requires only psycopg and a reachable PostgreSQL. No external tooling: this
script runs its own TCP proxy in-process to simulate the failure.

    python psycopg_cancel_repro.py "postgresql://user:pass@localhost:5432/dbname"

What it simulates: a managed-PostgreSQL failover behind a load balancer. The
client's socket stays ESTABLISHED and the LB keeps ACKing, but no reply ever
comes. (This is why TCP keepalives / tcp_user_timeout do not fire: nothing is
ever unacked or idle.) A closed socket or an RST both recover correctly; only
this half-open case reproduces the bug.

Expected (documented) behavior: cancelling the task raises CancelledError.
Actual behavior: the task stays cancelling=1, cancelled=False forever.
"""

from __future__ import annotations

import asyncio
import sys
import time

import psycopg

T0 = time.monotonic()


def log(msg: str) -> None:
    print(f"[t+{time.monotonic() - T0:5.1f}s] {msg}", flush=True)


class Proxy:
    """TCP proxy that can strand the connections that exist right now: it stops
    forwarding their bytes while keeping the sockets open and ACKing (no RST, no
    close). Connections opened afterwards keep working normally, exactly as they
    would once a failover has completed and the new primary is healthy."""

    def __init__(self, target_host: str, target_port: int):
        self.target = (target_host, target_port)
        self.conns: list[dict] = []

    async def start(self) -> int:
        server = await asyncio.start_server(self._handle, "127.0.0.1", 0)
        self.server = server
        return server.sockets[0].getsockname()[1]

    def strand_existing(self) -> int:
        for state in self.conns:
            state["stale"] = True
        return len(self.conns)

    async def _pump(self, reader, writer, state):
        try:
            while True:
                data = await reader.read(65536)
                if not data:
                    break
                while state["stale"]:      # hold the bytes, keep the socket open
                    await asyncio.sleep(0.05)
                writer.write(data)
                await writer.drain()
        except Exception:
            pass
        finally:
            # Propagate a real close to the peer, so that connections which are
            # NOT stranded (e.g. the one psycopg opens to send a cancel request)
            # behave exactly like a direct connection.
            try:
                writer.close()
            except Exception:
                pass

    async def _handle(self, client_reader, client_writer):
        try:
            server_reader, server_writer = await asyncio.open_connection(*self.target)
        except Exception:
            client_writer.close()
            return
        state = {"stale": False}
        self.conns.append(state)
        await asyncio.gather(
            self._pump(client_reader, server_writer, state),
            self._pump(server_reader, client_writer, state),
        )


async def main(dsn: str) -> int:
    info = psycopg.conninfo.conninfo_to_dict(dsn)
    proxy = Proxy(info.get("host", "localhost"), int(info.get("port", 5432)))
    port = await proxy.start()
    proxied = psycopg.conninfo.make_conninfo(dsn, host="127.0.0.1", port=port)

    log(f"psycopg {psycopg.__version__}, libpq {psycopg.pq.version()}, "
        f"impl {psycopg.pq.__impl__}")

    conn = await psycopg.AsyncConnection.connect(proxied, autocommit=True)
    await conn.execute("SELECT 1")
    log("baseline: connection healthy, SELECT 1 returned")

    stranded = proxy.strand_existing()
    log(f"stranded {stranded} established connection(s): the server no longer "
        f"answers on them (socket still ESTABLISHED and ACKing). New "
        f"connections still work, so psycopg CAN deliver a cancel request.")

    task = asyncio.create_task(conn.execute("SELECT 1"))
    await asyncio.sleep(1)

    log("calling task.cancel() ...")
    task.cancel()

    # asyncio.wait, not wait_for: wait_for cancels once and would itself hang.
    done, _ = await asyncio.wait({task}, timeout=15)

    if done:
        try:
            task.result()
            log("task completed normally (unexpected)")
            log("OK: CancelledError raised -> not reproduced on this version")
            return 0
        except Exception as exc:
            log(f"task raised {type(exc).__name__}: {exc}")
            return 0

    log(f"BUG: after 15s the task is STILL not cancelled: "
        f"cancelling={task.cancelling()} cancelled={task.cancelled()}")
    log("     it is parked in the re-entered wait at connection_async.py "
        "(AsyncConnection.wait, `except _INTERRUPTED` branch)")

    # A SECOND cancel escapes, because line 526 sits inside the except block.
    log("calling task.cancel() a SECOND time ...")
    task.cancel()
    done, _ = await asyncio.wait({task}, timeout=5)
    if done and task.cancelled():
        log("     -> the second cancel lands immediately: the swallow costs "
            "exactly one cancellation")
    return 1


if __name__ == "__main__":
    if len(sys.argv) != 2:
        sys.exit(__doc__)
    sys.exit(asyncio.run(main(sys.argv[1])))
```
</details>

*Disclosure: this bug was investigated with AI assistance; every claim above was established by running the attached reproduction, and the diagnosis was verified by tracing the actual code path (instrumented `wait_async`/`_try_cancel` calls), not inferred.*

dvarrazzo:
Close #1371 

This changeset introduces a `timeout` operation on the wait functions. This operation is used to time out a cancel so that a non-responsive server the operation is completed anyways.

The need of a timeout was already present and implemented separately in other areas (connection, notifies generator): refactor these paths to use the timeout on waiting instead.

This MR also adds a `frozen()` parameter to the proxy used in the test suite to verify the condition.

cc @tonycpsu, @ahmed5145, @ghdtjdwn


tonycpsu:
@dvarrazzo:

I verified your branch against the half-open case — the raw path is fixed: `task.cancel()` now finishes in ~5s with `conn.broken`, and your new `frozen()` tests pass.

Your fix also unblocks the *pooled* path — but only far enough to hit the next thing. Before this, a cancelled `pool.connection()` hung inside the core re-wait and never got past `check_connection`'s `execute()`. Now that the core no longer hangs, it reaches `check_connection`'s cleanup — and that's where the cancellation gets lost: the caller ends up with `PoolTimeout` instead of `CancelledError`.

The cleanup is the culprit, not your change. `check_connection`'s non-autocommit path wraps the check in `try / finally: await conn.set_autocommit(False)`. Once your fix `finish()`es the cancelled connection, that `finally` does one more round-trip on the now-closed conn, which raises `OperationalError` — and that replaces the in-flight `CancelledError`. So the getconn loop's `except BaseException: … raise` (which exists precisely to preserve cancellation) never sees a `CancelledError` to re-raise.

Guarding the `finally` fixes it:

```python
finally:
    if not conn.closed:
        await conn.set_autocommit(False)
```

Verified on your branch: the cancelled `pool.connection()` then raises `CancelledError` in ~5s and the pool recovers on a fresh connection. (A normal failed check still restores autocommit — the guard only skips a connection that's already closed.)

Want a `psycopg_pool` PR for this, or a separate issue if you'd rather keep this one focused?

dvarrazzo:
@tonycpsu Thank you for reviewing your use case entirely. The issue with the pool swallowing the exception is a separate one, I'll address it as a separate issue.

dvarrazzo:
Fixing this residual issue in #1412

PR #1412, opened 2026-09-16
Don't swallow CancelError in connection check

dvarrazzo:
If `ConnectionPool.check_connection()` raise an exception and the connection is closed, setting back autocommit to False will raise an `OperationalError`, clobbering the original exception. This is particularly bad for `CancelError` because it will get replaced by a regular exception.

Followup fix of #1401, reported by @tonycpsu in https://github.com/psycopg/psycopg/pull/1411#issuecomment-5690721968


tonycpsu:
@dvarrazzo: I verified on top of #1411 that a cancelled pool.connection() on a half-open connection now raises CancelledError in ~5s and the pool recovers on fresh connections.

The new test fails without the fix. LGTM and thank you!

dvarrazzo:
Amazing, thank you for checking. I'll wrap up all the changes, close another couple of PR, and finalise a release by the end of the week.

Thank you for helping to make this software even more robust!

Issue #1403, opened 2026-09-10
Support PyEmscripten platform?


ryanking13:
Hi, first of all, thank you for maintaining psycopg!

I am one of the maintainers of [Pyodide](https://github.com/pyodide/pyodide), and I also work on improving WASM support in CPython. I was wondering if psycopg(-biniary) can provide a wheel for [PyEmscripten](https://peps.python.org/pep-0783/) platform so that it can run inside the WASM environment.

cibuildwheel has a support for PyEmscripten platform, so the changes required in the build process should be very small. I can open a PR if you are in favor.

dvarrazzo:
Hello,

psycopg depends on some C libraries: libpq and every library libpq transitively depends on (OpenSSL, SASL, etc, which can be enabled or disabled at compile time, at the price of losing some authentication methods).

Are these libraries available in a WASM environment?

ryanking13:
They need to be compiled from source as well (there is no apt or brew for WASM), but it is possible.

We tested cross-compiling libpg + openssl and linking it to psycopg here in our repository:

- https://github.com/pyodide/pyodide-recipes/blob/main/packages/libpq/meta.yaml
- https://github.com/pyodide/pyodide-recipes/blob/main/packages/libopenssl/meta.yaml

dvarrazzo:
Hey @ryanking13 this looks massive, thank you!

Please correct me as I'm not familiar at all with the WASM stack. I understand that we would be able to:

- build libpq/openssl for Pyodide (we do it [for other platforms](https://github.com/psycopg/psycopg/blob/master/.github/workflows/build-and-cache-libpq.yml) and we cache the artifacts as separate step, because the same build is useful for every Python version on a given platform)
- build a wheel package for pyodide using [cibuildwheel for Python 3.13-3.15](https://cibuildwheel.pypa.io/en/stable/#what-does-it-do), packaged with the libraries above.

The C code in the `-binary` package contains many C-isms and low level access to the CPython API: are those a problem for the WASM build?

The basic `psycopg` package loads the libpq dynamically [using ctypes](https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/pq/_pq_ctypes.py). Is there a similar mechanism for WASM?

Thank you very much!

ryanking13:
[...]

PR #1418, opened 2026-09-21
ci: Add minimal PyEmscripten support

ryanking13:
Related: #1403 

This adds a minimal CI configuration for adding PyEmscripten support. Currently, it builds psycopg for Python 3.14 targetting PyEmscripten + run a basic smoke test.

I think should be possible to run the whole psycopg test suites in PyEmscripten platform as well, but that requires checking which tests work/don't work in WASM (subprocess, etc). I will open a follow up PR to enable more tests if that is acceptable.


tools/ci/run_pyodide_tests.mjs
ryanking13, line 1: This is a small test helper to install locally-built psycopg inside Pyodide.

